PR #62, opened 2015-02-09
Add Properties object

### LINKED ISSUE #61: Improve Property Management
We need a better way to differentiate between removing an entity's property and setting that property to a NULL value. See #25 for previous discussion.

blowmage:
Properties allows for properties to be removed from an entity,
not simply setting the value to NULL.
Properties is a hash-like data structure that does several things for us.
It ensures that only strings are stored as keys, although it allows
symbols to be used for getting and setting values.

Fixes #61

### DIFF
--- a/lib/gcloud/datastore/entity.rb
+++ b/lib/gcloud/datastore/entity.rb
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 require "gcloud/datastore/key"
+require "gcloud/datastore/properties"
 require "gcloud/datastore/proto"
 
 module Gcloud
@@ -33,8 +34,7 @@ module Gcloud
       ##
       # Create a new Entity object.
       def initialize
-        @_entity = Proto::Entity.new
-        @_entity.property = []
+        @properties = Properties.new
         @key = Key.new
         @_exclude_indexes = {}
       end
@@ -43,12 +43,10 @@ module Gcloud
       # Retrieve a property value.
       #
       #   puts entity["name"]
+      #
+      # If a property doesn't exist then nil will be returned.
       def [] prop_name
-        prop_name = prop_name.to_s
-        prop = Array(@_entity.property).find { |p| p.name == prop_name }
-        Proto.from_proto_value prop.value
-      rescue
-        nil
+        @properties[prop_name]
       end
 
       ##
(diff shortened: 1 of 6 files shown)


blowmage:
@pcostell Could you take a look at this?

lib/gcloud/datastore/entity.rb
```diff
@@ -216,9 +222,9 @@ def map_exclude_flag_to_value flag, value #:nodoc:
...
       ##
       # Update the exclude data after a new object is created.
-      def update_exclude_indexes! #:nodoc:
+      def update_exclude_indexes! entity #:nodoc:
         @_exclude_indexes = {}
-        Array(@_entity.property).each do |property|
+        Array(entity.property).each do |property|
           @_exclude_indexes[property.name] = property.value.indexed
           unless property.value.list_value.nil?
             exclude = Array(property.value.list_value).map(&:indexed)
```
pcostell: Technically each value in a list can choose to be indexes or not.

pcostell:
LGTM - though just FYI I haven't read/written ruby in a long time.

lib/gcloud/datastore/entity.rb
blowmage: Yep. We support that now actually. Take a look at [test/gcloud/datastore/test_entity_exclude.rb](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/test/gcloud/datastore/test_entity_exclude.rb) for the covered cases. You can set one exclude value to be used for each value the list:

```ruby
entity["tags"] = ["ruby", "code"]
entity.exclude_from_indexes! "tags", true
entity.exclude_from_indexes?("tags") #=> [true, true]
```

Or you can set a list of exclude values:

```ruby
entity["tags"] = ["ruby", "code"]
entity.exclude_from_indexes! "tags", [true, false]
entity.exclude_from_indexes?("tags") #=> [true, false]
```

Or, my favorite, you can pass a  block with the exclude logic to be evaluated on each value in the list:

```ruby
entity["tags"] = ["ruby", "code"]
entity.exclude_from_indexes! "tags" do |tag|
  tag =~ /r/
end
entity.exclude_from_indexes?("tags") #=> [true, false]
```

pcostell: Oh that's sweet! I clearly need to brush up on my ruby skills :-)

PR #69, opened 2015-02-10
Update README

### LINKED ISSUE #68: Update README.md to fit consistent style
See https://github.com/GoogleCloudPlatform/gcloud-meta/issues/3 for the format.

blowmage:
Changes to bring the README into alignment with the proposed format.

Closes #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,12 @@
 # gcloud
 
+Idiomatic Ruby client for [Google Cloud Platform](https://cloud.google.com/) services.
+
 [![Travis Build Status](https://travis-ci.org/GoogleCloudPlatform/gcloud-ruby.svg)](https://travis-ci.org/GoogleCloudPlatform/gcloud-ruby/)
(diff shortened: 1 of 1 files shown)


blowmage:
@jgeewax So far this PR adds links to the homepage and API docs. The only sections in the proposal that are missing are "example applications" and "versioning". We don't have any example applications that I'm aware of, and we haven't discussed versioning. (I assume semver.) What would you like me to put for these? Are there other sections that are missing?

jgeewax:
Are we OK with doing http://semver.org/ ?

if so, we can use the wording from gcloud-python:

> Versioning
> 
> This library follows Semantic Versioning.
> 
> It is currently in major version zero (0.y.z), which means that anything may change at any time and the public API should not be considered stable.

blowmage:
I'm :+1: on using semver. I've updated the README to include a section on versioning. Anything else needed for the README?

jgeewax:
If we're in line with https://github.com/GoogleCloudPlatform/gcloud-meta/issues/3 then we're all good.

PR #100, opened 2015-02-25
Add Support section to README

blowmage:
Direct folks to StackOverflow for questions.

Brazenly stolen from [google-auth-library-ruby](https://github.com/google/google-auth-library-ruby).


blowmage:
@jgeewax Do you like this? Is this something that should be added to the unified README standard?

PR #101, opened 2015-02-25
Simplify Convenience Methods

blowmage:
Move the convenience methods to the top-level Gcloud namespace. This reduces the typing needed to get started.


blowmage:
@jgeewax Thoughts?

PR #108, opened 2015-03-30
Remove Gcloud.devserver

blowmage:
Remove convenience method and env vars for local devserver support.
The only way to specify a devserver is through DATASTORE_HOST.
Update CONTRIBUTING guide with demonstration of using DATASTORE_HOST.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,14 @@ $ export STORAGE_TEST_KEYFILE=/path/to/other/keyfile.json
 $ rake test:regression
 ```
 
+### Local Datastore Devserver
+
+You can run the Datstore regression tests against a devserver running locally. To switch to the devserver set the `DATASTORE_HOST` environment variable with the location of the local devserver.
+
+``` sh
+$ DATASTORE_HOST=http://127.0.0.1:8080 rake test:regression:datastore
+```
+
 ## Coding Style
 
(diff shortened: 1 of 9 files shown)


pcostell:
LGTM -- I wonder if it would be useful to optionally provide the host as an argument to `gcloud.datastore`. Then you could write a unittest setUp method that forced a localhost. But totally up to you.

blowmage:
@pcostell I like the symmetry we have with datastore and storage and pubsub all taking the same parameters, project/dataset id and credentials/keyfile. In addition to setting the `DATASTORE_HOST` environment variable we could expose and support the `http_host` attribute on Dataset. Right now it isn't exposed on the public API.

```ruby
dataset = Gcloud.datastore "test123", "path/to/keyfile.json"
dataset.http_host = "http://localhost:8080"
entity = dataset.find "Thing", 789
```

pcostell:
@blowmage Sounds fine to me.

blowmage:
Recording some additional thoughts here. What I don't like about that approach is that changing the host is mutable state and that could have some adverse consequences later. We could stretch to support the following signature:

```ruby
dataset = Gcloud.datastore
dataset = Gcloud.datastore "test123"
dataset = Gcloud.datastore "test123", "path/to/keyfile.json"
dataset = Gcloud.datastore "test123", "path/to/keyfile.json", host: "http://localhost:8080"
dataset = Gcloud.datastore "test123", host: "http://localhost:8080"
dataset = Gcloud.datastore host: "http://localhost:8080"
```

blowmage:
I'm gonna accept this as-is. If we want to support more options for specifying the host beyond `DATASTORE_HOST` then we should treat that as a separate enhancement.

PR #110, opened 2015-04-16
Save file with binary mode

premist:
When using Google Cloud Storage, File#download opens local file with [non-binary mode](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/storage/file.rb#L143).
I tried downloading sample image from the bucket and it raised an encoding conversion error.

This simply adds binary switch to ensure that binary files stored on GCS can also be downloaded.


blowmage:
Can you rebase on master? CI is failing because rubocop had some new rules that needed to be accounted for.

premist:
Ok, I will open another PR from master branch.

Issue #109, opened 2015-04-16
Keyfile from environment variables


premist:
Do you have any plans to allow setting json key file from environment variables?

Right now I toss all credentials via environment variables, and gcloud ruby library is the only exception which requires json file to be placed somewhere in the filesystem.

blowmage:
Currently Gcloud only supports the project identifier and keyfile path in environment variables. The path must point to an actual file.

```ruby
require "gcloud/datastore"

ENV["DATASTORE_PROJECT"] = "project-identifier"
ENV["DATASTORE_KEYFILE"] = "/path/to/keyfile.json"

@datastore = Gcloud.datastore
@entity = @datastore.find "Task", 123
```

Here is what I think you are asking for:

```ruby
require "gcloud/datastore"

ENV["DATASTORE_PROJECT"] = "project-identifier"
ENV["DATASTORE_KEYFILE_JSON"] = "{\"private_key_id\": \"...\", \"private_key\": \"...\", \"client_email\": \"...\", \"client_id\": \"...\", \"type\": \"...\"}"

@datastore = Gcloud.datastore
@entity = @datastore.find "Task", 123
```

Is that right?

premist:
Yes, that's right. Or since JSON file contains string values, it would be possible if gcloud allows setting them separately via environment variable.

PR #111, opened 2015-04-16
Allow JSON in Environment Variables

blowmage:
This change allows the keyfile JSON contents to be accessed from environment variables. It allows for configuration without writing the file to disk. This is useful when using pre-packaged docker containers and other deployment scenarios.

Currently Gcloud only supports the project identifier and keyfile path in environment variables. The path must point to an actual file.

```ruby
require "gcloud/datastore"

ENV["DATASTORE_PROJECT"] = "project-identifier"
ENV["DATASTORE_KEYFILE"] = "/path/to/keyfile.json"

@datastore = Gcloud.datastore
@entity = @datastore.find "Task", 123
```

This change allows for the JSON to be in the environment variable instead of being read from a file.

```ruby
require "gcloud/datastore"

ENV["DATASTORE_PROJECT"] = "project-identifier"
ENV["DATASTORE_KEYFILE_JSON"] = "{\"private_key_id\": \"...\", \"private_key\": \"...\", \"client_email\": \"...\", \"client_id\": \"...\", \"type\": \"...\"}"

@datastore = Gcloud.datastore
@entity = @datastore.find "Task", 123
```

Addresses #109

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -1,0 +1,71 @@
+# Authentication
+
+## Creating a Service Account
+
+Gcloud aims to make authentication as simple as possible. Google Cloud requires a **Project ID** and **Service Account Credentials** to connect to the APIs. To create a service account:
+
+1. Visit the [Google Developers Console](https://console.developers.google.com/project).
+2. Create a new project or click on an existing project.
+3. Navigate to **APIs & auth** > **APIs section** and turn on the following APIs (you may need to enable billing in order to use these services):
+  * Google Cloud Datastore API
+  * Google Cloud Storage
+  * Google Cloud Storage JSON API
+4. Navigate to **APIs & auth** > **Credentials** and then:
+  * If you want to use a new service account, click on **Create new Client ID** and select **Service account**. After the account is created, you will be prompted to download the JSON key file that the library uses to authorize your requests.
+  * If you want to generate a new key for an existing service account, click on **Generate new JSON key** and download the JSON key file.
+
+You will use the **Project ID** and **JSON file** to connect to services with gcloud.
+
+## Project and Credential Lookup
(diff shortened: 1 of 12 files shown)


blowmage:
@jgeewax @silvolu @thagomizer Thoughts?

thagomizer:
I like it. More flexibility is good and not having to write out files makes some scenarios easier.

jgeewax:
Looks like the explicit JSON overrides the file path silently. I foresee some frustrating debugging unless we log a warning that you're providing two explicit keys...

Any thoughts on that?

blowmage:
I don't think it overrides as much as it adds a new step in the lookup process for authentication. Here is the old process:

1. Specify credentials file provided in code
2. Specify credentials file provided in env var
3. Discover credentials file in the Cloud SDK's path
4. Discover GCE credentials

The new process with this change is now:

1. Specify credentials file provided in code
2. Specify credentials file provided in env var
3. Specify credentials json provided in env var
4. Discover credentials file in the Cloud SDK's path
5. Discover GCE credentials

I do think that this needs to be spelled out very clearly in the documentation. But as long as it is communicated well I don't believe it will be confusing or frustrating.

jgeewax:
Cool - yes no problem at all with the order of evaluation. LGTM then, so long as we document this somewhere :)

See https://github.com/GoogleCloudPlatform/gcloud-python/pull/830/files for my first pass at outlining how auth works for gcloud-python. If that looks OK, we should open an issue to do likewise for gcloud-ruby.

blowmage:
I have updated this to work on the recent pubsub additions, and added an Authentication guide spelling out the order followed. Thoughts?

thagomizer:
Thanks @blowmage

Issue #137, opened 2015-05-11
Is there a way to globally configure Gcloud and have that configuration trickle down?


jgeewax:
The examples I see are things like

```ruby
require 'gcloud/storage'

storage = Gcloud.storage
bucket = storage.find_bucket "my-bucket"
```

If I wanted to configure this, I have to configure each item individually:

```ruby
require 'gcloud/storage'
require 'gcloud/datastore'

storage = Gcloud.storage "my-project-id"
datastore = Gcloud.datastore "my-project-id"
```

Is there a way to configure things at the Gcloud level? that is:

```ruby
require 'gcloud'

gcloud = Gcloud "my-project-id"
storage = gcloud.storage
dataster = gcloud.datastore
```

blowmage:
An issue I see with making all services available on the initial require is that gcloud could be loading many, many services that are never used. What happens when gcloud supports Datastore, Storage, Cloud SQL, BigQuery, Pub/Sub, Compute Engine, DNS, Load Balancing, Monitoring, Logging, etc? Do we really want to load all that if the user only wants to make calls to Storage? If the services are scoped to a require then the user gets to control which code is loaded, which should be beneficial in a production environment.

That's my thinking so far, but I'm happy to change those thoughts. :)

jgeewax:
Maybe you can help me understand how much overhead it is to "load" all these service.

In Python and Node it's not that big of a deal to import the module.

The idea would be that someone doing this would be using **at least 2** of the services -- because the whole point is that you want to configure all of your stuff one time.

Is it reasonable to at least make this *possible*, while saying "if you're only using one service, this is dumb and will be slower..." ?

blowmage:
My concern about loading services is about dependencies. Datastore requires the beefcake library for protobuf, while Storage requires the mime-types and digest-crc libraries for dealing with files. IMO gcloud shouldn't load those dependencies unless they are needed. (And shouldn't load `Gcloud::Datastore` or `Gcloud::Storage` namespaces until needed either.) The easiest, most idiomatic way to manage dependencies is to put them behind a require. That said, we can make your proposed API work, but we want to make sure we aren't loading things before they are needed. I'll give it some thought.

PR #139, opened 2015-06-04
Create top-level Gcloud object

blowmage:
Allows us to create a top level gcloud object instead of creating service objects on the modules/namespace. The top-level gcloud instance object creates service objects using the shared credentials.

Before:

```ruby
require "gcloud/datastore"
require "gcloud/storage"

datastore = Gcloud.datastore
entity = datastore.find "Task", 1234

storage = Gcloud.storage
bucket = storage.bucket "my-todos-bucket"
```

After:

```ruby
require "gcloud"

gcloud = Gcloud.new

datastore = gcloud.datastore
entity = datastore.find "Task", 1234

storage = gcloud.storage
bucket = storage.bucket "my-todos-bucket"
```
For now the various namespaces and requires still exist. This change is additive and optional.

Thoughts?

[refs #137]

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -17,10 +17,10 @@ Follow the [activation instructions](https://cloud.google.com/datastore/docs/act
 See the [gcloud-ruby Datastore API documentation](rdoc-ref:Gcloud::Datastore) to learn how to interact with the Cloud Datastore using this library.
 
 ```ruby
-require "gcloud/datastore"
+require "gcloud"
 
-dataset = Gcloud.datastore "my-todo-project-id",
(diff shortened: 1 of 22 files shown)


jgeewax:
Can I still use the old syntax as well or do we require the new one ?

For people doing configuration, I like the idea, but I can certainly see
the logic for wanting to have the "shorter" syntax....

blowmage:
You can still use the old syntax. The old vs. new is purely optional. Resources are loaded as needed. For example, if you never call `Gcloud.new.datastore` then you will never load the protobuf code.

jgeewax:
LGTM :+1:

PR #140, opened 2015-06-08
readme: point service links to readme intro documentation

stephenplusplus:
Hey, guys!

RE: GoogleCloudPlatform/gcloud-node#637

@jgeewax thinks it makes more sense to point the Service links (Google Cloud Datastore and Google Cloud Storage) to their respective sections in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,8 @@ Idiomatic Ruby client for [Google Cloud Platform](https://cloud.google.com/) ser
 
 This client supports the following Google Cloud Platform services:
 
-* [Google Cloud Datastore](https://cloud.google.com/datastore/) ([docs](https://cloud.google.com/datastore/docs))
-* [Google Cloud Storage](https://cloud.google.com/storage/) ([docs](https://cloud.google.com/storage/docs/json_api/))
+* [Google Cloud Datastore](#datastore)
+* [Google Cloud Storage](#storage)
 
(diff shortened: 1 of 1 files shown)


blowmage:
Before we can accept this we need to make sure that the generated documentation matches GitHub's markdown output. Currently, GitHub and RDoc generates the HTML anchors differently:

https://github.com/GoogleCloudPlatform/gcloud-ruby#datastore
vs.
http://googlecloudplatform.github.io/gcloud-ruby/docs/master/#label-Datastore

We need to configure RDoc to use the GitHub markdown parser.

jgeewax:
For other libraries, the README.md isn't slurped up into the gh-pages generated documentation. 

Could the fix here be to keep those two isolated from one another?

blowmage:
That is certainly an option. The default is show the README, but we can change that. Node shows the `gcloud` documentation on the documentation homepage, and python shows a separate Getting Started guide (not the README) on the documentation homepage. Which would you prefer ruby shows?

jgeewax:
I think some sort of quick start would be pretty useful, but ... dealer's choice here. We don't know which works best, so I have no problem with trying things and changing them based on feedback.

blowmage:
K, I'll try to get this done today.

blowmage:
README was removed from the documentation in 0b0ad8b93b1bd73c53a61162dbd950fe027f9298, so this can now be accepted. Thanks!

PR #172, opened 2015-06-19
Update Pubsub Documentation

blowmage:
Overhaul the documentation with parameter definitions and several code examples.

### DIFF
--- a/lib/gcloud/pubsub.rb
+++ b/lib/gcloud/pubsub.rb
@@ -20,18 +20,31 @@ require "gcloud/pubsub/project"
 # Google Cloud Pub/Sub
 module Gcloud
   ##
-  # Create a new Pubsub project.
+  # Creates a new object for connecting to the Pubsub service.
+  # Each call creates a new connection.
   #
-  #   pubsub = Gcloud.pubsub "my-todo-project",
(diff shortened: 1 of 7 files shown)


lib/gcloud/pubsub.rb
```diff
@@ -43,14 +56,168 @@ def self.pubsub project = nil, keyfile = nil
...
+  #   pubsub = Gcloud.pubsub
+  #
+  #   topic = pubsub.topic "my-topic"
+  #   msg = topic.publish "new-message"
+  #
+  # Messages can also be published with attributes:
+  #
+  #   require "glcoud/pubsub"
+  #
+  #   pubsub = Gcloud.pubsub
```
jgeewax: Can we condense these a bit? Feels like they're extra spaced apart -- almost needlessly so.

blowmage: You mean the empty line between `pubsub =` and `topic =`?

jgeewax: Yessir.

Particular for the 3-line examples, where each line gets a newline.

blowmage: K, will update.

PR #176, opened 2015-06-29
Add BigQuery

blowmage:
Work in progress for BigQuery...

### Datasets

- [x] Create a dataset
- [x] List datasets
- [x] Get a dataset
- [x] Update a dataset (name, description, expiration)
- [x] Delete a dataset
- [x] Represent extended attributes from get

### Tables

- [x] Create a table
- [x] List tables
- [x] Get a table
- [x] Update a table (name, description, schema)
- [x] Delete a table
- [x] Create a view
- [x] Update the query used by the view
- [x] Represent extended attributes from get

### Jobs

- [x] Create a job
- [x] List jobs
- [x] List jobs with total count on list
- [x] Get a job
- [x] Knows the job's state (pending/running/done)
- [x] Knows if there are errors (status/errorResult and status/errors)
- [x] Represents a Copy job
- [x] Represents an Export job (destinationUriFileCounts)
- [x] Represents a Load job (inputFiles, inputFileBytes, outputRows, outputBytes)
- [x] Represents a Query job (totalBytesProcessed)
- [x] Represent Job status error data
- [x] Represent Job statistics data
- [x] Refresh a job

### Insert Data

- [x] Load records to a table directly (insertAll)
- [x] Upload data to a table
- [x] Import Storage file to a table

### Query Data

- [x] Create an query
- [x] Create a Batch query
- [x] Query with query cache disabled
- [x] Query to a table
- [x] Query with default dataset

### Query Results

- [x] Get query results from a job (pagination, etc.)

### Export Data

- [x] Export to single file
- [x] Export to multiple files

### Copy Data

- [x] Copy data to new table
- [x] Copy data to new dataset and table
- [x] Copy data to new project and dataset and table

### Other

- [x] Have entities grab the full resource data when initially created by a list call (`gcloud.bigquery.datasets.first.tables.first.description` vs. `gcloud.bigquery.dataset("test").table("test").description`) (Look for creationTime in the google api object)
- [x] Link Data?
- [x] Regression tests
- [x] Documentation
- [x] README
- [x] Documentation Left Nav


lib/gcloud/bigquery.rb
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "gcloud"
+require "gcloud/bigquery/project"
+
+##
+# Google Cloud Bigquery
```
jgeewax: BigQuery* -- capital Q (unlike Bigtable -- lowercase T... confusing I know...)

lib/gcloud/bigquery/connection.rb
```diff
@@ -0,0 +1,200 @@
...
+      def get_dataset dataset_id
+        @client.execute(
+          api_method: @bigquery.datasets.get,
+          parameters: { projectId: @project, datasetId: dataset_id }
+        )
+      end
+
+      ##
+      # Creates a new empty dataset.
+      def insert_dataset options = {}
```
jgeewax: I know the API method is called "insert", but doesn't "create" make more sense?

lib/gcloud/bigquery/dataset.rb
```diff
@@ -0,0 +1,161 @@
...
+          Table.from_gapi resp.data, connection
+        else
+          # TODO: Handle ALREADY_EXISTS and NOT_FOUND
+          fail ApiError.from_response(resp)
+        end
+      end
+
+      ##
+      # Retrieves a table by name.
+      def table table_name
```
jgeewax: should this be "table" or "get_table" ? Thought "get" was our way of saying "this will make an API call" ?

lib/gcloud/bigquery/connection.rb
blowmage: This is low level stuff and not part of the gcloud public API. I agree "create" makes more sense for the gcloud public API.

lib/gcloud/bigquery.rb
blowmage: I had found some Java code that used "Bigquery". I originally had "BigQuery", but changed it to be more inline with the other code I was seeing. Easy enough to change back.

lib/gcloud/bigquery/dataset.rb
blowmage: It will have a `get_table` alias. WIP and all that. :)

blowmage:
The main documentation page is not 100% complete, but the rest of the documentation is. We're accepting this PR so we can get the docs showing up on the gcloud-ruby site.

PR #190, opened 2015-07-16
Pub/Sub Documentation Clarification

blowmage:


quartzmo:
I've signed the CLA, I'm good to go.

blowmage:
@jgeewax Have you seen googlebot do this before? Can we both push commits to a PR?

jgeewax:
I can confirm that @blowmage and @quartzmo both signed the CLA. Good to merge if only those two authors are on the PR.

Double check that you've signed with the right e-mail addresses and all that too (commit author e-mail needs to match)

tmatsuo:
LGTM

PR #223, opened 2015-08-17
Improve tests

blowmage:
Several improvements to the tests:

* Add autotest as a development dependency
* Add minitest-focus as development dependency
* Add red/green test output
* Rename tests so they are suffix named. e.g. `test/gcloud_test.rb` instead of `test/test_gcloud.rb`
* Rename regression tests to acceptance tests


quartzmo:
Bravo, this should really improve the testing, looks great.

PR #233, opened 2015-08-20
Allow use of custom scopes when connecting to service APIs

blowmage:
Allow users to assert a specific OAuth 2.0 scope when connecting to the API.

### DIFF
--- a/lib/gcloud.rb
+++ b/lib/gcloud.rb
@@ -70,11 +70,26 @@ module Gcloud
   # Creates a new object for connecting to the Datastore service.
   # Each call creates a new connection.
   #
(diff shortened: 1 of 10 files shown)


quartzmo:
For the [doc of the `scope` option](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/233/files#diff-38e0b358ff5015275e482c57036c4327R77), I found a nice description in the [Using OAuth guide](https://developers.google.com/identity/protocols/OAuth2)

> controls the set of resources and operations that an access token permits

Would you like to add that?

blowmage:
:+1: Wording is updated.

PR #240, opened 2015-08-25
Add Datastore backup format option to Table#load

### LINKED ISSUE #205: Import data from Datastore backup
The BigQuery API can import data from a Datastore backup, but this is not exposed in the gcloud API yet.

quartzmo:
This includes adding the `projection_fields` option to Table#load
for specifying which entity fields to load from the backup, as
well as detection of format from the file extention `.backup_info`.

[closes #205]


blowmage:
Yay! Looks good to me.

Issue #242, opened 2015-08-25
BigQuery table refresh


blowmage:
In BigQuery we added `Job#refresh!` to update the data for the job. This has proven to be very useful, we should extend this to Table and View.

blowmage:
I kinda wonder if we should refresh all the things... Thoughts?

quartzmo:
The question is where to stop... I know Table makes sense because I find myself trying it in console, and feeling frustrated that it's not there. I say we add it to Table for now, and see where else we wish we had it.

blowmage:
I meant being able to refresh a Storage Bucket and File, a Pub/Sub Topic and Subscription, and a Datastore Entity. Extending this pattern to more than just BigQuery.

quartzmo:
Agreed. If you know where it's needed, let's do it.

Question: Any chance you want to go with `reload!` instead of `refresh!` if we are going to do this widely? Fewer characters and similar to http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-reload

blowmage:
Currently Storage ACL and BigQuery Job use `refresh!`. Nothing in the public API uses `reload!`, although we can add an alias for it.

quartzmo:
OK, I vote for adding a `reload!` alias.

PR #247, opened 2015-08-25
Add id and refresh! to Bigquery Table

### LINKED ISSUE #227: BigQuery Table#id
I would like to ask a table what its ID is. It would be useful for using the fully namespaced table name in queries.

```ruby
dataset.query "SELECT name, age FROM [#{table.id}]"
```

quartzmo:
closes #227, #242

### DIFF
--- a/lib/gcloud/bigquery/table.rb
+++ b/lib/gcloud/bigquery/table.rb
@@ -127,6 +127,31 @@ module Gcloud
       end
 
       ##
+      # The combined Project ID, Dataset ID, and Table ID for this table, in the
+      # format specified by the {Query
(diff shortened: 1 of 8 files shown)


quartzmo:
I changed the Project ID in the BigQuery tests to make it more realistic by adding a dash. An id containing dashes must be surrounded with square brackets in a query. Sorry, should have noted in commit message.

quartzmo:
Here's how it's working locally:
```
irb(main):004:0> table.id
=> "nodal-almanac-725:my_dataset.baby_names"
irb(main):005:0> table.query_id
=> "[nodal-almanac-725:my_dataset.baby_names]"
```

PR #248, opened 2015-08-26
Add mechanism to block on a job until it is done

### LINKED ISSUE #228: BigQuery Job#wait_until_complete
I would like a helper method that will block on a job until the state moves to `DONE`.

```ruby
job = dataset.query_job "SELECT name, age FROM [users]"
job.wait_until_complete # continually refreshes until the job is DONE.
job.done? #=> true
```

blowmage:
The common idiom is to refresh a job as it cycles from pending to running to done.
The `Job#wait_until_done!` convenience method implements this logic.

[closed #228]


quartzmo:
It will be so great to finally have this. Thank you.

PR #251, opened 2015-08-27
Add inspect override to Connection

### LINKED ISSUE #208: Add inspect
Some objects are overly large when displayed in IRB or test output. Overriding `inspect` on the classes will make them to be easier to view.

quartzmo:
Improve user experience in IRB and elsewhere by hiding the details of Google API Client, the inspect output of which is incredibly verbose.

[closes #208]

### DIFF
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -315,6 +315,10 @@ module Gcloud
         result
       end
 
+      def inspect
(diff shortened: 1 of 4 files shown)


quartzmo:
Do these `inspect` overrides need `nodoc`?

blowmage:
Yes, `inspect` is a public method but not something we should document as part of the public API. I think we should `nodoc` it.

PR #254, opened 2015-08-27
Invoke RDoc task in clean environment

### LINKED ISSUE #215: Error building v0.1.0 docs with `rake pages:rebuild`
Calling `bundle exec rake pages:rebuild` will fail on v0.1.0 because of missing hoe dependency. It would be good if that first release played nice with bundler...

We don't want to add hoe as a dependency because it was only used on the v0.1.0 release and we'd rather not force all future developers to install it.

blowmage:
The 0.1.0 tag has some old dependencies that were not installed when
running the rake pages:tag[v0.1.0] command using bundle exec.
This was due to Bundler being too clevar and applying the parent
scope to all sub shells. To resolve this call the commands using the
Bundler.with_clean_env method.

[closes #215]

### DIFF
--- a/rakelib/gh-pages.rake
+++ b/rakelib/gh-pages.rake
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 gem "rdoc"
+require "bundler"
 require "rdoc/task"
 require "fileutils"
 require "pathname"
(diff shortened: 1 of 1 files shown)


jgeewax:
I'll leave this for @quartzmo

quartzmo:
I'll merge this after #255. Hopefully by merging #255 first, #254 will pass travis more quickly.

PR #255, opened 2015-08-27
Tighten Pub/Sub topic names

blowmage:
We see many more acceptance test failures for Pub/Sub tests.
We think this is due to how Travis-ci starts the tests.
Add a little more randomness to the topic names to avoid
accidental conflicts between the different versions of ruby
being tested concurrently.


jgeewax:
LGTM

Issue #217, opened 2015-08-13
Improve API for configuring the schema for a BigQuery Table


quartzmo:
The current JSON-as-hash mechanism for setting a Table schema is needlessly verbose and exposes the all-caps convention of the underlying Rest API. Here is the example from the docs.

```ruby
require "gcloud"

gcloud = Gcloud.new
bigquery = gcloud.bigquery
dataset = bigquery.dataset "my_dataset"

schema = {
  "fields" => [
    {
      "name" => "first_name",
      "type" => "STRING",
      "mode" => "REQUIRED"
    },
    {
      "name" => "cities_lived",
      "type" => "RECORD",
      "mode" => "REPEATED",
      "fields" => [
        {
          "name" => "place",
          "type" => "STRING",
          "mode" => "REQUIRED"
        },
        {
          "name" => "number_of_years",
          "type" => "INTEGER",
          "mode" => "REQUIRED"
        }
      ]
    }
  ]
}
table = dataset.create_table "my_table",
                             name: "My Table",
                             schema: schema
```

I believe it would improve the API to use something more structured and native to Ruby, such as a builder inside a block:

```ruby
table = dataset.create_table "my_table", name: "My Table" do |schema|
  schema.string :first_name, mode: :required
  schema.record :cities_lived, mode: :repeated do |repeated_schema|
    repeated_schema.string :place, mode: :required
    repeated_schema.integer :number_of_years, mode: :required
  end
end
```

Both Dataset#create_table and Table#schema= would need to be updated for the new approach.

jgeewax:
Likely related to https://github.com/GoogleCloudPlatform/gcloud-python/issues/1058 (Python)

PR #257, opened 2015-08-28
Add BigQuery Access Control

### LINKED ISSUE #202: BigQuery Access Control Settings
Display and update the access rules on a BigQuery table. The proposed API is:

- [x] `Dataset#access` getter
- [x] `Dataset#access=` setter
- [x] `Dataset#access` DSL with block
- [x] `Project#create_dataset` with access DSL block
- [x] Updated documentation

The `access` getter/setter will return and set the raw data as represented in the API. The access block will yield an access object that will enable easy configuration of the access rules:

```ruby
dataset.access do |acl|
  acl.add_owner_group "owners@example.com"
  acl.add_writer_user "writer@example.com"
  acl.remove_writer_user "readers@example.com"
  acl.add_reader_special :all
  acl.add_reader_view other_dataset_view_object
end # All changes made in a single API call at the end of the block
```

The same DSL can be used when creating a new dataset:

```ruby
bigquery.create_dataset "new-dataset" do |acl|
  acl.add_owner_group "owners@example.com"
  acl.add_writer_user "writer@example.com"
  acl.remove_writer_user "readers@example.com"
  acl.add_reader_special :all
  acl.add_reader_view other_dataset_view_object
end #=> New dataset object created with the access rules
```

blowmage:
Work in progress.

[closes #202]

### DIFF
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -316,7 +316,7 @@ module Gcloud
       # from strings in the formats: "my_table", "my_dataset.my_table", or
       # "my-project:my_dataset.my_table". Then merges project_id and
       # dataset_id from the default table if they are missing.
-      def table_ref_from_s str, default_table_ref
+      def self.table_ref_from_s str, default_table_ref
         str = str.to_s
         m = /\A(((?<prj>\S*):)?(?<dts>\S*)\.)?(?<tbl>\S*)\z/.match str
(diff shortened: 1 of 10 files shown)


quartzmo:
The dual mode setter/getter method `Dataset#access` is not my preference, but otherwise I love this and we'll stick with that design for now. I will borrow heavily from this PR for #217, `Table#schema`.

PR #261, opened 2015-08-31
Allow BigQuery Table/View Representation as a String

### LINKED ISSUE #258: Allow BigQuery Table/View Representation as a String
Allow tables to be represented as a string in BigQuery methods that accept table/view objects. Use the `table_id` format: `project:dataset.table` If the project_id or dataset_id values are missing they should use the current project and dataset when available.

One problem we have with copying tables in the current Ruby API is that the destination table must already exist.

```ruby
require "gcloud"

gcloud = Gcloud.new
bigquery = gcloud.bigquery
table = dataset.table "my_table"
destination_table = dataset.table "my_destination_table"

copy_job = table.copy destination_table
```

If we could represent the table as a string, we could easily allow for the destination table to be created using the `create` option:

```ruby
require "gcloud"

gcloud = Gcloud.new
bigquery = gcloud.bigquery
table = dataset.table "my_table"
destination_table = dataset.table "my_destination_table"

copy_job = table.copy "new-table-to-be-created", create: :needed
```

If we could accept a table string, we could also use that for granting access to a view using the access controls:

```ruby
require "gcloud"

gcloud = Gcloud.new
bigquery = gcloud.bigquery

dataset.access do |access|
  access.add_reader_view "external-project:other-dataset.view-to-access-this-dataset"
end
```

quartzmo:
Obtaining an object reference to a BigQuery table in a project other than the one loaded via auth credentials is not easy, so allowing string references to tables makes it easier to perform operations on tables in other projects.

[closes #258]

### DIFF
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -315,6 +315,22 @@ module Gcloud
         result
       end
 
+      ##
+      # Extracts at least +tbl+ group, and possibly +dts+ and +prj+ groups,
+      # from strings in the formats: "my_table", "my_dataset.my_table", or
+      # "my-project:my_dataset.my_table". Then merges project_id and
+      # dataset_id from the default table if they are missing.
+      def table_ref_from_s str, default_table_ref
+        str = str.to_s
+        m = /\A(((?<prj>\S*):)?(?<dts>\S*)\.)?(?<tbl>\S*)\z/.match str
+        unless m
+          fail ArgumentError, "unable to identify table from #{str.inspect}"
+        end
+        default_table_ref.merge("projectId" => m["prj"],
+                                "datasetId" => m["dts"],
+                                "tableId" => m["tbl"])
+      end
+
       def inspect #:nodoc:
         "#{self.class}(#{@project})"
       end
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -461,16 +461,8 @@ module Gcloud
         {
           "configuration" => {
             "copy" => {
-              "sourceTable" => {
-                "projectId" => source["tableReference"]["projectId"],
-                "datasetId" => source["tableReference"]["datasetId"],
-                "tableId" => source["tableReference"]["tableId"]
-              }.delete_if { |_, v| v.nil? },
-              "destinationTable" => {
-                "projectId" => target["tableReference"]["projectId"],
-                "datasetId" => target["tableReference"]["datasetId"],
-                "tableId" => target["tableReference"]["tableId"]
-              }.delete_if { |_, v| v.nil? },
+              "sourceTable" => source,
+              "destinationTable" => target,
               "createDisposition" => create_disposition(options[:create]),
               "writeDisposition" => write_disposition(options[:write])
             }.delete_if { |_, v| v.nil? },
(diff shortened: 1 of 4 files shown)


lib/gcloud/bigquery/connection.rb
```diff
@@ -462,14 +462,14 @@ def copy_table_config source, target, options = {}
           "configuration" => {
             "copy" => {
               "sourceTable" => {
```
blowmage: If we only accept the table hash here, these lines could be:

```ruby
"sourceTable" => source,
"destinationTable" =>  target,
```

quartzmo: Good eye, will do.

lib/gcloud/bigquery/table.rb
```diff
@@ -127,6 +127,19 @@ def project_id
       end
 
       ##
+      # The Project ID, Dataset ID, and Table ID as a hash.
+      #
+      # :category: Attributes
+      #
+      def to_id_hash
```
blowmage: This should be hidden from the public API. I think this name could be more descriptive. What about `to_tbl_hash` or `to_tbl_ref` instead?

lib/gcloud/bigquery/connection.rb
quartzmo: the problem is the difference between `projectId` and `project_id`. Now the hash arg could always contain camelcase keys instead, but that isn’t very rubyish, and leaks the API details out of connection into Table

lib/gcloud/bigquery/table.rb
quartzmo: I'll use `to_tbl_ref` and yes, agreed, will add :nodoc:.

lib/gcloud/bigquery/connection.rb
blowmage: Makes sense. I would almost prefer that we pass the table string instead, but converting a hash to a string, and then a string to a hash would be silly choice for performance reasons.

The table does know API details in that it knows how to use the Google Client API object. It knows the project, dataset, and table are found in `@gapi["tableReference"]`, for example.

quartzmo: Hmmm, good point. This method could be seen as "to gapi tableReference". Should I change it to camel case keys?

lib/gcloud/bigquery/connection.rb
```diff
@@ -315,6 +315,22 @@ def load_resumable table, file, chunk_size = nil, options = {}
         result
       end
 
+      ##
+      # Extracts at least +tbl+ group, and possibly +dts+ and +prj+ groups,
+      # from strings in the formats: "my_table", "my_dataset.my_table", or
+      # "my-project:my_dataset.my_table". Then merges project_id and
+      # dataset_id from the default table if they are missing.
+      def table_ref_from_s str, default_table_ref
+        str = str.to_s
```
blowmage: Nice. I was looking for this. :+1:

lib/gcloud/bigquery/connection.rb
```diff
@@ -315,6 +315,22 @@ def load_resumable table, file, chunk_size = nil, options = {}
...
+        m = /\A(((?<prj>\S*):)?(?<dts>\S*)\.)?(?<tbl>\S*)\z/.match str
+        unless m
+          fail ArgumentError, "unable to identify table from #{str.inspect}"
+        end
+        default_table_ref.merge("projectId" => m["prj"],
```
blowmage: I think we need to add this line here:

```ruby
default_table_ref = default_table_ref.to_hash if default_table_ref.respond_to? :to_hash
```

blowmage:
:metal:

PR #262, opened 2015-09-01
Fix hash syntax in Table example

quartzmo:
The invalid, JSON-style syntax can prevent proper Ruby syntax highlighting in some generated documentation.

### DIFF
--- a/lib/gcloud/bigquery/table.rb
+++ b/lib/gcloud/bigquery/table.rb
@@ -70,12 +70,12 @@ module Gcloud
     #     "first_name" => "Alice",
     #     "cities_lived" => [
     #       {
-    #         "place": "Seattle",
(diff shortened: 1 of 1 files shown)


blowmage:
Not technically invalid syntax. It works just fine in ruby 2.2. :trollface:

quartzmo:
True, but with surprising results (at least for me). To be avoided!

```
irb(main):001:0> { "a": 1 }
=> {:a=>1}
```

blowmage:
Totally agree. Converting the string to a symbol is a bit surprising and worth avoiding. Thanks for fixing this.

quartzmo:
It only got in because I copied/pasted some JSON from the official docs...

blowmage:
I've been bitten by that more times than I can count.

PR #270, opened 2015-09-04
Check Google Compute Engine for project_id

### LINKED ISSUE #211: Support getting project id from GCE
In the same way we retrieve the access token from the GCE environment, we should also pull the project info from GCE.

The info should be available at http://metadata.google.internal/computeMetadata/v1/project/project-id. Here is how [googleauth](https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth/compute_engine.rb) accesses the access token.

### LINKED ISSUE #221: Uploading a file to a bucket
I successfully uploaded a file in my bucket with the help of the following script:

require "gcloud"

gcloud = Gcloud.new "g-storage-103621",
                    'C:\Users\Kanishka\GoogleCloudStorage\keyfile\google-storage-756c0950e8bc.json'
storage = gcloud.storage

bucket = storage.bucket "g-storage-bucket"

bucket.create_file "C:/Users/Kanishka/GoogleCloudStorage/dataset/language_id.csv"

When I checked my bucket, I found that the bucket had a path called "C:/Users/Kanishka/GoogleCloudStorage/dataset/language_id.csv".

Instead of directly uploading the file to the bucket, it made directories similar to my computer and uploaded it there. 

How do I upload the file so that it doesn't create directories such has mentioned above and directly uploads it in my bucket like: 'BUCKET_NAME' -> 'FILE_NAME'?

blowmage:
The googleauth library detects when running on Google Compute Engine and
retrieves the authentication from the GCE environment.
Likewise, we should look to GCE to find the project_id.

[closes #211]

### DIFF
--- a/lib/gcloud/bigquery/project.rb
+++ b/lib/gcloud/bigquery/project.rb
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require "gcloud/gce"
 require "gcloud/bigquery/connection"
 require "gcloud/bigquery/credentials"
 require "gcloud/bigquery/errors"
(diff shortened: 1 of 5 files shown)


quartzmo:
Looks good! Interesting how this works. Will merge soon.

PR #271, opened 2015-09-08
Change Table#schema option default

quartzmo:
It is more likely to add to a schema than to replace it, so default for the replace option should be false. Also, change acceptance test to use new schema builder.


blowmage:
This looks good to me. Will merge when the build passes.

Issue #293, opened 2015-09-09
DNS Zone add Records


blowmage:
A zone object must allow for new records to be added to the zone.

PR #311, opened 2015-09-15
Add Zone#import

### LINKED ISSUE #300: DNS import records from a file

quartzmo:
Adds a dependency on boesemar/zonefile.

[closes #300]


quartzmo:
This is NOT ready to be merged. I am working on replacing the zonefile lib dependency with a small amount of internal code.

quartzmo:
I just noticed that the `SOA` line is not added by the zonefile lib as a resource record, which is the way that Cloud DNS API expects it. Fixing.

quartzmo:
This PR is now waiting on #293, in order to resolve this [TODO to add the imported records](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/311/files#diff-5129440d3f3f29bfaa3fbb10b634f787R377).

quartzmo:
@blowmage Working on both of these good suggestions now.

quartzmo:
@blowmage Please review the current state of the PR, which I intended as merge-ready, but note that exploratory testing just revealed a couple of new issues:

1. The Record `name` value has to be fully-qualified, which isn't the norm for zone files. I'm hoping to automate the substitution.
  ```
  irb(main):007:0> change = zone.import "acceptance/data/db.example.com"
Gcloud::Dns::ApiError: Invalid value for 'entity.change.additions[1].name': '@'
  ```

1. A new Zone is given an NS record in addition to a SOA record. I think I noticed this before but didn't make the connection. It probably makes sense for the Importer to also exclude NS records by default.
  ```
  irb(main):018:0> change = zone.import "acceptance/data/db.example.com"
Gcloud::Dns::ApiError: The resource 'entity.change.additions[6]' named 'example.net. (NS)' already exists
  ```

blowmage:
Yeah, the name needs to be fully qualified. The import process should use `zone.dns` for the base DNS name. However, it does make me wonder if we could create a new zone during the import process, using `$ORIGIN` for the dns name?

```ruby
zone = dns.create_zone "example-zone", "example.net."
zone.import "/dns/example.net.zone"
```

vs.

```ruby
zone = dns.import "/dns/example.net.zone"
```

Just a thought, definitely not 100% sold on that approach.

I think it makes sense to treat NS records like SOA records and not create them during import. But to also include an option to import from the zonefile.

_Sorry the response is late, it was somehow didn't get posted over the weekend..._

quartzmo:
I'm going to open a separate issue regarding the fully-qualified addresses requirement from the service. For now, I have added [this passage](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/311/files#diff-5129440d3f3f29bfaa3fbb10b634f787R361) to the doc for `Zone#import`.

This PR is ready for review, and pending approval, to be merged.

PR #313, opened 2015-09-17
Add DNS Record::List#all

blowmage:
Helper method to retrieve all pages of DNS records. This may be useful. Feedback wanted.

### DIFF
--- a/lib/gcloud/dns/record/list.rb
+++ b/lib/gcloud/dns/record/list.rb
@@ -45,8 +45,18 @@ module Gcloud
         end
 
         ##
-        # Retrieve all records and add to the current list.
-        # This changes the contained records.
+        # Retrieves all records by repeatedly loading pages until #next? returns
+        # false. Returns the list instance for method chaining.
+        #
+        # === Example
+        #
+        #   require "gcloud"
+        #
+        #   gcloud = Gcloud.new
+        #   dns = gcloud.dns
+        #   zone = dns.zone "example-zone"
+        #   records = zone.records.all # Load all pages of records
+        #
         def all
           while next?
             next_records = self.next
--- a/lib/gcloud/dns/record/list.rb
+++ b/lib/gcloud/dns/record/list.rb
@@ -45,6 +45,18 @@ module Gcloud
         end
 
         ##
+        # Retrieve all records and add to the current list.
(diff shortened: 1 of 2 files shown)


quartzmo:
Looks handy. I'd merge it except for the 'don't merge' label.

lib/gcloud/dns/record/list.rb
```diff
@@ -45,6 +45,18 @@ def next
         end
 
         ##
+        # Retrieve all records and add to the current list.
+        # This changes the contained records.
```
quartzmo: I think "This changes the contained records." sounds a little scary, and I don't think it's really accurate either. The contained records are unchanged, while any un-fetched records are added to them.

blowmage: Agreed. I'm struggling on how to best explain it. But at this point I don't know if this is the public API or the private API...

blowmage: @quartzmo: Thoughts on a better documentation for this method? I'm stumped on what to say.

test/gcloud/dns/zone_test.rb
```diff
@@ -282,6 +282,33 @@ def find_change_json change_id
     second_records.next?.must_equal false
   end
 
+  it "paginates all records with all" do
```
quartzmo: `it "loads all records with all"`

lib/gcloud/dns/record/list.rb
quartzmo: Maybe something like:
```ruby
##
# Retrieves all records by repeatedly loading pages until #next? returns false. 
# Returns the list instance for method chaining.
#
# Example
#
#  records.all.each do |record|
#    ...
```

test/gcloud/dns/zone_test.rb
blowmage: Updated!

lib/gcloud/dns/record/list.rb
blowmage: You make it look so easy...

quartzmo:
Looks good, ready to merge?

blowmage:
Yeah, should be good to go now.

PR #318, opened 2015-09-21
Add Change DSL

blowmage:
Add mechanism for adding, removing, and replacing multiple records in the same API call.

```ruby
require "gcloud"

gcloud = Gcloud.new
dns = gcloud.dns
zone = dns.zone "example-zone"
change = zone.update do |tx|
  tx.add     "example.com.",     "A",  86400, "1.2.3.4"
  tx.add     "www.example.com.", "A",  86400, "1.2.3.4"
  tx.remove  "example.com.",     "TXT"
  tx.replace "example.com.",     "MX", 86400, ["10 mail1.example.com.",
                                               "20 mail2.example.com."]
end
```

### DIFF
--- a/lib/gcloud/dns/zone.rb
+++ b/lib/gcloud/dns/zone.rb
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 require "gcloud/dns/change"
+require "gcloud/dns/zone/transaction"
 require "gcloud/dns/zone/list"
 require "gcloud/dns/record"
 require "time"
@@ -347,7 +348,24 @@ module Gcloud
       #
       # Gcloud::Dns::Change
       #
-      # === Example
+      # === Examples
+      #
+      # The recommended way to make changes is call +update+ with a block. See
+      # Zone::Transaction.
+      #
+      #   require "gcloud"
+      #
+      #   gcloud = Gcloud.new
+      #   dns = gcloud.dns
+      #   zone = dns.zone "example-zone"
+      #   zone.update do |tx|
+      #     tx.add     "example.com.", "A",  86400, "1.2.3.4"
+      #     tx.remove  "example.com.", "TXT"
+      #     tx.replace "example.com.", "MX", 86400, ["10 mail1.example.com.",
+      #                                              "20 mail2.example.com."]
+      #   end
+      #
+      # Or you can provide the record objects to add and remove.
       #
       #   require "gcloud"
       #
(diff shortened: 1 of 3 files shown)


lib/gcloud/dns/zone.rb
```diff
@@ -355,24 +373,25 @@ def record name, type, ttl, data
...
+          deletions += updater.deletions
         end
+
+        create_change additions, deletions
       end
 
       ##
-      # Adds records to the Zone. In order to update existing records, or add
+      # Adds a record to the Zone. In order to update existing records, or add
       # and delete records in the same transaction, use #update.
```
quartzmo: Could this still support adding multiple records? What about adding Record instance(s) as well as passing just the property values? (I'm sure there's a reason you changed it, just missed the story somehow..)

quartzmo: Sorry just realized this is just a doc typo fix, not something changed in this PR. Off topic...

lib/gcloud/dns/zone/transaction.rb
```diff
@@ -0,0 +1,142 @@
...
+        #   dns = gcloud.dns
+        #   zone = dns.zone "example-zone"
+        #   zone.update do |tx|
+        #     tx.add     "example.com.", "A",  86400, "1.2.3.4"
+        #     tx.remove  "example.com.", "TXT"
+        #     tx.replace "example.com.", "MX", 86400, ["10 mail1.example.com.",
+        #                                              "20 mail2.example.com."]
+        #   end
+        #
+        def replace name, type, ttl, data
```
quartzmo: What about passing Record instance(s) as well as passing just the property values? For me, the preferred way to change an existing record would be to retrieve it, modify the value I want to change, and then pass it to this method.

```ruby
mx_record = zone.records(name: "example.com.", type: "MX").all.first
mx_record.ttl = 3600 # change the TTL
zone.update do |tx|
  # do some other adds, replaces, and removes...
   tx.replace mx_record
 end
``` 

Using the Record instance is a convenient way to "pass back" the values I don't want to change. And it makes sense that I would have the instance, otherwise how could I know what the current values are?

lib/gcloud/dns/zone.rb
blowmage: Yeah, just clarification in the docs. This will only add a single record object, but the single record object can include multiple data values. It depends on how we count records I guess.

lib/gcloud/dns/zone/transaction.rb
blowmage: Yeah, we had a discussion about modify vs. replace. I have in my notes that you were going to add a separate issue for a `modify` method that would retrieve the existing record(s) and allow you to make changes that would then be persisted.

```ruby
zone.update |tx|
  tx.modify "example.com", "MX" do |mx|
    mx.ttl = mx.ttl*2 # double the TTL
    mx.data.map! { |r| r.upcase } # upper case the data
  end
end
```

quartzmo: Ah, thanks for refreshing my memory! Do you still like that idea? Or is it maybe a simpler, better approach (since this "read for update" is not really atomic anyway) to just let the user retrieve the record outside of `update`, and have `modify` accept the changed record?

blowmage: I think its a very good idea, just a slightly different use case than replace. With `replace` we don't have to care what was there, just specify what it needs to be. But if you just want to update TTL it would be simpler to use `modify`.

Do you want me to open the issue?

quartzmo: Yes please!

PR #321, opened 2015-09-22
Add DNS Modify

### LINKED ISSUE #319: DNS Modify records
We have the ability to replace records (if they exist), but we also want to be able to modify an existing record. For instance, we want to be able to update the `ttl` without caring what `data`. Currently, we need to do something like this:

```ruby
existing_mx_record = zone.records(name: "example.com.", type: "MX").all.first
updated_mx_record = existing_mx_record.dup
updated_mx_record.ttl = 3600 # change the TTL
zone.update updated_mx_record, existing_mx_record
``` 

We would prefer it be this instead:

```ruby
zone.modify "example.com.", "MX" do |mx|
 mx.ttl = 3600 # change the TTL
end
```

blowmage:
Enables the following syntax to update existing records:

```ruby
zone.modify "example.com.", "MX" do |mx|
 mx.ttl = 3600 # change the TTL
end
```

[closes #319]

### DIFF
--- a/lib/gcloud/dns/record.rb
+++ b/lib/gcloud/dns/record.rb
@@ -111,19 +111,20 @@ module Gcloud
         [name, type, ttl, data].hash
       end
 
-      def eql?(other) #:nodoc:
+      def eql? other #:nodoc:
         return false unless other.is_a? self.class
         name == other.name && type == other.type &&
           ttl == other.ttl && data == other.data
       end
 
-      def ==(other) #:nodoc:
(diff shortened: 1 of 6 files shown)


blowmage:
@quartzmo: Don't accept this just yet. I didn't see the update DSL PR was accepted, and I want to add a modify method to the DSL as well.

blowmage:
Okay, that change has been made. Thoughts?

PR #324, opened 2015-09-24
Fully Qualified Domain Names

blowmage:
This change allows you to use "@" and "www" names when creating records. So this:

```ruby
zone.record "@", "A", 18600, "1.2.3.4"
```

Is equivalent to this:

```ruby
zone.record "example.net.", "A", 18600, "1.2.3.4"
```

And this:

```ruby
zone.record "www", "A", 18600, "1.2.3.4"
```

Is equivalent to this:

```ruby
zone.record "www.example.net.", "A", 18600, "1.2.3.4"
```

### DIFF
--- a/lib/gcloud/dns/importer.rb
+++ b/lib/gcloud/dns/importer.rb
@@ -46,7 +46,8 @@ module Gcloud
       #   The path to a zone file on the filesystem, or an IO instance from
       #   which zone file data can be read. (+String+ or +IO+)
       #
-      def initialize path_or_io
+      def initialize zone, path_or_io
+        @zone = zone
         @grouped_zf_records = {}
         @records = []
         @zonefile = if path_or_io.respond_to? :read
@@ -114,7 +115,7 @@ module Gcloud
           data = zf_records.map do |zf_record|
             data_from_zonefile_record(key[1], zf_record)
           end
-          Gcloud::Dns::Record.new key[0], key[1], ttl, data
+          @zone.record key[0], key[1], ttl, data
         end
       end
 
@@ -122,7 +123,7 @@ module Gcloud
         zf_soa = @zonefile.soa
         ttl = ttl_to_i(zf_soa[:ttl]) || ttl_to_i(@zonefile.ttl)
         data = data_from_zonefile_record :soa, zf_soa
-        Gcloud::Dns::Record.new zf_soa[:origin], "SOA", ttl, data
+        @zone.record zf_soa[:origin], "SOA", ttl, data
       end
 
       ##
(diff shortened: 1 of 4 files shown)


blowmage:
@quartzmo What are your thoughts about this?

quartzmo:
Sometime today I'll check the Cloud DNS web UI and the gcloud cli tool to see if they offer anything like this. A comment in [this issue](http://stackoverflow.com/questions/29318141/website-not-working-with-www) makes me think maybe so: 

> make a new RR set of type A at 'www', and... create a CNAME at www that...

quartzmo:
The gcloud command-line tool makes no attempt to validate or qualify names:

```sh
Christophers-MacBook-Air:gcloud-ruby quartzmo$ gcloud dns record-sets transaction start --zone example-zone
Transaction started [transaction.yaml].
Christophers-MacBook-Air:gcloud-ruby quartzmo$ gcloud dns record-sets transaction add  --name @ --ttl 1234 --type A "1.2.3.4" --zone example-zone
Record addition appended to transaction at [transaction.yaml].
Christophers-MacBook-Air:gcloud-ruby quartzmo$ gcloud dns record-sets transaction add  --name www --ttl 1234 --type A "1.2.3.4" --zone example-zone
Record addition appended to transaction at [transaction.yaml].
Christophers-MacBook-Air:gcloud-ruby quartzmo$ gcloud dns record-sets transaction execute --zone example-zone
ERROR: (gcloud.dns.record-sets.transaction.execute) 
Invalid value for 'entity.change.additions[1].name': '@.' (code: 400)
Christophers-MacBook-Air:gcloud-ruby quartzmo$ gcloud dns record-sets transaction execute --zone example-zone 
ERROR: (gcloud.dns.record-sets.transaction.execute) 
Invalid value for 'entity.change.additions[1].name': 'www.' (code: 400)
```

quartzmo:
The developer console makes it easy to use fragments when adding records. It prohibits you from using names with dots in them! See screenshot below, in which I added `@` and `www`.

dns_add_record_dev_console

blowmage:
That's interesting. It is making an attempt at a FQDM by adding ".", because I suspect most folks say "example.com" and not "example.com.".

I think the proper values need to be highlighted in the gcloud-ruby documentation. We can do better in that area.

blowmage:
The developer console seems to reinforce the thought that it only allows one subdomin level in a zone. You can't have "relay01.mail.example.com." in a zone for "example.com.", for instance. Does that change our thinking about what values should be supported?

blowmage:
Only allowing one subdomain level in a zone also has implications for import. It seems reasonable that a zonefile can have more than one subdomain level, or more than one domain.

blowmage:
Perhaps a better algorithm is to add he zone's dns name if the name given for the record doesn't contain a "."... Hmmm...

quartzmo:
Here's [some background on nested subdomains](http://www.zytrax.com/books/dns/ch9/delegate.html). I think in Cloud DNS it will require a new Zone for the subdomain, but I haven't tried it.

PR #328, opened 2015-09-25
Add forceful deletion of Zone

blowmage:
A zone cannot be deleted while there are non-essential records on it. This option will forcefully delete a zone by removing those records. Also includes the clear! method to remove all non-essential records.

### DIFF
--- a/lib/gcloud/dns/zone.rb
+++ b/lib/gcloud/dns/zone.rb
@@ -116,11 +116,20 @@ module Gcloud
       ##
       # Permanently deletes the zone.
       #
+      # === Parameters
+      #
+      # +options+::
+      #   An optional Hash for controlling additional behavior. (+Hash+)
+      # <code>options[:force]</code>::
+      #   If +true+, forces the deletion of the zone by delete all records. If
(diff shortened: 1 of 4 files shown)


lib/gcloud/dns/zone.rb
```diff
@@ -116,11 +116,20 @@ def created_at
       ##
       # Permanently deletes the zone.
       #
+      # === Parameters
+      #
+      # +options+::
+      #   An optional Hash for controlling additional behavior. (+Hash+)
+      # <code>options[:force]</code>::
+      #   If +true+, forces the deletion of the zone by delete all records. If
```
quartzmo: "*ensures* the deletion of the zone by *deleting* all"

lib/gcloud/dns/zone.rb
```diff
@@ -140,6 +160,25 @@ def delete
       end
 
       ##
+      # Removes non-essential records the zone. Only NS and SOA records will be
```
quartzmo: "records the zone" -> "records *from* the zone"

quartzmo:
Except for the typos noted above, LGTM, feel free to merge!

PR #332, opened 2015-09-28
Fix bug with string table names

### LINKED ISSUE #326: BigQuery: Table copy is not working if destination table does not exist
Samples from the docs: http://googlecloudplatform.github.io/gcloud-ruby/docs/master/Gcloud/Bigquery/Table.html#method-i-copy

```ruby
      ds = bigquery.dataset 'logging'
      # destination_table = ds.table "logs_week"

      5.times do |i|
        src = "logs2015092#{i}"
        table = ds.table src
        table.copy('logging.logs_week_test', create: :needed, write: :append)
      end
```
/Users/lz/.rvm/gems/ruby-2.2.2@bq_uploader/gems/gcloud-0.3.0/lib/gcloud/bigquery/table.rb:428:in `copy': undefined method `gapi' for "logging.logs_week_test":String (NoMethodError)

If I uncomment line 2 and use that variable I get 
/Users/lz/.rvm/gems/ruby-2.2.2@bq_uploader/gems/gcloud-0.3.0/lib/gcloud/bigquery/table.rb:428:in `copy': undefined method `gapi' for nil:NilClass (NoMethodError)

blowmage:
The intent is that the current project and dataset values are used when copying
to a table identified by a string if the string does not contain them.
The bug here is that the nil values were kept when merged.
Removing the keys with nil values before the merge corrects the behavior.

[fixes #326]

### DIFF
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -332,9 +332,10 @@ module Gcloud
         unless m
           fail ArgumentError, "unable to identify table from #{str.inspect}"
         end
-        default_table_ref.merge("projectId" => m["prj"],
-                                "datasetId" => m["dts"],
-                                "tableId" => m["tbl"])
+        str_table_ref = { "projectId" => m["prj"],
+                          "datasetId" => m["dts"],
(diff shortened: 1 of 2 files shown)


quartzmo:
Fast work @blowmage!

PR #334, opened 2015-09-28
Add DNS aliases

blowmage:


lib/gcloud/dns/zone.rb
```diff
@@ -360,6 +363,7 @@ def records options = {}
           fail ApiError.from_response(resp)
         end
       end
+      alias_method :find_record, :records
```
quartzmo: Rather than or in addition to this alias, I would love to have a `find_record_by name, type` convenience method that returns the first record found (or nil) rather than the array:

```ruby
record = zone.find_record_by "www", "A"
record.name #=> "www.example.com."
```

blowmage: That will return multiple records. Its not likely IMO, but possible with how the DNS API is set up.

quartzmo: I know. I think we should have a finder that returns a single record.

blowmage: @quartzmo I've addressed this in a [separate PR](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/335).

Issue #370, opened 2015-10-08
Suport for fieldDelimiter in the Bigquery table.load_job


gramos74:
Right now there is no a option to set a delimiter for the file to be uploaded, so only csv files can be uploaded.

Our processes generate tsv files ( tab delimiter ) and would be great to upload those files directly because the tab delimiter is already supported by Bigquery. [ https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load ]

Could be implemented this option for the next release ? 

I could try to do it myself and create a pull request...

quartzmo:
@gramos74 You are right, this option is missing from Table#load. As far as I can tell, it can and should be added.

Adding it should be fairly simple if you are interested in contributing the change, and I would be happy to help you do it today if you have time. Have you read our [CONTRIBUTING](CONTRIBUTING.md) guide?

gramos74:
Hi
Yes, Will be great for me do it. 

Yesterday I changed the Connection file locally and I tested the upload and worked fine. The thing is that I had problems running the tests.
Let me try again and I will reach you later.

quartzmo:
Great! You read about the [contributor license agreement](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/CONTRIBUTING.md#contributor-license-agreements), right?

If you need to open a separate issue to get help with running tests, go ahead.

gramos74:
Hi,

Pull request created.

quartzmo:
Fixed by PR #371, thanks @gramos74.

PR #372, opened 2015-10-08
Missing BigQuery Options

blowmage:
The BigQuery load and extract methods are missing some options that are supported on the API.

Missing load options:

* allowJaggedRows
* allowQuotedNewlines
* encoding
* fieldDelimiter
* ignoreUnknownValues
* maxBadRecords
* quote
* schema
* skipLeadingRows

Missing extract options:

* compression
* fieldDelimiter
* printHeader

See also #370.


blowmage:
@quartzmo This PR has all of the options identified except for `schema` on `Table#load`. It looks like this is there to load a file and create a table in one operation. Gcloud doesn't have a way to create a table and load a file in one operation, so I wonder if `schema` is needed. Thoughts?

quartzmo:
@blowmage The optional parameter [configuration.load.schema](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load.schema) is definitely not needed in this context, as the destination table is the resource represented here and already exists. At least that's how I read it:

> [Optional] The schema for the destination table. The schema can be omitted if the destination table already exists, or if you're loading data from Google Cloud Datastore.

quartzmo:
Although, best would be to see if sending a schema allows a table that previously had no schema to be updated.

blowmage:
Yeah, I can also see an argument for something like `Dataset#load` or `Project#load` that can be used to create and load in one API call.

blowmage:
That said, I'm pretty comfortable with not supporting the option for the time being. Since we require two API calls the schema can be specified on table creation before the load.

quartzmo:
Is this PR ready to merge then?

blowmage:
Yeah.

quartzmo:
Actually I just noticed something that should have been done in an earlier PR, but would be natural to fix here. Do you mind doc'ing the default value for `delimiter`?

blowmage:
Yeah, no problem.

PR #374, opened 2015-10-08
Add pages feature task

blowmage:
This task is intended to be used from a feature branch. It will update the docs in the gh-pages branch from the feature branch. It can optionally push the change if the push argument is specified. The intended use is to push the updated gh-pages branch to an individual's github repo, so the docs can be seen without accepting the feature branch to master.

To automatically push to a repo, specify the remote when calling the task:

    $ rake pages:feature[blowmage]


quartzmo:
This will be really helpful! Thanks a lot for doing it!

Issue #375, opened 2015-10-08
Manage Bucket CORS with block-based DSL


quartzmo:
Support setting and updating a bucket's CORS settings with a block-based DSL. This issue was extracted from #273.

- [ ] Allow setting CORS using the DSL during bucket creation: (`add_rule` method signature is just a suggestion)

```ruby
storage.bucket_create "new_todo_bucket" do |cors|
  cors.add_rule 300, [ "http://example.org",
                       "https://example.org" ],
                [ "*" ],
                [ "Content-Type" ]
end
```

- [x] Allow the CORS settings on an existing bucket to be set using a DSL: (`add_rule` method signature is just a suggestion)

```ruby
bucket.cors do |cors|
  cors.add_rule 300, [ "http://example.org",
                       "https://example.org" ],
                [ "*" ],
                [ "Content-Type" ]
end
```

- [x] In the update method described in #279, I think it would feel natural to also support a nested block `cors` update, like so:

```ruby
bucket.update do |b|
  b.website_main = "index.html"
  b.website_404 = "not_found.html"
  b.cors do |cors|
     # possible method signature, not definite yet
     cors.add_rule 300, [ "http://example.org", "https://example.org" ], [ "*" ], [ "Content-Type" ] 
  end
end
```

quartzmo:
After reading [Troubleshooting CORS-Related-Problems](https://cloud.google.com/storage/docs/cross-origin#Troubleshooting%20CORS-Related-Problems), I think the best signature for the proposed `add_rule` method would *require* `origin` and `method` arguments, but make the `response_headers` and `max_age` arguments optional, since they are not needed for simple use cases. These options will default to none and `1800` seconds, respectively. This default for `max_age` is based on the following statement, from the above document; but the value will be sent in the request, just in case.

> The default setting for <MaxAgeSec> is 1800 seconds (30 minutes).

quartzmo:
Thus, update examples:

Rule for requests that do *not* require pre-flight.

```ruby
bucket.cors do |cors|
  cors.add_rule [ "http://example.org", "https://example.org" ], [ "GET", "POST" ]
end
```

Strings as well as arrays should be supported.

```ruby
bucket.cors do |cors|
  cors.add_rule "http://example.org", "GET"
end
```

Rule for requests that require pre-flight.

```ruby
bucket.cors do |cors|
  cors.add_rule "http://example.org", 
                "*", 
[...]

PR #376, opened 2015-10-09
Add block-based Bucket#update and Bucket#cors

### LINKED ISSUE #279: Storage Bucket Update
Allow multiple attributes to be changed in a single API request.

```ruby
bucket.update do |b|
  b.website_main = "index.html"
  b.website_404 = "not_found.html"
end
```

quartzmo:
[closes #279, #375]

### DIFF
--- a/lib/gcloud/storage/bucket.rb
+++ b/lib/gcloud/storage/bucket.rb
@@ -213,6 +213,33 @@ module Gcloud
       end
 
       ##
+      # Updates the bucket with changes made in the given block in a single
+      # PATCH request. Nested objects accessible in the block, such as the
+      # values in the CORS configuration, are also mutable and will be included
+      # in the request.
+      #
+      # === Example
+      #
+      #   require "gcloud"
+      #
+      #   gcloud = Gcloud.new
+      #   storage = gcloud.storage
+      #
+      #   bucket = storage.bucket "my-bucket"
+      #   bucket.update do |b|
+      #     b.website_main = "index.html"
+      #     b.website_404 = "not_found.html"
+      #     b.cors[0]["method"] = ["PUT"]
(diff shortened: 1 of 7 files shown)


quartzmo:

blowmage:
How are CORS rules removed or updated?

quartzmo:
I thought about that, of course, but couldn't come up with anything. As far as I know, the rules are completely non-unique. If you want to change or remove one, it's a matter of "you'll know it when you see it". So you need to inspect all attributes of each record to be sure of what you're doing. I think the best approach is to work with the mutable "raw" (array of hashes) representation within a `Bucket#update` block. But feel free to suggest a concrete design if you have one, I'd be more than happy to add it.

blowmage:
I understand that finding a CORS rule to replace or remove is difficult, because there is no unique identifiers. BigQuery's schema is also difficult to modify because, by definition, the schema can only be added to, but we still have the option in the schema DSL to replace the schema by a new schema defined in the DSL. But CORS doesn't even have that.

I think it would be good to have a mutable array of hashes in the DSL, that you can use to find and possibly remove a rule, as well as the `add_rule` helper method.

```ruby
bucket.cors do |cors|
  # Remove the first CORS hash in the array
  cors.shift
  # Remove all existing rules that include example.org
  cors.delete_if { |cors_hash| cors_hash["origin"].includes? "http://example.org" }
  # Add new rule
  cors.add_rule 300, [ "http://example.org", "https://example.org" ],
                [ "*" ],
                [ "Content-Type" ]
end
```

quartzmo:
@blowmage The array is a great idea, I'll do it. Update coming soon. Thanks!

blowmage:
Love it! The added functionality and tests is terrific!

Issue #280, opened 2015-09-09
Storage File Update


blowmage:
Allow multiple attributes to be changed in a single API request.

```ruby
file.update do |f|
  f.content_type = "text/ext"
  f.content_encoding = "gzip"
end
```

Issue #278, opened 2015-09-09
Manage File Metadata


blowmage:
Support management of a file's [metadata](https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata#custom-metadata). _This issue is rudimentary, needs details to be filled in as we learn more._

- [x] Allow setting metadata during file creation:

```ruby
bucket.create_file "hello.txt", metadata: { title: "Hello world!" }
```

- [x] Allow getting values using the `metadata` method.

```ruby
file.metadata["title"] #=> "Hello world!"
```

- [x] Allow setting a new value using a setter method.

```ruby
file.metadata = { title: "Goodbye cruel world!" }
```

Calling this method makes an API call to update the value. A caveat is that this won't update the API:

```ruby
file.metadata["title"] = "Goodbye cruel world!" # updates local hash, does not make an API call.
```

Calling this method inside an update block should update the API:

```ruby
file.update do |f|
  f.metadata["title"] = "Goodbye cruel world!"
end
```

quartzmo:
@blowmage For returned hashes like this (everywhere in gcloud eventually, Bucket#cors for example), I am in favor of [freezing](http://ruby-doc.org/core-2.2.3/Object.html#method-i-freeze) the return value so that modification is not possible. There is no sense in taking the chance that a user is misled into thinking a remote update was made.

> Calling this method makes an API call to update the value. A caveat is that this won't update the API:

What's your feeling on this?

blowmage:
I agree with freezing.

quartzmo:
Thanks @blowmage, will do.

quartzmo:
Modifying the hash without sending a request to the service won't be possible. The returned hash will be frozen.

PR #378, opened 2015-10-15
Storage File attribute management

### LINKED ISSUE #92: Manage File Attributes
Support management (create, read, update) of the following:

- [x] contentType
- [x] cacheControl
- [x] contentDisposition
- [x] contentEncoding
- [x] contentLanguage

In addition, support create and read only of the following:

- [x] md5Hash
- [x] crc32c

Allow providing values during file creation:

```ruby
bucket.create_file "hello.txt", cache_control: "private, max-age=0, no-cache"
```
Allow getting values using a well named method.

```ruby
file.content_type #=> "text/ext"
```

Allow setting new values using a setter method that makes an API call.

```ruby
file.content_type = "text/ext"
```

Making multiple changes in a single API call will be addressed in a future issue to add the update method which can make multiple at the same time:

```ruby
file.update do |f|
  f.content_type = "text/ext"
  f.content_encoding = "gzip"
  f.metadata["title"] = "Goodbye cruel world!"
end
```

quartzmo:
[closes #92, #278, #280]


lib/gcloud/storage/connection.rb
```diff
@@ -285,7 +285,7 @@ def download_file bucket_name, file_path
 
       ##
       # Updates a file's metadata.
-      def patch_file bucket_name, file_path, options = {}
+      def patch_file bucket_name, file_path, options = {}, body_options = {}
```
blowmage: I would prefer if `options` and `body_options` were the same hash. Having two makes it more difficult to use.

```ruby
def patch_file bucket_name, file_path, options = {}
```

quartzmo: Done, good catch. I don't know what I was thinking, adding that extra param.

quartzmo:

PR #379, opened 2015-10-16
Add checks on multipart encoded data

blowmage:
Multipart encoded data is a pain to work with, which is why the tests didn't have
any assertions for them. Add a simple parse method to structure the multipart
encodede data into something that can be asserted on.


blowmage:
@quartzmo Here is what I came up with for asserting the multipart encoded data. This approach can be applied to the other tests if it you. Thoughts?

quartzmo:
Thank you so much for solving this!

PR #383, opened 2015-10-20
Improve test runtime

blowmage:
The Google API Client makes a discovery request ever time a service is instantiated.
Store the service objects into global variables to avoid the network request.
The result is fast tests.


quartzmo:
This will sure be nice. Unit tests running the way they were always meant to run!

PR #387, opened 2015-10-20
Acceptance Retry

blowmage:
The acceptance tests work against the production services. Sometimes the
services present their eventually consistent nature and return results that
are not yet correct. Try rerunning failed tests instead of rerunning the
entire test suite.


quartzmo:
Worth a try! Even a retry, haha.

Issue #360, opened 2015-10-01
Final review of Resource Manager API by Google Cloud team


blowmage:
We need someone from the Google Cloud Resource Manager team to perform a review of the Resource Manager implementation in gcloud-ruby.

* [Code](https://github.com/GoogleCloudPlatform/gcloud-ruby/tree/resource-manager)
* [Documentation](http://blowmage.com/gcloud-ruby/docs/master/Gcloud/ResourceManager.html)

blowmage:
@jgeewax Do you know anyone from the Resource Manager team who can review the ruby implementation?

blowmage:
@jgeewax My account is whitelisted and I've been able to verify the implementation behaves correctly. Is there someone to look at the docs and implementation and give us feedback?

blowmage:
@jgeewax Can we plan on releasing next week unless we hear from someone on the Google Cloud Resource Manager team? Say the 28th?

rae42:
two comments:
- get and list are now available to all users
- we have also added get and setiampolicy methods (only accessible to whitelisted customers as create and delete)

charvesf:
Do you need us to review the code as well? I've reached out to my team to see if there are any Ruby enthusiasts that can take a look.

quartzmo:
@charvesf That would be great! But in case no one responds, even someone unfamiliar with Ruby can review our [`connection.rb`](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/resource-manager/lib/gcloud/resource_manager/connection.rb) file to see if we've misunderstood something, or missed an opportunity to add something useful.

@rae42 We will update the docs for the general availability of `get` and `list`. Will we see the get and setiampolicy methods soon in [the REST API docs](https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects)?

quartzmo:
@rae42 Looking at [APIs Explorer](https://developers.google.com/apis-explorer/#p/cloudresourcemanager/v1beta1/), I also noticed the new (?) Organization resource.

blowmage:
@rae42/@charvesf: Thank you for helping us. A few of other quick questions...

[...]

PR #394, opened 2015-10-22
Resource manager overview, readme, and guide docs

quartzmo:
Includes clarification of API access.

[refs #360]

### DIFF
--- a/lib/gcloud/resource_manager.rb
+++ b/lib/gcloud/resource_manager.rb
@@ -84,9 +84,13 @@ module Gcloud
   #
   # == Accessing the Service
   #
-  # The Resource Manager API *cannot* be enabled in the developer console, and
-  # can only be enabled by your account manager or a member of the Google Cloud
-  # sales team.
+  # Currently, the full functionality of the Resource Manager API is available
+  # only to whitelisted users. (Contact your account manager or a member of the
+  # Google Cloud sales team if you are interested in access.) Read-only methods
+  # such as ResourceManager::Manager#projects and
+  # ResourceManager::Manager#project are accessible to any user who enables the
+  # Resource Manager API in the {Developers
+  # Console}[https://console.developers.google.com].
   #
   # == Authentication
   #
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -120,6 +120,39 @@ sub = pubsub.subscription "my-topic-sub"
 msgs = sub.pull
 ```
 
+# Resource Manager
+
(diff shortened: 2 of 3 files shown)


lib/gcloud/resource_manager.rb
```diff
@@ -84,9 +84,13 @@ def self.resource_manager keyfile = nil, options = {}
   #
   # == Accessing the Service
   #
-  # The Resource Manager API *cannot* be enabled in the developer console, and
-  # can only be enabled by your account manager or a member of the Google Cloud
-  # sales team.
+  # Currently, the full functionality of the Resource Manager API is available
+  # only to whitelisted users. (Contact your account manager or a member of the
+  # Google Cloud sales team if you are interested in access.) The read-only methods
+  # {projects.list}[https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list] and
```
blowmage: What do you think about referencing the gcloud methods instead of the API methods here? So linking to `Manager#projects` here instead of `projects.list`?

quartzmo: That's what I did at first, then I changed it to the REST methods to indicate the origin of the access error.

quartzmo: Happy to change it back, I was on the fence about this.

lib/gcloud/resource_manager.rb
```diff
@@ -84,9 +84,13 @@ def self.resource_manager keyfile = nil, options = {}
...
+  # {projects.get}[https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/get]
+  # are accessible to any user who enables the Resource Manager API in the
```
blowmage: Thoughts on "are accessible to any user who enables" -> "are accessible to any user who manages a project that has enabled"?

lib/gcloud/resource_manager.rb
blowmage: Should we need to add a note to all the other methods that an ApiError will be raised if they don't have access granted?

blowmage: I think I'd prefer to link to the local methods here if we can make it clear they need to gain access to use the full ruby API.

quartzmo: Certainly not yet. Let's let user feedback drive that.
I'll push an updated commit in a minute. I agree, better to point to the immediate API.

blowmage:
LGTM :+1+

Issue #395, opened 2015-10-23
Pub/Sub Topic Policy is returning nil


jeffmendoza:
Looks like this isn't needed:
https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/d7870d074645a9b3b38c9d8dda133b60a8b66def/lib/gcloud/pubsub/topic.rb#L427

Here is what I get:
```
irb(main):051:0* resp.data
=> #<Google::APIClient::Schema::Pubsub::V1::Policy:0xe7fb44 DATA:{"etag"=>"BwUiutJSWn8=", "bindings"=>[{"role"=>"roles/viewer", "members"=>["serviceAccount:404066594503-j4825a3hq5k361sqvill049gc72qt3uf@developer.gserviceaccount.com"]}]}>
irb(main):052:0> resp.data["policy"]
=> nil
irb(main):053:0> resp.data.to_hash
=> {"etag"=>"BwUiutJSWn8=", "bindings"=>[{"role"=>"roles/viewer", "members"=>["serviceAccount:404066594503-j4825a3hq5k361sqvill049gc72qt3uf@developer.gserviceaccount.com"]}]}
irb(main):054:0> 
```

blowmage:
Hi @jeffmendoza, thank you for opening this issue. It certainly looks like the IAM policy data structure has changed. We have been looking deeper into this and there may be even more changes as a result.

blowmage:
This was closed when the PR was accepted, but there are more changes coming that will relate to this. Thanks again!

PR #397, opened 2015-10-26
Pub/Sub IAM

blowmage:
The Pub/Sub IAM Policy JSON structure has changed. (See #395). Update the code to use the latest Pub/Sub JSON and behavior. Add acceptance tests for getting and setting the policy.

### DIFF
--- a/lib/gcloud/pubsub/subscription.rb
+++ b/lib/gcloud/pubsub/subscription.rb
@@ -562,6 +562,46 @@ module Gcloud
       end
 
       ##
+      # Tests the specified permissions against the {Cloud
+      # IAM}[https://cloud.google.com/iam/] access control policy. See
+      # {Managing Policies}[https://cloud.google.com/iam/docs/managing-policies]
+      # for more information.
+      #
+      # === Parameters
+      #
+      # +permissions+::
+      #   The set of permissions to check access for. Permissions with wildcards
+      #   (such as +*+ or +storage.*+) are not allowed.
+      #   (String or Array of Strings)
+      #
+      # === Returns
+      #
+      # The permissions that have access. (Array of Strings)
+      #
+      # === Example
+      #
+      #   require "gcloud"
+      #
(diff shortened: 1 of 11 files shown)


jeffmendoza:
+1

quartzmo:
Thanks again @jeffmendoza for surfacing the IAM change! Apparently there is no versioning on the IAM methods in the REST API(s)?

blowmage:
I have acceptance tests for the `testIamPermissions` endpoint, but they are failing right now with the following:

```yaml
error:
  code: 503
  message: The service was unable to fulfill your request. Please try again.
  errors:
  - message: The service was unable to fulfill your request. Please try again.
    domain: global
    reason: backendError
  status: UNAVAILABLE
```

AFAICT this doesn't look to be a problem with how the ruby code is calling the endpoint. Do we want to hold off on the acceptance tests for now? Or should I add `skip` to them until they start to pass?

quartzmo:
If you already have them, keeping them with `skip` sounds ok to me. We just need to remember to try them again periodically!

lib/gcloud/pubsub/subscription.rb
```diff
@@ -476,11 +476,11 @@ def delay new_deadline, *messages
       # A hash that conforms to the following structure:
       #
       #   {
+      #     "etag"=>"CAE=",
```
quartzmo: What about the `version` attribute?

blowmage: The `version` attribute is in the documentation, but not in the data being returned by the service. I decided to trust the service's behavior over the documentation so I left `version` out.

quartzmo: Oh! OK, good call.

quartzmo:
This looks good to me. @blowmage, can I merge?

blowmage:
Yep. Merge away. :shipit:

PR #398, opened 2015-10-26
Resource Manager IAM

blowmage:
Update the IAM Policy structure to match what the service is using.


blowmage:
Tested the policy and permissions calls with my user account. Works without issue.

Issue #393, opened 2015-10-22
Question: Most Ruby way to create a Pub/Sub topic?


jeffmendoza:
This library allows a lot of flexibility

```ruby
# I think I'll go with
topic = pubsub.get_topic $topic_name
if topic.nil?
  topic = pubsub.create_topic $topic_name
end

#----

topic = pubsub.create_topic $topic_name unless pubsub.topic($topic_name).exists?
topic = pubsub.topic $topic_name

#---

if pubsub.topic($topic_name).exists?
  topic = pubsub.topic $topic_name
else
  topic = pubsub.create_topic $topic_name
end

#---

# use autocreate
topic = pubsub.topic $topic_name
topic.publish "message"
# need to create explicitly first to add a subscription though...

#---
```
cc @thagomizer

thagomizer:
What you have is fine but I try to avoid nil? unless necessary since it is an extra method call. 

I'd do it this way:

```ruby
topic ||= pubsub.create_topic $topic_name
```

You could also do

```ruby
topic = pubsub.create_topic($topic_name) unless topic
```

jeffmendoza:
Thanks =)

blowmage:
FWIW, I agree with @thagomizer that the most idiomatic way to do this in ruby is:

```ruby
topic = pubsub.get_topic($topic_name) || pubsub.create_topic($topic_name)
```

However this is not my preferred way. Further in your comments you indicate that you cannot use autocreate because you also need to add a subscription. This is not true. Autocreate will just work in this instance, and both the topic and subscription resources will be created when needed.

```ruby
topic = pubsub.topic "this-topic-does-not-exist-yet"
sub = topic.subscribe "this-subscription-does-not-exist-yet-either"
sub.name # Now both the topic and subscription exist.
```

After the second line has run both the topic and subscription will exist. Hope that helps!

jeffmendoza:
topic.subscribe will throw an exception if the subscription already exists though.

It's a bit confusing which operations will create the underlying topic/subscription. Since I can't lazily create the subscription, I need to make sure I have a real topic, then check weather the subscription exists. Now I have:

```ruby
  topic = pubsub.get_topic $topic_name
  topic ||= pubsub.create_topic $topic_name

  sub = topic.get_subscription $sub_name
  sub ||= topic.subscribe $sub_name
```

jgeewax:
This seems like a good place for `get_or_create`, no?

[...]

PR #402, opened 2015-10-27
Pub/Sub Refactor

### LINKED ISSUE #400: Having to choose between methods for retrieving a Topic or Subscription is confusing
In #393, @jeffmendoza wrote:

> This library allows a lot of flexibility

I share that impression, and in particular I was confused by the presence of two methods for retrieving topics and two for retrieving subscriptions:

* [Project#get_topic](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.4.1/lib/gcloud/pubsub/project.rb#L103)
* [Project#topic](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.4.1/lib/gcloud/pubsub/project.rb#L173)

and

* [Project#get_subscription](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.4.1/lib/gcloud/pubsub/project.rb#L298)
* [Project#subscription](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.4.1/lib/gcloud/pubsub/project.rb#L329)

The method names do little to suggest a distinction, so I looked at the docs, [which state](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.4.1/lib/gcloud/pubsub/project.rb#L117-L118):

> The difference between this method and Project#get_topic is that this method does not make an API call to Pub/Sub to verify the topic exists.

To me the behavior of "avoiding the API call to verify the Topic exists" feels like it should be an option to `Project#get_topic`, not a separate method.

Digging into the code, however, I found a rather involved mechanism for lazy loading. In the case of Topic, does this mechanism serve any goal other than just letting the user publish a message without needlessly retrieving a topic? If not, it seems to me that much internal complexity (and potential user confusion about state) could be avoided with a method that simply accepts the topic name and message:

```ruby
pubsub = gcloud.pubsub
msg = pubsub.publish  "existing-topic", "This is the first API call to Pub/Sub."
```

In my opinion, this simple design makes it very clear that the Topic is not being retrieved, and that "this method does not make an API call to Pub/Sub to verify the topic exists."

Clearly there could be more to this lazy loading than I was able to find on my own, so please inform me!

### LINKED ISSUE #401: Pub/Sub Topic auto-create is confusing
One of the issues with the current Pub/Sub design is the behavior of auto-creating topic objects. Currently, it is possible for the topic to create itself while publishing:

```ruby
topic = gcloud.pubsub.topic("not-yet-created")
topic.publish "Some message" # Topic is created in this call
```

A more verbose version of the same code is:

```ruby
topic = gcloud.pubsub.topic("not-yet-created", autocreate: true)
topic.publish "Some message" # Topic is created in this call
```

Having the creation occur on `Topic#publish` instead of `Project#topic` is confusing. Along with the proposed `Project#publish` in #400, we also propose streamlining the auto-create functionality by making the following changes:

* Set the default behavior for `autocreate` to `false`. The current default value is `true`.
* Create the record on the method that the `autocreate` option is specified on:

```ruby
topic = gcloud.pubsub.topic("not-yet-created", autocreate: true) # Topic is created in this call
topic.publish "Some message" # Message is published in this call
```

This will change performance, but `Project#publish` as defined in #400 will maintain existing performance. But the behavior is more inline with where we believe more developers expectations are.

blowmage:
This PR provides a number of changes.

### Reworked `autocreate` option

The `autocreate` option now applies to the scope of the `#topic` method the option is on instead of the topic object it returns. This means that instead of this behavior:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "not-yet-created" # autocreate used to default to true here
topic.publish "Some message" # Topic is created in this call
```

We now have this behavior:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "not-yet-created", autocreate: true # Topic is created in this call
topic.publish "Some message"
```

The `autocreate` option now defaults to `false`, where previously it was `true`.

### Add `Project#publish`

Since we lose the ability to create a topic resource after the topic object has been instantiated, we have introduced `Project#publish` that also supports the `autocreate` option and maintains the number of API calls made. Previously the following code made 1 API call.

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "existing-topic" # No API call to retrieve topic
topic.publish "Some message" # API call made in this call
```

Now this can be accomplished with the following:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.publish "existing-topic", "Some message" # API call made in this call
```

Alternatively, the previous code could auto-create a missing topic with 3 API calls:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "new-topic" # The autocreate default used to be true here
topic.publish "Some message" # 3 API calls made to create topic and publish
```

Now this can be accomplished with the following:

```ruby
pubsub = gcloud.pubsub
# 3 API calls made to create topic and publish
topic = pubsub.publish "new-topic", "Some message", autocreate: true
```

### Add `Project#subscribe`

Similar to `Project#publish`, `Project#subscribe` accepts the `autocreate` option and maintains the previous number of API calls:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "new-topic" # The autocreate default used to be true here
topic.subscribe "new-sub" # 3 API calls made to create topic and subscription
```

Now this can be accomplished with the following:

```ruby
pubsub = gcloud.pubsub
# 3 API calls made to create topic and subscription
topic = pubsub.subscribe "new-topic", "new-sub", autocreate: true
```

### Added `skip_lookup` option

The differences between `#topic` and `#get_topic` calling the API to verify the existance of the resource on the service have been removed. `#get_topic` is now an alias of `#topic` and the default behavior is to lookup the resource on the service. The ability to skip this lookup has been added with the `skip_lookup` option on the method.

This means it is still possible to publish using a topic object while skipping the lookup API call, but it is no longer possible to auto-create the topic during the publish. Before we had:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "existing-topic" # Used to skip lookup by default
topic.publish "Some message" # API call made here
```

We can match this with the following:

```ruby
pubsub = gcloud.pubsub
topic = pubsub.topic "existing-topic", skip_lookup: true
topic.publish "Some message" # API call made here
```

The same applies to `#subscription` and `#get_subscription`.

[closes #400, closes #401]


blowmage:
@stephenplusplus @tmatsuo @jeffmendoza This PR is ready for consideration. Thoughts?

quartzmo:
@blowmage Nice example-rich description that you wrote up for this PR, I appreciate it!

blowmage:
The Pub/Sub documentation still needs to be refreshed to ensure it encouraging the proper usage, but that will happen in a separate PR.

quartzmo:
Just want to note that in addition to addressing #400 and #401, this PR also refs #393.

tmatsuo:
I'm not a rubyist, so I'd rather not review the code, but from reading the description, the changes make sense to me. I'm also impressed with the great description!

Just one question. In the first section (autocreate), I don't understand why the second example will `autocreate` the topic, even if now it defaults to false. Am I miss something?

blowmage:
@tmatsuo Please refresh the issue. The option was missing from the code sample but it was added ~5 minutes ago.

quartzmo:
@blowmage Didn't you once tell me that clients should be able to create subscriptions without providing a subscription name, in order to receive a generated name? Shouldn't the `subscription_name` argument to `Project#subscribe` be optional?

blowmage:
@quartzmo I did say that to you, but I was mistaken. I confused Subscription with something else. The [current documentation](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.4.1/Gcloud/Pubsub/Topic.html#method-i-subscribe) makes it clear that subscription name is required.

lib/gcloud/pubsub/project.rb
```diff
@@ -125,15 +92,22 @@ def get_topic topic_name
...
+      #   exists on the Pub/Sub service. Calls made on this object will raise
+      #   errors if the topic resource does not exist. Default is +false+.
+      #   (+Boolean+)
       #
       # === Returns
       #
-      # Gcloud::Pubsub::Topic
+      # Gcloud::Pubsub::Topic or nil if topic does not exist. Can return a newly
+      # created Gcloud::Pubsub::Topic if the topic does not exist and
+      # +autocreate+ is not set to +true+
```
quartzmo: "is not set" should be "is set"?

lib/gcloud/pubsub/project.rb
```diff
@@ -125,15 +92,22 @@ def get_topic topic_name
       #   An optional Hash for controlling additional behavior. (+Hash+)
       # <code>options[:autocreate]</code>::
       #   Flag to control whether the topic should be created when needed.
```
quartzmo: "the topic should be created when needed" -> "the requested topic should be created if it does not exist"

quartzmo:
I think you also told me that `autocreate` and `skip_lookup` options should not be allowed together? I don't see that mentioned in the docs for `Project#topic`.

tmatsuo:
@blowmage Now I see it. Thanks!

blowmage:
@quartzmo Yeah, the `skip_lookup` option will override `autocreate`. If the lookup is skipped, then it can't know whether to auto-create or not. I toyed with the idea of raising an exception if both options were specified, but I found that to be somewhat confusing. I'd like to address this in the upcoming documentation changes, including raising any errors.

lib/gcloud/pubsub/project.rb
blowmage: Updated.

lib/gcloud/pubsub/project.rb
blowmage: Changed.

blowmage:
After further discussion with @quartzmo, we've decided to update the docs on this PR. The latest changes represent our current best thinking for design and documentation. The [docs are hosted here](http://blowmage.com/gcloud-ruby/docs/master/Gcloud/Pubsub.html).

Issue #404, opened 2015-10-27
Query namespace and kind


jondot:
Hi guys,
Is there a way to specify a namespace for a query (like with the Node variant of gcloud)? I didn't find such a thing anywhere in the official docs.

Thanks

blowmage:
Hey again @jondot. I've looked through the code and gcloud-ruby is not exposing the ability to specify namespace on a query. I'm confident we can add this and get it included in this week's release however. Here is what I was thinking:

```ruby
require "gcloud"

gcloud = Gcloud.new
dataset = gcloud.datastore
query = Gcloud::Datastore::Query.new
query.kind("List").where("active", "=", true)
active_lists = dataset.run query, namespace: "custom-ns"
```

Would that work for you? Or would you rather see a different design?

Thanks for opening this issue!

jondot:
Yes, that should be great. This is what I hacked out myself just to get some data flowing and to verify that this is possible.
Thanks!

jgeewax:
Is there a way we can throw in some hierarchy here so that people can pass around a namespace object that knows how to run queries? 

That is, the code might look like...

```ruby
require "gcloud"
gcloud = Gcloud.new
dataset = gcloud.datastore
namespace = dataset.namespace "mynamespace"
query = Gcloud::Datastore::Query.new
query.kind("List").where("active", "=", true)
active_lists = namespace.run query
```

Or is that not very ruby-esque ?

jondot:
I think the namespace object you are referring to is a mirror of the protobuf PartitionId, in addition @blowmage suggested something that will be mirroring the API for gcloud Java and Node drivers, at least these were the ones we have evaluated.

blowmage:
IMO its neither more or less ruby-esque that the previous example, just different. I'm somewhat on the fence about adding a separate `Namespace` class, but I'll think on it.

It looks like gcloud-node allows a default namespace to be set on the dataset object, so all queries run on the dataset will have the namespace applied. So it looks like you create separate Dataset instances with different default namespaces.

blowmage:
Since we are discussing Datastore, I have had a thought on some expanded functionality that namespace might fit into very well. Currently we create a separate Query object and pass that to Dataset. What if we allow the Query to be built by chaining methods instead? So this:

```ruby
require "gcloud"

gcloud = Gcloud.new
dataset = gcloud.datastore
query = Gcloud::Datastore::Query.new
query.kind("List").where("active", "=", true)
[...]

PR #409, opened 2015-10-28
Datastore Namespace

blowmage:
Refactor some of the internals for namespace support on queries. Add some documentation as well.

[refs #404]

### DIFF
--- a/lib/gcloud/datastore/dataset.rb
+++ b/lib/gcloud/datastore/dataset.rb
@@ -232,6 +232,9 @@ module Gcloud
         true
       end
 
+      # rubocop:disable Metrics/AbcSize
+      # Disabled rubocop because the level of abstraction is not violated here
+
       ##
       # Retrieve entities specified by a Query.
       #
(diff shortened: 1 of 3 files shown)


lib/gcloud/datastore/dataset.rb
```diff
@@ -239,19 +242,31 @@ def delete *entities_or_keys
...
+      # === Examples
       #
       #   query = Gcloud::Datastore::Query.new.kind("Task").
       #     where("completed", "=", true)
       #   tasks = dataset.run query
       #
-      def run query, opts = {}
-        response = connection.run_query query.to_proto, opts
+      # The query can optionally run within namespace when the +namespace+
+      # option is provided:
```
quartzmo: Maybe: "The query can be scoped to a namespace with the +namespace+ option." ?
And link namespace to external doc if there's a good explanation of it somewhere?

quartzmo:
Other than the typo above ("within namespace"), LGTM and ready to merge.

lib/gcloud/datastore/dataset.rb
blowmage: This has been changed.

blowmage:
That isn't a typo! :laughing:

lib/gcloud/datastore/dataset.rb
blowmage: I'm not aware of a Google Cloud link that explains namespaces. They are mentioned throughout the Datastore docs, but namespaces and multitenancy are not explicitly defined AFAICT.

PR #410, opened 2015-10-28
Datastore Helpers

blowmage:
This change intends to make it easier to write (and maintain) Datastore code using gcloud. It adds three small convenience methods that will save on characters needed to be typed and read.

Before:

```ruby
dataset = gcloud.datastore

# Create a new task to demo datastore
demo_task = Gcloud::Datastore::Entity.new
demo_task.key = Gcloud::Datastore::Key.new "Task", "datastore-demo"
demo_task["description"] = "Demonstrate Datastore functionality"
demo_task["completed"] = false

# Save the new task
dataset.save demo_task
```

After:

```ruby
dataset = gcloud.datastore

# Create a new task to demo datastore
demo_task = dataset.entity "Task", "datastore-demo" do |t|
  t["description"] = "Demonstrate Datastore functionality"
  t["completed"] = false
end  

# Save the new task
dataset.save demo_task
```

Before:

```ruby
# Create a query for all completed tasks
query = Gcloud::Datastore::Query.new.
  kind("Task").
  where("completed", "=", true)

# Run the query
completed_tasks = dataset.run query
```

After:

```ruby
# Create a query for all completed tasks
query = dataset.query("Task").
  where("completed", "=", true)

# Run the query
completed_tasks = dataset.run query
```

### DIFF
--- a/lib/gcloud/datastore/dataset.rb
+++ b/lib/gcloud/datastore/dataset.rb
@@ -39,7 +39,7 @@ module Gcloud
     #   gcloud = Gcloud.new
     #   dataset = gcloud.datastore
     #
-    #   query = Gcloud::Datastore::Query.new.kind("Task").
+    #   query = dataset.query("Task").
     #     where("completed", "=", true)
     #
     #   tasks = dataset.run query
@@ -246,7 +246,7 @@ module Gcloud
       #
       # === Example
       #
-      #   query = Gcloud::Datastore::Query.new.kind("Task").
+      #   query = dataset.query("Task").
       #     where("completed", "=", true)
       #   tasks = dataset.run query
       #
@@ -319,6 +319,38 @@ module Gcloud
           tx.rollback
           raise TransactionError.new("Transaction failed to commit.", e)
         end
+      end
+
+      ##
+      # Create a new Query instance. This is a convenience method to make the
+      # creation of Query objects easier.
+      #
+      # === Parameters
+      #
+      # +kinds+::
+      #   The kind of entities to query. This is optional. (+String+)
+      #
+      # === Returns
+      #
+      # Gcloud::Datastore::Query
+      #
+      # === Example
+      #
+      #   query = dataset.query("Task").
+      #     where("completed", "=", true)
+      #   tasks = dataset.run query
+      #
+      # This code is equivalent to the following:
+      #
+      #   query = Gcloud::Datastore::Query.new.
+      #     kind("Task").
+      #     where("completed", "=", true)
+      #   tasks = dataset.run query
+      #
+      def query *kinds
+        query = Query.new
+        query.kind(*kinds) unless kinds.empty?
+        query
       end
 
       protected
--- a/lib/gcloud/datastore/dataset.rb
+++ b/lib/gcloud/datastore/dataset.rb
@@ -368,11 +368,11 @@ module Gcloud
       #
       # === Example
       #
-      #   key = dataset.key "User", "username"
+      #   key = dataset.key "User", "heidi@example.com"
       #
       # This code is equivalent to the following:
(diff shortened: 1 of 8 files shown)


lib/gcloud/datastore/dataset.rb
```diff
@@ -321,6 +319,133 @@ def transaction
...
+      #   entity = dataset.entity
+      #
+      # This code is equivalent to the following:
+      #
+      #   entity = Gcloud::Datastore::Entity.new
+      #
+      # The key can also be passed in as an object:
+      #
+      #   key = dataset.key "User", "username"
+      #   entity = dataset.entity
```
quartzmo: argument `key` is missing

blowmage: Thanks for catching that. Fixed.

lib/gcloud/datastore/dataset.rb
```diff
@@ -321,6 +319,133 @@ def transaction
...
+      #
+      # Gcloud::Datastore::Key
+      #
+      # === Example
+      #
+      #   key = dataset.key "User", "username"
+      #
+      # This code is equivalent to the following:
+      #
+      #   key = dataset.key "User", "username"
```
quartzmo: replace with constructor example

lib/gcloud/datastore/dataset.rb
```diff
@@ -321,6 +319,133 @@ def transaction
...
       end
 
+      ##
+      # Create a new Query instance. This is a convenience method to make the
+      # creation of Query objects easier.
+      #
+      # === Parameters
+      #
+      # +kinds+::
+      #   The kind of entities to query. This is optional. (+String+)
```
quartzmo: Doc is for singular kind, but parameter supports multiple/list?
"kind" -> "kind or kinds"
String -> String or multiple

lib/gcloud/datastore/dataset.rb
blowmage: Ugh, I have literally fixed this 5 times this morning. Good catch. Fixed.

lib/gcloud/datastore/dataset.rb
blowmage: You can query multiple kinds. This is valid:

```ruby
gcloud.run gcloud.query("Users", "Admins)".where("active", "=", true)
```
 But the common usage is to provide a single `kind` value.

lib/gcloud/datastore/dataset.rb
```diff
@@ -321,6 +319,133 @@ def transaction
...
+      #
+      #   entity = dataset.entity
+      #
+      # This code is equivalent to the following:
+      #
+      #   entity = Gcloud::Datastore::Entity.new
+      #
+      # The key can also be passed in as an object:
+      #
+      #   key = dataset.key "User", "username"
```
quartzmo: The `"username"` example value for `id` really had me confused. On the Key docs I found this example, which I think is clearer.
```ruby
key = Gcloud::Datastore::Key.new "User", "heidi"
```

quartzmo: Or maybe the `id` example value `"heidi@example.com"` would be more realistic?

quartzmo:
I think these added methods must set a project record for docs to code ratio. Nice work. Everything is perfectly clear.

lib/gcloud/datastore/dataset.rb
blowmage: Totes agree. Added a new commit to replace the vague `"username"` with the concrete `"heidi@example.com"` in the documentation.

blowmage:
Well there is almost no implementation to them. They are just shortcuts. All I can do is add documentation! :)

PR #412, opened 2015-10-29
Rubocop Cleanup

### LINKED ISSUE #408: Replace rubocop:disable in code with Rubocop configuration
I consider the `rubocop:disable` statements an anti-pattern that adds unwelcome noise to the codebase. We should embrace Rubocop's flexibility and configure it (sparingly, of course) to support our own notion of good style.

blowmage:
Update Rubocop configuration to set slightly higher levels for method complexity and line length.
Many of the methods connecting to Google Cloud services are complex because
the Google Cloud service is complex. Relax these metrics to ease development.

[refs #408]

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -19,3 +19,7 @@ Metrics/ClassLength:
   Enabled: false
 Style/EmptyElse:
   Enabled: false
+Metrics/AbcSize:
+  Max: 20
+Metrics/MethodLength:
+  Max: 15
--- a/lib/gcloud/bigquery/project.rb
+++ b/lib/gcloud/bigquery/project.rb
@@ -266,10 +266,6 @@ module Gcloud
         end
       end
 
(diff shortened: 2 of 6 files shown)


quartzmo:
@blowmage Does this PR miss any cleanup in Resource Manager that we might want to do?

quartzmo:
Also, does this leave any remaining `rubocop:disable` statements in the codebase?

blowmage:
Nope. There is only one disable in Resource Manager and it is for [line length](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/ca83bbac3b0c2552283a4cfeeb026779a79a0bad/lib/gcloud/resource_manager.rb#L64-L65) because of the length of URLs in the documentation.

Any unnecessary disable will be flagged by Rubocop. So as long as rubocop is passing there are no unnecessary exceptions in the comments. I went through the dozen or so places we disable and they seem fine to me.

Issue #415, opened 2015-10-30
List Indexes


quartzmo:
Indexes are identified by a unique name (`indexId` in the REST API).

- [x] Use the [indexes.list](https://cloud.google.com/search/reference/rest/v1/projects/indexes/list) method to retrieve the indexes in the project:

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

indexes = search.indexes  # API call
indexes.each do |index|
  index.name #=> "store_products" 
end
```

- [x] The REST API accepts a filter parameter named `indexNamePrefix`. I'm not sure what these prefixes look like in practice, but here's a guess:

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

indexes = search.indexes prefix: "store_"  # API call
indexes.each do |index|
  index.name #=> "store_products" 
end
```

- [x] Support pagination:

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

indexes = search.indexes  # API call
indexes.next? # Are there more pages?
indexes.next # Get the next page of indexes
indexes.all # Get all pages of indexes
```

Issue #414, opened 2015-10-30
Create new Index object


quartzmo:
Indexes are searchable collections of documents.

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

new_index = search.index "products"
```

**Important**: Indexes cannot be created, updated, or deleted directly on the server: they are derived from the documents which are created “within” them. The cannot even be retrieved except by `indexes.list` in the REST API.

Personally, I find it very easy to mistake this method for a lookup/retrieval, but if we follow the [implementation in gcloud-python](https://github.com/GoogleCloudPlatform/gcloud-python/blob/master/gcloud/search/client.py#L96-L105), it only initializes an `Index` value object that holds a reference to the "Client" so that it can execute searches. Indeed, the REST API does not offer [`indexes.get`](https://cloud.google.com/search/reference/rest/v1/projects/indexes).

jgeewax:
This will be changed in the API at some point (before it becomes generally available to to the public), so... eventually we'll have this as a thing...

blowmage:
If the indexes are going to CRUD in the future then I think it would be wise to make the ruby API behave like they are connecting to a CRUD service. In #421 we simulate a `get` API by performing a list with the prefix, and then returning the index that matches the id exactly. This is not be as performant as a real `get`, but it does allow us to build the ruby API as if it were there.

PR #420, opened 2015-11-02
Search Setup

blowmage:
Add basic boilerplate for Cloud Search. Because we are unable to use the Google API Client for this service, we have to call the API directly. The setup has changed to reflect this. The plumbing will change as we continue development, but for now this works.

This also includes preliminary implementations of the methods to return a list of all existing indexes, as well as mocking out a method to find an index by name, which the Cloud Search API does not yet allow.

[refs #414, refs #415]

### DIFF
--- a/lib/gcloud.rb
+++ b/lib/gcloud.rb
@@ -350,4 +350,35 @@ module Gcloud
   end
 
   # rubocop:enable Metrics/LineLength
+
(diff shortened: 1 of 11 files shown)


blowmage:
Oops, this PR is trying to merge into the wrong branch. Disregard.

PR #424, opened 2015-11-04
Add http verbs to Search Connection

quartzmo:
This is largely a refactoring of the existing Search Connection to use the API (but not the actual implementation) of Google API Client. This refactoring should make it much easier to convert to using the real Google API Client once the Search REST API is discoverable.

### DIFF
--- a/lib/gcloud/search/api_client.rb
+++ b/lib/gcloud/search/api_client.rb
@@ -1,0 +1,128 @@
+#--
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "gcloud/version"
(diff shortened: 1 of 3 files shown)


lib/gcloud/search/connection.rb
```diff
@@ -34,44 +34,32 @@ class Connection #:nodoc:
...
-        options[:uri] = uri path
-        if @client.nil?
-          @connection.get do |req|
-            req.url options[:uri]
-            req.params = options[:params] if options[:params]
-            req.body = options[:body] if options[:body]
-          end
-        else
-          options[:connection] = @connection
-          @client.fetch_protected_resource options
```
quartzmo: @blowmage Which additional options (or other method calls) are required here to support POST, DELETE in production?

lib/gcloud/search/api_client.rb
```diff
@@ -0,0 +1,128 @@
...
+          @connection.send method do |req|
+            req.url uri
+            req.params = options[:parameters] if options[:parameters]
+            req.body = options[:body_object] if options[:body_object]
+          end
+        else
+          # TODO: Support POST, DELETE
+          options[:uri] = uri
+          options[:connection] = @connection
+          authorization.fetch_protected_resource options
```
quartzmo: @blowmage Have you already figured out which additional options (or other method calls) are required with Signet's `fetch_protected_resource` to support POST, DELETE in production?

blowmage: I think it is `:method`, but I'll have to test for sure.

```ruby
options[:method] = method
```

blowmage:
This looks great! I love this approach. I say we merge this and then I'll rebase my WIP to use it. Once we get to POST and DELETE we can make those changes.

PR #425, opened 2015-11-06
Add Index#search

### LINKED ISSUE #418: Search an Index
The `Index#search` method should:

- [x] accept a [query string](https://cloud.google.com/search/query) parameter ("If the query is missing or empty, all documents are returned.")
- [x] return search result objects, each containing `fields` as specified in #413, and the `document_id`
- [x] return paginated results controlled by `pageSize`

Basic usage:

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

index = search.index "products"
results = index.search "red cotton"  # API call
results.size #=> 10
results.next? #=> true
result_1 = results.first
fields = result_1.fields # all or selected document fields; and computed fields from fieldExpressions
fields[0].name #=> "price"
fields[0].type #=> :number
fields[0].value # or just fields[0] to get value? See Issue #413
#=> 24.95 
result_1.document_id #=> "product-sku-000001"
document_1 = result_1.document # API call
document_1.fields # all document fields; no computed fields from the query fieldExpressions
```

In addition, this method must support the following query parameters described in [indexes.search](https://cloud.google.com/search/reference/rest/v1/projects/indexes/search):

- [x] fieldExpressions
```ruby
expressions = [
  { name: "TotalPrice", expression: "(Price+Tax)" },
  { name: "snippet", expression: "snippet('good times', content)" }
]
documents = index.search "red cotton", expressions: expressions
```
- [x] matchedCountAccuracy (integer)
- [x] orderBy
```ruby
documents = index.search "red cotton", order: ["price", "-avg_review"]
```
- [x] scorer
```ruby
documents = index.search "red cotton", scorer: :generic
```
- [x] scorerSize (integer)
- [x] returnFields
```ruby
documents = index.search "red cotton", fields: ["sku", "description"]
```

quartzmo:
[closes #418]

### DIFF
--- a/lib/gcloud/search/connection.rb
+++ b/lib/gcloud/search/connection.rb
@@ -49,7 +49,7 @@ module Gcloud
                    indexNamePrefix: options.delete(:prefix),
                    pageSize: options.delete(:max),
                    pageToken: options.delete(:token)
-                 }.delete_if { |_, v| v.nil? }
+        }.delete_if { |_, v| v.nil? }
 
         @client.execute(
           api_method: @search.indexes.list,
@@ -106,8 +106,33 @@ module Gcloud
         )
       end
 
+      def search index_id, query, options = {}
+        @client.execute(
+          api_method: @search.indexes.search,
+          parameters: search_request(index_id, query, options)
+        )
+      end
+
       def inspect #:nodoc:
         "#{self.class}(#{@project})"
+      end
+
+      protected
+
+      def search_request index_id, query, options = {}
+        { projectId: @project,
+          indexId: index_id,
+          query: query,
+          fieldExpressions: options[:expressions],
+          matchedCountAccuracy: options[:matched_count_accuracy],
+          offset: options[:offset],
+          orderBy: options[:order],
+          pageSize: options[:max],
+          pageToken: options[:token],
+          returnFields: options[:return_fields],
+          scorerSize: options[:scorer_size],
+          scorer: options[:scorer]
+        }.delete_if { |_, v| v.nil? }
       end
     end
   end
--- a/lib/gcloud/search/index.rb
+++ b/lib/gcloud/search/index.rb
@@ -238,7 +238,7 @@ module Gcloud
         ensure_connection!
         resp = connection.search index_id, query, options
         if resp.success?
-          Result::List.from_response resp, self
+          Result::List.from_response resp, self, query, options
         else
           fail ApiError.from_response(resp)
         end
--- a/lib/gcloud/search/index.rb
+++ b/lib/gcloud/search/index.rb
@@ -15,8 +15,7 @@
 
 require "gcloud/search/document"
 require "gcloud/search/index/list"
-# TODO: replace with /result, require list in Result
-require "gcloud/search/result/list"
+require "gcloud/search/result"
 
 module Gcloud
   module Search
@@ -193,7 +192,7 @@ module Gcloud
       #
       #   results = index.search "dark stormy"
       #   results.each do |result|
-      #     puts result.document_id
+      #     puts result.doc_id
       #   end
       #
       # If you have a significant number of search results, you may need to
@@ -208,7 +207,7 @@ module Gcloud
       #   results = index.results
       #   loop do
       #     results.each do |result|
-      #       puts result.document_id
+      #       puts result.doc_id
       #     end
       #     break unless results.next?
       #     results = results.next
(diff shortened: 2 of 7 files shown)


quartzmo:
I still need to add Search::Result, after which I'll remove the `don't merge` label.

lib/gcloud/search/connection.rb
```diff
@@ -57,9 +57,34 @@ def list_indexes options = {}
...
         "#{self.class}(#{@project})"
       end
+
+      protected
+
+      def search_request index_id, query, options = {}
+        { projectId: @project,
+          indexId: index_id,
+          query: query,
+          fieldExpressions: options.delete(:expressions),
```
blowmage: The code isn't using `options` after this method is called, so I don't think we need to call `.delete` to get the values. We should be able to just do this:

```ruby
fieldExpressions: options[:expressions],
```

lib/gcloud/search/index.rb
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "gcloud/search/index/list"
+require "gcloud/search/result/list"
```
blowmage: On the other services we are requiring the model, and let the model require the list. So I think the requires here should be:

```ruby
require "gcloud/search/index/list"
require "gcloud/search/result"
```

lib/gcloud/search/project.rb
```diff
@@ -17,6 +17,7 @@
 require "gcloud/search/connection"
 require "gcloud/search/credentials"
 require "gcloud/search/index"
+require "gcloud/search/index/list"
```
blowmage: As stated in the other require comment, I don't think the list require is needed here. It should be required with index.

test/helper.rb
```diff
@@ -772,6 +772,15 @@ def mock_connection
     @connection
   end
 
+  def random_index_hash seed = nil, index_id = nil
```
blowmage: I think the args should be just `index_id`. If index_id is nil then it should generate a random id.

```ruby
def random_index_hash index_id = nil
  index_id ||= "example-index-#{rand 9999}"
  {
    "projectId" => project,
    "indexId" => index_id
  }
end
```

lib/gcloud/search/index.rb
quartzmo: Will do, as soon as the model is added.

lib/gcloud/search/connection.rb
quartzmo: Agreed. We can also do this for `Connection#list_indexes` I think, but I won't do it here.

test/helper.rb
quartzmo: Good, will do.

lib/gcloud/search/result/list.rb
```diff
@@ -0,0 +1,93 @@
...
+        rescue JSON::ParserError
+          ApiError.from_response_status resp
+        end
+
+        protected
+
+        ##
+        # Raise an error unless an active connection is available.
+        def ensure_index!
+          fail "Must have @index" unless @index
```
blowmage: Users don't create list instances, so this error message isn't helpful to them. They shouldn't know what internals to set to make the list work. This is why we've used the connection error message in other lists that keep hold of a resource object instead of a connection object.

quartzmo: Good catch. Should it be "Must have active connection" instead? That's what's used in ResourceManager::Project::List.

blowmage: Yeah.

lib/gcloud/search/result.rb
```diff
@@ -0,0 +1,70 @@
...
+        @index = index
+      end
+
+      ##
+      # The ID of the Document referenced by this result.
+      def doc_id
+        @raw["docId"]
+      end
+
+      # TODO: Uncomment after PR #427 is merged
```
quartzmo: This PR should wait for #427 to be merged in order to complete this TODO.

quartzmo:
@blowmage This should be ready to merge, but please take a careful look at it, of course.

blowmage:
Looking.

lib/gcloud/search/result.rb
```diff
@@ -0,0 +1,61 @@
...
+  module Search
+    ##
+    # = Result
+    #
+    # See Gcloud#search
+    class Result
+      ##
+      # Creates a new Result instance.
+      def initialize raw #:nodoc:
+        @raw = raw
```
blowmage: In Document `@raw` is set to an empty hash in the initializer, and the way to convert to/from hash is using `Document.from_hash` and `Document#to_hash`. I think Result and Document should behave the same way.

PR #429, opened 2015-11-09
Remove parallel assignment

blowmage:
Remove instances of parallel assignment in Storage::File. This is needed because the latest version of Rubocop has flagged parallel assignment as an offense.


quartzmo:
I like parallel assignment, so I'd be open to configuring Rubocop to allow this. However, let's merge this and conform for the time being.

blowmage:
FWIW, I like it as well. But I would also like to keep the Rubocop exceptions to a minimum. Since parallel assignment was only used in the one method I figured it was just as easy to change the code than add to the configuration.

Issue #413, opened 2015-10-29
Manage Document fields


quartzmo:
Document fields are probably the most complex part of the Search v1 REST API. We need to decide the best way for the gcloud-ruby API to manage fields and their values.

## Background

The following excerpts were **both** taken from the Cloud Search guide to [Documents and Indexes](https://cloud.google.com/search/documents_indexes):

In the first mention of fields, `name`, `type`, and `value` are attributes of each field.

> A document is an object with a unique ID and a list of fields containing user data. Each field has a name and a type. There are several types of fields, identified by the kinds of values they contain:
> 
> Atom Field - an indivisible character string
> Text Field - a plain text string that can be searched word by word
> HTML Field - a string that contains HTML markup tags, only the text outside the markup tags can be searched
> Number Field - a floating point number
> Date Field - a date object with year/month/day and optional time
> Geopoint Field - a data object with latitude and longitude coordinates  

In the second mention of fields, `name` is a unique attribute of each field, which may contain multiple values of different types.

> A document contains a list of one or more fields, with each field having a name and a list of values. Each field name consists of only ASCII characters, must be unique within the document and is case sensitive. A field name must start with a letter and can contain letters, digits, or underscore, with a maximum of 500 characters. A field can have multiple values with same or different types; however, it cannot have multiple Timestamp or number values.

Obviously, the data in both cases can be the same. It is just a question of how best to manage it.

#### Option 1: Grouped by name

In option 1, multiple values are set on a Field object that is created (or retrieved?) by name. This is the
approach taken by the REST API as well as gcloud-python.

```ruby
document = index.document "product-sku-000001"

field = document.field "description"
field.add_value "100% organic cotton ruby gem T-shirt",
                type: :text,
                lang: "en"
field.add_value "<p>100% organic cotton ruby gem T-shirt</p>",
                type: :html,
                lang: "en"
field.values.count #=> 2
```

#### Option 2: Flat

In option 2, multiple fields are added to a single list. Field `name` is not unique.

```ruby
document = index.document "product-sku-000001"

document.add_field "description",
                   "100% organic cotton ruby gem T-shirt",
                   type: :text,
                   lang: "en"

document.add_field "description",
                   "<p>100% organic cotton ruby gem T-shirt</p>",
                   type: :html,
                   lang: "en"
document.fields_by(name: "description").count #=> 2
```

[...]

PR #433, opened 2015-11-12
Add Search::Fields

quartzmo:
This PR is currently a work in progress, exploring solutions for the [FieldValueList](https://cloud.google.com/search/reference/rest/google/cloudsearch/v1/FieldValueList) representation suggested by @blowmage in #413.

### DIFF
--- a/lib/gcloud/search/document.rb
+++ b/lib/gcloud/search/document.rb
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 require "gcloud/search/document/list"
+require "gcloud/search/fields"
 
 module Gcloud
   module Search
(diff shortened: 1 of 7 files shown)


quartzmo:
@blowmage When you look at this, don't worry too much about the implementation. I'm coding it test-first and so the implementation is only what it needs to be until we nail down the API. Look at and please provide feedback on the tests.

blowmage:
I really like the direction this is going. I've had similar thoughts re implementation. As long as we have a great API defined in tests we should be free to change the internals as needed.

quartzmo:
@blowmage This is an incomplete solution, but I feel it would be good to merge what I have done so far, so that we can more easily add to this foundation in parallel.

blowmage:
Agreed. This direction is the way to to IMO.

PR #434, opened 2015-11-16
Add Vagrant

blowmage:
Using vagrant is optional, but it will allow an easier way to get the
gcloud SDK and GRPC installed on the development environment.


quartzmo:
Thanks @blowmage this looks great.

PR #437, opened 2015-11-19
Fix Datastore::Transaction unit tests

quartzmo:
A change in minitest exposed what appears to be redundant mock setup. This setup is already made in lines 21-5:

```ruby
let(:connection)  do
   mock = Minitest::Mock.new
   mock.expect :begin_transaction, begin_transaction_response
   mock
end
```


blowmage:
Makes sense. I've been following the discussion on changes Minitest::Mock that causes this. :+1:

PR #445, opened 2015-11-24
Search Value Add

blowmage:
Only allow one datetime or number value to be added to a field.

### DIFF
--- a/lib/gcloud/search/field_values.rb
+++ b/lib/gcloud/search/field_values.rb
@@ -143,7 +143,11 @@ module Gcloud
       #   document["price"].add 24.95
       #
       def add value, options = {}
-        @values << FieldValue.new(value, options.merge(name: @name))
+        new_field = FieldValue.new(value, options.merge(name: @name))
+        if [:datetime, :number].include? new_field.type
+          @values.delete_if { |v| v.type == new_field.type }
+        end
+        @values << new_field
       end
 
       # rubocop:enable Metrics/LineLength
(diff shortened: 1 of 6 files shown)


lib/gcloud/search/fields.rb
```diff
@@ -94,7 +94,10 @@ def [] name
       # Disabled because there are links in the docs that are long.
 
       ##
-      # Add a new value. If the field name does not exist it will be added.
+      # Add a new value. If the field name does not exist it will be added. If
+      # the field value is a DateTime or Numeric, or the type is set to
+      # +:datetime+ or +:number+, then the added value will replace any existing
+      # values of the same type.
```
quartzmo: "any existing values of the same type" -> "an existing value of the same type" (since there can be only one)

lib/gcloud/search/field_values.rb
```diff
@@ -86,7 +86,10 @@ def [] index
       # Disabled because there are links in the docs that are long.
 
       ##
-      # Add a new value. The field name will be added to the value object.
+      # Add a new value. The field name will be added to the value object. If
+      # the field value is a DateTime or Numeric, or the type is set to
+      # +:datetime+ or +:number+, then the added value will replace any existing
+      # values of the same type.
```
quartzmo: "any existing values of the same type" -> "an existing value of the same type" (since there can be only one)

quartzmo:
Can you also add new documentation to `Document#add`?

blowmage:
Updated.

Issue #416, opened 2015-10-30
Inspect Index schema


quartzmo:
Every index has a schema that shows all the field names and field types that appear in the documents it contains. See [Index schemas](https://cloud.google.com/search/documents_indexes#index_schemas) for discussion. 

The [indexes.list FieldNames](https://cloud.google.com/search/reference/rest/v1/projects/indexes/list#google.cloudsearch.v1.FieldNames) REST API representation organizes the field names by type, but personally I would prefer a flat structure containing **all** fields in order to facilitate flexible, functional-style transformations: 

```ruby
require "gcloud"

gcloud = Gcloud.new
search = gcloud.search

indexes = search.indexes  # API call
indexes.each do |index|
  puts index.name
  index.fields.each do |field|
    puts "#{field.name} - #{field.type}"
  end

  # This would not be as easy if fields returned a Hash containing 
  # type keys and array of name values.
  index.fields.select { |f| f.name == "description" }.map(&:type) #=> [:text, :html]
end
```

The [gcloud-python example](https://googlecloudplatform.github.io/gcloud-python/latest/search-usage.html#indexes) also shows an `index.fields` array that presumably contains **all** fields.

blowmage:
The Search indexes.list API isn't returning this data. The [API docs](https://cloud.google.com/search/reference/rest/v1/projects/indexes/list) state that this should be returned when setting [`view`](https://cloud.google.com/search/reference/rest/v1/projects/indexes/list#body.query_parameters.google.cloudsearch.v1.ListIndexesRequest.view) to [`FULL`](https://cloud.google.com/search/reference/rest/v1/projects/indexes/list#google.cloudsearch.v1.ListIndexesRequest.IndexInfoView.enum_values), but the service isn't returning the data.

I'm removing this issue from the Search milestone. I will leave this open for when the API behaves as documented and we can support this in gcloud-ruby.

quartzmo:
Closed by #444, #446.

PR #446, opened 2015-11-24
Add helper methods for index schema

blowmage:
These helper methods will make it easier to inspect the indexed fields on an index.

[refs #416]


quartzmo:
I'm pretty sure folks will find this helpful, very nice!

PR #447, opened 2015-11-24
Update Search documentation

quartzmo:


lib/gcloud/search.rb
```diff
@@ -170,32 +167,40 @@ def self.search project = nil, keyfile = nil, options = {}
   #   document = index.find "product-sku-000001"
   #
   #   document.rank = 12345
-  #   document.add "price", 24.95
+  #   document["price"].delete # clear existing value
```
blowmage: This probably needs to be:

```ruby
document.delete "price" # clear existing value
```

This will call `Fields#delete` which removed the field, where the way it is now it is calling `FieldList#delete` which is removing the field value.

Also, #445 will change this and adding a new value will remove the old one, so if that PR is accepted this line could just be removed.

blowmage:
This is looking good! :+1:

lib/gcloud/search.rb
quartzmo: Good catch, thanks!! I'll update it.

PR #451, opened 2015-12-02
Fix search URL generation

blowmage:
The search querystring was not being added when running authed requests.
Add query parameters to the URI to force both code paths to be the same.


quartzmo:
Thanks for figuring this out, and sorry you had to dig in to ApiClient. It should have just worked!

PR #467, opened 2015-12-11
Search Named Parameters

blowmage:
Update the search branch to named parameters. This brings it inline with where master is.

### DIFF
--- a/lib/gcloud/search/index.rb
+++ b/lib/gcloud/search/index.rb
@@ -243,10 +243,10 @@ module Gcloud
       #
       # === Parameters
       #
-      # <code>options[:token]</code>::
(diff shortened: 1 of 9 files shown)


lib/gcloud/search/index.rb
```diff
@@ -286,8 +284,9 @@ def document doc_id = nil, rank = nil
       #     documents = documents.next
       #   end
       #
-      def documents options = {}
+      def documents token: nil, max: nil, view: nil
```
quartzmo: Is `view` missing from the docs?

blowmage: No, it is undocumented.

Issue #466, opened 2015-12-10
Generate docs to static HTML using gcloud-yard template


blowmage:
We may need a YARD template, similar to the gcloud-rdoc template used for the current RDoc formatted documentation. This would help us bridge the gap between generating HTML documentation files that we do now, and moving to the JSON-backed angular application sometime in the future. Tags before the YARD conversion would use gcloud-rdoc, and tags after the conversion would use gcloud-yard.

quartzmo:
@blowmage I agree this would be great, as it would allow us to merge our YARD branch into master without having to wait on the Angular app. How much work do you think it would be?

blowmage:
¯\_(ツ)_/¯

blowmage:
:laughing: I'm not sure how much time it would take, but I bet it can be done by the time the YARD conversion is merged to master.

quartzmo:
I think it's a good idea, and will pay off in terms of the time it saves us maintaining the branch.

blowmage:
This is complete.

PR #477, opened 2015-12-18
Add YARD doc template

blowmage:
Supports the new yard-gcloud template gem for building static HTML site with
the same structure as the gcloud-rdoc template for RDoc.
When the YARD conversion is complete the yard-gcloud template will be the default.
Older releases will have their documentation built using the existing gcloud-rdoc
template, but newer releases will use the yard-gcloud template.

[refs #466]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,15 +1,15 @@
 Gemfile.lock
 keyfile.json
-coverage
-doc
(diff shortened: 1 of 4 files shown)


quartzmo:
Awesome progress!

PR #515, opened 2016-01-12
Logging setup

blowmage:
Create core Logging entities such as Project, Connection, Credentials, Resource, Sink, and Metric. Most resources and API calls are accounted for, except for Entries.

### DIFF
--- a/lib/gcloud.rb
+++ b/lib/gcloud.rb
@@ -313,4 +313,38 @@ module Gcloud
     require "gcloud/search"
     Gcloud.search @project, @keyfile, scope: scope
   end
+
+  ##
+  # Creates a new object for connecting to the Logging service.
+  # Each call creates a new connection.
+  #
+  # @param [String, Array<String>] scope The OAuth 2.0 scopes controlling the
+  #   set of resources and operations that the connection can access. See [Using
+  #   OAuth 2.0 to Access Google
+  #   APIs](https://developers.google.com/identity/protocols/OAuth2).
+  #
+  #   The default scope is:
+  #
+  #   * `https://www.googleapis.com/auth/logging.admin`
+  #
+  # @return [Gcloud::Logging::Project]
+  #
+  # @example
+  #   require "gcloud"
+  #
+  #   gcloud = Gcloud.new
+  #   logging = gcloud.logging
+  #   # ...
+  #
+  # @example The default scope can be overridden with the `scope` option:
+  #   require "gcloud"
+  #
+  #   gcloud  = Gcloud.new
+  #   platform_scope = "https://www.googleapis.com/auth/cloud-platform"
+  #   logging = gcloud.logging scope: platform_scope
+  #
+  def logging scope: nil
+    require "gcloud/logging"
+    Gcloud.logging @project, @keyfile, scope: scope
+  end
 end
--- a/lib/gcloud/logging/project.rb
+++ b/lib/gcloud/logging/project.rb
@@ -72,6 +72,52 @@ module Gcloud
           Gcloud::GCE.project_id
       end
 
+      ##
+      # Retrieves the list of monitored resources that are used by Google Cloud
+      # Logging.
+      #
+      # @param [String] token A previously-returned page token representing part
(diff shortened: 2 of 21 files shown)


lib/gcloud.rb
```diff
@@ -313,4 +313,39 @@ def search scope: nil
...
+  #   * `https://www.googleapis.com/auth/logging.admin`
+  #
+  # @return [Gcloud::Logging::Project]
+  #
+  # @example
+  #   require "gcloud"
+  #
+  #   gcloud = Gcloud.new
+  #   logging = gcloud.logging
+  #   topic = logging.topic "my-topic"
```
quartzmo: Change from this pubsub example to `logging.resources`, etc.

blowmage: Ugh, I thought I caught all of those. Updated and removed. These code examples will need to be revisited once we get more of the logging service implemented.

lib/gcloud/logging/project.rb
```diff
@@ -0,0 +1,362 @@
...
+      # @return [Gcloud::Logging::Sink]
+      #
+      # @example
+      #   require "gcloud"
+      #
+      #   gcloud = Gcloud.new
+      #   logging = gcloud.logging
+      #   sink = logging.create_sink "my-sink"
+      #
+      def create_sink name, destination: nil, filter: nil, version: :unspecified
```
quartzmo: The API docs state: "An unspecified version format will default to V2." If this is the case, do we need/want to include `unspecified` as an option?

lib/gcloud/logging/project.rb
```diff
@@ -0,0 +1,362 @@
...
+        version = Sink::VERSIONS[version] if Sink::VERSIONS[version]
```
quartzmo: In other services we have done this type of conversion differently, in protected methods in Connection. I'm good with either way, just pointing this out.

lib/gcloud/logging/project.rb
blowmage: It seems like the default behavior is when unspecified it defaults to V2. That could change over time, so the way I read it is that the default value is `unspecified` which will then become `V2`. So I thought it made sense to make the default `unspecified`. But right now I'm just trying to establish the service API defaults, we can always refine the ruby code defaults as we polish the ruby API. Getting some acceptance tests would help clarify this I bet.

lib/gcloud/logging/resource/list.rb
```diff
@@ -0,0 +1,77 @@
...
+      ##
+      # Resource::List is a special case Array with additional values.
+      class List < DelegateClass(::Array)
+        ##
+        # If not empty, indicates that there are more records that match
+        # the request and this value should be passed to continue.
+        attr_accessor :token
+
+        ##
+        # Create a new Zone::List with an array of Zone instances.
```
quartzmo: All three List docs have residual Zone/zone mentions.

lib/gcloud/logging/project.rb
blowmage: Yeah, we can make those changes in a future refactor and not change the existing ruby API. Sometimes I wait until I see something duplicated 3 or more times before refactoring.

lib/gcloud/logging/resource.rb
```diff
@@ -0,0 +1,81 @@
...
+      # documentation.
+      def description
+        @gapi["description"]
+      end
+
+      ##
+      # A set of labels that can be used to describe instances of this monitored
+      # resource type.
+      def labels
+        # TODO: Make a proper Label class to represent this structure...
```
quartzmo: Yes, curious to see how we can represent the somewhat awkward `LabelDescriptor`...

blowmage: Yeah, not looking forward to that.. :)

quartzmo:
What a strong start to Logging. I know you put a lot of work into this! Great progress.

blowmage:
Thanks for being so diligent in the code review @quartzmo. I'll try to keep future PRs smaller and easier to review.

PR #521, opened 2016-01-14
Add Project#delete_log

### LINKED ISSUE #497: Delete Log
From the API docs: "Deletes a log and all its log entries. The log will reappear if it receives new entries."

quartzmo:
[closes #497]

### DIFF
--- a/test/gcloud/logging/project_test.rb
+++ b/test/gcloud/logging/project_test.rb
@@ -19,4 +19,26 @@ describe Gcloud::Logging::Project, :mock_logging do
     logging.must_be_kind_of Gcloud::Logging::Project
     logging.project.must_equal project
   end
+
+  it "deletes a log" do
+    log_name = "syslog"
+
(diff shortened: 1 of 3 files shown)


test/gcloud/logging/project_test.rb
```diff
@@ -19,4 +19,15 @@
     logging.must_be_kind_of Gcloud::Logging::Project
     logging.project.must_equal project
   end
+
+  it "deletes a log" do
```
blowmage: It might be good to have a test that accepts a full log path instead of just the log name. Seeing as Entry will return the [full log path](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry#FIELDS.log_name).

blowmage: Similar to how Pub/Sub accepts a topic name or the full topic path. See `Gcloud::Pubsub::Connection#topic_path`.

PR #522, opened 2016-01-14
Cap rubocop at last version (0.35)

quartzmo:
Capping rubocop at last release (0.35) due to trouble with
0.36 release and older Ruby versions.

### DIFF
--- a/gcloud.gemspec
+++ b/gcloud.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency      "minitest-rg", "~> 5.2"
   gem.add_development_dependency      "autotest-suffix", "~> 1.1"
   gem.add_development_dependency      "rdoc", "~> 4.0"
-  gem.add_development_dependency      "rubocop", "~> 0.27"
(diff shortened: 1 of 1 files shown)


gcloud.gemspec
```diff
@@ -34,7 +34,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency      "minitest-rg", "~> 5.2"
   gem.add_development_dependency      "autotest-suffix", "~> 1.1"
   gem.add_development_dependency      "rdoc", "~> 4.0"
-  gem.add_development_dependency      "rubocop", "~> 0.27"
+  gem.add_development_dependency      "rubocop", "<= 0.35"
```
blowmage: I think we want to make it 0.35, but not any higher. I'd rather configure it as:

```ruby
gem.add_development_dependency      "rubocop", "~> 0.35.0"
```

blowmage:
The latest [rubocop release](https://github.com/bbatsov/rubocop/releases/tag/v0.36.0) added a large number of new rules, and several of them break on our code. I 100% support the move to pin the dependency on the previous release so we can take time to evaluate the new rules and make the changes needed, or disable them.

Issue #502, opened 2016-01-05
Resource type


quartzmo:
See [MonitoredResource](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/MonitoredResource).

Issue #501, opened 2016-01-05
Operation type


quartzmo:
See [LogEntryOperation](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry#LogEntryOperation).

Issue #500, opened 2016-01-05
HttpRequest type


quartzmo:
See [HttpRequest](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry#HttpRequest).

Issue #499, opened 2016-01-05
Severity type


quartzmo:
See [LogSeverity](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/LogEntry#LogSeverity).

PR #523, opened 2016-01-25
Add Logging Entries

### LINKED ISSUE #492: List Entries
Include support for the following:

- [ ] All
- [ ] Filter
- [ ] Order
- [ ] Pages

blowmage:
Add listing and writing entries.

Entry, Resource, HttpRequest, and Operation are implemented as simple classes, with only attributes being on the public API. They hold no connection to the logging service. Entry has a large surface area, and there are little affordances in the API to make populating it any easier. This will be addressed in future pull requests.

[closes #492, #493, #498, #499, #500, #501, #502]

### DIFF
--- a/lib/gcloud/logging/connection.rb
+++ b/lib/gcloud/logging/connection.rb
@@ -38,6 +38,39 @@ module Gcloud
         @logging = @client.discovered_api "logging", API_VERSION
       end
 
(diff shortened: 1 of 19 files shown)


quartzmo:
Phew! Nice job getting through all this.

PR #536, opened 2016-02-03
Set YARD to use markdown

blowmage:
This will allow sites such as rubydoc.info to display the docs correctly.


quartzmo:
Cool, thanks for this.

PR #541, opened 2016-02-07
Fix NameError Bigquery::Dataset::Access#validate_special_group

yhirano55:
I found undefined local variable `scope` in [bigquery/dataset/access.rb#L441](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/bigquery/dataset/access.rb#L441)

### DIFF
--- a/lib/gcloud/bigquery/dataset/access.rb
+++ b/lib/gcloud/bigquery/dataset/access.rb
@@ -438,7 +438,7 @@ module Gcloud
         def validate_special_group value
           good_value = GROUPS[value.to_s]
           return good_value unless good_value.nil?
-          scope
+          value
         end
 
         # @private


quartzmo:
@yhirano55 Good find!

@blowmage The parallel validation methods in this class (`validate_role`, `validate_scope`) raise an error if a mapped value is not found. Should this one do the same?

blowmage:
I'm not sure if that validation method should raise an error yet. I was hoping to dive deeper into this today and figure out what is the right behavior.

blowmage:
I don't think this method is intending to raise if the value is not be validated. It is intended to pass the value though to the API. Thanks for finding this @yhirano55!

blowmage:
This fix included in the [0.6.2 release](https://github.com/GoogleCloudPlatform/gcloud-ruby/releases/tag/v0.6.2).

PR #547, opened 2016-02-09
Add missing YARD tags to methods that yield

quartzmo:
These tags are used to add yield and yield params to the JSON docs.


blowmage:
Great job finding these. I know it was tedious.

PR #548, opened 2016-02-12
Update jsondoc to generate multiple JSON files

quartzmo:
This PR makes output of the jsondoc tool on the `gcloud-jsondoc` branch compatible with the current gcloud-common site JSON. It also changes its markdown library from Redcarpet to Kramdown, to keep up with the times and for consistency with the current YARD-generated docs.


blowmage:
Great work! This is a huge step towards getting the docs on the angular app!

PR #559, opened 2016-02-26
Fix Entity#update_exclude_indexes!

### LINKED ISSUE #537: Cloud datastore find entities exclude_from_indexes is set to true
If I retrieve an entity from a local gcd-emulator using Gcloud::Datastore with the find method as follows:
```ruby
require 'gcloud/datastore'
dataset = Gcloud.datastore

key = Gcloud::Datastore::Key.new('Person', 4697113673859072)
key.parent = Gcloud::Datastore::Key.new('AccountPerson', 5742310809665536)
entity = dataset.find(key)
```
the _exclude_indexes is set to true for all properties.

Example:
```ruby
@_exclude_indexes={"name"=>true}
```
If I change the value of a property on the entity and save it back to the datastore:
```ruby
entity['name'] = 'different'
dataset.save entity
```
I can no longer query for the entity as all properties have been set to unindexed.

quartzmo:
This PR adds breaking unit and acceptance tests for the correct conversion of `indexed` properties to `Entity#exclude_from_indexes?`. It fixes the broken tests by flipping the boolean value of `indexed`, both for individual values and list values.

Thank you @bmclean for reporting this issue and indicating how to fix it in #537.

[closes #537]

### DIFF
--- a/acceptance/datastore/datastore_test.rb
+++ b/acceptance/datastore/datastore_test.rb
@@ -56,6 +56,11 @@ describe "Datastore", :datastore do
 
     it "should save/find/delete with a key name" do
       post.key = Gcloud::Datastore::Key.new "Post", "post1"
+      post.exclude_from_indexes! "author", true
+      # Verify the index excludes are set properly
+      post.exclude_from_indexes?("title").must_equal false
+      post.exclude_from_indexes?("author").must_equal true
+
       dataset.save post
 
       refresh = dataset.find post.key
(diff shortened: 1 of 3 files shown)


bmclean:
You're welcome @quartzmo.

blowmage:
This fix included in the [0.6.2 release](https://github.com/GoogleCloudPlatform/gcloud-ruby/releases/tag/v0.6.2).

PR #560, opened 2016-02-26
Logging Acceptance Tests

blowmage:
Here are the acceptance tests for Logging. There is a permissions issue modifying sinks. I was able to modify sinks using my user account, but not a service account.

### DIFF
--- a/acceptance/logging/logging_test.rb
+++ b/acceptance/logging/logging_test.rb
@@ -1,0 +1,136 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "logging_helper"
+
+# This test is a ruby version of gcloud-node"s logging test.
+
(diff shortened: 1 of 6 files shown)


blowmage:
Looks like the permissions issue is a known issue, as gcloud-node states in their documentation. Only Sinks are affected. Metrics and Entries can be managed with a service account.

I'll leave the Sink tests skipped for now, until we can run them with a service account.

acceptance/logging/logging_test.rb
```diff
@@ -0,0 +1,172 @@
...
+      logging.metrics(max: 1).length.must_equal 1
+
+      metric.delete
+
+      logging.metric("#{prefix}-metric").must_be :nil?
+    end
+  end
+
+  describe "Entries" do
+    let(:proto_payload) { Google::Protobuf::Any.new type_url: "example.com/Greeter/SayHello",
```
quartzmo: Would it be better to use a more specific and realistic Protobuf type here? (Note: I'm not sure what that would be, just asking!)

blowmage: I don't think we need a more realistic protobuf object, any `Google::Protobuf::Any` object will do. If you look at what is returned from `Project#entries`, you'll see plenty of log entries that have protobuf payloads. AFAICT the Logging service just passes those values around, it doesn't care what they are.

quartzmo:
Would it be fairly easy to retrieve and verify the entries created with the Ruby Logger impl?

acceptance/logging/logging_test.rb
quartzmo: Ok

quartzmo:
Great to see how doing these tests led you to implement GRPCUtils.

blowmage:
Updated to verify the log entries are written. The check has a retry with an incremental backoff, so it should be more stable on the eventual nature of the cloud services.

quartzmo:
Great! Merging!

PR #564, opened 2016-02-26
Add Error subclasses

blowmage:
The gRPC project identifies 16 known error types. Currently, each service creates their own error classes and subclasses. We have an opportunity to consolidate all service errors to the main gRPC error subclasses. This PR adds the gRPC error subtypes and converts the logging service to use them.

What do you think of this approach?


blowmage:
Ping @quartzmo

blowmage:
Some examples, Pub/Sub defines a [`Gcloud::Pubsub::AlreadyExistsError`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Pubsub/AlreadyExistsError.html) and [`Gcloud::Pubsub::NotFoundError`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Pubsub/NotFoundError.html), but those could be replaced with the `Glcoud::AlreadyExistsError` and `Gcloud::NotFoundError` defined in this PR.

quartzmo:
I think this is a great idea.

blowmage:
Cool. I like this idea, so feel free to merge.

PR #569, opened 2016-03-01
Update Pub/Sub IAM

blowmage:
The build is failing due to a permissions change in getting and setting the IAM policy. This change adds some additional checks to the tests to ensure that the account has permissions to access and update the policy.

Correct the permissions values used in the code examples to use the correct permissions, and list the permissions values that can be checked on `test_permissions` documentation.


quartzmo:
Looks like everything has been straightened out! Great.

PR #572, opened 2016-03-02
Add location to Bigquery::Project#create_dataset

### LINKED ISSUE #571: Setting location when creating a dataset
Right now we can retrieve the location of a dataset but are not able to set it when we create a dataset.
Could be possible to add it as an option ?

quartzmo:
Although our current policy is not to include experimental Cloud Platform API features in gcloud-ruby, this was not our strict policy when we added support for BigQuery, and the ability to read `dataset.location` at that time, making it seem inconsistent to allow setting this property. Also, as @gramos74 [commented](https://github.com/GoogleCloudPlatform/gcloud-ruby/issues/571#issuecomment-191330515) today, adding the ability to set location will bring gcloud in line with other tools:

> In fact, choosing a location is not only available via API, it is also a option when you create manually a dataset in the BigQuery web console.

[closes #571]

### DIFF
--- a/lib/gcloud/bigquery/connection.rb
+++ b/lib/gcloud/bigquery/connection.rb
@@ -355,7 +355,8 @@ module Gcloud
           "friendlyName" => options[:name],
           "description" => options[:description],
           "defaultTableExpirationMs" => options[:expiration],
-          "access" => options[:access]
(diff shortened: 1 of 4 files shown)


blowmage:
:metal:

PR #573, opened 2016-03-03
Update rake ghpages jsondoc

quartzmo:
Updates Rake tasks for new gcloud-jsondoc behavior, most notably multiple JSON doc files.


blowmage:
Pulled this locally and it seems to work. :+1:

PR #579, opened 2016-03-10
Update jsondoc generator

quartzmo:
While the initial commits for this PR fix tests and an issue with default parameter values, I don't want to merge immediately as I anticipate more changes related to GoogleCloudPlatform/gcloud-common#102.


quartzmo:
@blowmage We can merge this once GoogleCloudPlatform/gcloud-common#115 is accepted.

lib/gcloud/jsondoc/markup_helper.rb
```diff
@@ -30,12 +30,20 @@ def linkify(name, title)
         code_obj = YARD::Registry.resolve object.namespace, name, true
         if code_obj.nil?
           name
+        elsif code_obj.type == :method
+          path = code_obj.namespace.path.gsub("::", "/").downcase
+          method_type = get_method_type code_obj
+          method_id = "#{code_obj.name}-#{method_type}"
```
blowmage: This is a purely cosmetic thing, but I don't love appending `-instance` and `-class` to the method ids. Could we change this so class methods were prepended by `.`, and instance methods were left unchanged? This should work since `.` isn't a legal character for a method, so it would still be unique.

```ruby
method_id = code_obj.name
method_id = ".#{method_id}" if method_type == "class"
```

That would give us URLs like `#/docs/master/gcloud?method=bigquery` and `#/docs/master/gcloud?method=.bigquery` instead of `#/docs/master/gcloud?method=bigquery-instance` and `#/docs/master/gcloud?method=bigquery-class`

quartzmo: I was trying to hold closely to prior art such as rDoc and YARD. Yard for example: [Array#count-instance_method](http://www.rubydoc.info/stdlib/core/Array#count-instance_method) and [Hash#try_convert-class_method](http://www.rubydoc.info/stdlib/core/Hash#try_convert-class_method). But most importantly, if we just use words, we never have to worry about which symbol to use (or not use). It's very simple and human-readable.

blowmage: Understood. Its just that this this approach is so far removed from RDoc and YARD that we have the opportunity to improve on it.

PR #581, opened 2016-03-14
Convert Pub/Sub to GRPC

blowmage:
Move Pub/Sub from using Google API Client to GRPC.

The only API changes in this PR are that the following error classes are removed in favor of the top-level error classes that were introduced in the Logging GRPC conversion:

* `Gcloud::Pubsub::Error` -> `Gcloud::Error`
* `Gcloud::Pubsub::AlreadyExistsError` -> `Gcloud::AlreadyExistsError`
* `Gcloud::Pubsub::NotFoundError` -> `Gcloud::NotFoundError`


blowmage:
All Pub/Sub calls are converted to GRPC, but there are a few IAM calls that remain on the Google API Client.

blowmage:
Not sure if we want to wait for the known GRPC issues to be resolved before merging this or not. What do you think @quartzmo?

quartzmo:
I think we should wait.

blowmage:
The Pub/Sub GRPC Conversion is complete. Ready to merge whenever you feel is appropriate.

quartzmo:
So awesome @blowmage! I will look at it tonight!

quartzmo:
I pulled this branch and ran acceptance and unit tests locally, and also did a fair amount of manual testing in console; found nothing wrong. I also searched the code for any lingering "connection", "gapi" or similiar... found none. I read through everything. Still, this is a big PR, I may have missed something, but for now, LGTM! Excellent work @blowmage! Merging!

arparker:
Hi -- If I am following correctly, all Gcloud gem versions starting with 0.7.0 use gRPC for PubSub.  Only, gRPC support on the PubSub side still appears to be in alpha (https://cloud.google.com/pubsub/reference/rpc/) and 'not recommended for production use'.

I was curious if I had that right, and/or if you had other knowledge of the state of support in Pub/Sub.

I am still on version 0.6.3, but making efforts to upgrade to the last version 0.10.0 to gain all recent benefits, including the speed gains of this change.  I just don't want to steer in a direction that takes me away from production-izing my code.

blowmage:
Hi @arparker, the direction we've gotten from those inside Google is to use gRPC where it is supported. I don't know about the Pub/Sub RPC, but I'd assume that was using the protobuf definitions on HTTP 1.1, where gRPC is using HTTP 2. There is a tremendous amount of energy going into supporting gRPC, and while there have been some issues it is getting better. 

I see no reason not to use the latest releases, but I can understand being cautious and staying on the old release as well. Up to you.

PR #582, opened 2016-03-14
Skip Search tests

blowmage:
The search acceptance tests are unstable and may not be supported soon.
Comment out the search acceptance tests for now, so setup and teardown
isn't attempted. Comment out cleanup as well.


quartzmo:
Sad that we have to do this!

PR #596, opened 2016-03-24
Add links to Service guides

blowmage:
Some links to the classes and methods documented in the top-level service guides were not converted to the YARD syntax. This should add those links back to the documentation.

### DIFF
--- a/OVERVIEW.md
+++ b/OVERVIEW.md
@@ -6,7 +6,7 @@ The `gcloud` library is installable through rubygems:
 $ gem install gcloud
 ```
 
-Gcloud aims to make authentication as simple as possible. Google Cloud requires a **Project ID** and **Service Account Credentials** to connect to the APIs. You can learn more about various options for connection on the [Authentication Guide](AUTHENTICATION).
+Gcloud aims to make authentication as simple as possible. Google Cloud requires a **Project ID** and **Service Account Credentials** to connect to the APIs. You can learn more about various options for connection on the <a ui-sref="docs.guides({ guideId: 'authentication' })" href="AUTHENTICATION">Authentication Guide</a>.
 
 # BigQuery
 
--- a/lib/gcloud.rb
+++ b/lib/gcloud.rb
@@ -35,6 +35,10 @@ module Gcloud
   ##
   # Creates a new object for connecting to Google Cloud.
   #
+  # For more information on connecting to Google Cloud see the <a
+  # ui-sref="docs.guides({ guideId: 'authentication' })"
+  # href="AUTHENTICATION">Authentication Guide</a>.
+  #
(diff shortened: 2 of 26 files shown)


quartzmo:
Fantastic, we no longer have any YARD warnings about unresolved links, which is so great. Thank you @blowmage! Amazing to see this all done in a day.

OVERVIEW.md
```diff
@@ -6,7 +6,7 @@ The `gcloud` library is installable through rubygems:
 $ gem install gcloud
 ```
 
-Gcloud aims to make authentication as simple as possible. Google Cloud requires a **Project ID** and **Service Account Credentials** to connect to the APIs. You can learn more about various options for connection on the [Authentication Guide](AUTHENTICATION).
+Gcloud aims to make authentication as simple as possible. Google Cloud requires a **Project ID** and **Service Account Credentials** to connect to the APIs. You can learn more about various options for connection on the <a ui-sref="docs.guides({ guideId: 'authentication' })" href="AUTHENTICATION">Authentication Guide</a>.
```
jgeewax: This makes me sad -- why do we need to have HTML markup in here? Shouldn't that be a last resort?

blowmage:
It is a last resort. From what I found the node docs don't link to the guides from their code documentation, so the mechanisms for creating a link to guides from doesn't exist yet. Because of the urgency of a Q1 release we felt it was better to get it working first, and then make better later. 

Our links to guides are always problematic. They only ever worked on our generated documentation, and were broken on sites like rubydoc.info.

quartzmo:
@jgeewax @blowmage @callmehiphop In order to avoid using `ui-sref` in our static markdown documents, would it work to add routing logic in the Angular app that would redirect the full URL `/#/docs/guides/authentication` to `/#/docs/{version}/guides/authentication`? The problem we have is setting the correct version in a completely static document.

quartzmo:
Essentially I'm suggesting: Hide the ugliness of the raw Angular directive usage in these links behind a pretty route.

blowmage:
@quartzmo The only guide we link to in the code comment documentation is the authentication guide. We could leave the links the way they were, and add the `ui-sref` attribute to the HTML when building the JSON docs.

callmehiphop:
Actually you can omit the version from the URL, when not supplied it redirects to the latest version, I *believe* it should inherit it when available - so I think it would be safe to completely omit the version and use a plain path.

blowmage:
But we should keep them on the version they were looking at when they click the link, right?

callmehiphop:
I haven't tested it, but I *think* so. If not, I don't think it would be very much work to add that logic in.

blowmage:
If the non-versioned URL will redirect and maintain the version then we'll hardcode the link to that and trust that the angular app will behave correctly.

PR #598, opened 2016-03-28
Use Prerelease GRPC

blowmage:
Verify that the prerelease version of GRPC fixes the issues identified with version. 0.13.0.


quartzmo:
@murgatroid99 We're excited to give `0.13.1.pre1` a try right now in our CI environment!  (@blowmage reports that it appears to work great in his development environment.)

PR #600, opened 2016-03-29
Convert Datastore to GRPC

blowmage:
WIP for the Datastore GRPC conversion. Waiting to get access to the v1beta3 service to run acceptance tests.


lib/gcloud/datastore/key.rb
```diff
@@ -226,41 +224,43 @@ def incomplete?
...
+          if pe_id_or_name.is_a? Integer
+            path_args[:id] = pe_id_or_name
+          else
+            path_args[:name] = pe_id_or_name
           end
-          k.partition_id = Proto.new_partition_id dataset_id, namespace
+          Google::Datastore::V1beta3::Key::PathElement.new(path_args)
+        end
+        grpc = Google::Datastore::V1beta3::Key.new(path: grpc_path)
+        if dataset_id || namespace
```
pcostell: The dataset_id is now just project_id everywhere. Is it possible to deprecate the dataset_id field and prefer a project_id field?

pcostell:
LGTM -- thanks for getting this done!

lib/gcloud/datastore/key.rb
blowmage: @pcostell Yes, absolutely. So far this PR has no breaking ruby API changes, and is only swapping out the underlying communication to the service. We have a few small changes to the ruby API we'd like to make, and renaming `dataset_id` on Key is one of them. Do you want us to include them in this PR or should we make them in a separate PR?

pcostell: Separate is fine!

lib/gcloud/datastore/dataset/query_results.rb
```diff
@@ -60,24 +60,21 @@ class QueryResults < DelegateClass(::Array)
...
         end
 
         ##
         # Convenience method for determining id the more_results value
         # is "MORE_RESULTS_AFTER_LIMIT"
         def more_after_limit?
-          more_results == Proto.to_more_results_string(
-            Proto::QueryResultBatch::MoreResultsType::MORE_RESULTS_AFTER_LIMIT)
+          more_results == :MORE_RESULTS_AFTER_LIMIT
         end
```
pcostell: There is one more enum value now:
more_results_after_cursor. This gets set if you do a query with an end_cursor

blowmage: Sorry I missed that. I've added a new helper method.

pcostell:
@blowmage are we able to get this working with a version of protobuf from master?

Datastore is ready for you to merge this into master. One caveat is that the emulator available in the cloud SDK doesn't support grpc. However, the gcd.sh tool has a grpc version which users can get from [here](storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip).

blowmage:
@pcostell asks:

> @blowmage are we able to get this working with a version of protobuf from master?

Yes. This branch does pass acceptance tests when using a version of google-protobuf from the master branch of their repository. Unfortunately, the JSON conventions on that version has changed, and our unit tests that use JSON to create protobuf objects don't pass. So we will need to rework some test data when we switch to the next version of google-protobuf.

> Datastore is ready for you to merge this into master.

I would love to accept this PR and get the branch merged into master, but the CI build will fail because of the [known issue](https://github.com/google/protobuf/issues/1204) on the current version of google-protobuf. I think we should wait for a new release of google-protobuf before we merge this.

> One caveat is that the emulator available in the cloud SDK doesn't support grpc. However, the gcd.sh tool has a grpc version which users can get from [here](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip).

Thanks! We'll pull that and start testing this locally.

pcostell:
Sounds good, definitely better to play it safe.

blowmage:
AFAIAC this PR is complete and ready to be evaluated for merging into the `datastore` feature branch. We are still dependent on a newer release of google-protobuf gem that will contain the fix for `oneof`, but I think this PR is as code complete as we can get.

blowmage:
I just realized that this pull request is merging into master. We should be merging into a topic branch for the Datastore v1beta3 milestone instead. I will close this PR and create a new one targeting the correct branch.

PR #619, opened 2016-04-12
Convert Datastore to gRPC

blowmage:
Convert Datastore to use gRPC and Google's Protobuf library instead of Faraday and Beefcake. Also upgrade to using v1beta3. There are no API changes in this conversion, except for the addition of GQL and Cursor.

We are blocked on an issue in Google's Protobuf library's support of `oneof` used in the datastore proto3 definitions. The issue is fixed in master, but has not yet been released. We need this to be released before this releasing a new version of gcloud-ruby.

This PR replaces #600 which was created to merge into master, and not the datastore topic branch.


quartzmo:
So great to see this finally moving toward master!

Issue #618, opened 2016-04-07
Support setting ReadOptions


quartzmo:
An example requested in #601 demonstrates setting eventual consistency in [ReadOptions](https://cloud.google.com/datastore/reference/rpc/google.datastore.v1beta3#google.datastore.v1beta3.ReadOptions). We do not currently expose `ReadOptions` in either `lookup` or `run_query`.

Should we support setting `ReadOptions` in gcloud-ruby?

/cc @pcostell @jgeewax

quartzmo:
@blowmage @pcostell @jgeewax I have finished all of the code example snippets requested in #601, except for `eventual_consistent_query`, which depends on this feature. Thoughts?

blowmage:
I would like to know this as well. 😄

quartzmo:
From the [example on cloud.google.com/datastore](https://cloud.google.com/datastore/docs/concepts/structuring_for_strong_consistency#datastore_read_policy), it appears that only gcloud-java and gcloud-golang support this currently.

pcostell:
We should support setting it (it's in the API after all!).

In the meantime though, it would be ok for the snippets to just be a comment that says "gcloud-ruby doesn't currently support setting read options."

quartzmo:
@pcostell Thanks, sounds good.

blowmage:
We can probably get this implemented before the next release.

blowmage:
So the way I read this we can implement this with minimal changes to the public API by modifying `Dataset#lookup` and `Dataset#run` to add a `consistency` optional argument that can be either `:strong` or `:eventual`. The ruby code will take care of setting the transaction since it manages that anyway. Thoughts?

blowmage:
```ruby
require "gcloud"

gcloud = Gcloud.new
dataset = gcloud.datastore
query = dataset.query("List").
  where("active", "=", true)
active_lists = dataset.run query, consistency: :eventual
```

pcostell:
Seems reasonable. One thing to note is that when you set the consistency
you can only go down (so you can make a strongly consistent (ancestor)
query eventually consistent, but you can't force an eventually consistent
query to be strongly consistent.

blowmage:
@pcostell Can you set eventual consistency if you don't have a transaction?

pcostell:
[...]

PR #620, opened 2016-04-12
Datastore changes

blowmage:
This PR contains the changes to the public API to reflect the differences in v1beta3. For the most part, these changes are additive and backwards compatible with previous releases.

The changes include:

- [x] Add `QueryResults#next?`, `QueryResults#next`, and `QueryResults#all`
- [x] Renaming `Key#dataset_id` to `Key#project`
- [x] Updating `Dataset#key` and `Dataset#entity` convenience methods to make creation of nested keys easier
- [x] Removal of `ApiError` (Breaking change)
- [x] Remove old Proto file
- [x] Blob support (#653)
- [x] Location support (Geo Point, Lat/Lon)
- [x] Support Date and DateTime when setting a Timestamp (#654)
- [x] Support ReadOptions (#618, #655)
- [x] Don't set project_id on PartitionId (#651)
- [x] Batch commits (#652)


vyper:
Connection don't work using [emulator](https://cloud.google.com/datastore/docs/tools/datastore-emulator).

I received this error and console stay freezed after error:

```
> dataset.save entity
E0428 17:56:05.215965000 123145303928832 resolve_address_posix.c:126] getaddrinfo: nodename nor servname provided, or not known
```

blowmage:
The current emulator doesn't support GRPC. You can download the updated emulator [here](https://storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip). This is still a WIP.

vyper:
Oh! Thanks @blowmage. I'm using your branch in tests. :grin:

blowmage:
You should probably build the google-protobuf gem from their master branch then. There is an unreleased fix coming from them that we are waiting on. It's been a thing...

blowmage:
I believe this PR is ready for consideration.

quartzmo:
@blowmage Is "Renaming Key#dataset_id to Key#project" a breaking change? If so, can you indicate in PR description for later inclusion in release notes?

blowmage:
It is not a breaking change. There is an alias named `dataset_id` so old code will continue to work.

quartzmo:
Time to get this into `datastore` branch! Merging! Thanks for this big bundle of additions @blowmage!

PR #622, opened 2016-04-14
Support Pub/Sub Emulator

### LINKED ISSUE #491: Support PUBSUB_EMULATOR_HOST environment variable
gcloud-ruby should support the ~~`PUBSUB_HOST`~~ `PUBSUB_EMULATOR_HOST` environment variable and connect to service specified in it. This will enable connecting to an emulator running locally, similar to how the Datastore emulator works.

See GoogleCloudPlatform/gcloud-common#48

blowmage:
Use the PUBSUB_EMULATOR_HOST environment variable to indicate the target service API is running an insecure emulator.

Add documentation for installing the emulator, and for configuring gcloud-ruby to use the emulator.

[closes #491]

### DIFF
--- a/lib/gcloud/pubsub.rb
+++ b/lib/gcloud/pubsub.rb
@@ -49,6 +49,11 @@ module Gcloud
   #
   def self.pubsub project = nil, keyfile = nil, scope: nil
     project ||= Gcloud::Pubsub::Project.default_project
+    if ENV["PUBSUB_EMULATOR_HOST"]
+      ps = Gcloud::Pubsub::Project.new project, :this_channel_is_insecure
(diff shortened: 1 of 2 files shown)


blowmage:
This is the simplest implementation I could come up with. I've tested it against the current Google Cloud SDK Pub/Sub Emulator. The conversion to GRPC made this significantly easier to implement, even though I did my best to make it as complicated as possible initially. 😄

The only caveat is that the current version of the Google Cloud SDK Pub/Sub Emulator doesn't support the IAM calls. But the ruby code should work if/when the emulator supports them.

quartzmo:
@blowmage Thank you for pushing this to the front of the queue!

PR #626, opened 2016-04-18
Add Translate Service

blowmage:
WIP implementation of the Translate service.


blowmage:
Closing this PR so we can use #627 which will merge into a topic branch for translate instead.

PR #631, opened 2016-04-22
Make Translation a delegate of String

blowmage:
Adding the method `Translate#to_str` so we can use a Translate instance object as a String might have been a bit too clever. It might be clearer in the documentation if the object is an actual delegate of String. This approach is already in use on all the `List` classes and `Gcloud::ResourceManager::Project::Updater`.

Thoughts?

### DIFF
--- a/lib/gcloud/translate/translation.rb
+++ b/lib/gcloud/translate/translation.rb
@@ -13,19 +13,13 @@
 # limitations under the License.
 
 
+require "delegate"
+
 module Gcloud
   module Translate
     ##
     # TODO
-    class Translation
-      ##
-      # The translated text.
-      #
-      # @return [String]
-      attr_reader :text
-      alias_method :to_s, :text
-      alias_method :to_str, :text
-
+    class Translation < DelegateClass(::String)
       ##
       # The that was translated.
       #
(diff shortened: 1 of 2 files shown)


quartzmo:
I think this solution is probably better than just implementing `to_str`, so I think we should merge this PR.

However, I also think that treating `Translation` as a string obscures the API a bit. Although the string behavior is a great convenience feature that we can demonstrate in some examples, I would prefer that our headline examples also explicitly call `Translation`'s attributes:

```ruby
translation = translate.translate "Hello world!", to: "la"
puts translation #=> Salve mundi!
translation.text #=> "Salve mundi!"
translation.from #=> "en"
translation.to #=> "la"
translation.origin #=> "Hello world!"
```

This suggestion is purely for docs changes and does not impact this PR.

blowmage:
If treating `Translation` as a string obscures the API then we shouldn't do it.

quartzmo:
It's also a very nice convenience.

blowmage:
So I'm still 50/50 on this PR. `DelegateClass` makes sense when you want to add values to an `Array`, but I'm not fully convinced it is useful for other primitive types like `String`. I think it is super convenient to use a `Translation` as a `String`, but we don't _need_ the other `String` methods on `Translation`. For example, I can see folks wanting to call `map` on a `List` way more often than call `upcase` on a Translation.

quartzmo:
I'm happy with the API behavior as it currently is.

PR #636, opened 2016-04-26
Remove Search

### LINKED ISSUE #634: Remove search API ?
Looks like we're phasing search into a different API surface eventually, but won't be continuing the private alpha as planned. Can we chop search out of gcloud-ruby?

blowmage:
The Search service is going to change, and the alpha is ending.

[closes #634]


quartzmo:
LGTM, @blowmage, thanks for the quick turnaround.

PR #641, opened 2016-04-28
Refresh access tokens sooner

### LINKED ISSUE #638: Implement auth token refresh
I'm not sure if this problem affects services other than Datastore. What I see is after 4 hrs (I think it is auth token ttl) after server start, I sometimes (not always) see `401 "Invalid Credentials"` responses.  My theory is that it is because of race condition here: https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/datastore/credentials.rb#L39

Probably check for expiration time should be more conservative, and consider token expired few seconds or minutes before actual date. 

Other option is to retry request at library level.

At application level it is hard for me to decide if I should retry the request, normally 4XX requests should not be repeated.

blowmage:
Be more proactive in refreshing access tokens to avoid a potential race condition where the access token expires before an API request is received. The [`expires_within?`](http://www.rubydoc.info/github/google/signet/Signet%2FOAuth2%2FClient%3Aexpires_within%3F) method returns `true` if the access token has expired or expires within the next _n_ seconds.

[fixes #638]

### DIFF
--- a/lib/gcloud/datastore/credentials.rb
+++ b/lib/gcloud/datastore/credentials.rb
@@ -36,7 +36,7 @@ module Gcloud
       # Sign OAuth 2.0 API calls.
       def sign_http_request request
         if @client
-          @client.fetch_access_token! if @client.expired?
+          @client.fetch_access_token! if @client.expires_within? 30
           @client.generate_authenticated_request request: request
         end
         request


quartzmo:
Worth a try for sure! Thanks @blowmage!

PR #632, opened 2016-04-25
Vision setup

blowmage:
Implementation of the Vision service. Still a work in progress.

PR #643, opened 2016-04-29
Vision setup

blowmage:
Implementation of the Vision service.


blowmage:
This is the updated PR for #632.

quartzmo:
This is a very solid foundation that I think we should merge immediately into the `vision` topic branch. I'm very impressed with the usability of this implementation. It was obviously a lot of work. Bravo, @blowmage!

PR #647, opened 2016-05-02
Fix Vision Error Handling

blowmage:
Update ApiError to use the error format returned by the Vision service.
Add the missing check and raise and error when the API call is not successful.


quartzmo:
Makes sense to me. Merging!

PR #649, opened 2016-05-02
Enhance Datastore TransactionError

### LINKED ISSUE #645: Datastore: when rolling back failed transaction exception may be lost if there is secondary exception in `rollback` rpc
If exception happens at line https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/datastore/dataset.rb#L287 it is poped up to user code, masking real exception in transaction block rescued one line higher.

My real world example is that `commit` fails with timeout error (why, this I will be investigating), and then this rescue code tries to unwind transaction with `rollback` and it fails as transaction duration of 30s is expired by that time. I can imagine more error causes in `rollback`.

I'm not sure what ideal implementation should be, but if Datastore automatically rolls back uncommitted transaction (which I think is implied by 30 seconds limit), then secondary exception in rollback can probably be just muted.

_Since you asked for more bug reports ;)_

blowmage:
It was possible to have errors on both commit and rollback. When this happened the rollback error would be thrown, with no way to report the commit error to the user. Add commit_error and rollback_error to TransactionError to communicate both error objects.

[closes #645]

### DIFF
--- a/lib/gcloud/datastore/dataset.rb
+++ b/lib/gcloud/datastore/dataset.rb
@@ -284,8 +284,14 @@ module Gcloud
           yield tx
           tx.commit
         rescue => e
-          tx.rollback
-          raise TransactionError.new("Transaction failed to commit.", e)
+          begin
+            tx.rollback
+          rescue => re
+            msg = "Transaction failed to commit and rollback."
+            raise TransactionError.new(msg, commit_error: e, rollback_error: re)
+          end
+          raise TransactionError.new("Transaction failed to commit.",
+                                     commit_error: e)
         end
       end
 
--- a/lib/gcloud/datastore/errors.rb
+++ b/lib/gcloud/datastore/errors.rb
@@ -66,12 +66,18 @@ module Gcloud
     class TransactionError < Gcloud::Datastore::Error
       ##
       # An error that occurred within the transaction. (optional)
-      attr_reader :inner
+      attr_reader :commit_error
+      alias_method :inner, :commit_error # backwards compatibility
+
+      ##
+      # An error that occurred within the transaction. (optional)
+      attr_reader :rollback_error
 
       # @private
-      def initialize message, inner = nil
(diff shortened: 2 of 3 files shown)


timanovsky:
Well, first I wouldn't wrap transactions inside `yield tx` at all, these are transactions in user code and for ease of handling would be good to expose original exception class. For `tx.commit` `tx.rollback` pair I thing wrapping them into single `TransactionError` is ok, but may be make ruby's `Exception.cause = commit_error` – again for simplicity of common case. I don't think many people need to handle details of `rollback` exception.

timanovsky:
That's leaves question how to handle exceptions in `yield tx` and `rollback` open though  :)

blowmage:
If you don't want to wrap transactions inside of a block (`yield tx`), then you can use the transaction object and make calls to `#commit` and `#rollback` on your own. You don't have to use the block syntax. Here is the example from the [Datastore documentation](https://googlecloudplatform.github.io/gcloud-ruby/#/docs/v0.8.0/gcloud/datastore).

```ruby
tx = dataset.transaction
begin
  if tx.find(user.key).nil?
    tx.save user
  end
  tx.commit
rescue
  tx.rollback
end
```

Glad you think having both `commit_error` and `rollback_error` is an acceptable solution. We chose to use `commit_error` and `rollback_error` so users can differentiate between the two. The design of the Error classes predates ruby 2.1, which is the first version to add `Exception#cause`. We need to support `TransactionError#inner` to maintain backwards compatibility with existing code. I do agree it would be nice to convert all `#inner` methods `#cause` in the future.

timanovsky:
Makes sense.

As of inner, I missed `alias_method :inner, :commit_error` line. Now seeing it I don't have other objections.

quartzmo:
I looked around for prior art on how to handle more than one nested error in Ruby, but couldn't find anything. I'm going to merge this solution. We have #650 to remind us to transition to `Exception#cause` as soon as we transition to Ruby >= 2.1.

PR #651, opened 2016-05-02
Fix run query with namespace.

toots:
Otherwise, I get this response from the API:

```
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "INVALID_ARGUMENT",
    "message": "app s~foo cannot access app foo's data"
   }
  ],
  "code": 400,
  "message": "app s~foo cannot access app foo's data"
 }
}
```

### DIFF
--- a/lib/gcloud/datastore/dataset.rb
+++ b/lib/gcloud/datastore/dataset.rb
@@ -447,7 +447,6 @@ module Gcloud
         return nil if namespace.nil?
         Proto::PartitionId.new.tap do |p|
           p.namespace = namespace
-          p.dataset_id = project
         end
       end
     end
--- a/test/gcloud/datastore/dataset_test.rb
+++ b/test/gcloud/datastore/dataset_test.rb
@@ -297,7 +297,6 @@ describe Gcloud::Datastore::Dataset do
 
 
   it "run_query will fulfill a query with a namespace" do
-    dataset.connection.expect :dataset_id, project
     dataset.connection.expect :run_query,
                               run_query_response,
                               [Gcloud::Datastore::Proto::Query, Gcloud::Datastore::Proto::PartitionId]


toots:
CLA signed now.

blowmage:
@toots Thanks for this PR!

@pcostell Can you weigh in on what the proper data for PartitionId is? Can you explain this error?

pcostell:
LGTM.

Caveat -- This could be indicative of a user setting the project ID to the application ID (e.g. `"s~<project-id>"` instead of just `"<project-id>"`). Note that in Datastore v1beta2 this sometimes worked, but in v1beta3 it will always fail.

When the project in the partition ID is unset, it gets populated from the request URL with the proper `project_id`. So unset here is always OK.

blowmage:
@pcostell Should we add the option to set the project on a query? We currently don't expose that as an option.

blowmage:
Is it possible to have the `project_id` in the URL and the `PartitionId` be different?

blowmage:
@toots How are you setting your `project_id` when you authenticate?

toots:
`Gcloud.new <project-id>, <path-to-credentials.json>`

Please note that the error also occurs if I make a direct API call through the page at: https://cloud.google.com/datastore/docs/apis/v1beta2/datasets/runQuery

blowmage:
@toots Can you comment on what @pcostell says about the difference between the application id and the project id? Does your string start with `"s~"`?

pcostell:
No, you shouldn't be able to specify a project-id for a query that is different than the request.

blowmage:
Great! Thank you for the clarification. :heart:

blowmage:
@toots I see no reason why we can't accept this PR. Please amend your commit to fix the test coverage so the build passes and we'll get it merged. Thanks!

blowmage:
@toots Can you squash those two commits into one commit?

blowmage:
@toots Thank you for your contribution!

toots:
Thanks for the quick response!

blowmage:
@toots This has been released in [0.8.1](https://rubygems.org/gems/gcloud/versions/0.8.1). Please give it a try. Thanks again for your contribution!

toots:
It's working great, thanks for the quick release

blowmage:
Very glad to hear it!

quartzmo:
@toots We just released [0.8.2](https://rubygems.org/gems/gcloud/versions/0.8.2) to fix a bug in 0.8.1 (see #667 for details.) Can you please update to 0.8.2 as soon as possible? Thank you.

toots:
Thanks for the update. Gem updated and working fine!

PR #653, opened 2016-05-03
Support Datastore Blob values

### LINKED ISSUE #644: Datastore finds of blob_value entities return nil properties
I have entities that are a JSON serialization stored as a blob in Datastore.
Using the dataset.find like this:

    key = dataset.key('Digest', '5740703418155118')
    entity = dataset.find(key)

Results in a retrieved proto results object passed to self.from_proto (within entity.rb) that looks like this (it is correct so far):

    <Gcloud::Datastore::Proto::Entity key:
      <Gcloud::Datastore::Proto::Key partition_id:
        <Gcloud::Datastore::Proto::PartitionId dataset_id: "s~xxxx-xxx">, 
        path_element: [<Gcloud::Datastore::Proto::Key::PathElement kind: "Digest", name: "5740703418155118">]>,
        property: [<Gcloud::Datastore::Proto::Property name: "values", value:
        <Gcloud::Datastore::Proto::Value blob_value: "{\"transaction\": {\"car\": 1993322165, \"version\": 154}}", indexed: false>>]>

However, after the proto results object is processed by Proto.from_proto_value it is turned into this (the values property is now nil):

    <Gcloud::Datastore::Entity:0x007fc7b3b9aa00 @properties=#   <Gcloud::Datastore::Properties:0x007fc7b3b9a9d8 @hash={"values"=>nil}>,
    @key=#<Gcloud::Datastore::Key:0x007fc7b3b98e80 @kind="AccountDigest",   @name="5740703418155118", @dataset_id="s~xxxx-xxx", @namespace=nil>,
    @_exclude_indexes={}>

The problem appears to be that Proto.from_proto_value doesn't support blob_property.

    Array(proto.property).each do |p|
      entity[p.name] = Proto.from_proto_value p.value
    end

Could we consider adding:

    elsif !proto_value.blob_value.nil?
      return proto_value.blob_value

to [Proto.from_proto_value](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/1c128baf4cf6903ab491d2c7652477556ab2923d/lib/gcloud/datastore/proto.rb#L50)?

blowmage:
Byte arrays are represented in Google API Client as base64 encoded strings. Allow IO-like objects to be saves as blobs, and blobs to be returned as StringIO objects.

[closes #644]

### DIFF
--- a/lib/gcloud/datastore/proto.rb
+++ b/lib/gcloud/datastore/proto.rb
@@ -15,6 +15,8 @@
 
 require "gcloud/proto/datastore_v1.pb"
 require "gcloud/datastore/errors"
+require "stringio"
+require "base64"
 
 module Gcloud
   module Datastore
@@ -53,6 +55,8 @@ module Gcloud
           return Array(proto_value.list_value).map do |item|
             from_proto_value item
           end
+        elsif !proto_value.blob_value.nil?
+          return StringIO.new(Base64.decode64(proto_value.blob_value))
         else
           nil
         end
@@ -82,6 +86,8 @@ module Gcloud
           v.string_value = value
         elsif Array === value
           v.list_value = value.map { |item| to_proto_value item }
+        elsif value.respond_to?(:read)
+          v.blob_value = Base64.encode64(value.read)
         else
           fail PropertyError, "A property of type #{value.class} is not supported."
         end
--- a/lib/gcloud/datastore/proto.rb
+++ b/lib/gcloud/datastore/proto.rb
@@ -88,7 +88,7 @@ module Gcloud
           v.list_value = value.map { |item| to_proto_value item }
         elsif value.respond_to?(:read) && value.respond_to?(:rewind)
           value.rewind
-          v.blob_value = Base64.encode64(value.read)
+          v.blob_value = Base64.strict_encode64(value.read)
         else
           fail PropertyError, "A property of type #{value.class} is not supported."
         end
--- a/lib/gcloud/datastore/proto.rb
+++ b/lib/gcloud/datastore/proto.rb
@@ -86,7 +86,8 @@ module Gcloud
           v.string_value = value
         elsif Array === value
           v.list_value = value.map { |item| to_proto_value item }
-        elsif value.respond_to?(:read)
+        elsif value.respond_to?(:read) && value.respond_to?(:rewind)
+          value.rewind
           v.blob_value = Base64.encode64(value.read)
         else
           fail PropertyError, "A property of type #{value.class} is not supported."
--- a/acceptance/datastore/datastore_test.rb
+++ b/acceptance/datastore/datastore_test.rb
@@ -177,7 +177,8 @@ describe "Datastore", :datastore do
 
       entity = dataset.find post.key
 
-      entity["avatar"].rewind
+      # Rewind not needed because the StringIO poistion is always at the beginning whe retrieved from Datastore.
+      # entity["avatar"].rewind
       post["avatar"].rewind
       entity["avatar"].size.must_equal post["avatar"].size
       entity["avatar"].read.must_equal post["avatar"].read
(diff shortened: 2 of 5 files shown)


blowmage:
@quartzmo Can you give this a look?

@bmclean Any input?

blowmage:
Basically, any IO object (IO, StringIO, Temple, etc...) in an entity's property will be stored as a blob value, and a blob property will be returned as a StringIO object.

blowmage:
@pcostell Any input on this?

lib/gcloud/datastore/proto.rb
```diff
@@ -82,6 +86,9 @@ def self.to_proto_value value
           v.string_value = value
         elsif Array === value
           v.list_value = value.map { |item| to_proto_value item }
+        elsif value.respond_to?(:read) && value.respond_to?(:rewind)
+          value.rewind
+          v.blob_value = Base64.encode64(value.read)
```
timanovsky: `encode64` seems to add linefeeds. Better use variant `strict_encode64` which does not

timanovsky:
Really looking forward this PR, as I need to write binary (compressed) data

lib/gcloud/datastore/proto.rb
blowmage: Decoding base64 will skip the line feeds in the encoded string, but it will save a few bytes over the wire. I'll make the change.

blowmage:
@timanovsky How do you feel about the requirement that blob fields need to be wrapped in an IO-like object when setting, and returned as a StringIO when getting? Is that an intuitive API?

timanovsky:
Looks fine to me, how else would you do that in ruby. People who have
string data in SttingIO will have a problem, but I don't see what can be
done other than documentation. My only concern is about rewind, but if it
is a common practice so be it

@timanovsky <https://github.com/timanovsky> How do you feel about the
requirement that blob fields need to be wrapped in an IO-like object when
setting, and returned as a StringIO when getting? Is that an intuitive API?

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/653#issuecomment-216433481>

blowmage:
Well, if you try to set a StringIO object into an entity property now it will raise a PropertyError with a message that "A property of type StringIO is not supported." So this change is purely additive.

pcostell:
I think whatever the appropriate ruby blob structure is fine. My only preference would be that the input and output types are the same -- I think it's dangerous when users can write one type but then read it later as a different type. It can lead to a weird API, for example (excuse the none ruby-esque pseudocode) :
```
e = Entity()
e.prop = MyType
k = ds.put(e)
e.prop is a MyType
e = ds.get(k)
e.prop is OtherType
```

blowmage:
I agree it would be best to maintain types, but I think the closest we can get is IO-like. Here are some examples of code we'd like to support:

```ruby
# loading a file from the local file system
entity["avatar"] = open "/avatars/blowmage.png"
# File, inherits IO

# loading a file from the web
require "open-uri"
entity["avatar"] = open "https://avatars.githubusercontent.com/blowmage"
# Tempfile, does not inherit IO

# loading a file from in memory
require "stringio"
entity["avatar"] = StringIO.new avatar_bytes
# StringIO, does not inherit IO
```

All of these objects are IO-ish, but they all have different ancestors.

When retrieving the data from Datastore the bytes are all in memory, and AFAICT the most recommended approach for in memory IO objects is StringIO. I think this approach is the way to go.

pcostell:
Fair enough -- I may then just add the requirement that:

```
entity['avatar'] = open(...) # Tempfile input
entity['avatar'] # returns a StringIO
```

We can accept a few different types, but they are immediately serialized to
a blob so you don't run into the different types issue.

blowmage:
That's a good idea. I had some issues with having to rewind the objects in the tests based on whether the object was read or not.

bmclean:
@blowmage This looks fantastic. Thanks!

quartzmo:
This approach seems great to me, way to go @blowmage! Merging!

PR #654, opened 2016-05-03
Add support for Date and DateTime

blowmage:
Allow Date and DateTime objects to be persisted in Entity properties.

### DIFF
--- a/lib/gcloud/datastore/proto.rb
+++ b/lib/gcloud/datastore/proto.rb
@@ -64,8 +64,8 @@ module Gcloud
 
       def self.to_proto_value value
         v = Gcloud::Datastore::Proto::Value.new
-        if Time === value
-          v.timestamp_microseconds_value = self.microseconds_from_time value
+        if value.respond_to?(:to_time)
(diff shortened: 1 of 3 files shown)


lib/gcloud/datastore/proto.rb
```diff
@@ -60,8 +60,8 @@ def self.from_proto_value proto_value
 
       def self.to_proto_value value
         v = Gcloud::Datastore::Proto::Value.new
-        if Time === value
-          v.timestamp_microseconds_value = self.microseconds_from_time value
+        if value.respond_to?(:to_time)
```
timanovsky: Be careful here. Rails adds `.to_time` to string class:
```
"2016-01-01".to_time
=> 2016-01-01 00:00:00 +0000
```

blowmage: Great catch. I'll move the `to_time` check to be after the check for String.

quartzmo:
@blowmage I like the duck type detection! But we owe @timanovsky a big thanks for posting the observation about Rails adding `.to_time` to string! Really appreciate your help @timanovsky. Nice solution @blowmage.

PR #657, opened 2016-05-03
Fix typo in consistency check

blowmage:
Yeah... That thing where your tests aren't as good as you thought...


quartzmo:
Got it. Merging.

PR #658, opened 2016-05-04
Add dependency on google-protobuf

blowmage:
Pin dependency on 3.0.0.alpha.5.0.5 to include the oneof fix.
Remove well known protobuf files, since they are provided by google-protobuf now.
Keep `lib/google/protobuf/descriptor.rb` since it is not provided by google-protobuf yet.


blowmage:
@quartzmo The issue with the yanked gem is resolved and this PR is ready to be reviewed.

quartzmo:
I can't wait to try this out!

Issue #662, opened 2016-05-04
Allow ImageContext to be set on analysis


blowmage:
The [ImageContext](https://cloud.google.com/vision/reference/rest/v1/images/annotate#ImageContext) structure includes a location (longitude/latitude) and language hints for text processing. But it may include more in the future. Support this data so can be provided to the annotate call.

My current thinking is that context can be added to an Image object, and the Image object can be sent to annotate. This means if you want to provide context you must first create an Image object.

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
analysis = vision.annotate filepath, faces: 10 # no ImageContext possible

image = vision.image filepath
image.context.location = { longitude: -122.0862462, latitude: 37.4220041 }
analysis = vision.annotate image, faces: 10 # ImageContext sent
```

I feel putting this data on an image object will keep the call to annotate as clean as possible, considering annotate already has so many optional arguments. So I feel being a little more verbose in the use of an Image object is warranted. Thoughts?

Issue #661, opened 2016-05-04
Set default value on Image analysis helper methods


blowmage:
Currently the Image analysis helper methods `faces`, `landmarks`, `logos`, and `labels` require a count. Calling those methods without providing an upper limit of the number of matches to find will raise an error.

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
image = vision.image filepath

analyses = image.faces #raises an error because count is not set
```

Change those helper methods to have a sensible default, so users can change the value if they want, but get sensible behavior if they don't.

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
image = vision.image filepath

analyses = image.faces

# same as calling
analyses = image.faces 10
```

Issue #660, opened 2016-05-04
Run full analysis with empty options


blowmage:
If annotate is run without any options, an empty analysis is returned. Instead, detect that no options were set and run all options, with sensible defaults.

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
analyses = vision.annotate img1

# same as:
analyses = vision.annotate img1, faces: 10, landmarks: 10, logos: 10, labels: 10,
                                 text: true, safe_search: true, properties: true
```

Issue #659, opened 2016-05-04
Batch mode annotations


blowmage:
Allow a batch-mode annotation where several images can be processed with different options.

Currently multiple images can be processed in a single annotation, but they all share the same options:

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
analyses = vision.annotate img1, img2, faces: 10, landmarks: 20, labels: 30
```

Allow the images to have different analysis options by using a block syntax:

```ruby
require "gcloud/pubsub"

vision = Gcloud.vision
analyses = vision.annotate do |a|
  a.annotate img1, faces: 10, landmarks: 20, labels: 30
  a.annotate img2, text: true, safe_search: true, properties: true
end
```

PR #663, opened 2016-05-04
Add additional functionality to Vision

blowmage:
Additional functionality for Vision service implementation.

- [x] Allow multiple images to be processed with different annotate options. (#659)
- [x] Run full analysis with empty options (#660)
- [x] Set default value on Image analysis helper methods (#661)
- [x] Allow ImageContext to be set on analysis (#662)


blowmage:
@quartzmo Can you take a lookout this functionality? I believe this brings the Vision implementation to feature completeness. Obviously missing lots of documentation though...

quartzmo:
@blowmage Pulling your branch and looking at it now!

blowmage:
The main tradeoff on setting ImageContext is that you have to create an Image object first, and update the context on the Image object, as opposed to adding additional values to the `annotate` call.

```ruby
image = vision.image filepath
image.context.location = { longitude: -122.0862462, latitude: 37.4220041 }
analysis = vision.annotate image, landmarks: 10
```

vs. 

```ruby
analysis = vision.annotate filepath, landmarks: 10,
  location: { longitude: -122.0862462, latitude: 37.4220041 }
```

And...

```ruby
image = vision.image filepath
image.context.languages = ["en", "es"]
analysis = vision.annotate image, text: true
```

vs. 

```ruby
analysis = vision.annotate image, text: true, language_hints: ["en", "es"]
```

Setting those values on the Image object is a little more work for users, but `annotate` already has so many optional arguments and I didn't feel comfortable adding even more. Also, keeping it on the Image object allows the ImageContext API to expand if it needs to add more options beyond location and image hints. Feels a bit more OO and comfortable to me. Thoughts?

blowmage:
Another reason to only allow ImageContext to be configured on an Image object is that it avoids adding additional optional arguments not just on `Project#annotate`, but `Image#texts`, and `Image#text`, and `Image#landmarks`, and `Image#landmark`, etc...

quartzmo:
@blowmage, I agree with you: I prefer the first examples in which the `Image` parameter collects the metadata  configuration (`latLongRect ` and `languageHints` currently, possibly more in the future, as you say.) I consider calling `annotate` with just a filepath is a convenience/shortcut, and it is OK that it doesn't support this more advanced configuration.

However, now that I've looked at this, I have a couple of observations:

1. Can we accept the context args in `Project#image`, to avoid forcing the need to call nested setters on the Image/Context after construction? (Saves a line...)

  ```ruby
  image = vision.image filepath, language_hints: ["en", "es"]
  analysis = vision.annotate image, text: true
  ```

2. Shouldn't the location parameter be a type that enforces the contract of [`latLongRect `](https://cloud.google.com/vision/reference/rest/v1/images/annotate#LatLongRect), meaning a pair of lat/long values, `minLatLng ` and `maxLatLng `?

blowmage:
> 1. Can we accept the context args in `Project#image`

We _could_, but I'd like to not. The reason for this is kinda involved. When we've had lots of options on object constructors its because we are calling an API method to create the resource. [`Gcloud::Storage::Project#create_bucket`](https://googlecloudplatform.github.io/gcloud-ruby/#/docs/master/gcloud/storage/project?method=create_bucket-instance) is a great example of this. There are so many optional arguments in the constructor that are also methods on the Bucket object, because those values are needed during the resource creation. But for vision's Image we don't create a resource on the service, and so I'd rather we didn't. It keeps the API surface areas smaller.

> 2. Shouldn't the location parameter be a type that enforces the contract of `latLongRect`?

That is correct. 100% my bad. I'll update the code to not be busted.

quartzmo:
@blowmage Sounds good. On 1., I prefer immutable objects and directly calling constructors, but that's just my personal preference, I don't think it's an important issue here.

quartzmo:
How do you feel about renaming `Color#rgb` to `Color#hex`, since most people use "hex" to refer to this encoding of the rgb values. (See endless "rgb to hex" converters online, also [wikipedia](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet).)

blowmage:
Perhaps I'm showing my age, but that six character representation of a color is `rgb`, although I'm fine aliasing it as `hex`. The Color object already has `red`, `green`, and `blue` integer values, so the object is inherently the decimal format already. The only conversion we make is to the hex code, and I think its fine to name it `rgb`. But that's just me...

I also pronounce "gif" with a hard G.

quartzmo:
Area looks good to me, merging!

PR #665, opened 2016-05-04
Fix Commit#save docs

quartzmo:


blowmage:
Thanks for catching this!

PR #667, opened 2016-05-04
Fix bug where blob values were double Base64 encoded

### LINKED ISSUE #666: Blob values created as ndb JsonProperty return as Base64 decoded
We have entities in our datastore that were previously created using the ndb [JsonProperty](https://cloud.google.com/appengine/docs/python/ndb/entity-property-reference). For those entities the Gcloud::Datastore::Proto::Value `blob_value` doesn't seem to be returned as Base64 encoded. For example, adding this debugging to Proto.[from_proto_value](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/f944f6c4cd8cbff6f2a6432999ac36288ad5a89f/lib/gcloud/datastore/proto.rb#L58):

    elsif !proto_value.blob_value.nil?
      puts "*" * 40
      puts proto_value.blob_value
      return StringIO.new(Base64.decode64(proto_value.blob_value))  

returns this:

    ****************************************
    {"transaction": {"crc": 2325900754, "entity_version": 1225}, ...

Which is very strange, as when I look at the value for those entities using the cloud.google.com DataStore Viewer they are encoded. Why do these properties return already decoded?

Should we test that contents of `blob_value` are in fact base64 encoded before attempting to base64 decode?

Should we instead raise an exception if `blob_value` isn't Base64 encoded?

blowmage:
Force blob values to ASCII-8BIT instead of Base64 encoding. The Google API Client already performs Base64 encoding, so gcloud was double encoding and storing encoded bytes in Datastore.

[fixes #666]

### DIFF
--- a/lib/gcloud/datastore/properties.rb
+++ b/lib/gcloud/datastore/properties.rb
@@ -101,11 +101,9 @@ module Gcloud
         elsif value.respond_to?(:to_time)
           return value
         elsif value.respond_to?(:read) && value.respond_to?(:rewind)
-          # shortcut creating a StringIO if it already is one.
-          return value if value.is_a? StringIO
           # Always convert an IO object to a StringIO when storing.
           value.rewind
-          return StringIO.new(value.read)
+          return StringIO.new(value.read.force_encoding("ASCII-8BIT"))
         elsif defined?(BigDecimal) && BigDecimal === value
           return value
         end
--- a/lib/gcloud/datastore/proto.rb
+++ b/lib/gcloud/datastore/proto.rb
@@ -16,7 +16,6 @@
 require "gcloud/proto/datastore_v1.pb"
 require "gcloud/datastore/errors"
 require "stringio"
-require "base64"
 
 module Gcloud
   module Datastore
@@ -56,7 +55,7 @@ module Gcloud
             from_proto_value item
           end
         elsif !proto_value.blob_value.nil?
-          return StringIO.new(Base64.decode64(proto_value.blob_value))
+          return StringIO.new(proto_value.blob_value.force_encoding("ASCII-8BIT"))
         else
           nil
         end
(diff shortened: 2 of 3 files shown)


blowmage:
@bmclean Are you able to try this code and see if you can read the existing JsonProperty/blob values?

bmclean:
@blowmage Sure, I will try it right away...

blowmage:
@bmclean Thanks! AFAICT this plays nice with blob values already in Datastore, but I'd love to get confirmation from you.

Once we are sure we have fixed this issue we'll ship 0.8.2 and then yank 0.8.1.

bmclean:
@blowmage Yes, I am able to read the existing values and the JSON parses fine with ActiveSupport::JSON.decode.

bmclean:
@blowmage Does this impact images coming back as blob values though?

blowmage:
@bmclean This impacts any blob value saved with 0.8.1. Anything saved with 0.8.1 is base64 encoded in Datastore.

We are just waiting on a green build to make the release.

blowmage:
We have [acceptance tests](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/acceptance/datastore/datastore_test.rb#L196-L224) that save an image as a blob in an entity's property, save the entity, retrieve the entity, write the image data to a temp file, and then check that the temp file contents are the same as the original image file contents.

bmclean:
Perfect!

blowmage:
@bmclean We just released 0.8.2. Can you try that gem version and let us know if it fixes this problem?

bmclean:
Yes it does. Thanks @blowmage!

blowmage:
@bmclean Fantastic. We just yanked 0.8.1. Thank you so much for your help with this!

PR #669, opened 2016-05-05
Force encoding of binary data

blowmage:
This matches the newer behavior as of 0.8.2. Binary strings should be binary.


quartzmo:
I pulled this branch and acceptance tests pass. @blowmage reports that manual testing retrieving blob values from 0.8.2 inserts works. Merging.

PR #671, opened 2016-05-05
Remove sign_http_request

blowmage:
The GRPC code takes care of this, and this method is no longer needed.


quartzmo:
Right on, way to keep things lean and mean.

PR #675, opened 2016-05-06
Remove calls to instance_eval

blowmage:
This is a performance optimization. Use `instance_variable_set` instead of `instance_eval`. We don't want to bust the method cache every time we create an object. Thanks tenderlove!


quartzmo:
LGTM, merging.

PR #678, opened 2016-05-09
Support Datastore Emulator

### LINKED ISSUE #604: Test GRPC Datastore against updated emulator
The updated emulator can be downloaded [here](https://storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip).

### LINKED ISSUE #615: Document usage with Datastore emulator
As [reported](https://github.com/GoogleCloudPlatform/gcloud-ruby/issues/610#issuecomment-205424753) by @samkelly:

> would also be great if the documentation specifically talks about there even being an emulator and how to connect to it from ruby -- was sort of buried

Currently the only documentation is in [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/CONTRIBUTING.md#local-datastore-devserver). It uses the earlier name "Datastore Devserver", and does not link to [usage instructions](https://cloud.google.com/datastore/docs/tools/datastore-emulator) for the emulator.

### LINKED ISSUE #676: Datastore library requires valid Google authentication token even when using datastore emulator
If i set JSON credentials `gcloud.datastore()` call tries to authenticate against google (and fails if machine clock is out of sync for example)

If I don't set JSON credentials code falls to https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/credentials.rb#L81 and fails.

This is undesired behavior in both cases. In integration tests we WebMock and do not expect real HTTP requests to live service. I expect all requests only hit emulator.

blowmage:
- [x] Support `DATASTORE_EMULATOR_HOST` environment variable
- [x] Skip authentication when connecting to emulator
- [x] Document emulator usage

[closes #676, closes #615, closes #604]

### DIFF
--- a/lib/gcloud/datastore.rb
+++ b/lib/gcloud/datastore.rb
@@ -520,6 +520,49 @@ module Gcloud
   # end
   # ```
   #
+  # ## The Datastore Emulator
+  #
+  # As of this release, the Datastore emulator that is part of the gcloud SDK is
+  # no longer compatible with gcloud-ruby. This is because the gcloud SDK's
+  # Datastore emulator does not yet support gRPC as a transport layer.
+  #
+  # A gRPC-compatible emulator is available until the gcloud SDK Datastore
+  # emulator supports gRPC. To use it you must [download the gRPC
+  # emulator](https://storage.googleapis.com/gcd/tools/gcd-grpc-1.0.0.zip) and
+  # use the `gcd.sh` script.
+  #
+  # When you run the gRPC emulator you will see a message similar to the
+  # following printed:
+  #
+  # ```
+  # If you are using a library that supports the DATASTORE_EMULATOR_HOST
+  # environment variable, run:
+  #
+  # export DATASTORE_EMULATOR_HOST=localhost:8978
+  # ```
+  #
+  # Now you can connect to the emulator using the `DATASTORE_EMULATOR_HOST`
+  # environment variable:
+  #
+  # ```ruby
+  # require "gcloud"
+  #
+  # # Make Datastore use the emulator
+  # ENV["DATASTORE_EMULATOR_HOST"] = "localhost:8978"
+  #
+  # gcloud = Gcloud.new "emulator-project-id"
+  # datastore = gcloud.datastore
+  #
+  # task = datastore.entity "Task", "emulatorTask" do |t|
+  #   t["type"] = "Testing"
+  #   t["done"] = false
+  #   t["priority"] = 5
+  #   t["description"] = "Use Datastore Emulator"
+  # end
+  #
+  # datastore.save task
+  # ```
+  #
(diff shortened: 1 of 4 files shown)


lib/gcloud/datastore.rb
```diff
@@ -515,6 +520,49 @@ def self.datastore project = nil, keyfile = nil, scope: nil
   # end
   # ```
   #
+  # ## The Datastore Emulator
+  #
+  # As of this release, the Datastore emulator that is part of the gcloud SDK is
+  # not compatible with gcloud-ruby. This is because the gcloud SDK's Datastore
```
quartzmo: "not compatible" -> "no longer compatible"

lib/gcloud/datastore.rb
```diff
@@ -515,6 +520,49 @@ def self.datastore project = nil, keyfile = nil, scope: nil
...
+  # export DATASTORE_EMULATOR_HOST=localhost:8978
+  # ```
+  #
+  # Now you can connect to the emulator using the `DATASTORE_EMULATOR_HOST`
+  # environment variable:
+  #
+  # ```ruby
+  # require "gcloud"
+  #
+  # Make Datastore use the emulator
```
quartzmo: Add `#` to comment out these two lines?

lib/gcloud/datastore.rb
blowmage: Updated.

lib/gcloud/datastore.rb
blowmage: Updated.

quartzmo:
Thanks for taking care of this @blowmage!

timanovsky:
Does it effectively replace `DATASTORE_HOST` with `DATASTORE_EMULATOR_HOST`? It breaks backward compatibility :(

blowmage:
@timanovsky 

> Does it effectively replace `DATASTORE_HOST` with `DATASTORE_EMULATOR_HOST`?

Yes. There has been lots of activity on how these environment variables should be named. For instance, Pub/Sub went from `PUBSUB_HOST` to `PUBSUB_HOSTPORT` to `PUBSUB_EMULATOR_HOST`. But the gcloud SDK team seems to have settled on `<service>_EMULATOR_HOST`.

> It breaks backward compatibility :(

Yes. This next release breaks compatibility in some minor ways. For example, `Gcloud::Datastore::ApiError` is removed, and most errors that Datastore will raise now inherit directly from `Gcloud::Error`, and not `Gcloud::Datastore::Error`. Most API changes maintain backwards-compatibility however.

If gcloud-ruby was already 1.0 this next release would be a major upgrade to 2.0. But since gcloud-ruby is still pre-1.0 we can make this change [without violating semver](http://semver.org/#spec-item-4).

PR #679, opened 2016-05-09
Update Vision docs

quartzmo:
- [x] faces
- [x] landmarks
- [x] logos
- [x] labels
- [x] text
- [x] safe_search
- [x] properties
- [x] Guide/Overview/Readme

### DIFF
--- a/lib/gcloud/vision.rb
+++ b/lib/gcloud/vision.rb
@@ -78,25 +78,15 @@ module Gcloud
   #
   # ## Creating images
   #
-  # Currently, the Google Cloud Vision API supports the following image types:
-  #
-  # * JPEG
-  # * PNG8
-  # * PNG24
-  # * GIF
-  # * Animated GIF (first frame only)
-  # * BMP
-  # * WEBP
-  # * RAW
-  # * ICO
-  #
-  # Image files sent to the Google Cloud Vision API should not exceed 4 MB.
+  # The Cloud Vision API supports a variety of image file formats, including
+  # JPEG, PNG8, PNG24, Animated GIF (first frame only), and RAW. See [Best
+  # Practices - Image Types](https://cloud.google.com/vision/docs/image-best-practices#image_types)
+  # for the list of formats. Be aware that Cloud Vision sets upper limits on
+  # file size as well as on the total combined size of all images in a request.
   # Reducing your file size can significantly improve throughput; however, be
-  # careful not to reduce image quality in the process. If you are batching
-  # images and sending them in one request, also note that the Vision API
-  # imposes an 8 MB per request limit. (These limits may change, see [Best
-  # Practices](https://cloud.google.com/vision/docs/image-best-practices) for
-  # the most current information.)
+  # careful not to reduce image quality in the process. See [Best Practices -
+  # Image Sizing](https://cloud.google.com/vision/docs/image-best-practices#image_sizing)
+  # for current file size limits.
   #
   # Use {Vision::Project#image} to create images for the Cloud Vision service:
   #
@@ -185,7 +175,8 @@ module Gcloud
   # ```
   #
   # It is even possible to configure different features for multiple images in
-  # the same request:
+  # a single call using a block. The following example results in a single
+  # request to the Cloud Vision API:
   #
   # ```ruby
   # require "gcloud"
--- a/lib/gcloud/vision.rb
+++ b/lib/gcloud/vision.rb
@@ -41,7 +41,11 @@ module Gcloud
   #
   #   gcloud = Gcloud.new
   #   vision = gcloud.vision
-  #   # ...
+  #
+  #   image = vision.image "./acceptance/data/landmark.jpg"
+  #
+  #   landmark = image.landmark
+  #   landmark.description #=> "Mount Rushmore"
   #
   def self.vision project = nil, keyfile = nil, scope: nil
     project ||= Gcloud::Vision::Project.default_project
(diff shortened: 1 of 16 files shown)


blowmage:
I can pull the commits and review locally.

blowmage:
@quartzmo Can you load the doc changes on to your github pages branch so we can review them there as well?

quartzmo:
@blowmage The demo is up on my fork. It's a work in progress.

I've been working from the Analysis types upward, so `Project` and `Image` are still bare. You should be able to navigate to everything I've done from [`Analysis`](http://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud/vision/analysis).

quartzmo:
@blowmage @jgeewax I have updated the [demo](https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud/vision) for this PR, which is ready for review.

blowmage:
That is a ton of work. Great job!

jgeewax:
Definitely. A few notes, some of which may turn into separate issues...

- can we call the result `annotation` and `annotations` instead of `analysis` and `analyses`? seems weird to call "annotate" and get back something that isnt "annotations"

- we have the "faces" and "labels" fields, but these are really the "max" values, right? if so, what's stopping us from calling this `max_faces: 10`?

- the 4mb/8mb limits as well as the list of image types seem like they're going to be tough to maintain. how do we make sure those numbers don't get stale?

- additionally, if we keep the list of image types, can we actually show the bullet points in the UL ? 

- the last example shows doing a bunch of requests in a block -- does that do things in batch? ie, one http request? if so, it might be useful to clarify that in the text.

quartzmo:
@jgeewax:

Thanks for the great suggestions, I’ll address these issues today with docs changes (no need for GitHub issues):

> - the 4mb/8mb limits as well as the list of image types seem like they're going to be tough to maintain. how do we make sure those numbers don't get stale?

> - additionally, if we keep the list of image types, can we actually show the bullet points in the UL ? 

> - the last example shows doing a bunch of requests in a block -- does that do things in batch? ie, one http request? if so, it might be useful to clarify that in the text.

@blowmage:

What is your feeling about:

> - can we call the result `annotation` and `annotations` instead of `analysis` and `analyses`? seems weird to call "annotate" and get back something that isnt "annotations"

> - we have the "faces" and "labels" fields, but these are really the "max" values, right? if so, what's stopping us from calling this `max_faces: 10`?

blowmage:
> * can we call the result `annotation` and `annotations` instead of `analysis` and `analyses`? seems weird to call "annotate" and get back something that isnt "annotations"

I am in favor of this change. I think we should create a separate issue for this, since the change will be renaming the classes as well as updating the documentation.

> * we have the "faces" and "labels" fields, but these are really the "max" values, right? if so, what's stopping us from calling this `max_faces: 10`?

I am not in favor of changing those optional argument names. I typically prefer shorter names to longer as long as it doesn't introduce confusion. I also like the symmetry between `vision.annotate image, faces: 20` and `image.faces 20`.

quartzmo:
@blowmage Sounds good, I agree on both. Please create the issue if you want, otherwise I'll do it in a little bit. Regarding shorter vs longer names, my personal preference is for longer names, but the style of gcloud-ruby is very consistent in using shorter names, so I agree that we should sustain this style.

jgeewax:
> I am not in favor of changing those optional argument names. I typically prefer shorter names to longer as long as it doesn't introduce confusion. I also like the symmetry between vision.annotate image, faces: 20 and image.faces 20.

My concern is that I *was* confused by the parameter name. "Do I have to tell you exactly how many faces are in the picture? If I knew that, I wouldn't need this API... So what exactly is that field for... ?"

blowmage:
I don't think the problem is the argument name, I think it is the way the documentation is presented. IMO this is one area where the Angular doc app fails for ruby code. When using RDoc or YARD one of the first things you will see is:

```ruby
annotate(*images, faces: 0, landmarks: 0, logos: 0, labels: 0, text: false, safe_search: false, properties: false) { |a| ... }
```

When I see this I understand that I should provide a different value for the analysis I want run. But in the angular app we don't ever show what the default value for those options are. We will have to add that information to the argument/parameter documentation. It means we have to parse a big wall of text to get that most basic information.

blowmage:
> * additionally, if we keep the list of image types, can we actually show the bullet points in the UL

The image types are already in a bulleted list. For some reason the style is set to not display the bullets.

lib/gcloud/vision.rb
```diff
@@ -41,7 +41,11 @@ module Gcloud
   #
   #   gcloud = Gcloud.new
   #   vision = gcloud.vision
-  #   # ...
+  #
+  #   image = vision.image "./acceptance/data/landmark.jpg"
```
blowmage: All the code examples are using the images files in the acceptance directory. This directory is not included in the released gem, and these paths won't resolve. Because of this I think these paths should use `/path/to/<image>` instead.

We use `/path/to` in the [Storage code examples](https://quartzmo.github.io/gcloud-ruby/#/docs/master/gcloud/storage) as well.

quartzmo:
@blowmage OK, I'll change to `/path/to` for the image files.

lib/gcloud/vision/image.rb
```diff
@@ -42,15 +74,17 @@ def initialize
         @context = Context.new
       end
 
-      # Determines if the Image has content.
+      ##
```
blowmage: I'd be okay making this @private.

lib/gcloud/vision/image.rb
```diff
@@ -42,15 +74,17 @@ def initialize
...
+      # Whether the Image has content.
       #
       # @see {#url?}
       #
       def content?
         !@io.nil?
       end
 
-      # Determines if the Image is a URL.
+      ##
```
blowmage: `@private`?

lib/gcloud/vision/image.rb
```diff
@@ -60,79 +94,168 @@ def url?
...
+      #
+      # @return [String]
+      #
       def content
         @content ||= Base64.encode64 @io.read
       end
 
       ##
-      # The URL of the image
+      # The URL of the image.
```
blowmage: `@private`?

lib/gcloud/vision/image.rb
```diff
@@ -60,79 +94,168 @@ def url?
 
       ##
```
blowmage: `@private`?

lib/gcloud/vision/image.rb
```diff
@@ -60,79 +94,168 @@ def url?
...
 
+      ##
+      # Performs the `FACE_DETECTION` feature on the image.
+      #
+      # @see https://cloud.google.com/vision/docs/pricing Cloud Vision Pricing
+      #
+      # @param [Integer] count the maximum number of results
+      #
+      # @return [Array<Analysis::Face>] the results of face detection
+      #
```
blowmage: Should all of these analysis methods have code examples?

lib/gcloud/vision/project.rb
```diff
@@ -76,11 +83,146 @@ def self.default_project
...
+      # For multiple images in a request, the Vision API currently imposes an 8
+      # MB per request limit.
+      #
+      # @see https://cloud.google.com/vision/docs/requests-and-responses Cloud
+      #   Vision API Requests and Responses
+      # @see https://cloud.google.com/vision/reference/rest/v1/images/annotate#AnnotateImageRequest
+      #   AnnotateImageRequest
+      # @see https://cloud.google.com/vision/docs/pricing Cloud Vision Pricing
+      #
+      # @param [Image, Object] images The image or images to annotate. This can
```
blowmage: The casing of the arguments are inconsistent. Some start with an upper case letter, some with a lower case letter.

lib/gcloud/vision/project.rb
```diff
@@ -76,11 +83,146 @@ def self.default_project
...
+      #
+      # Note that an object in Google Cloud Storage is a single entity;
+      # permissions affect only that object. "Directory permissions" do not
+      # exist (though default bucket permissions do exist). Make sure the code
+      # which performs your request has access to that image.
+      #
+      # @see https://cloud.google.com/vision/docs/image-best-practices Best
+      #   Practices
+      #
+      # @param [Gcloud::Storage::File, IO, StringIO, String] source a Google
```
blowmage: I think the most common types given here will be String (local path and gs URL), followed by File, followed by other IO types (Tempfile, StringIO) that respond to `#rewind` and `#read`, and then Gcloud::Storage::File. Would it make sense to order the types by likelihood?

lib/gcloud/vision/analysis.rb
```diff
@@ -199,142 +255,192 @@ def labels
...
+      #   image = vision.image "./acceptance/data/face.jpg"
+      #
+      #   analysis = vision.annotate image, properties: true
       #   analysis.properties? #=> true
       #
       def properties?
         !properties.nil?
       end
 
+      ##
```
blowmage: Does it make sense to streamline the docs a bit by marking the `to_h` and `to_hash` methods as `@private`?

lib/gcloud/vision/project.rb
quartzmo: These should all be upper case. Sorry, I was trying to adopt YARD conventions, but that should be all or nothing, so I'll stick to how we've done it before, with capitalized full sentences.

quartzmo:
@blowmage This is ready to go, unless you want me to squash commits.

Issue #601, opened 2016-03-29
Update Datastore examples with snippets for cloud.google.com


quartzmo:
For consistency with GoogleCloudPlatform/gcloud-common#105, update all Datastore API doc examples to use the Ruby snippet examples for cloud.google.com/datastore. Hopefully, all examples can be adapted from [gcloud-node samples](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/datastore/concepts.js).

Since the snippets have not yet been created for Ruby, I am thinking that we should write them first as API docs, so that they can be included in the gcloud-ruby update for Datastore v1beta3 (PR #600). This means merging the examples in gcloud-ruby before they are accepted into [GoogleCloudPlatform/ruby-docs-samples](https://github.com/GoogleCloudPlatform/ruby-docs-samples), at the risk of some rework if changes are requested later.

/cc @pcostell

quartzmo:
@pcostell Status update: With PR #621, I added all but a couple of the requested snippets, including the [GQL examples](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/datastore/lib/gcloud/datastore/gql_query.rb#L32-L34), to gcloud-ruby API docs. I am now preparing a PR to add them as tested code to [GoogleCloudPlatform/ruby-docs-samples](https://github.com/GoogleCloudPlatform/ruby-docs-samples).

pcostell:
Thanks for the update @quartzmo.

/cc @WalterHub

PR #680, opened 2016-05-10
Add Datastore insert and update support

### LINKED ISSUE #648: Allow Datastore calls to insert vs. update
The original design goal for Datastore was to design a humane API at the expense of not exposing all features. This is why `Datastore#save` calls `upsert`, because it has the lowest conceptual overhead. But now the thinking has changed and gcloud-ruby should expose all Datastore features. To that end gcloud-ruby should add `Dataset#insert` and `Dataset#update`, so that users can decide of they want the API call to fail if the entities exist or not. IMO `Datastore#save` should be aliased as `Datastore#upsert` for consistency.

blowmage:
Allow end users to specify if entities should be inserted or updated.
Allow the Datastore API to raise an error when the entities can't be saved or updated.

[closes #648]

### DIFF
--- a/lib/gcloud/datastore/commit.rb
+++ b/lib/gcloud/datastore/commit.rb
@@ -34,14 +34,15 @@ module Gcloud
       # @private Create a new Commit object.
       def initialize
         @shared_upserts = []
+        @shared_inserts = []
+        @shared_updates = []
         @shared_deletes = []
       end
 
       ##
       # Saves entities to the Datastore.
       #
-      # @param [Entity, Key] entities_or_keys One or more Entity or Key
-      #   objects to remove.
+      # @param [Entity] entities One or more Entity objects to save.
       #
       # @example
       #   gcloud = Gcloud.new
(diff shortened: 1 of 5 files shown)


lib/gcloud/datastore/commit.rb
```diff
@@ -51,10 +52,47 @@ def initialize
...
+      ##
+      # Updates entities to the Datastore.
+      #
+      # @param [Entity] entities One or more Entity objects to update.
+      #
+      # @example
+      #   gcloud = Gcloud.new
+      #   dataset = gcloud.datastore
+      #   dataset.commit do |c|
+      #     c.insert task1, task2
```
jgeewax: c.update* ?

blowmage: Oops. Good catch. Updated.

blowmage: (The code example for `Commit#update` was showing `insert` instead of `update`, but this comment was made on `Commit#insert`, which was already correct.)

quartzmo:
Thanks a lot for adding these methods before we release. I will also update the Datastore snippets (#601) for cloud.google.com.

PR #685, opened 2016-05-11
Ensure NullValue values are stored properly

blowmage:
When an Entity property is set to `nil`, we want the Value to be `NULL_VALUE`.
Otherwise the value will be skipped and not persisted.
`NULL_VALUE` is a valid value and should be persisted across saves.


quartzmo:
@blowmage Glad you could figure this out!

PR #692, opened 2016-05-12
Fixing references to logging instead of datastore.

bmclean:


bmclean:
Signed it with secondary email as well.

blowmage:
D'oh! My shame as a copy/paste programmer was discovered!

Thanks for the fix @bmclean!

PR #693, opened 2016-05-12
Update QueryResults#all docs and code example

blowmage:
The original docs were take from Logging and not adapted to Datastore.
Copy/Paste FTW!

[refs #692]


quartzmo:
@blowmage Thanks for catching these lingering copy/paste docs errors.

@bmclean Thanks for the original catch.

PR #696, opened 2016-05-18
Vision changes

blowmage:
Implement some requested changes to the existing Vision implementation. Including renaming Analysis to Annotation and adding support for default max results when giving a truthy value to `#annotate`.

```ruby
require "gcloud"
gcloud = Gcloud.new

vision = gcloud.vision

annotation = vision.annotate "path/to/faces.jpg", faces: true
# As opposed to having to pass in the numeric value
# annotation = vision.annotate "path/to/faces.jpg", faces: 10
```


quartzmo:
I think changing the syntax from `faces: 10` to `faces: true` is a huge improvement in usability. It considerably lowers the friction to trying out the Vision service. For bulk (production) users, setting global configuration is also better. Win-win in my opinion.

PR #700, opened 2016-05-19
Pin google-protobuf dependency

blowmage:
We didn't expect "~> 3.0.0.alpha.5.0.5" to match "3.0.0.alpha.6.0.0".
Make the dependency explicit on "3.0.0.alpha.5.0.5.1" as a result.


quartzmo:
The reasons we need to pin the google-protobuf dependency (and not depend on `3.0.0.alpha.6.0.0`) are provided in google/protobuf#1563 and google/protobuf#1564.

Issue #594, opened 2016-03-23
Add IAM convenience methods to Pub/Sub Project


blowmage:
The current Pub/Sub API has convenience methods to get/set a Topic or Subscription's policy, but not for a Project. With the conversion to GRPC we can add convenience methods to Project as well.

blowmage:
After discussing this further we decided not to go in this direction.

Issue #588, opened 2016-03-18
Pub/Sub Get/Set/Test Project IAM


blowmage:
It is possible to work with IAM policies and permissions for topics and subscriptions, but not the project. For instance, there is no way to check for permissions for creating a topic, since that is done on the project. We should allow _all_ roles to be get/set and tested on a Pub/Sub project.

quartzmo:
@blowmage I implemented this using gRPC. But when I try to get the Policy for the current project, I get this error:

```sh
irb(main):002:0> policy = pubsub.policy
Gcloud::InvalidArgumentError: 3:This service cannot be used to manage the permissions of resources of type project. Please find the appropriate service to do so. See https://cloud.google.com/iam/docs/#supported_cloud_platform_services.
	from /Users/quartzmo/code/google/codez/gcloud-ruby/lib/gcloud/pubsub/project.rb:517:in `rescue in policy'
	from /Users/quartzmo/code/google/codez/gcloud-ruby/lib/gcloud/pubsub/project.rb:512:in `policy'
	from (irb):2
```

blowmage:
Right. The Pub/Sub service allows only Topic and Subscription to get/set an IAM Policy, and testing permissions. That is why the current Pub/Sub implementation does not have policy implemented on Project. Users would have to use either the IAM service, or Cloud Resource Manager, to manage a project's policy. I was hoping that the move to gRPC would allow these calls to be made on the IAM service, but this turned out not to be the case.

PR #716, opened 2016-05-27
Add Pub/Sub Policy

### LINKED ISSUE #587: Create IAM Policy DSL
Both Pub/Sub and Resource Manager have IAM calls to get and set a Policy as part of their API. The API is very simple, either returning a hash or setting a hash. We should offer a DSL for modifying the Policy in place, rather than entirely replacing it. Similar to how permissions are modified in Storage.

### LINKED ISSUE #593: Protect against overwriting IAM Policies
`SetIamPolicy` will override the resource's policy if setting without an etag value. This is a fundamentally dangerous operation. Make this much safer by providing a mechanism to get, modify, and set a policy in one operation. Perhaps similar to how DNS `Zone#update` works when called with a block.

quartzmo:
This PR adds an object, Policy, that holds the `etag` value for use in updates.

[closes #587, closes #593]

### DIFF
--- a/lib/gcloud/pubsub/policy.rb
+++ b/lib/gcloud/pubsub/policy.rb
@@ -1,0 +1,101 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+require "gcloud/errors"
+
+module Gcloud
+  module Pubsub
+    ##
+    # # Policy
+    #
+    # Represents a Cloud IAM Policy for the Pub/Sub service.
+    #
+    # A common pattern for updating a resource's metadata, such as its Policy,
+    # is to read the current data from the service, update the data locally, and
+    # then send the modified data for writing. This pattern may result in a
+    # conflict if two or more processes attempt the sequence simultaneously. IAM
+    # solves this problem with the {Gcloud::Pubsub::Policy#etag} property, which
+    # is used to verify whether the policy has changed since the last request.
+    # When you make a request to with an `etag` value, Cloud IAM compares the
+    # `etag` value in the request with the existing `etag` value associated with
+    # the policy. It writes the policy only if the `etag` values match.
+    #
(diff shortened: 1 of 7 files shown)


lib/gcloud/pubsub/policy.rb
```diff
@@ -0,0 +1,98 @@
+# Copyright 2015 Google Inc. All rights reserved.
```
blowmage: 2016

quartzmo: Thanks, good catch!

lib/gcloud/pubsub/topic.rb
```diff
@@ -299,85 +300,77 @@ def publish data = nil, attributes = {}
...
-      #       "role" => "roles/viewer",
-      #       "members" => ["serviceAccount:your-service-account"]
-      #     }]
-      #   }
       #   topic = pubsub.topic "my-topic"
-      #   topic.policy = viewer_policy
+      #
+      #   policy = topic.policy # API call
+      #
+      #   policy.roles["roles/owner"] << "user:owner@example.com"
```
blowmage: I think it would be better to use a top-level policy method to make changes like this.

```ruby
policy.add "roles/owner", "user:owner@example.com"
```

quartzmo:
@blowmage I think this is ready. I don't think #588 or #594 are actionable (see error posted on #588).

blowmage:
I don't think this is ready. To close #587 and #593 I would like to see the policy object yielded to the method's block so users can make minor changes in a declarative way. This should also use the latest etag to ensure that only the changes made in the block are made.

```ruby
require "gcloud"

gcloud = Gcloud.new
pubsub = gcloud.pubsub
topic = pubsub.topic "my-topic"

policy = topic.policy do |p|
  p.add "roles/owner", "user:newowner@example.com"
end
```

I also think while we are making these changes in Pub/Sub we should also make the same changes in Resource Manager. Use the same Policy API in both services.

quartzmo:
I'm adding the block to `#policy` now.

quartzmo:
@blowmage I'll do the Resource Manager change in a separate PR--right on the heels of this one.

blowmage:
Some additional thoughts after taking another look. I'd like to see a way to inspect all the roles on a policy. Like `Hash#keys`, so perhaps `Policy#roles`? I'd also like to see the block syntax as the main use-case. So the `#policy` code examples should show the block syntax first, and I'd like to see all Policy code examples use the block syntax.

Here are some thoughts based on my biases: I would also prefer `Profile#[]` to `Profile#role`, but that's debatable. I'm a little suspicious of allowing users to modify the members array directly. I'd rather the returned array was frozen so they had to use the mutator methods (`add`/`remove`) instead. But again that's just a preference. Lastly I'd prefer to see the Policy initializer not take any arguments, since it isn't intended to be created by users, similar to most other classes.

blowmage:
Gonna accept this now and we can make some of these tweaks after.

PR #719, opened 2016-06-01
Update Storage Verifier

blowmage:
Avoid reading very large files into memory when calculating digests. Fix CRC32c errors message.


blowmage:
I've run the digest verifier against a 6GB file and the ruby process pegs a single core but doesn't grow memory usage above 50KB.

quartzmo:
Thanks for the quick turnaround on this, @blowmage!

PR #720, opened 2016-06-02
Add BigQuery standard SQL support

yancya:
Add argument of use_legacy_sql for #query and #query_job .
Can use BigQuery's [standard SQL](https://cloud.google.com/bigquery/sql-reference/) if use_legacy_sql is false(default true).

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -578,10 +578,12 @@ module Google
         # @!group Data
         #
         def query_job query, priority: "INTERACTIVE", cache: true, table: nil,
-                      create: nil, write: nil, large_results: nil, flatten: nil
+                      create: nil, write: nil, large_results: nil, flatten: nil,
+                      use_legacy_sql: true
           options = { priority: priority, cache: cache, table: table,
                       create: create, write: write,
-                      large_results: large_results, flatten: flatten }
+                      large_results: large_results, flatten: flatten,
+                      use_legacy_sql: use_legacy_sql }
           options[:dataset] ||= self
           ensure_service!
           gapi = service.query_job query, options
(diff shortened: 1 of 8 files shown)


yancya:
I signed it!

quartzmo:
@yancya Thank you for your PR! 

The [`useLegacySql`](https://cloud.google.com/bigquery/docs/reference/v2/jobs/query#request-body) option is currently marked `[Experimental]`:

> [Experimental] Specifies whether to use BigQuery's legacy SQL dialect for this query. The default value is true. If set to false, the query will use BigQuery's standard SQL: https://cloud.google.com/bigquery/sql-reference/. When useLegacySql is set to false, the values of allowLargeResults and flattenResults are ignored; query will be run as if allowLargeResults is true and flattenResults is false.

Our current policy is to **not** add support for `[Experimental]` options to gcloud-ruby, but if you need this feature, can you perhaps explain why, and make a strong case for adding it?

yancya:
> if you need this feature, can you perhaps explain why, and make a strong case for adding it?

I am transplanting my analytics system on BigQuery from PostgreSQL.  I want to use standard SQL on BigQuery. Because PostgreSQL's SQL is like the standard SQL. 
But, I can wait [Experimental] to be released. Because I can select using forked version in the meantime.

> Our current policy is to not add support for [Experimental] options to gcloud-ruby,

I understood it. I don't have to be merged necessarily right now.

Thank you.

quartzmo:
@yancya Thank you for the additional detail.

@blowmage @jgeewax Any thoughts on this?

blowmage:
If only we had the option to add experimental features via some sort of plugin mechanism...

quartzmo:
@blowmage For simple values like strings and booleans, such as this flag, couldn't that be implemented just by being flexible and passing along any and all options to the service?

yancya:
@quartzmo Hi. The [useLegacySql](The useLegacySql option is currently marked [Experimental]) option is not currently marked [Experimental] now.

quartzmo:
In that case, I don't see any reason why we can't add this feature.

@blowmage PTAL

blowmage:
@yancya, would you be willing to add the unit and acceptance tests for this new parameter, and documenting it, as @quartzmo mentioned?

PR #721, opened 2016-06-03
Add Backoff to all google-api-client requests 

### LINKED ISSUE #714: Backoff for all (or selected) storage methods
As per [0] we're getting rate limited on e.g. file copy call etc. It seems to make sense to have backoff on all methods, instead of just bucket ones? At least updates are documented to be rate limited [1] but its probably safe to apply universally?

[0] https://stackoverflow.com/questions/37431917/gcloud-ruby-backoff-on-copying-files-and-other-operations/37439948#37439948
[1] https://cloud.google.com/storage/docs/best-practices#buckets-object

quartzmo:
Since we apply backoff to all gRPC requests, and also because a lack of backoff retry logic has been raised as an issue with Storage (#714), apply this behavior to all google-api-client requests.

This PR also introduces a breaking change by removing the `retries` option from Storage `Project#create_bucket` and `Bucket#delete`. The number of retries should now be configured directly on `Backoff`.

[closes #714]

### DIFF
--- a/lib/gcloud/backoff.rb
+++ b/lib/gcloud/backoff.rb
@@ -33,26 +33,26 @@ module Gcloud
       ##
       # The number of times a retriable API call should be retried.
       #
-      # The default value is 3.
+      # The default value is `3`.
       attr_accessor :retries
 
       ##
       # The GRPC Status Codes that should be retried.
       #
-      # The default values are 14.
+      # The default values are `14`.
       attr_accessor :grpc_codes
 
       ##
       # The HTTP Status Codes that should be retried.
       #
-      # The default values are 500 and 503.
+      # The default values are `500` and `503`.
       attr_accessor :http_codes
 
       ##
       # The Google API error reasons that should be retried.
       #
-      # The default values are rateLimitExceeded and
-      # userRateLimitExceeded.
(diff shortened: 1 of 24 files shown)


blowmage:
:+1:

quartzmo:
@blowmage How do you feel about [a single unit test for backoff](https://github.com/quartzmo/gcloud-ruby/blob/c33cefb295048b349f4f29eecac1dd0f05c7a172/test/gcloud/bigquery/backoff_test.rb) per service? Any suggestions?

blowmage:
I'm in favor of _not_ doubling or tripling the number of tests per service in the name of proper test coverage. 😏

blowmage:
I think one test per service is fine. I don't think the backoff code is much at risk of being unintentionally broken.

quartzmo:
@blowmage Any suggestions for [this discussion of Backoff](https://github.com/GoogleCloudPlatform/gcloud-ruby/pull/721/commits/805ebbf54c97f70616dcbabe8125b912078423d1) at the end of the BigQuery guide? Once it's good I'll adapt it to the other guides.

blowmage:
Awesome work!

Issue #711, opened 2016-05-23
Per-result cursor support (v1beta3)


blowmage:
Datastore v1beta3 adds a [cursor for every result](https://cloud.google.com/datastore/reference/rest/v1beta3/EntityResult#FIELDS.cursor) returned. We don't currently expose this, and only show the cursor for the last result on the page. The QueryResults object should provide cursors for every result.

```ruby
tasks = datastore.run query
tasks.size #=> 3

# iterate through results and cursors
tasks.each_with_cursor do |task, cursor|
  # ..
end
```

timanovsky:
Is it related to #690 ?

blowmage:
Kinda. I'm working on both.

blowmage:
Resolved in the 0.11.0 release.

Issue #707, opened 2016-05-20
Pagination seems a bit ... confusing (specifically Datastore)


jgeewax:
Doc: https://googlecloudplatform.github.io/gcloud-ruby/#/docs/v0.10.0/gcloud/datastore#paginating-records

This discussion has also been happening over at https://github.com/GoogleCloudPlatform/gcloud-python/issues/895

Basic gist is that we have a few different scenarios when "listing things":

1. Give me all the things
  - Kind of link infiniscroll
  - I don't care about how many API calls are made under the hood
  - I don't care about how many items there are

1. Give me all the things, up to N
  - Same as above, but stop at N things

1. Give me all the things, but no more than N API calls
  - Same as (1) except limit this to 100 API calls so I don't pay out the ass...

1. Give me page X of things, where each page is exactly N items (unless I'm out of items)
  - I start with no token, which gives me "page 1"
  - I get a token with that response, which I can pass in later to get "the next page"
  - The only reason why I should get < N items in a page is if there are no more to give me

---

Some examples of what I'd want... (assumes we start with a Datastore `query` object):

```ruby
# Give me all the things forever
results = dataset.run query
results.each do |entity|
  puts "Got entity #{entity}"
end

# And in another list method....
topics = pubsub.topics
topics.each do |topic|
  puts "Got topic: #{topic.name}"
end

# Give me up to 100 things
first_100_topics = pubsub.topics limit: 100
topics.each do |topic|
  puts "Got topic: #{topic.name}"
end

# Give me all the things up to 100 API calls
topics = pubsub.topics max_api_calls: 100  # Maybe a different name?

# Give me paginated things.
first_page_of_topics = pubsub.topics page_size: 10
first_page_of_topics.each do |topic|
  puts "Got topic: #{topic.name}"
end

# Give me the next page of 10
next_page_of_topics = pubsub.topics page_size: 10, page_token: first_page_of_topics.next_page_token
first_page_of_topics.each do |topic|
  puts "Got topic: #{topic.name}"
end
```

quartzmo:
@jgeewax I edited the docs link in your description to point to `googlecloudplatform.github.io` (was `quartzmo` fork.) Thanks for all these suggestions! I'll keep an eye on GoogleCloudPlatform/gcloud-python#895.

blowmage:
I think we can get most of the way there with lazy enumerables. See #690. The only thing that won't do is limit the number of total API calls.

[...]

Issue #690, opened 2016-05-12
Datastore feature request: implement Enumerator interface in QueryResults 


timanovsky:
Something like `QueryResults.to_enum` should return enumerator with lazy page loading contrary to `.all` eager loading.

blowmage:
I 100% agree. I had a spike that implemented this many moons ago, but it was before gcloud-ruby dropped 1.9 support. Now that gcloud-ruby only supports 2.0+ we can safely add this. When we do we will add this to all result objects, across all services.

Also, the method `QueryResults#to_enum` already exists. The object is a delegate to an array, and you can already enumerate over the page of results. So the future enumerator method will need to be named something new.

blowmage:
Resolved in the 0.11.0 release.

PR #724, opened 2016-06-07
Add Enumerators to List classes

blowmage:
This adds Enumerators to the many List classes in gcloud-ruby. This will enable more rubyish pagination options as described in #690 and #707. It also adds support for result cursors as described in #711. Lots of test coverage and documentation because that is how we roll.


quartzmo:
Awesome, awesome feature!

Let the bikeshedding begin: How do you feel about `max_requests` instead of `max_api_calls`? ("API calls" to me is a bit informal and loose. It could mean local calls, or any level of interface. The specific concern here is the number of HTTP requests.)

blowmage:
It should come as no surprise to you that I consider `max_api_requests` to be too long and overly explicit, but that was the name that JJ used in his sample code. 

I don't know that `max_requests` is much of an improvement. I'd prefer `max` or `requests` myself. I did not look to see what other libs use for this type of feature. I bet having some prior art will be useful in deciding.

blowmage:
What about `limit`?

blowmage:
[Rack::Throttle](https://github.com/bendiken/rack-throttle) uses `max`.

blowmage:
Most other libs refer to this concept as `limit`.

quartzmo:
I think `limit` is too commonly used for paginating records ([`limit` and `offset`](https://www.postgresql.org/docs/8.3/static/queries-limit.html)).

quartzmo:
I'm with @jgeewax's example then: If you change `max_api_calls` to `max_api_requests`, it I think it would be more precise. But like I said, just a bikeshed...

blowmage:
The more I think of this the more I think `limit` is semantically correct. What do you think of `api_limit` as a compromise?

quartzmo:
`request_limit` ? The problem I have with "api" is that it's not used correctly (precisely). The definition of API is broad and can be local.

blowmage:
Well, I think it's better than `max_api_calls`.

lib/gcloud/storage/bucket/list.rb
```diff
@@ -51,6 +51,67 @@ def next
...
         ##
+        # Retrieves all buckets by repeatedly loading {#next} until {#next?}
+        # returns `false`. Calls the given block once for each bucket, which is
+        # passed as the parameter.
+        #
+        # An Enumerator is returned if no block is given.
+        #
+        # This method may make several API calls until all buckets are
+        # retrieved. Be sure to use as narrow a search criteria as possible.
+        # Please use with caution.
```
quartzmo: Need `@param` doc for `max_api_calls`?

lib/gcloud/storage/bucket/list.rb
```diff
@@ -51,6 +51,67 @@ def next
...
+          max_api_calls = max_api_calls.to_i if max_api_calls
+          unless block_given?
+            return enum_for(:all, max_api_calls: max_api_calls)
+          end
+          results = self
+          loop do
+            results.each { |r| yield r }
+            if max_api_calls
+              max_api_calls -= 1
+              break if max_api_calls < 0
```
quartzmo: Because this `break` is after the `yield`, this corner case is possible:

```ruby
irb(main):015:0> storage.buckets.all(max_api_calls: 0).each {|x| puts x.name};nil
my-bucket123321
=> nil
```

quartzmo:
Yay! 

Moving on, I wonder if the block usage of `#all` increases the mental overhead of the API? I wonder if new users (familiar with Ruby` Enumerable` but not with gcloud) might be better served by just being guided to use `Enumerable`? 

Meaning: `storage.buckets.all.each {|x| puts x.name}` rather than `storage.buckets.all {|x| puts x.name}`.

As usual, my biases are: 1) Keep the API surface area small. 2) Leverage core APIs. 3) Optimize for reducing new users' mental overhead, not expert users' keystrokes.

blowmage:
Added new commits to rename `max_api_calls` to `request_limit`. Also added lots of param documentation, and a couple missing tests.

lib/gcloud/storage/bucket/list.rb
blowmage: Just now saw this comment. The missing `@param` and `@yieldparam` docs in the latest commits.

blowmage:
> I wonder if new users (familiar with Ruby `Enumerable` but not with `gcloud`) might be better served by just being guided to use `Enumerable`?

> Meaning: `storage.buckets.all.each {|x| puts x.name}` rather than `storage.buckets.all {|x| puts x.name}`.

I actually started out with that approach, but the Datastore cursor requirements made me change my mind. For Datastore we added `each_with_cursor`, which sits along side Enumerable's `each` and `each_with_index`. I don't prefer to call `results.each_with_cusor.each { |res, cur| ... }` as opposed to `results.each_with_cusor { |res, cur| ... }`, but you certainly can. Similarly, I don't prefer `buckets.all.each {|x| puts x.name}` to `buckets.all {|x| puts x.name}`, even though both work.

But you know how much of a minimalist I am. And that's, like, just my opinion man.

quartzmo:
Wow, you're right,  `results.each_with_cusor.each { |res, cur| ... }`  is pretty rough on the eyes! Ok, let's leave it as-is!

blowmage:
Crazy thing is that this might actually work: `results.each_with_cusor.each_with_index { |res, cur, idx| ... }`

quartzmo:
I made a comment earlier that I think got overwritten:

Because `yield` occurs before the `break` if `request_limit` is satisfied, I believe you get one iteration more than you specify:

```ruby
irb(main):017:0> storage.buckets.all(request_limit: 0) {|x| puts x.name};nil
my-bucket123321
=> nil
```

lib/gcloud/storage/bucket/list.rb
blowmage: The code is the way it is so the limit count works properly. If you specify no additional calls (`all(max_api_calls: 0)`) the enumerator should not make any additional calls. The initial api call made in `storage.buckets` remains, as it was made before accessing the `all` enumerator.

blowmage: `storage.buckets.all(max_api_calls: 0)` behaves the same as `storage.buckets.all(max_api_calls: -99)`, where no additional api calls are made.

quartzmo: OK, a little funky due to the initial call for sure, but I see no way around it. Thanks for explaining!

blowmage:
The initial yield happens on the initial list object. If the list's `next?` returns false the enumerator will never make another call, but the list should still yield all the initial results before ever making that check.

lib/gcloud/datastore/dataset/query_results.rb
```diff
@@ -110,38 +110,43 @@ def no_more?
...
-        # Whether there are more results available for subsequent API calls.
-        # Tests the value of {#more_results}.
+        # Whether there are more results available.
+        #
+        # @example
+        #   require "gcloud"
+        #
+        #   gcloud = Gcloud.new
+        #   datastore = gcloud.datastore
+        #   query = datastore.query "Tasks"
```
quartzmo: The examples for http://cloud.google.com/datastore consistently use `"Task"`.

quartzmo:
@blowmage Incredible job on the docs, they are super clean. You even managed to correct some old copy/paste errors. Fantastic work.

I've read through everything, and except for the two little docs issues above (line notes), I'm ready to merge.

lib/gcloud/datastore/dataset/query_results.rb
blowmage: Updated.

lib/gcloud/bigquery/job/list.rb
```diff
@@ -33,22 +33,145 @@ class List < DelegateClass(::Array)
...
+        #     puts job.state
+        #   end
+        #
+        # @example Using the enumerator by not passing a block:
+        #   require "gcloud"
+        #
+        #   gcloud = Gcloud.new
+        #   bigquery = gcloud.bigquery
+        #
+        #   all_names = bigquery.jobs.all.map do |job|
```
blowmage: Changed.

blowmage:
K, I've addressed the two doc issues you caught, and a third for good measure.

quartzmo:
There are still a few `"Tasks"` in the Datastore examples that need to change to `"Task"`, but we can follow up with a small docs PR. Merging. Fantastic new addition to gcloud-ruby!!

PR #726, opened 2016-06-08
Update Pub/Sub publish

blowmage:
A couple of changes to Pub/Sub. The first addresses a bug when publishing multi-byte strings. The second is an enhancement allowing users to publish message data using an IO-ish object like File, StringIO, Tempfile, etc.


quartzmo:
Great to have this bug fixed! Thanks @blowmage.

Issue #718, opened 2016-05-31
downloading large file: failed to allocate memory (NoMemoryError)


Aguasvivas22:
Simple script to download file:
```
require 'gcloud'

gcloud = Gcloud.new "my-bucket"
storage = gcloud.storage

bucket = storage.bucket "my-bucket"
file = bucket.file "myfile.zip"
file.download "/tmp/myfile.zip"
```

When downloading 1.8GB file:

```
/opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file/**verifier.rb:55:in `read': failed to allocate memory (NoMemoryError)**
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file/verifier.rb:55:in `block in md5_for'
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file/verifier.rb:54:in `open'
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file/verifier.rb:54:in `md5_for'
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file/verifier.rb:31:in `verify_md5!'
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file.rb:688:in `verify_file!'
	from /opt/lang/ruby-2.2.4/lib/ruby/gems/2.2.0/gems/gcloud-0.10.0/lib/gcloud/storage/file.rb:353:in `download'
	from pull.rb:11:in `<main>'
```

If I change 'verify' to 'none', I don't encounter this error.

free mem + cached > 3GB

quartzmo:
@Aguasvivas22 Thanks for opening this issue! I'll see if I can reproduce it myself today.

quartzmo:
I'm hoping that the solution will be to change the implementation to something that uses less memory (see "Compute digest by chunks" in examples for [Digest](http://ruby-doc.org/stdlib-2.2.3/libdoc/digest/rdoc/Digest.html)), but this is just an initial guess.

[...]

PR #727, opened 2016-06-08
Vision Updates

blowmage:
Small correction to allow Vision to use Tempfiles. Brings Vision inline with how gcloud treats IO-ish objects in the other services.

Another small correction to avoid reading IO contents into memory.

### DIFF
--- a/lib/gcloud/vision/image.rb
+++ b/lib/gcloud/vision/image.rb
@@ -72,26 +72,17 @@ module Gcloud
       #
       # @see {#url?}
       #
-      def content?
+      def io?
         !@io.nil?
       end
 
       ##
       # @private Whether the Image is a URL.
       #
-      # @see {#content?}
+      # @see {#io?}
       #
       def url?
         !@url.nil?
-      end
-
-      ##
-      # @private The contents of the image, encoded via Base64.
-      #
-      # @return [String]
-      #
-      def content
-        @content ||= Base64.encode64 @io.read
       end
 
       ##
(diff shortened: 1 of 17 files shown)


lib/gcloud/vision/image.rb
```diff
@@ -359,8 +356,9 @@ def inspect
       ##
       # @private The Google API Client object for the Image.
       def to_gapi
-        if content?
-          { content: content }
+        if io?
+          @io.rewind
+          { content: Base64.strict_encode64(@io.read) }
```
quartzmo: Would it be safer for memory usage (see #718) to [do this instead](https://github.com/GoogleCloudPlatform/gcloud-ruby/commit/19c47947b22dc0a09c9ac0a2a8ed30a7b2cbdbc3)?

blowmage: Not really. Calculating cryptographic hashes are different than encoding the IO data.

The Vision image object needs to read the entire string in order to encode it, and then the encoded string is given to the API call. Not every image object's IO object will be persisted on the filesystem, some could be a StringIO object which are already in memory.

But once the image object reads from the IO and encrypts to Base64, the image object no longer holds on to a reference to the contents, and so the contents can be garbage collected. Previously the image object put the contents into an ivar and so the contents stayed around as long as the vision image object did.

quartzmo: OK, great, I just wanted to be cautious. That sounds good.

PR #728, opened 2016-06-08
Fix Datastore examples Kind name

quartzmo:
Use singular form of `Kind` name, as in the official guide snippets.

This PR is just a small docs follow-up to #724.


blowmage:
I feel bad for missing those. Thanks for cleaning up after me! 😏

Issue #687, opened 2016-05-12
Datastore question: how to get estimation of on-the-wire packet size?


timanovsky:
As you know each operation is limited to 10MB on the wire. I want to pack as many entities into single batch insert as possible but with 100% guarantee it fits into the limit. In *v1beta2* I used the the following workaround: `entity.to_proto.encode.length` which worked fine and I have never seen size limit errors. Now with *v1beta3* this hack does not work, and I'm not sure how to get correct estimate given base64 etc.

timanovsky:
Worked around again by `entity.to_grpc.to_proto.length` But would be good to have public API

blowmage:
I think that this is a great feature. It is obviously important for folks trying to maximize the effectiveness of their calls to Datastore. We will add a method to the public API to get you this data.

blowmage:
Hey @timanovsky, we've been discussing this and we have a couple paths we can go. One is to add this estimated size and let users like yourself manage updates. The other is to attempt to manage this within gcloud-ruby, and make multiple updates when the size is bigger than 10MB. 

What are your thoughts on this? Do we choose option 1 and grow the API? Or choose option 2 and make the library do more with the existing API? Which would you prefer?

timanovsky:
Well in my case (batch ops) it wouldn't be much of a difference, but if you
consider transactions, the user-level control/visibility is a must. I don't
want my huge transaction be silently broken down to smaller ones.

API growing is not a big problem, as it is only for advanced use, and
people does not need to learn it from the start.





Best regards,
Alexey Timanovsky.

blowmage:
Thanks for your input!

PR #729, opened 2016-06-08
Add Datastore serialized_size

blowmage:
Add the method `serialized_size` to give the number of bytes the Entity or Key take when being serialized during API calls. (See #687)


quartzmo:
Thanks @blowmage. And thanks @timanovsky for raising the issue.

PR #731, opened 2016-06-11
Add customer-supplied encryption keys to Storage

### LINKED ISSUE #722: Support custom encryption keys for storage?
Can we support custom encryption keys for files uploaded?

Spec is here: https://cloud.google.com/compute/docs/disks/customer-supplied-encryption#top_of_page

quartzmo:
This implementation does not address a known issue in which the
service returns 411 when request sent with httpclient.

[closes #722]

### DIFF
--- a/lib/gcloud/storage/bucket.rb
+++ b/lib/gcloud/storage/bucket.rb
@@ -416,9 +416,23 @@ module Gcloud
       ##
       # Retrieves a file matching the path.
       #
+      # If a [customer-supplied encryption
+      # key](https://cloud.google.com/storage/docs/encryption#customer-supplied)
+      # was used with {#create_file}, the `encryption_key` and
+      # `encryption_key_sha256` options must be provided or else the file's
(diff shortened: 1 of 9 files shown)


lib/gcloud/storage/bucket.rb
```diff
@@ -433,9 +446,9 @@ def files prefix: nil, delimiter: nil, token: nil, max: nil, versions: nil
       #   file = bucket.file "path/to/my-file.ext"
       #   puts file.name
       #
-      def file path, generation: nil
+      def file path, generation: nil, key: nil, key_sha256: nil
```
jgeewax: encryption_key * ?

quartzmo: @blowmage? (My personal preference is for long, explicit names that hold closely to the underlying API...)

blowmage: My preference is for `key`.

quartzmo:
@blowmage Option names updated, ready for review.

Issue #201, opened 2015-08-04
Stream downloads to disk


blowmage:
Storage downloads are read into memory and then written to the file system. We want the files to be streamed to the file instead of storing all contents into memory. The Google API Client does not support streaming downloads, so we might want to handle downloads outside of it. See #67.

Issue #490, opened 2016-01-04
Replace optional params with yielded object?


blowmage:
We have a couple methods in Storage that have a long list of optional parameters.

* [`Project#create_bucket`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Storage/Project.html#method-i-create_bucket)
* [`Bucket#create_file`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Storage/Bucket.html#method-i-create_file)

About half of these parameters could be removed if we yielded the object to be created before the API call was made. This means that instead of making a call like this:

```ruby
require "gcloud"

gcloud = Gcloud.new
storage = gcloud.storage

bucket = storage.create_bucket "my-bucket",
                               location:       "US",
                               logging_bucket: "example-logs-bucket",
                               logging_prefix: "log_object_prefix",
                               storage_class:  "NEARLINE",
                               versioning:     true,
                               website_main:   "index.html",
                               website_404:    "404.html" do |c|
  c.add_rule ["http://example.org", "https://example.org"],
             "*",
             response_headers: ["X-My-Custom-Header"],
             max_age: 300
end
```

We could make a call like this instead:

```ruby
require "gcloud"

gcloud = Gcloud.new
storage = gcloud.storage

bucket = storage.create_bucket "my-bucket" do |b|
  b.location       = "US"
  b.logging_bucket = "example-logs-bucket"
  b.logging_prefix = "log_object_prefix"
  b.storage_class  = "NEARLINE"
  b.versioning     = true
  b.website_main   = "index.html"
  b.website_404    = "404.html"
  b.cors do |c|
    c.add_rule ["http://example.org", "https://example.org"],
               "*",
               response_headers: ["X-My-Custom-Header"],
               max_age: 300
  end
end
```

I think the latter call is a significant improvement. And functionally equivalent. All data is sent on the single API call to create the resource. And this approach is used in [`Bucket#update`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Storage/Bucket.html#method-i-update) and [`File#update`](http://googlecloudplatform.github.io/gcloud-ruby/docs/v0.6.1/Gcloud/Storage/File.html#method-i-update) so it should be familiar to users. But this would break the existing API. What do you think? Is this a change we should make?

quartzmo:
[...]

Issue #269, opened 2015-09-04
Upgrade to Google API Client 0.9


blowmage:
The [next version](https://github.com/google/google-api-ruby-client/milestones/0.9) of Google API Client makes some major changes, including a new ruby API for connecting to services and a replacing Faraday for Hurley. This transition will require changes in tests to use Hurley's new test API.

blowmage:
We will want to make sure that lostisland/hurley#18 is resolved as all our tests depend on this behavior.

blowmage:
[Migrating notes](https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md)

edwinkcw:
I think https://github.com/lostisland/hurley/issues/18 is resolved, could you please try and test it? Thanks

blowmage:
Yes, lostisland/hurley#18 is resolved and Google Client API 0.9 has been released. But while that was happening GRPC has become viable and so we are in the process of migrating services to use GRPC. Once the GRPC-compatible services are migrated we will look at upgrading to Google Client API 0.9 and converting our tests to use the hurley mocks.

blowmage:
This should have been closed by #733.

PR #733, opened 2016-06-14
Upgrade google api client

blowmage:
WIP of the Google API Client upgrade.


quartzmo:
I'm going to start work on Translate.

lib/gcloud/resource_manager/service.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      def service
+        return mocked_service if mocked_service
+        @service
+      end
+      attr_accessor :mocked_service
+
+      ##
+      # Returns API::ListProjectsResponse
+      def list_project filter: nil, token: nil, max: nil
```
quartzmo: Should this be `list_projects` instead?

blowmage: It could be. I wasn't trying to make improvements in that commit, just convert the client.

quartzmo: Gotcha!

quartzmo:
Can `lib/gcloud/resource_manager/errors.rb` be deleted?

quartzmo:
I think `ApiError` will still be needed, although it may not have HTTP attributes such as `status` In Vision I have encountered an error object nested in the response object, like this: 

```sh
#<Google::Apis::VisionV1::AnnotateImageResponse:0x007fa6b65a8048 @error=#<Google::Apis::VisionV1::Status:0x007fa6b6619d10 @code=3, @message="image-annotator::Malformed request.: Image processing error!">>
```

blowmage:
We will have additional work to handle errors appropriately in the 0.9 client. That particular response is just how vision API works. https://cloud.google.com/vision/reference/rest/v1/images/annotate#annotateimageresponse

quartzmo:
I have signed the CLA.

quartzmo:
This is a big PR and although all tests are passing, we need to follow up with extensive manual testing. Should we do that before merging?

blowmage:
I agree. I think there is an argument that can be made to merge this while we test so that the acceptance tests are running on CI.

quartzmo:
That's a great point. Merging now.

blowmage:
Here are the issues this PR closes: #201, #269, #490, #734, #735

PR #747, opened 2016-06-27
Fix Logging acceptance test

quartzmo:
Remove protocol buffer payload from test due to consistent failure.


blowmage:
Can you leave the original test code commented out? And update the commit to give more background on the issue? I'd like to be able to revert this change, and think having comments in the test would encourage that.

blowmage:
The updated commit (d607cd6d529428caf56c17c0d7a605c1ff47a28b ) was pushed to master. Looks good. Closing.

PR #757, opened 2016-07-04
Added chunk_size option to the bigquery::table::load method

gramos74:
I was using version 0.10 where the chunk_size option was introduced for the bigquery::table::load method. But there we have an error in the method bigquery:table:load_resumable ( line 920).

The calculation for the file size should be File.size(file) instead of file.length.

I was checking the version 0.11 and I realised that the chunk_size option was removed again. So I added it again and added a test in order to ensure that if is removed accidentally in the future a test will fail.


blowmage:
I'm closing this PR as we are unable to support this option. This PR as written when BigQuery was using Google API Client 0.8, which allowed setting of `chunk_size`. Since then, BigQuery has moved to Google API Client 0.9, which does not allow `chunk_size` to be set. We don't have access to these settings, as they are all internal to Google API Client now.

PR #762, opened 2016-07-11
Move retries to parameter

### LINKED ISSUE #463: Rethink Incremental Backoff Integration
Right now incremental back off is configured in two different ways:

1) Globally by configuring Gcloud::Backoff
2) On any method that supports back off by using the method's three backoff optional arguments in the options hash

Now that we are dropping Ruby 1.9.3, and moving from option hashes to named parameters, we cannot continue to have these options remain undocumented. We will either need to add these additional named parameters to all the methods that will support incremental backoff, or change how it is implemented. We may consider one of the following options:

1) Propagate the three back off named parameters to all methods.
2) Remove all back off named parameters on methods, leaving only the global configuration.
3) Pass in one named parameter of the Gcloud::Backoff object instance instead of the three named parameters.

Storage is the only service making use of incremental back off. It should probably be used on more API calls, but we have had a hard time getting information on which API calls should be automatically retried.

blowmage:
Remove Gcloud::Backoff and add retries as optional argument when creating
new service objects.

[closes #463]

### DIFF
--- a/test/gcloud/gcloud_test.rb
+++ b/test/gcloud/gcloud_test.rb
@@ -20,10 +20,11 @@ require "gcloud/pubsub"
 describe Gcloud do
   it "calls out to Gcloud.datastore" do
     gcloud = Gcloud.new
-    stubbed_datastore = ->(project, keyfile, scope: nil) {
+    stubbed_datastore = ->(project, keyfile, scope: nil, retries: nil) {
       project.must_equal nil
       keyfile.must_equal nil
       scope.must_be :nil?
+      retries.must_be :nil?
       "datastore-project-object-empty"
     }
     Gcloud.stub :datastore, stubbed_datastore do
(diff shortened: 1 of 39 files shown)


quartzmo:
I like this pushing down of `retries` to the services.

So this is backward-compatible with existing usage of `Backoff#retries=` ? That method just gets removed from the docs? (Appears so to me, and that seems good.)

blowmage:
Not backwards compatible. `Glcoud::Backoff` should go away in the upcoming release. Only the GRPC services will use what used to be `Gcloud::Backoff`, the REST services will perform backoff retries through Google API Client. In the future, GRPC back off retries will be performed by GAX.

test/gcloud/gcloud_test.rb
```diff
@@ -20,10 +20,11 @@
 describe Gcloud do
   it "calls out to Gcloud.datastore" do
     gcloud = Gcloud.new
-    stubbed_datastore = ->(project, keyfile, scope: nil) {
+    stubbed_datastore = ->(project, keyfile, scope: nil, retries: nil) {
       project.must_equal nil
       keyfile.must_equal nil
```
quartzmo: Currently failing acceptance test for Ruby 2.2:

```
 1) Failure:
Gcloud#test_0001_calls out to Gcloud.datastore [/home/travis/build/GoogleCloudPlatform/gcloud-ruby/test/gcloud/gcloud_test.rb:25]:
Expected: nil
  Actual: 1
```

blowmage: That is failing on only one ruby version and only on Travis-CI. I can't recreate the failure locally. Frustrating.

PR #775, opened 2016-07-15
Split main gcloud test into separate files

blowmage:
The current Travis-CI build is erroring on the test/gclcoud/gcloud_test.rb file,
but only on ruby 2.2.0. We cannot recreate the failure locally. Also, the failure
is referencing line numbers that the file does not have. Split the test into
several files to force Travis-CI to flush its cache, and becasue it will be
easier to identify services that are missing these tests.


quartzmo:
While we're at it, can we fix the copy/pasted variable name `dataset` that is used everywhere, replacing it with `project`?

blowmage:
Hah! didn't even notice that! Will update.

quartzmo:
Yay for fixing the build! I wonder about Travis though...

PR #776, opened 2016-07-15
Use current ruby versions

blowmage:


quartzmo:
Let's try it! Good to update these versions in any case.

PR #783, opened 2016-07-20
Fix issue with freezing BigQuery Access

### LINKED ISSUE #782: Error updating frozen Dataset::Access
Because calling `Dataset#access` without a block [freezes the underlying `@gapi.access` object](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/bigquery/dataset/access.rb#L97), a subsequent call to `Dataset#access` with a block that attempts to modify `@gapi.access` results in an error.

```ruby
dataset = bigquery.dataset dataset_id
assert dataset.access.reader_special? :all # Freezes gapi.access

dataset.access do |acl|
    acl.remove_reader_special :all # Attempts to modify gapi.access
end
```

```sh
RuntimeError: can't modify frozen Array
    /Users/quartzmo/code/google/codez/gcloud-ruby/lib/gcloud/bigquery/dataset/access.rb:491:in `reject!'
    /Users/quartzmo/code/google/codez/gcloud-ruby/lib/gcloud/bigquery/dataset/access.rb:491:in `remove_access_role_scope_value'
    /Users/quartzmo/code/google/codez/gcloud-ruby/lib/gcloud/bigquery/dataset/access.rb:237:in `remove_reader_special'
```

blowmage:
Ensure that freezing the access object does not freeze the gapi object.

[fixes #782]

### DIFF
--- a/lib/gcloud/bigquery/dataset.rb
+++ b/lib/gcloud/bigquery/dataset.rb
@@ -243,10 +243,13 @@ module Gcloud
       #
       def access
         ensure_full_data!
-        access_builder = Access.new @gapi
+        access_builder = Access.from_gapi @gapi
         if block_given?
           yield access_builder
-          patch_gapi! :access if access_builder.changed?
+          if access_builder.changed?
+            @gapi.update! access: access_builder.to_gapi
+            patch_gapi! :access
+          end
         end
         access_builder.freeze
       end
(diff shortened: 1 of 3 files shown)


quartzmo:
The asymmetry between `from_gapi` and `to_gapi` is a little confusing, although I understand why it is there. Do you want to add any comments explaining it? Otherwise I'll just merge, basically looks good to go!

blowmage:
I added those methods to make it clear that the `gapi.access` array was no longer being updated directly. Those values now need to be set outside of the `Access` object. Using this approach and adding those methods also brings it inline with how `Storage::Bucket::Cors` and `Bigquery::Schema` now work since the upgrade.

PR #787, opened 2016-07-21
Relax mime-types dependency

### LINKED ISSUE #765: Upgrade MIME Types dependency
Ensure we are using the [columnar store](https://github.com/mime-types/ruby-mime-types#columnar-store) and the same version as the latest [Google API Client](https://github.com/google/google-api-ruby-client/issues/424#issuecomment-228154814).

blowmage:
Allow mime-types 3 to be used, which addresses many performance issues.
Originally the dependency was pinned to 2.x, since we didn't know what changes
would be made in 3.x, but so far there haven't been any affecting gcloud.

[closes #765]


quartzmo:
Great, merging!

PR #788, opened 2016-07-21
Add timeout optional argument

blowmage:
This is an extension of the existing API. We don't currently allow users to specify a timeout. Part of the challenge is how we would support the feature as we move from Google API Client 0.8 to 0.9 and eventually to GRPC and GAX. I believe we should be able to support this API across all those implementations.

### DIFF
--- a/lib/gcloud.rb
+++ b/lib/gcloud.rb
@@ -44,6 +44,7 @@ module Gcloud
   #   path the file must be readable.
   # @param [Integer] retries Number of times to retry requests on server error.
   #   The default value is `3`. Optional.
+  # @param [Integer] timeout Default timeout to use in requests. Optional.
   #
   # @return [Gcloud]
   #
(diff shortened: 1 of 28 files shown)


blowmage:
I did not specify what the default timeout is when none is set, but I could not figure out what the timeout is in Hurley or HTTPClient. The gRPC timeout is _infinity_. So I felt it best to not say in the documentation exactly what the default timeout was.

blowmage:
This does work across GAPI and GRPC services:

```ruby
gcloud.storage(timeout: 0.1).buckets.count
#=> Gcloud::Error: execution expired
gcloud.pubsub(timeout: 0.1).topics.count
#=> Gcloud::DeadlineExceededError: 4:Deadline Exceeded
```

quartzmo:
I think moving the timeout configuration from a single global variable to the project instances is a great change.

gramos74:
Sorry... I was not at home.

I saw the code and looks really good. I agree with Chris that moving out the timeout variable from the global configuration is great.

But I have a question... maybe it's a bit stupid... I'm not able to see how Hurley use that timeout_sec. Could you let me know  ? ( It's simple interest in order to know a bit more the flow ).

PR #791, opened 2016-07-22
Allow for more difference in time values

blowmage:
The default delta of 0.001 is too small for Travis-CI.
Getting some failures on this test for some reason.
Bump up the delta to make it pass more reliably on CI.


quartzmo:
great, thanks @blowmage!!

PR #794, opened 2016-07-22
Upgrade Resource Manager to V1

blowmage:
We don't have acceptance tests because this service requires user credentials. I've manually tested this and it works as expected. Much easier to do since the Google API Client upgrade.


quartzmo:
Great that we were so conservative on the path to `v1`!

PR #800, opened 2016-07-22
Additional documentation changes

blowmage:
Update documentation on methods that are not part of the public API. Update BigQuery Table#schema documentation.


quartzmo:
So awesome that we could catch these little issues and leaks in our public API by diffing our previous version docs JSON with our current docs JSON. Great idea @blowmage!

/cc @callmehiphop @stephenplusplus @jgeewax

PR #802, opened 2016-07-23
Update GRPC and GAPI dependencies

blowmage:
Update to the most recent versions.


blowmage:
The google-protobuf dependency is not changed because the newer releases continue to be broken.

quartzmo:
Upgrades, woo hoo!

PR #804, opened 2016-07-25
Remove old gapi require

blowmage:
This file was added to aid in the conversion from Google API Client 0.8 to 0.9.
Now that the conversion is complete we no longer need this file.


quartzmo:
Good, thanks @blowmage.

PR #806, opened 2016-07-26
Rename Gcloud to Google::Cloud

blowmage:


quartzmo:
This is quite a large commit? Would it be possible to make the conversion incrementally, service-by-service, then finally the core? Or starting with the core?

blowmage:
I don't think an incremental, service-by-service approach would be any easier to consume. The problem is once you start pulling things apart things... fall apart. This isn't my first attempt at making this change, and it isn't an easy change to make. All files are renamed, and much of the contents have to be adjusted so their formatting passes the rubocop rules. In the end I figured it was better to make the change all at once.

quartzmo:
@blowmage Makes sense, and sounds good.

quartzmo:
I've reviewed this PR with @blowmage and am planning to accept it as soon as the build for his rebase on latest master is successful.

blowmage:
:metal:

PR #812, opened 2016-08-01
Fix Pub/Sub bug in pull timeout

### LINKED ISSUE #811: uninitialized constant Gcloud::Pubsub::Subscription::Hurley
Hit this just a few moments ago:

`rescued...
uninitialized constant Gcloud::Pubsub::Subscription::Hurley`

specific line with issue: https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/lib/gcloud/pubsub/subscription.rb#L237

general search link: https://github.com/GoogleCloudPlatform/gcloud-ruby/search?utf8=%E2%9C%93&q=Hurley

Not sure if this is just left over and still needs removing, or needs other special handling....

blowmage:
When pull times out, it should continue. Now that Pub/Sub is on gRPC the
timeout error is no longer a Faraday or Hurley error. Fix the rescue statement
to recover from the error class that will be raised in a gRPC timeout.

[fixes #811]

### DIFF
--- a/lib/gcloud/pubsub/subscription.rb
+++ b/lib/gcloud/pubsub/subscription.rb
@@ -234,7 +234,7 @@ module Gcloud
         end
         acknowledge messages if autoack
         messages
-      rescue Hurley::Timeout
+      rescue Gcloud::DeadlineExceededError
         []
       end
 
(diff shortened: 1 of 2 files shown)


quartzmo:
Thanks for the quick turnaround on this fix @blowmage!

PR #813, opened 2016-08-01
Modularize into separate gems

### LINKED ISSUE #341: Should we publish a separate module (gem) for each Google Cloud service?
This is the gcloud-ruby specific discussion of the broader conversation in gcloud-common, [#30, Should separate APIs be separately installable?](https://github.com/GoogleCloudPlatform/gcloud-common/issues/30). See also @jgeewax's [comment in gcloud-node #91](https://github.com/GoogleCloudPlatform/gcloud-node/issues/91#issuecomment-138522547).

blowmage:
Separate the monolithic package into individual packages by service.

[closes #341]


blowmage:
The coveralls decrease makes sense. We are switching from measuring coverage with unit tests and acceptance tests to just unit tests. Also, we have introduced many new files and some have lower coverage.

quartzmo:
@blowmage Thanks for explaining that. It shouldn't be hard to increase unit test coverage to a passing level, correct?

quartzmo:
@blowmage Do you intend to continue supporting the `rake console` task? I don't see it.

blowmage:
Each gem has a `console` task, but there isn't one for the top level yet. If there were one, it should just load the `google-cloud` gem, and let the service gems get loaded when called.

quartzmo:
I have a feeling we will need to change how docs json is generated. From GoogleCloudPlatform/gcloud-common#157, it appears that in addition to generating the docs json for the individual service packages (which we currently support), we will also have to generate a "bundled package" docs json containing all of the packages.

blowmage:
I'm happy with the state of the individual packages. There is some additional work around updating the documentation site and automating releases, but I think this PR is ready to be accepted.

blowmage:
I don't know that we will need to change the way the JSON docs are generated as much as add to it. We'll need the JSON files for the individual packages either way, right?

I have some reservations about merging all the docs into one "bundled" package. Viewers of the docs won't be shown which package the classes and methods are defined in, and more importantly which version of the package they are defined in.  I would prefer to see more linking across packages than duplicating the documentation into the "bundled" package.

quartzmo:
@blowmage I would really like to merge this now, so that I can better assist with the effort. Can you create issues now for the outstanding tasks, such as docs json generation?

blowmage:
Yep! Will add them as soon as this is accepted.

quartzmo:
OK, I will merge it now.

PR #809, opened 2016-07-29
Introduce gRPC and Gapic client for Cloud Language API.

jmuk:
Gapic client (language_service_api.rb) and its config (_config.json)
are the wrapper automatically generated from the proto definition
and API configs.

PR #815, opened 2016-08-04
Add empty natural language gem packages

blowmage:


blowmage:
@quartzmo Thoughts?

quartzmo:
It seems a little premature to add these now, but my understanding is that we are doing it to unblock @jmuk so that he can update or renew #809 for both language and speech. In this case I'm all for it since we should definitely prioritize his help with the Gapic client code!

I'm just looking it over for typos, then I'll merge. Thanks @blowmage!

google-cloud-speech/lib/google/cloud/speech.rb
```diff
@@ -0,0 +1,27 @@
...
+
+require "google-cloud-speech"
+require "google/cloud/speech/project"
+
+module Google
+  module Cloud
+    ##
+    # # Google Cloud Language
+    #
+    module Language
```
quartzmo: Critical c/p typo: change to `module Speech`.

quartzmo:
@blowmage I only found one typo. Other than that, ready to merge.

quartzmo:
I have signed the CLA and am okay with these commits being contributed to Google.

PR #823, opened 2016-08-09
Add BigQuery Schema methods on Table::Updater

blowmage:
The [`#create_table`](https://googlecloudplatform.github.io/gcloud-ruby/#/docs/v0.12.1/gcloud/bigquery/dataset?method=create_table-instance) syntax changed in the [v0.12.0 release](https://github.com/GoogleCloudPlatform/gcloud-ruby/releases/tag/v0.12.0), but the old syntax has been [held up as an example of the niceness of the library](http://google-opensource.blogspot.com/2016/08/making-rubyists-more-comfortable-on.html). Restore the original syntax by adding the Schema methods to the Table::Updater class.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
@@ -325,7 +325,7 @@ module Google
         #   When a table already contains data, schema changes must be additive.
         #   Thus, the default value is `false`.
         # @yield [schema] a block for setting the schema
-        # @yieldparam [Table::Schema] schema the object accepting the schema
+        # @yieldparam [Schema] schema the object accepting the schema
         #
         # @return [Google::Cloud::Bigquery::Schema]
         #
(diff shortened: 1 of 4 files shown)


blowmage:
If this is approved I will backport this change to the 0.12 branch so we can release a new version there as well.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -870,6 +870,37 @@ def initialize gapi
...
+          #
+          # @return [Google::Cloud::Bigquery::Schema]
+          #
+          # @example
+          #   require "google/cloud"
+          #
+          #   gcloud = Google::Cloud.new
+          #   bigquery = gcloud.bigquery
+          #   dataset = bigquery.dataset "my_dataset"
+          #   table = dataset.create_table "my_table" do |schema|
```
quartzmo: Although this is a great example, I don't see this method (`Updater#schema`) called anywhere in it. Should this example be added to the list on `Dataset#create_table`, and replaced here with a different one?

blowmage: Good catch. I'll update that code example to actually demonstrate the method described! :)

quartzmo:
I love the flexibility @blowmage. Good call.

PR #826, opened 2016-08-12
improve gcloud core login syntax and support for local keyfiles

pugsiman:


pugsiman:
added a switch statement, a bit faster(and more readable in this case) in ruby.
also added support for physical keyfiles, as per the google docs it is officially supported to be able to load the whole keyfile as a string (e.g. from an environment variable).

quartzmo:
@pugsiman Thanks for submitting this pull request!

blowmage:
To accept the keyfile's contents I think we should check for an IO object instead of a String, and leave the String for the file path. The [Google Auth library](https://github.com/google/google-auth-library-ruby) is now used as a wrapper for Signet, and it's service account object takes an [IO object in the constructor](http://www.rubydoc.info/gems/googleauth/0.4.1/Google/Auth/ServiceAccountCredentials#initialize-instance_method).

While we do allow the keyfile contents via environment variable, we don't allow them in the same environment variable. Specifically there are different environment variables for keyfile path vs. contents.  `GOOGLE_CLOUD_KEYFILE_JSON` vs. `GOOGLE_CLOUD_KEYFILE`.

pugsiman:
Ah, I see. I just read the code more thoroughly and realized it should work with one of those envars, but only as long as you don't send your keyfile's content as a parameter. Maybe it should be clarified? It definitely has caused some confusion at work :-).

quartzmo:
Sorry we never accepted this PR, but it just didn't rise above the threshold in terms of adding value for users. But thank you!

Issue #830, opened 2016-08-15
Automate Individual Package Documentation Build


blowmage:
With the switch from monolithic gem to individual gems, we need to replace the previous build process that identifies new releases and generates the documentation. The new mechanism should be aware of package-specifc git tags (google-cloud-datastore:v1.2.0 vs. v1.2.0) and create a new version for the appropriate gem.

quartzmo:
@blowmage To be clear, when a new version of an individual gem is released, a new version of the `gcloud-common` "bundled" gem will also be released, and new documentation published for both the individual and the bundled gems? 

See [previous discussion on gcloud-common](https://github.com/GoogleCloudPlatform/gcloud-common/pull/157#issuecomment-237376667):

> when a subpackage gets a feature, our umbrella package gets a feature... which is a feature, so we should tag it as one.

PR #833, opened 2016-08-15
Add jsondoc support for multiple packages

quartzmo:
This PR adds `docs` directories containing assets such as `manifest.json` `toc.json` to the root project and to each sub-project, and updates the `jsondoc` rake task to copy these assets. 

It also adds a rake task that copies JSON content from the sub-projects to the `google-cloud` umbrella project, for compatibility with the site app (see GoogleCloudPlatform/gcloud-common/pull/157.)

[refs #830]

[Demo](https://quartzmo.github.io/gcloud-ruby/#/docs/google-cloud/v0.12.2/google/cloud)

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -48,15 +48,15 @@ The **Project ID** and **Credentials JSON** can be placed in environment variabl
 
 Here are the environment variables that Datastore checks for project ID:
 
-1. DATASTORE_PROJECT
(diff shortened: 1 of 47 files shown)


quartzmo:
I think the `gh-pages` update logic is pretty much there, but I need to test it to be sure that older versions won't be removed and will still work.

quartzmo:

PR #836, opened 2016-08-17
Update to protobuf3 and grpc1.0.0.

jmuk:
Also grpc files and gapic files are re-genarated.
See googleapis/gax-ruby#37


google-cloud-core/Gemfile
```diff
@@ -4,9 +4,9 @@ gemspec
 
 gem "rake"
 gem "google-api-client", "~> 0.9.11"
-gem "grpc", "= 1.0.0.pre1"
-gem "google-protobuf", "= 3.0.0.alpha.5.0.5.1"
-gem "googleapis-common-protos", "~> 1.1"
+gem "grpc", "~> 1.0.0.pre2"
```
blowmage: Can you set all the grpc dependencies to be exact?

```ruby
gem "grpc", "= 1.0.0.pre2"
```

google-cloud-core/Gemfile
```diff
@@ -4,9 +4,9 @@ gemspec
...
+gem "google-protobuf", "~> 3.0.0"
```
blowmage: Can you update all the protobuf dependencies to allow not as narrowly defined?

```ruby
gem "google-protobuf", "~> 3.0"
```

google-cloud-core/Gemfile
jmuk: Done

google-cloud-core/Gemfile
jmuk: Done

jmuk:
Wait, on the meanwhile, grpc-1.0.0 (non 'pre') has been published in the rubygems.org. Let me update this (and others) to use that.

quartzmo:
@jmuk Thank you! Very exciting.

jmuk:
Done. Is "~> 1.0" is good enough rather than exact "=1.0.0"?

Travis still fails because I haven't pushed google-gax yet. If this version spec is fine, I'll push new google-gax and then everything should be fine.

quartzmo:
"~> 1.0" seems correct to me, as long as we have no issues with gRPC 1.0. @blowmage?

jmuk:
Hmm, it turns out that this is not enough at all. gRPC 1.0 seems working (so the version spec would be okay), but protobuf behavioral change hits several test cases on gcloud.

jmuk:
Now it passes the tests. For the record, I've modified some source code and test code for some reasons:

- logging API: a few field names have changed since the gRPC files were generated last time. Renamed several names to support the new name.
- datastore API: some fields started returning non-`nil` value (the default value) when missing. Needed to modify some source code to work on the new cases.

Please take another look at the patch. Thanks!

quartzmo:
@jmuk Thank you! We will look at this today. Much appreciated!

PR #841, opened 2016-08-23
Language setup

blowmage:
Stand up the Natural Language service implementation. I decided to drive the development against the live service using acceptance tests, hence the lack of mocked out unit tests. Unit tests will be added in a later PR, but for now it would be good to get this implementation accepted so we can start moving on other concerns like documentation. (There is some documentation, but it is no where near final.)

The service is most similar to Vision. There is a main `#annotation` method that takes content arguments, and a Document class that models the content and can also call an `#annotation` method.


google-cloud-language/acceptance/language/html_test.rb
```diff
@@ -0,0 +1,273 @@
...
+      annotation.sentiment.polarity.must_be_close_to 1.0
+      annotation.sentiment.magnitude.must_be_close_to 1.899999976158142
+
+      annotation.entities.must_be :empty?
+
+      annotation.sentences.must_be :empty?
+      annotation.tokens.must_be :empty?
+    end
+
+    it "runs only the sentiment feature" do
```
quartzmo: "sentiment" -> "entries"

quartzmo:
I'm really happy with this new approach to test-driving the development against a new service. The acceptance test coverage is just awesome, and the unit tests that follow are based on empirical fact, not speculation. This is the way to do it imo.

quartzmo:
Ugh, I would love to merge this, but Job #1403 seems stalled. I'll wait a bit longer.

google-cloud-language/acceptance/language/html_test.rb
blowmage: Ugh, great catch. Thanks.

PR #844, opened 2016-08-23
Ensure order of ENV variables

blowmage:
Make sure GOOGLE_CLOUD_* has higher priority than GCLOUD_*.


quartzmo:
Good catch, will merge when build completes.

Issue #760, opened 2016-07-11
Switch Logging to Gax


blowmage:
Update to use google-gax.

PR #845, opened 2016-08-24
Add Gapic-generated clients for Logging API.

jmuk:
- v2.rb is hand-written. This should be auto-generated, but
  not yet.
- manually modified .rubocop.yml, .yardopts, and gemspec.

### DIFF
--- a/google-cloud-logging/.rubocop.yml
+++ b/google-cloud-logging/.rubocop.yml
@@ -3,6 +3,7 @@ AllCops:
     - "acceptance/**/*"
     - "google-cloud-logging.gemspec"
     - "lib/google/logging/**/*"
+    - "lib/google/cloud/logging/v2/**/*"
     - "Rakefile"
     - "test/**/*"
 
--- a/google-cloud-logging/.yardopts
+++ b/google-cloud-logging/.yardopts
@@ -1,6 +1,7 @@
 --no-private
 --title=Google Cloud Logging
 --exclude lib/google/logging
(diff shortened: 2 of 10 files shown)


blowmage:
Looks great. I don't know if v2.rb needs to be generated, honestly.

blowmage:
One of the benefits for creating the v2.rb file manually is that we can add all the "Low Level API" documentation in it. I really wouldn't worry about generating it.

jmuk:
I haven't thought about that deeply, but if v2.rb will have some documentation in it, that would be mostly boilerplate -- explain how to load the module or initialize classes with links to the individual generated files, I guess.
Individual generated file would hold the major part of the documentation and examples.

Eitherway, that's mostly for my thing, it wouldn't matter to you whether or not it's automatically generated (sorry I put in the commit message).

blowmage:
No worries about the commit message. 😄 

We have sort of a dual approach to documentation. One half is the reference material that would be generated along with the code, the other half is the guide material. You can see the main Stackdriver Logging guide documentation [here](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/master/google-cloud-logging/lib/google/cloud/logging.rb#L22). We would want similar guide documentation for the "Low Level API" explaining the concepts and giving users an idea of what can be accomplished. I expect the guide docs are going to be doubly important since I don't expect the reference docs to have many code examples.

blowmage:
This is the first part of completing #760.

PR #846, opened 2016-08-24
Upgrade Datastore to V1

blowmage:
Move Datastore from V1beta3 to V1. Also make some small changes to the ruby code to handle `oneof` values consistently.


quartzmo:
Very awesome, thank you @blowmage.

PR #852, opened 2016-08-25
Use post-install message

blowmage:
The `post_install_message` will will be shown when the gem is first installed. We can use this message to direct users to migrate to google-cloud. What do you think?


quartzmo:
How about, "please change the gem name in your dependencies"

blowmage:
@quartzmo: Okay, updated.

quartzmo:
Should i merge?

blowmage:
Yeah, good to merge.

PR #864, opened 2016-08-27
Autoload all Google Cloud services

blowmage:
When calling `require "google/cloud"` all service gems should also be loaded.
This happens automatically if Bundler's autorequire feature is on.
But if users aren't using Bundler then load all gems with this code.


blowmage:
Ugh, I can't believe I forgot to add this code weeks ago. Sorry about that.

PR #871, opened 2016-08-30
Speech API: back to v1beta1

jmuk:


jmuk:
See: https://github.com/googleapis/googleapis/issues/103

PR #877, opened 2016-09-02
Regenerate files for delay loading grpc module.

jmuk:


blowmage:
Thanks @jmuk! I'm going to hold of on merging this for a bit because bumps the google-gax dependency outside of the current dependency settings. I'm thinking the v0.21 releases. Hopefully Google API Client for Ruby will have their v0.10 out very soon as well.

PR #884, opened 2016-09-07
Add Datastore Gapic files.

jmuk:


jmuk:
Note that the generated file is in the same structure as #877 but grpc-loading is not fully delayed because of google-gax's problem.

jmuk:
Cc: @swcloud

quartzmo:
Thanks @jmuk!

blowmage:
🤘

swcloud:
LGTM

Issue #890, opened 2016-09-09
Add clouderrorreporting API using Gapic generated code


swcloud:

quartzmo:
Closing, per @dazuma's [comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/891#issuecomment-246439662) on PR #891.

PR #891, opened 2016-09-09
Add clouderrorreporting API using Gapic generated code

swcloud:
This is for #890 , which provides a Ruby client library directly using Gapic generated code without hand-written layer. So the directory structure will be slightly different from other APIs that have both hand-written code and Gapic generated code such as 'pubsub'. 

@jmuk


blowmage:
Why is this set up as a separate gem? Why not add the GAPIC generated files to google-cloud-logging?

swcloud:
This is a new API called 'cloud errorreporting'. 

https://github.com/googleapis/googleapis/tree/master/google/devtools/clouderrorreporting

It has nothing to do with 'cloud-logging'

hxiong388:
Per one of the recent PRDs, Stackdriver APIs are favored to be published under Stackdriver brand, such as Stackdriver Error Reporting, and deprecated "Google Cloud Error Reporting".

Also, in several discussions and threads, folks were talking about creating separate stackdriver-ruby repo to host all Stackdirver related APIs. I feel like this generated code is probably better suited to go in the new repo.

@hexion

blowmage:
Thanks for the PR. We have one blocking issue and a couple suggestions.

1. The directory name doesn't match the gem name in the .gemspec file. Our release process requires that the directory name and the gem name match. I would strongly suggest changing both to conform to the [RubyGems naming convention](http://guides.rubygems.org/name-your-gem/) and use the name `google-cloud-error_reporting`.
2. Would it be possible to change the code generation process to output `Google::Cloud::ErrorReporting` as opposed to `Google::Cloud::Errorreporting`?
3. Would it be possible to change the code generation process to output `Google::Devtools::CloudErrorReporting` as opposed to `Google::Devtools::Clouderrorreporting`?

dazuma:
Please hold off on this PR. Our plan is to release this code integrated with the Stackdriver integration gem, which will land in a separate repository.

quartzmo:
Thanks for the timely update @dazuma! Will do.

swcloud:
@blowmage Thanks for the suggestions.

We can definitely do 1 and 2.

3 is probably unnecessary since the classes in google/devtools/clouderrorreporting are generated from the proto definitions in the googleapi/googleapi repo. They are mainly used internally by Google::Cloud::ErrorReporting

blowmage:
3 is not that internal. All the ServiceApi input parameters and return values are defined in that namespace. Users will be dealing with those classes and it would be better (IMO) if that constant could be improved. 

I wish there was more control on how protobuf generates the Ruby code. But I can't find anything about options for Ruby like like `java` and `objc` have.

bjwatson:
@omaray Base on our offline discussion with @dazuma, it's my understanding that this code should land here, so that we're keeping all of our client libraries in one place. Is that correct?

omaray:
Sorry for the delay. As things stand right now, we are putting all client library code in google-cloud repos. My goal is to resolve this issue asap with the Stackdriver stakeholders (of client library versus framework integrations).

swcloud:
What's the conclusion from the discussion with Stackdriver stakeholders? Shall I continue to deliver the errorreporting code here?

swcloud:
The latest update is the same as https://github.com/googleapis/api-client-staging/tree/master/generated/ruby/google-cloud-ruby/google-cloud-error_reporting

hxiong388:
Hi Song,

This Error Reporting VTK library looks pretty good. The Google Cloud vs
Stackdriver packaging strategy conversation is still going, but I think
it's slowly converging to the solution of having everything named
google-cloud and put under google-cloud-ruby repo. I'm currently working on
rebuilding the Error Reporting Veneer library back to using
google-cloud-error_reporting name. That said, I think we should sync up
offline on couple action items:
1. VTK library and Veneer library release date, and when to update the
Veneer library to utilize the VTK API.
2. Coordinate the releasing of the first google-cloud-error_reporting gem
on rubygems.org, so it includes everything from VTK and Veneer.

Thanks,
Heng Xiong

swcloud:
@blowmage @quartzmo For error reporting API, please merge this PR first. @hxiong388 pushed another PR #961 which will be similar to the veneer layer. In other words, PR #961 will also have a wrapper that calls into the GAPIC layer in this PR. This is similar to what you have been doing for APIs like pubsub.

blowmage:
@swcloud I think we are waiting for a passing build, right?

swcloud:
@blowmage I just added a Rakefile with empty tasks in order to pass the build. In the auto-generated code, we don't have the assumption on RoboCop or Yard. But PR #961 does have these components defined.

PR #892, opened 2016-09-11
Make Job#wait_until_done! flexible

Altech:
The default interval of Job#wait_until_done! is 5 seconds at least. I think it is a little long in some cases. So I just move the degrees to the arguments.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
@@ -239,8 +239,8 @@ module Google
         #                               format: "json"
         #   extract_job.wait_until_done!
         #   extract_job.done? #=> true
-        def wait_until_done!
-          backoff = ->(retries) { sleep 2 * retries + 5 }
+        def wait_until_done!(base: 2, offset: 5)
+          backoff = ->(retries) { sleep base * retries + offset }
           retries = 0
           until done?
             backoff.call retries
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
@@ -239,7 +239,7 @@ module Google
         #                               format: "json"
         #   extract_job.wait_until_done!
         #   extract_job.done? #=> true
-        def wait_until_done!(base: 2, offset: 5)
+        def wait_until_done! base: 2, offset: 5
           backoff = ->(retries) { sleep base * retries + offset }
           retries = 0
           until done?
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
@@ -227,6 +227,11 @@ module Google
         # Refreshes the job until the job is `DONE`.
         # The delay between refreshes will incrementally increase.
         #
+        # @param [Integer] min The minimum time in seconds for waiting.
+        # @param [Integer] max The maximum time in seconds for waiting.
+        #   The method call will time out and return within the time.
+        #   The default value is infinity, so will not time out.
+        #
         # @example
         #   require "google/cloud"
         #
@@ -239,12 +244,16 @@ module Google
         #                               format: "json"
         #   extract_job.wait_until_done!
         #   extract_job.done? #=> true
-        def wait_until_done! base: 2, offset: 5
-          backoff = ->(retries) { sleep base * retries + offset }
+        def wait_until_done! min: 5, max: Float::INFINITY
+          backoff = ->(retries) { 2 * retries + min }
           retries = 0
+          total = 0
           until done?
-            backoff.call retries
+            interval = backoff.call retries
+            return unless total + interval < max
+            sleep interval
             retries += 1
+            total += interval
             reload!
           end
         end


Altech:
@googlebot I signed it!

blowmage:
@Altech thanks for the PR!

@quartzmo what do you think of exposing more of the wait internals? Do we need a `min` or `max` value to wait? The current implementation will keep increasing the wait time indefinitely.

quartzmo:
I like the PR, and I also think adding `max` is a great idea.

@Altech Let us know if you have any trouble understanding how to fix the Rubocop error:

```Running RuboCop...
Inspecting 23 files
......................C
Offenses:
lib/google/cloud/bigquery/job.rb:242:29: C: Use def without parentheses.
        def wait_until_done!(base: 2, offset: 5)
                            ^^^^^^^^^^^^^^^^^^^^
23 files inspected, 1 offense detected
RuboCop failed!
```

blowmage:
I can see having a total max wait, like 24 hour, but I can also see having an interval max, like every 5 minutes. Currently, the delay between checks will increment indefinitely.

quartzmo:
Yes, both sound good.

quartzmo:
@Altech Can you remove the parentheses from the method definition so that Rubocop passes?

Altech:
@quartzmo @blowmage Thanks for the comments! I just removed the parentheses.

I also think `min`/`max` approach is better abstraction.  So I implemented it while keeping the original behavior. What do you think of this?

google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
```diff
@@ -239,12 +244,16 @@ def reload!
         #                               format: "json"
         #   extract_job.wait_until_done!
         #   extract_job.done? #=> true
-        def wait_until_done!
-          backoff = ->(retries) { sleep 2 * retries + 5 }
+        def wait_until_done! min: 5, max: Float::INFINITY
+          backoff = ->(retries) { 2 * retries + min }
```
blowmage: Since we are allowing these values to be set, we still have the magical number `2` which is the multiplier. I think this should also be exposed, so users can determine how they want the delay to increase.

google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
```diff
@@ -239,12 +244,16 @@ def reload!
         #                               format: "json"
         #   extract_job.wait_until_done!
         #   extract_job.done? #=> true
-        def wait_until_done!
-          backoff = ->(retries) { sleep 2 * retries + 5 }
+        def wait_until_done! min: 5, max: Float::INFINITY
```
blowmage: We have two "max" values that we've discusses to far. The one I'd like to see here is the max delay value. The maximum time between refresh calls.

The other "max" value we've discussed is the total amount of time until an error is wait until the job is completed. But we don't have this currently and haven't discussed the behavior of this value. In the current implementation you are just returning and therefore the job is not yet complete. So users will have to check the job status. I think this is wrong. If we have this value then it should raise an error, and raise it exactly when the timeout is reached. I would skip this max timeout value for now, since implementing it will likely take some additional effort.

blowmage: I _think_ I would like to see the method signature be the following:

```ruby
def wait_until_done! initial_delay: 5, delay_multiplier: 2, max_delay: 300
```

The current implementation could be considered to be the following:

```ruby
def wait_until_done! initial_delay: 5, delay_multiplier: 2, max_delay: Float::INFINITY
```

But I think it would be better to cap the max delay at 5 minutes. @quartzmo, what do you think?

blowmage:
Made some comments on the code. It would be really nice to have some additional test coverage ensuring that these values are being properly used. If this is something you can contribute you can use the [`it "can wait until done"`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/ff7d9ba2adff6b6d3b9cd75d6d655a5dccc4df93/google-cloud-bigquery/test/google/cloud/bigquery/job_test.rb#L173-L201) test as a basis.

google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
quartzmo: I like all your suggestions @blowmage. And I'm really glad @Altech opened this PR.

@Altech If this is going beyond what you had time for, you can also close this PR and reference it in a new issue, and we will do the work. We really appreciate your effort so far!

blowmage:
Looking at the code changes in the upcoming google-gax release, it looks like we can standardize on the following 4 values to control this behavior:

* Initial Delay - How long to wait before retrying. Default value is 10 seconds.
* Delay Multiplier - How much to increase the delay between retries. Default value is 1.3.
* Maximum Delay - The maximum amount of time to wait between retries. Default value is 5 minutes.
* Total Timeout - The maximum amount of time to wait for the job to be done. Default value is 1 hour.

I think the new signature of this method should be the following:

```ruby
def wait_until_done! delay: 10, multiplier: 1.3, max_delay: 300, timeout: 3600
```

blowmage:
The BigQuery team has asked for changes to this functionality (per GoogleCloudPlatform/google-cloud-ruby#1626) and so we are closing this PR.

Issue #226, opened 2015-08-19
Do we test our example code anywhere?


jgeewax:
Example code (`@example` in our doc strings) needs automated testing. This is done with [yard-doctest](https://github.com/p0deje/yard-doctest) and a new `doctest` Rake task.

Done for the following core packages:

- [x] BigQuery
- [x] Datastore
- [x] Logging 
- [x] Storage

Also needs to be completed soon for the ML packages:

- [x] Language
- [x] Speech
- [x] Translation
- [x] Vision 

Followed by Pub/Sub, which was originally in the core group:

- [ ] Pub/Sub

And finally, for these remaining packages:

- [x] DNS
- [ ] Error Reporting
- [ ] Resource Manager

blowmage:
We've been testing the example code manually. We don't have any automated tests for code examples. There isn't a generally accepted library for automating doc code like python's doctest.

jgeewax:
We thought that too for Node... and tossed something together ourselves.. Can we consider doing that? At least for testing to typos?

/cc @stephenplusplus @ryanseys @callmehiphop

blowmage:
The syntax highlighting and colors won't be applied to the code examples if the code can't be parsed by ruby. I've caught a couple of typos in code examples by just looking at the documentation. :)

There are a couple of projects that attempt to implement a doctest for ruby, but they are all pretty rough. I'll look into it a bit more. I agree it would be nice to catch code examples that can't be parsed by ruby.

quartzmo:
I agree, this would be great!

quartzmo:
Now that we are using YARD, I think extracting the examples and testing that they are valid Ruby should be pretty easy to do.

quartzmo:
Bump! Is yard doc test an option?

jgeewax:
https://github.com/p0deje/yard-doctest

quartzmo:
As discussed offline with @jgeewax, these tests will be unit tests against mocked service implementations.

quartzmo:
Adding a checklist to the description, since this will be done service-by-service.

quartzmo:
In order to test that all examples can be run without error, yard-doctest appears to require modifications to support the following:

1. Execute all examples, including examples without output assertions (`#=>`).
1. Execute output assertions (`#=>`) that occur within blocks (see [comment on PR #969](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/969#issuecomment-254089595)).

[...]

PR #901, opened 2016-09-15
Add example testing to Translate

quartzmo:
This PR adds [yard-doctest](https://github.com/p0deje/yard-doctest) to `google-cloud-translate` in order to test the example code in the Translate documentation. 

The PR contains a `doctest_helper.rb` containing configuration for minitest mocks, and makes a few small changes to the examples for compatibility with yard-doctest (specifically, removing `puts` calls, since they return `nil`.)
 
To run the tests, pull my branch, then:

```sh
cd google-cloud-translate
bundle install
bundle exec yard doctest
```

[refs #226]

### DIFF
--- a/google-cloud-translate/support/doctest_helper.rb
+++ b/google-cloud-translate/support/doctest_helper.rb
@@ -1,0 +1,176 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "google/cloud/translate"
+
+# TODO: Pull up for reuse across services
+module Google
+  module Cloud
+    def self.stub_service name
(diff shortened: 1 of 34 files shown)


blowmage:
Woo!

quartzmo:
Successful output looks like this:

```sh
google-cloud-translate quartzmo$ bundle exec yard doctest
Run options: --seed 53351

# Running:

..............

Finished in 0.016305s, 858.6107 runs/s, 2759.8200 assertions/s.

14 runs, 45 assertions, 0 failures, 0 errors, 0 skips
```

It's important to note that if an example lacks assertions (in the form of `#=>` followed by expected return value), it will appear in "runs", but the code won't actually be executed.

blowmage:
I'm loving how you got multiple code examples to work.

google-cloud-translate/support/doctest_helper.rb
```diff
@@ -0,0 +1,176 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "google/cloud/translate"
+
+# TODO: Pull up for reuse across services
```
quartzmo: @blowmage Any ideas about to reuse this code across service gems?

blowmage: I would duplicate it in each gem, similar to how all the rake tasks are duplicated in each gem.

quartzmo: OK, so it's good as-is, I should remove the TODO?

Issue #893, opened 2016-09-12
Code: duplicate files


timanovsky:
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-core/lib/google/cloud/core/backoff.rb 
and 
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-core/lib/google/cloud/core/grpc_backoff.rb

First should probably be deleted

quartzmo:
Thanks for catching this duplication! Our goal is to replace this functionality soon with GAX, at which time we will remove both files.

Issue #759, opened 2016-07-11
Expand acceptance testing


quartzmo:
The current acceptance tests do not provide sufficient coverage. Tests should be added for all services, until we have the confidence to release without additional manual testing after making project-wide changes, such as those in #733 .

quartzmo:
I mistakenly closed this issue in a commit message. Re-opening because the following services still need expanded acceptance testing:

- [ ] Datastore
- [ ] Logging
- [ ] Pub/Sub

bjwatson:
@quartzmo Please scope this more tightly.

quartzmo:
I reviewed the acceptance testing for the three packages listed above (Datastore, Logging, and Pub/Sub), and feel that the coverage is adequate at this time. (Logging is also covered by integration tests.) So I will not be opening more tightly-scoped issues to replace this issue.

Issue #808, opened 2016-07-27
Question / feature request: error handling / retries in Datastore


timanovsky:
**Question**: What is current error handling procedure and retry approach? Documentation says 

> You can configure how many times API requests may be automatically retried. When an API request fails, the response will be inspected to see if the request meets criteria indicating that it may succeed on retry, such as 500 and 503 status codes or a specific internal error code such as rateLimitExceeded. If it meets the criteria, the request will be retried after a delay. If another error occurs, the delay will be increased before a subsequent attempt, until the retries limit is reached.

However from the code (`service.rb`, `backoff.rb`) I see that operation is only retried in case of grpc code 14. What is this code? It is not documented. Does it cover rate limit back off? What about network errors?

In the past with API v2 I used the following list of exception to retry the operation (in my wrapper):

``` ruby
    repeatable = exception.is_a?(Faraday::TimeoutError) || exception.is_a?(Faraday::ConnectionFailed)
    repeatable ||= exception.is_a?(Faraday::SSLError)
    repeatable ||= exception.try(:response).present? &&
                   exception.response.try(:env).present? &&
                   ((500..599).include?(exception.response.env.status) || [401, 409].include?(exception.response.env.status))
    repeatable ||= exception.is_a?(SocketError)
    repeatable ||= exception.is_a?(Errno::EPIPE) || exception.is_a?(Errno::ECONNRESET) || exception.is_a?(Errno::ECONNABORTED)
```

Which all have been observed during running on production and all proven to be retriable. 

**Feature request**: Make `retries` configurable per request. There are different types of requests in application which require different handling of errors/retries. For example if error happening during web page load I'd better show error to user and let him retry rather than render page a minute later after background retries. Some operations are ok to fail as we can retry higher level application level transaction. But some operations are critical and must be retried heavily. So optional `retries` parameter to all API calls would be helpful.

quartzmo:
@timanovsky Thank you for yet another well-written and thoughtful issue. Here are some supporting links: First, [the retry handling for gRPC](https://github.com/GoogleCloudPlatform/gcloud-ruby/blob/v0.12.0/lib/gcloud/backoff.rb#L114), which as you note, only tests for one error code: `self.grpc_codes = [14]` (which is `UNAVAILABLE `.)

[...]

PR #903, opened 2016-09-16
Convert Datastore to GAPIC

quartzmo:
This PR converts all Datastore requests to the GAX client introduced in #884.

There is a breaking change: The `retries` parameter is no longer supported (although method signatures have not changed.) See #848 and #849 for proposals to add new configuration options for retry behavior.

Also, Datastore is one of the three services that still requires expanded acceptance testing (#759). I could add the needed tests as part of this PR if desired.

### DIFF
--- a/google-cloud-datastore/lib/google-cloud-datastore.rb
+++ b/google-cloud-datastore/lib/google-cloud-datastore.rb
@@ -38,8 +38,9 @@ module Google
     #   The default scope is:
     #
     #   * `https://www.googleapis.com/auth/datastore`
-    # @param [Integer] retries This option is not currently supported.
     # @param [Integer] timeout Default timeout to use in requests. Optional.
+    # @param [Hash] client_config A hash of values to override the default
+    #   behavior of the API client. See Google::Gax::CallSettings. Optional.
     #
     # @return [Google::Cloud::Datastore::Dataset]
     #
@@ -65,10 +66,10 @@ module Google
     #   platform_scope = "https://www.googleapis.com/auth/cloud-platform"
     #   datastore = gcloud.datastore scope: platform_scope
     #
-    def datastore scope: nil, retries: nil, timeout: nil
+    def datastore scope: nil, timeout: nil, client_config: nil
       Google::Cloud.datastore @project, @keyfile,
-                              scope: scope, retries: retries,
-                              timeout: (timeout || @timeout)
+                              scope: scope, timeout: (timeout || @timeout),
+                              client_config: client_config
     end
 
     ##
@@ -90,8 +91,9 @@ module Google
     #   The default scope is:
     #
     #   * `https://www.googleapis.com/auth/datastore`
-    # @param [Integer] retries This option is not currently supported.
     # @param [Integer] timeout Default timeout to use in requests. Optional.
+    # @param [Hash] client_config A hash of values to override the default
+    #   behavior of the API client. See Google::Gax::CallSettings. Optional.
     #
     # @return [Google::Cloud::Datastore::Dataset]
     #
@@ -106,16 +108,16 @@ module Google
     #     t["done"] = false
     #     t["priority"] = 4
     #     t["description"] = "Learn Cloud Datastore"
-    #   end
+    #   end                                                                  ``
     #
     #   datastore.save task
     #
-    def self.datastore project = nil, keyfile = nil, scope: nil, retries: nil,
-                       timeout: nil
+    def self.datastore project = nil, keyfile = nil, scope: nil, timeout: nil,
+                       client_config: nil
       require "google/cloud/datastore"
       Google::Cloud::Datastore.new project: project, keyfile: keyfile,
-                                   scope: scope, retries: retries,
-                                   timeout: timeout
+                                   scope: scope, timeout: timeout,
+                                   client_config: client_config
     end
   end
 end
--- a/google-cloud-datastore/lib/google-cloud-datastore.rb
+++ b/google-cloud-datastore/lib/google-cloud-datastore.rb
@@ -38,8 +38,7 @@ module Google
     #   The default scope is:
     #
     #   * `https://www.googleapis.com/auth/datastore`
-    # @param [Integer] retries Number of times to retry requests on server
-    #   error. The default value is `3`. Optional.
+    # @param [Integer] retries This option is not currently supported.
     # @param [Integer] timeout Default timeout to use in requests. Optional.
     #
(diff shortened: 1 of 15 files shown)


quartzmo:
This PR should also remove files that are no longer needed for backoff, to close #893.

blowmage:
Not all gems are converted to GAX, so the backoff code should stay until they are all converted.

quartzmo:
Oh right, I mistakenly thought it wasn't shared. Of course.

quartzmo:
The [fourth commit](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/903/commits/87178674eaf1914001e794acbc0d8d7c15c0503b) on this PR removes `retries` (no longer supported) from the factory methods, and adds a new `client_config` parameter accepting a hash of `google-gax` [`client_config`](http://www.rubydoc.info/gems/google-gax/0.5.1/Google%2FGax.construct_settings) values. Our intention with this change is to greatly increase client control over retry, timeout, and error handling behaviors, addressing issues such as #808.

Here is the default  [`datastore_client_config.json`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-datastore/lib/google/cloud/datastore/v1/datastore_client_config.json), which can be overridden via the new parameter:

```js
{
  "interfaces": {
    "google.datastore.v1.Datastore": {
      "retry_codes": {
        "retry_codes_def": {
          "idempotent": [
            "DEADLINE_EXCEEDED",
            "UNAVAILABLE"
          ],
          "non_idempotent": []
        }
      },
      "retry_params": {
        "default": {
          "initial_retry_delay_millis": 100,
          "retry_delay_multiplier": 1.3,
          "max_retry_delay_millis": 60000,
          "initial_rpc_timeout_millis": 60000,
          "rpc_timeout_multiplier": 1.0,
          "max_rpc_timeout_millis": 60000,
          "total_timeout_millis": 600000
        }
      },
      "methods": {
        "Lookup": {
          "timeout_millis": 60000,
          "retry_codes_name": "idempotent",
          "retry_params_name": "default"
        },
        "RunQuery": {
          "timeout_millis": 60000,
          "retry_codes_name": "idempotent",
          "retry_params_name": "default"
        },
        "BeginTransaction": {
          "timeout_millis": 60000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        },
        "Commit": {
          "timeout_millis": 60000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        },
        "Rollback": {
          "timeout_millis": 60000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        },
        "AllocateIds": {
          "timeout_millis": 60000,
          "retry_codes_name": "non_idempotent",
          "retry_params_name": "default"
        }
      }
    }
  }
}
``` 

As you can see, there are some interesting fields, such as `retry_codes`, which suggests the ability to configure retry logic based on error status codes. At this time there is not much documentation for the [`client_config`](http://www.rubydoc.info/gems/google-gax/0.5.1/Google%2FGax.construct_settings) fields, but @jmuk from `google-gax` is working with us and should be able to answer questions.

Please note that this leaks the GAPIC implementation to the `google-cloud-datastore` public API. If `google-gax` changes its implementation, this may result in a change to the `google-cloud-datastore` public API.

@timanovsky @bmclean Any feedback on this proposal? We hope to merge and release this PR next week.

google-cloud-datastore/lib/google-cloud-datastore.rb
```diff
@@ -38,8 +38,9 @@ module Cloud
     #   The default scope is:
     #
     #   * `https://www.googleapis.com/auth/datastore`
-    # @param [Integer] retries This option is not currently supported.
     # @param [Integer] timeout Default timeout to use in requests. Optional.
+    # @param [Hash] client_config A hash of values to override the default
+    #   behavior of the API client. See Google::Gax::CallSettings. Optional.
```
bmclean: It isn't obvious that the parameters are in Google::Gax::BackoffSettings.

bmclean:
@quartzmo Please correct me if I have any of this wrong:
For a datastore Lookup, since it has a retry_codes_name defined of "idempotent", if an UNAVAILABLE error was raised, it would retry using the default params.
For a datastore Commit,  since it has a retry_codes_name defined of "non_idempotent", any error raised would not be retried.

blowmage:
@bmclean I'll take a swing at answering your question. Yes, your understanding is correct. That is the default behavior. We asked several times which endpoints we should retry and which ones we should not. But we never got an answer so we did a blanket retry on them all. I believe that this new behavior is correct.

Of course, the good news is that you can now override how each endpoint is treated and use whatever behavior you wish. :)

google-cloud-datastore/lib/google-cloud-datastore.rb
```diff
@@ -108,12 +108,12 @@ def datastore scope: nil, retries: nil, timeout: nil
     #     t["done"] = false
     #     t["priority"] = 4
     #     t["description"] = "Learn Cloud Datastore"
-    #   end
+    #   end                                                                  ``
```
blowmage: What is this? 😆

google-cloud-datastore/lib/google/cloud/datastore/service.rb
```diff
@@ -170,6 +141,13 @@ def generate_read_options consistency, transaction
           end
           nil
         end
+
+        def execute
+          require "grpc" # Ensure GRPC is loaded before rescuing exception
+          yield
+        rescue GRPC::BadStatus => e
```
blowmage: Need to rescue `Google::Gax::GaxError` here.

bmclean:
@blowmage What do you mean by "blanket retry on them all"?

blowmage:
Correct. Not all errors, and not all endpoints should be retried. Of course, you are free to override the default configuration and make it behave as you want.

bmclean:
Thanks. I agree that the proposed default config is correct. For Lookup and RunQuery, if datastore returns either a DEADLINE_EXCEEDED or UNAVAILABLE error the request is retried as per the retry configuration. For all other endpoints, the default is not to invoke the retry logic.

blowmage:
This new configuration comes from Google. It has been hard for those of us outside of Google to get this information, so we are happy to finally get this level of guidance. :smile:

quartzmo:
@blowmage I believe I've updated this PR with everything it needs to match the merged GAPIC conversions for Vision and Pub/Sub. PTAL!

quartzmo:
oops wrong comment button

PR #904, opened 2016-09-16
Update language acceptance tests

blowmage:
Running the extract_document_sentiment feature used to not return sentences, but now it does.
This looks to be a change in the behavior of the API.
Update the acceptance tests to match the behavior of the API.


quartzmo:
I guess we need to just go with the flow! Thanks @blowmage!

PR #920, opened 2016-09-25
Add acceptance tests to AppVeyor

blowmage:
This change should allow for acceptance tests to run on AppVeyor, but only on the master branch. Instead of using the encrypted keyfile that Travis-CI uses, the keyfile contents should be added to the `GCLOUD_TEST_KEYFILE_JSON` environment variable.


quartzmo:
Oh boy, acceptance tests on Windows! Thank you @blowmage.

PR #922, opened 2016-09-26
Initial Speech implementation

blowmage:
I'm breaking the speech implementation apart into separate PRs so we can review them on GitHub. This initial PR has the basic boilerplate and implementation for the synchronous and asynchronous calls.


blowmage:
I cleaned up a good deal of churn for the PR, but I may have missed some rough edges. Future PRs will build on this base. Let me know what you think.

quartzmo:
`Project#recognize` and `Project#recognize_job` are so similar, with the exact same arguments, it seems that if one or the other is unlikely to be used much, it could be implemented as an option in a unified method. I have my doubts about the usefulness of the synchronous recognition after reading [this](https://cloud.google.com/speech/docs/basics):

> The Speech API will take roughly the same amount of time to process audio data sent synchronously as the duration of the supplied audio data. That is, if you send audio data of 30 seconds in length, expect the synchronous request to take approximately 30 seconds to return results.

quartzmo:
I'm going to merge this now as a base, so that we can keep going!

PR #925, opened 2016-09-27
Update vkit generated code for pubsub

swcloud:
update to the latest vkit auto-generated code for pubsub


blowmage:
Wonderful! Glad to have a GAPIC layer for the IAM calls! Thanks!

PR #932, opened 2016-09-27
Add Project#projects to BigQuery

### LINKED ISSUE #911: BigQuery client doesn't implement Projects.list
As far as I can tell, google-cloud-bigquery does not implement Projects.list

API reference: https://cloud.google.com/bigquery/docs/reference/v2/projects/list
Documentation: https://cloud.google.com/bigquery/docs/managing_jobs_datasets_projects#projects
Endpoint: https://www.googleapis.com/bigquery/v2/projects

For reference, see similar issue filed and fixed for google-cloud-python:
https://github.com/GoogleCloudPlatform/google-cloud-python/issues/2143

quartzmo:
The class name `Project` is already taken, so I used `ProjectReference` instead, which may or may not be the best choice. I'm open for suggestions and am happy to change it.

[closes #911]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
@@ -20,6 +20,7 @@ require "google/cloud/bigquery/credentials"
 require "google/cloud/bigquery/dataset"
 require "google/cloud/bigquery/job"
 require "google/cloud/bigquery/query_data"
+require "google/cloud/bigquery/project/list"
 
 module Google
   module Cloud
@@ -35,7 +36,12 @@ module Google
       # Google BigQuery. {Google::Cloud::Bigquery::Dataset} objects are created,
       # accessed, and deleted by Google::Cloud::Bigquery::Project.
       #
-      # See {Google::Cloud#bigquery}
+      # See {Google::Cloud#bigquery}.
+      #
+      # @attr_reader [String, nil] name The descriptive name of the project.
+      #   Can only be present if the project was retrieved with {#projects}.
+      # @attr_reader [Integer, nil] numeric_id The numeric ID of the project.
+      #   Can only be present if the project was retrieved with {#projects}.
       #
       # @example
       #   require "google/cloud"
@@ -49,6 +55,8 @@ module Google
         ##
         # @private The Service object.
         attr_accessor :service
+
+        attr_reader :name, :numeric_id
 
         ##
         # Creates a new Service instance.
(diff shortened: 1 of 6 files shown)


blowmage:
Just an idea, but what do you think about `Project::Reference` instead of `ProjectReference`?

quartzmo:
Can you think of anything that doesn't use "Reference"? In Datastore, the naming convention `<entity>Reference` is consistently used for unique identifier combos, not representations that include things like name...

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -468,6 +469,46 @@ def jobs all: nil, token: nil, max: nil, filter: nil
...
+        #
+        #   gcloud = Google::Cloud.new
+        #   bigquery = gcloud.bigquery
+        #
+        #   projects = bigquery.projects
+        #   projects.all do |project|
+        #     puts project.project_id
+        #   end
+        #
+        def projects all: nil, token: nil, max: nil
```
blowmage: Why is `all` an optional named argument? Its not listed in the documentation. The value is sent to `list_projects`, but `list_projects` does not use it.

blowmage:
I think I am following what you are asking. The Bigquery service API uses attributes like `datasetReference` and `tableReference`. So you think there is the potential for a future conflict if we use "Reference" in the name of this new construct. Is that correct?

quartzmo:
Yes, exactly, so I'd like to avoid using `ProjectReference` because this construct also includes `friendlyName ` (`name`). Any ideas? `Project::Info`??

blowmage:
We have something similar in Pub/Sub, where there is a partial and full representation for a Topic. We've chosen to hide the fact that we don't have a full resource, and lazily load the full resource when needed. But Logging uses "Descriptor" to denote this type of structure.

I don't think the risks are very high with using "Reference", but "Resource" or "Representation". I don't love "Info" since there is likely more information that can be provided for a Project.

blowmage:
I would love to see a `ProjectReference#project` method that returns an active `Project` object, so you can query the datasets and tables and views easily. Or, you know, have the list just return `Project` objects...

quartzmo:
Brilliant. Let's try for that. Then this will be possible:

```ruby
require "google/cloud"

gcloud = Google::Cloud.new
bigquery = gcloud.bigquery

projects = bigquery.projects

projects.all do |project|
  puts project.project_id
  project.datasets.each do |dataset|
    puts dataset.name
  end
end
```

I'm working on it now.

quartzmo:
@blowmage If we don't check for and exclude the current project from the list, it will also be included. Thoughts?

blowmage:
We should return all projects that the API returns.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -35,7 +36,15 @@ module Bigquery
       # Google BigQuery. {Google::Cloud::Bigquery::Dataset} objects are created,
       # accessed, and deleted by Google::Cloud::Bigquery::Project.
       #
-      # See {Google::Cloud#bigquery}
+      # See {Google::Cloud#bigquery}.
+      #
+      # @attr_reader [String, nil] name The descriptive name of the project.
+      #   Only present if the project was retrieved with {#projects}.
```
blowmage: We always have project_id.

quartzmo:
@blowmage Please note that the current project is *also* returned in the list. If you think this is the wrong behavior, I can add logic to skip adding it to the list.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
quartzmo: You meant `project.service.project`?

quartzmo: Oh i get it, always present. Thank you, updating.

quartzmo: I think it's best to remove `Project#project_id` and just rely on the existing `Project#project` instead, do you agree?

quartzmo: Updated, removed `Project#project_id`.

PR #936, opened 2016-09-27
Update vkit generated code for speech

swcloud:


blowmage:
Nice! It will be nice to have the option to expose `AsyncRecognizeMetadata` on long running operations.

PR #938, opened 2016-09-28
Add autodetect to bigquery load

AlfonsoUceda:
Hi folks:

I'm adding this experimental option because I need to [autodetect](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load.autodetect) the format for my AVRO files.

When I load my files I need to upload my schema too because some tables are dynamic so I need this feature. I open this PR because I don't know if you give support for experimental features or maybe because I'm doing something wrong.

The documentation says when `configuration.load.writeDisposition` is WRITE_TRUNCATE the schema is overwritten, I saw it in [`configuration.load.schemaUpdateOptions`](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load.schemaUpdateOptions) but when I load avro file and I use truncate for `write` option, the schema doesn't change :(
> For normal tables, WRITE_TRUNCATE will always overwrite the schema

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
@@ -322,7 +322,8 @@ module Google
             encoding: options[:encoding], field_delimiter: options[:delimiter],
             ignore_unknown_values: options[:ignore_unknown],
             max_bad_records: options[:max_bad_records], quote: options[:quote],
-            schema: options[:schema], skip_leading_rows: options[:skip_leading]
+            schema: options[:schema], skip_leading_rows: options[:skip_leading],
+            autodetect: options[:autodetect]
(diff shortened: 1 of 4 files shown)


blowmage:
Hi @AlfonsoUceda, thanks for the PR. Unfortunately the current policy is to not support experimental options, as they can be removed with no warning. So we can't say with any certainty when this can be accepted. PR #720 is in the same situation with `use_legacy_sql`.

AlfonsoUceda:
Hi @blowmage thanks for your answer, no problem, I was testing AVRO in my code and I understand don't support experimental features ;) so I you don't mind, I'll close the PR for now.

Cheers!

quartzmo:
@AlfonsoUceda Thank you for understanding. If it were not for this situation, we would have loved to accept your PR!

Issue #939, opened 2016-09-28
Way to create schema during the table's load


AlfonsoUceda:
Hi:

I'd like to know if there is a way to create the schema atomically when you load the file, an example is the google-cloud-sdk with the bq binary:

`bq load ds.new_tbl ./info.csv ./info_schema.json`
or you can use the `--schema` option:
`bq load ds.new_tbl ./info.csv --schema=./info_schema.json`

I ask this because this operation is atomic and I need this behaviour with the gem, why? some of my tables are dynamic.

Cheers

AlfonsoUceda:
I have a fork adding `schema` option to `#load` table method and I could do the following:
```ruby
csv_file = File.open("test.csv", "r")

schema = Google::Cloud::Bigquery::Schema.from_gapi(nil)
schema.string "first_name", mode: :nullable
schema.string "last_name", mode: :nullable
schema.integer "age", mode: :nullable
gapi_schema = schema.to_gapi

table.load csv_file, write: 'truncate', schema: gapi_schema
```
This works, I can modify the schema, the CSV and I can add more fields and works perfect but it isn't an elegant solution in my opinion.

Are you opened to accept a PR for adding `schema` options in `#load` table method?

quartzmo:
@AlfonsoUceda Yes! I see no reason why not to support [`configuration.load.schema`](https://cloud.google.com/bigquery/docs/reference/v2/jobs#configuration.load). (And apologies that it slipped by us!) Please proceed with your PR and let us know if we can help in any way.

blowmage:
One way we specify schema in `#create_table` is to specify the schema's fields:

```ruby
require "google/cloud"

gcloud = Google::Cloud.new
bigquery = gcloud.bigquery
dataset = bigquery.dataset "my_dataset"

schema_fields = [
  Google::Cloud::Bigquery::Schema::Field.new(
    "first_name", :string, mode: :required),
  Google::Cloud::Bigquery::Schema::Field.new(
    "cities_lived", :record, mode: :repeated
    fields: [
      Google::Cloud::Bigquery::Schema::Field.new(
        "place", :string, mode: :required),
      Google::Cloud::Bigquery::Schema::Field.new(
        "number_of_years", :integer, mode: :required),
      ])
]
table = dataset.create_table "my_table", fields: schema_fields
```

Can we make `Table#load` use the same API?

```ruby
schema_fields = [
  Google::Cloud::Bigquery::Schema::Field.new(
[...]

PR #941, opened 2016-09-29
Add schema option to load method in table for bigquery

AlfonsoUceda:
Please for more info, see issue #939 

There aren't tests yet, I need to be sure if this feature could be added. I tested with a real project and it works perfect.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
@@ -21,6 +21,7 @@ require "google/cloud/bigquery/dataset"
 require "google/cloud/bigquery/job"
 require "google/cloud/bigquery/query_data"
 require "google/cloud/bigquery/project/list"
+require "google/cloud/bigquery/schema"
 
 module Google
   module Cloud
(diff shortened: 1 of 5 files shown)


quartzmo:
👍  Cool!

AlfonsoUceda:
@quartzmo @blowmage I've made the PR, can you check when you have time and share your opinions? thanks guys.

I used the schema approach instead the fields one because I think build a schema with its DSL is less verbose than using `Field` class.

AlfonsoUceda:
@quartzmo @blowmage what do you think about this? thanks!

blowmage:
@AlfonsoUceda I think we want something slightly different. I [commented](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/939#issuecomment-251741124) on your original issue. If you are interested we would love to accept another PR from you on this. If not, we can implement it. Just let us know.

AlfonsoUceda:
@blowmage I've updated the PR.

About your [comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/939#issuecomment-251741124) with this `Project#create_table` you'd want to say `Dataset#create_table`, isn't it? I see you can use a block to config the schema, do you want to have this behaviour and passing the parameter?

About passing `fields` to `Table#load`, do we need to create a Schema from fields? I've seen we can pass fields to low level api in a easy way.

AlfonsoUceda:
@blowmage @quartzmo did you have time to take a look?

blowmage:
Closing for reasons outlined on #939.

PR #947, opened 2016-09-30
Speech Updates

blowmage:
Revises some class names, adds some convenience methods, and adds basic documentation.


quartzmo:
@blowmage This looks very solid. What do you feel is still outstanding for releasing Speech? Can you provide a checklist either here or on the original issue? Should I merge this PR now and then open one of my own for guides and higher-level docs?

blowmage:
Mostly docs. We need the overview documentation where we can talk about how to convert media to the formats used by Speech.

PR #948, opened 2016-09-30
Docs fixes

quartzmo:
The first two commits fix reported docs issues and should be merged before any release of new docs.

The third commit is a work in progress (with `TODO` steps for completion), that can be merged now, since in the current version these tasks are broken anyway.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -223,8 +223,8 @@ namespace :jsondoc do
     puts `git clone --quiet --branch=gh-pages --single-branch #{git_repo} #{gh_pages} > /dev/null`
   end
 
-  desc "Copies jsondoc to gh-pages repo in temp dir."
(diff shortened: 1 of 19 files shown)


Rakefile
```diff
@@ -320,13 +327,21 @@ namespace :jsondoc do
...
+    Rake::Task["jsondoc:umbrella"].invoke("master")
     Rake::Task["jsondoc:publish"].invoke("master")
   end
 
   desc "Publishes the jsondoc for the tag to the gh-pages branch"
   task :tag, :tag do |t, args|
     tag = args[:tag]
     fail "Missing required parameter 'tag'." if tag.nil?
+
+    fail "'tag' must be in the format <gem>/<version>" unless tag.include?("/")
```
blowmage: Looks like the right approach to me.

Issue #931, opened 2016-09-27
Storage object with spaces fails signed_url signature match


gsbucks:
Hi. I'm on google-cloud-storage 0.20.1 and just narrowed down what I think is an issue, but could be an error on my part.

When trying to view generated signed_urls, they were failing with SignatureDoesNotMatch, and returning a StringToSign that had the spaces in the object name escaped to %20. If I printed out the return value of Google::Cloud::Storage::File::Signer#signature_string it showed the same string but with actual spaces in it.

Manipulating the gem to escape the spaces inside the signature_string method did cause the signed_url method to return a url that opened my file as expected. Seems like the server expects the string being signed to be URL escaped prior to signing and that's not happening?

Actual object names that caused it:
/qa_photos/2102606-01,%201,%20RKT-16340017/step-11.jpg
/qa_photos/2102606-01, 01, RKT-16340017/step-11.jpg

blowmage:
Thanks for opening this issue and making us aware of this, @gsbucks. Hopefully we'll get a fix out for you this week.

blowmage:
Hi @gsbucks, I believe this is fixed in google-cloud-storage version 0.20.2. All my tests show this now working. Can you verify the new release fixes this for you. If it does, please close the issue.

Thanks again for opening the issue and reporting the problem!

gsbucks:
Looks good. Thanks for the quick turnaround.

PR #949, opened 2016-09-30
Properly escape path in Storage signed_url

blowmage:
This addresses a bug where the string used in the signature is not the same
as the string used on the server to grant access. This is because browsers
will escape characters like space. Escape before signing so the values match.

[refs #931]


quartzmo:
@blowmage Thanks for getting to this, I'll merge when green.

PR #950, opened 2016-09-30
Add client_config to Speech factory methods

blowmage:
Add client_config to Speech since it uses GAPIC/GAX, and remove non-functional retries.


quartzmo:
Appreciate the quick action @blowmage! LGTM, will merge when green.

PR #958, opened 2016-10-07
Switch Vision to use GAPIC/gRPC/GAX

blowmage:
Move Vision from the JSON API to the gRPC API, using GAPIC as provided by GAX.

### DIFF
--- a/google-cloud-vision/lib/google/cloud/vision/service.rb
+++ b/google-cloud-vision/lib/google/cloud/vision/service.rb
@@ -15,7 +15,7 @@
 
 require "google/cloud/errors"
 require "google/cloud/vision/version"
-require "google/apis/vision_v1"
+require "google/cloud/vision/v1"
 
 module Google
   module Cloud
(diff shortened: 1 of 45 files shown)


google-cloud-vision/lib/google/cloud/vision/service.rb
```diff
@@ -66,9 +77,13 @@ def inspect
         protected
 
         def execute
+          require "grpc" # Ensure GRPC is loaded before rescuing exception
           yield
-        rescue Google::Apis::Error => e
+        rescue GRPC::BadStatus => e
```
quartzmo: Is this `GRPC::BadStatus` needed? I didn't add it in conversions for Datastore and Pub/Sub. I can, however.

google-cloud-vision/lib/google/cloud/vision/v1/image_annotator_api.rb
```diff
@@ -46,7 +46,7 @@ class ImageAnnotatorApi
           # The default port of the service.
           DEFAULT_SERVICE_PORT = 443
 
-          CODE_GEN_NAME_VERSION = "gapic/0.1.0".freeze
+          CODE_GEN_NAME_VERSION = "grpcc/0.1.0".freeze
```
quartzmo: Looks like this was mistakenly changed in a `/s gapi grpc` search and replace!

quartzmo:
In `vision.rb`, there is a docs section "Configuring retries and timeout". At the least, the `retries` mentions need to be removed. If you're feeling brave, you can add docs for `client_config`.

In my upcoming Pub/Sub conversion PR, this is what I have:

```
    # ## Configuring timeout
    #
    # You can configure the request `timeout` value in seconds.
```

google-cloud-vision/lib/google/cloud/vision/service.rb
blowmage: I was trying to be cautious, but not rescuing `GRPC::BadStatus` would actually simplify things a bit. Updated.

google-cloud-vision/lib/google/cloud/vision/v1/image_annotator_api.rb
blowmage: Dang it, I checked for this many, many times. Can't believe this one got through. Updated.

blowmage:
Great catches. I've removed the `retries` documentation.

quartzmo:
This is so awesome! Thanks for figuring out the little details with gRPC requires in `Service`, I will add those to #903 and #959. Merging!

PR #963, opened 2016-10-10
Convert Logging to GAPIC using GAX.

blowmage:
The big difference between Logging and Pub/Sub is in where the GAX PagedEnumerable is handled. In Logging it is handled in Service, but in Pub/Sub it is handled in the *List classes. By handling it in Service we can simplify the mocking needed for the unit tests, but we don't make use of the PagedEnumerable object at all.

@quartzmo, thoughts?


quartzmo:
@blowmage Regarding the handling of PagedEnumerable, I believe you took a better approach with this implementation, and that we should update the Pub/Sub implementation to match. PagedEnumerable strikes me as likely to be a great convenience for end users of GAX, but using it internally in google-cloud libraries seems to add more complexity than is warranted.

quartzmo:
@blowmage I opened a PR against your branch to share some fixes for the acceptance tests. 

https://github.com/blowmage/gcloud-ruby/pull/6

quartzmo:
LGTM! Merging!

PR #964, opened 2016-10-10
Add gapic generated code for long running operation API

swcloud:


swcloud:
Sorry. There is some misunderstanding here. LRO should be part of google-gax gem. Close this PR.

PR #970, opened 2016-10-13
Refactor Google::Cloud::Core::GCE to Google::Cloud::Core::Environment

hxiong388:
Hi Mike, Chris,

For all the instrumentation libraries I'm going to add soon, they require utility functions to check which environment code is running from and retrieve metadata. The existing Google::Cloud::Core::GCE module already does something like this, but only for GCE VM environment and queries GCP project_id. I think it makes sense to re-purpose and expand this module to support all metadata management for various GCP environments.

In this PR, I'm renaming Google::Cloud::Core::GCE to Google::Cloud::Core::Environment. I'm also adding several utility functions: 
* original "gce?" function renamed to "gce_vm?", because this function is essentially checking if environment is on a GCE VM.
* "gae?" funcion checks if running in GAE environment
* "gke?" function checks if running in GKE environment
* new "gce?" function checks if running on GCE deployed environment and make sure it's not GAE or GKE
* Finally, various other metadata retrieving functions that gets information from environment variables or internal metadata service

I'm adding more elaborate integration tests in later PR.

Please let me know what you think.

Thanks,
heng Xiong

### DIFF
--- a/google-cloud-core/lib/google/cloud/core/environment.rb
+++ b/google-cloud-core/lib/google/cloud/core/environment.rb
@@ -20,10 +20,11 @@ module Google
     module Core
       ##
       # @private
-      # Represents the Google Compute Engine environment.
-      module GCE
+      # Represents the Google Cloud Platform environments.
+      module Environment
         CHECK_URI = "http://169.254.169.254"
         PROJECT_URI = "#{CHECK_URI}/computeMetadata/v1/project/project-id"
+        INSTANCE_METADATA_URI = "#{CHECK_URI}/computeMetadata/v1/instance"
 
         def self.gce? options = {}
           conn = options[:connection] || Faraday.default_connection
--- a/google-cloud-core/lib/google/cloud/core/environment.rb
+++ b/google-cloud-core/lib/google/cloud/core/environment.rb
@@ -29,22 +29,29 @@ module Google
         ##
         # Check if running from a gce vm, which actually hosts GCE, GKE, and
         # GAE environments.
-        def self.gce_vm? options = {}
-          conn = options[:connection] || Faraday.default_connection
+        def self.gce_vm? connection: nil
+          @metadata ||= {}
+          return @metadata[:gce_vm] if @metadata.key? :gce_vm
+
+          conn = connection || Faraday.default_connection
           resp = conn.get CHECK_URI do |req|
             req.options.timeout = 0.1
           end
-          return false unless resp.status == 200
-          return false unless resp.headers.key? "Metadata-Flavor"
-          return resp.headers["Metadata-Flavor"] == "Google"
+          @metadata[:gce_vm] = if resp.status != 200 ||
+                                  !resp.headers.key?("Metadata-Flavor")
+                                 false
+                               else
+                                 resp.headers["Metadata-Flavor"] == "Google"
+                               end
         rescue Faraday::TimeoutError, Faraday::ConnectionFailed
-          return false
+          @metadata ||= {}
+          @metadata[:gce_vm] = false
         end
 
         ##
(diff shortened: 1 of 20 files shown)


google-cloud-core/lib/google/cloud/core/environment.rb
```diff
@@ -0,0 +1,141 @@
...
+        ##
+        # Helper method to send HTTP request to GCP metadata service and
+        # retrieve environment information
+        def self.get_metadata_attribute uri, attr_name, connection: nil
+          @metadata ||= {}
+          return @metadata[attr_name] if @metadata.key? attr_name
+
+          conn = connection || Faraday.default_connection
+
+          puts conn.get uri
```
quartzmo: Is this `puts` statement left over from debugging during development? Currently, google-cloud has no `puts` output, except for some retries in acceptance tests.

google-cloud-core/lib/google/cloud/core/environment.rb
```diff
@@ -0,0 +1,141 @@
...
+        # Check if running from a gce vm, which actually hosts GCE, GKE, and
+        # GAE environments.
+        def self.gce_vm? options = {}
+          conn = options[:connection] || Faraday.default_connection
+          resp = conn.get CHECK_URI do |req|
+            req.options.timeout = 0.1
+          end
+          return false unless resp.status == 200
+          return false unless resp.headers.key? "Metadata-Flavor"
+          return resp.headers["Metadata-Flavor"] == "Google"
```
quartzmo: Would it make sense to memoize the result?

google-cloud-core/lib/google/cloud/core/environment.rb
```diff
@@ -0,0 +1,141 @@
...
+        rescue Faraday::TimeoutError, Faraday::ConnectionFailed
+          return false
+        end
+
+        ##
+        # Check if running from Google Compute Engine, and not using GAE or GKE
+        #
+        # @return [Boolean] True iff running from GCE and not GAE or GKE
```
quartzmo: iff -> if

google-cloud-core/lib/google/cloud/core/environment.rb
```diff
@@ -0,0 +1,141 @@
...
+        def self.gce?
+          gce_vm? && !gae? && !gke?
+        end
+
+        ##
+        # Check if running from Google Container Engine by querying for
+        # GKE cluster name and VM instance_zone
+        #
+        # @return [Boolean] True iff self.gke_cluster_name() and
```
quartzmo: iff -> if

google-cloud-core/lib/google/cloud/core/environment.rb
```diff
@@ -0,0 +1,141 @@
...
+# limitations under the License.
+
+
+require "faraday"
+
+module Google
+  module Cloud
+    module Core
+      ##
+      # @private
```
quartzmo: Just to confirm, this module should remain undocumented?

blowmage: Yes, this module is not intended for public use.

google-cloud-core/lib/google/cloud/core/environment.rb
hxiong388: Fixed

google-cloud-core/lib/google/cloud/core/environment.rb
hxiong388: Dohhhhhhhhhh~~

Fixed

quartzmo:
I searched the codebase (master branch) for existing calls to the original `gce?` method that might need to be updated, but I couldn't find any.

@hxiong388 Thank you for this great update. Merging.

PR #973, opened 2016-10-14
Add Speech Streaming

blowmage:
This PR is an implementation of Speech's Streaming support.

A stream object represents the bi-directional requests and responses. Users will not need to create IO objects or pipes or Enumerators, but can simply call `Stream#send` and receive notifications using blocks. The intention here is to lower the barrier to using the streaming API.

```ruby
require "google/cloud/speech"

speech = Google::Cloud::Speech.new

stream = audio.stream encoding: :raw, sample_rate: 16000

# register callback for when a result is returned
stream.on_result do |results|
  result = results.first
  result.transcript #=> "how old is the Brooklyn Bridge"
  result.confidence #=> 0.8099
end

# Stream 5 seconds of audio from the microhone
# Actual implementation of microphone input varies by platform
5.times.do
  stream.send MicrophoneInput.read(32000)
end

stream.stop
```

### DIFF
--- a/google-cloud-speech/acceptance/speech/stream_test.rb
+++ b/google-cloud-speech/acceptance/speech/stream_test.rb
@@ -1,0 +1,272 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "speech_helper"
+
+describe "Streaming Recognition", :speech do
+  let(:filepath) { "acceptance/data/audio.raw" }
+
+  it "default params" do
+    counters = Hash.new { |h, k| h[k] = 0 }
+
+    stream = speech.stream encoding: :raw, sample_rate: 16000
+
+    stream.on_interim      { counters[:interim] += 1 }
+    stream.on_result        { counters[:result] += 1 }
+    stream.on_speech_start { counters[:speech_start] += 1 }
+    stream.on_speech_end   { counters[:speech_end] += 1 }
+    stream.on_complete     { counters[:complete] += 1 }
+    stream.on_utterance    { counters[:utterance] += 1 }
+
+    stream.send File.read(filepath, mode: "rb")
+
+    stream.stop
+
+    while counters[:complete] == 0 do
+      sleep 1
+    end
+    sleep 3 # give more callbacks time to finish
+
+    results = stream.results
+
+    results.count.must_equal 1
+    results.first.transcript.must_equal "how old is the Brooklyn Bridge"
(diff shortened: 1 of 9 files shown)


blowmage:
I have some examples of using this Speech Streaming implementation with actual microphone inputs. The following example works on OS X and will stream audio until an utterance is found, and then stop.

```ruby
require "coreaudio"

input_device = CoreAudio.default_input_device
input_buffer = input_device.input_buffer 1024

require "google/cloud/speech"

speech = Google::Cloud::Speech.new

# Stream audio until the first utterance is found.
stream = speech.stream encoding: :raw, 
                       sample_rate: input_device.actual_rate,
                       max_alternatives: 10,
                       interim: true
                       utterance: true

stream.on_interim do |final_results, interum_results|
 puts "#"*72
 puts "FINAL RESULTS:"
 final_results.each do |result| 
   puts result.transcript.inspect
   result.alternatives.each do |alt| 
     puts "* #{alt.transcript.inspect}"
   end
 end
 puts "INTERUM RESULTS:"
 interum_results.each do |result| 
   puts result.transcript.inspect
   result.alternatives.each do |alt| 
     puts "* #{alt.transcript.inspect}"
   end
 end
 puts "#"*72
 puts ""
end

stream.on_result do |results|
  puts "*"*72
  puts "RESULTS:"
  results.each do |result| 
    puts result.transcript.inspect
    result.alternatives.each do |alt| 
      puts "* #{alt.transcript.inspect}"
    end
  end
  puts "*"*72
  puts ""
end

stream.on_speech_start do
  puts "-"*72
  puts "SPEECH STARTED"
  puts "-"*72
  puts ""
end

stream.on_speech_end do
  puts "="*72
  puts "SPEECH ENDED"
  puts "="*72
  puts ""
end

stream.on_complete do
  puts "/"*72
  puts "AUDIO ENDED"
  puts "/"*72
  puts ""
end

stream.on_utterance do
  puts "\\"*72
  puts "UTTERANCE OCCURRED"
  puts "\\"*72
  puts ""
  
  stream.stop
end

stream.start
input_buffer.start

loop do
  break if stream.stopped?
  # Collect data from only the first channel
  bits = input_buffer.read(4096).to_a.map(&:first)
  # Convert the bits to 16-bit signed little-endian samples
  sample = bits.map { |bit| Array(bit).pack("s<") }.join.encode("ASCII-8BIT")
  # Send the sample to the Speech stream
  stream.send sample
end

input_buffer.stop
```

blowmage:
@jmuk @bjwatson Do you guys have any input on how we are wrapping the streaming API in this PR? Is there anyone else that should take a look at this?

quartzmo:
Do you think that the background thread support could benefit from some of the features that @dazuma used in the Logging `AsyncWriter`, such as the `last_exception` field (just to mention one)?

quartzmo:
Should the docs mention that a background thread is created?

blowmage:
Dunno. Open to suggestions.

blowmage:
What do you think of an `on_error` callback?

quartzmo:
I think we probably want a consistent approach for background threads, maybe with some shared support in `google-cloud-core`. But that can be a discussion launched from any real-world issues arising from this implementation and/or Logging `AsyncWriter`. I'm going to merge this PR now.

bjwatson:
Sorry for the delay in responding. @swcloud can you take a look at this PR to make sure it fits with what you have in mind for streaming?

blowmage:
We went ahead and merged this PR to get it running on CI. We found a couple issues, and have an update on #977. Any feedback here or there is super appreciated.

PR #979, opened 2016-10-17
Fix import for all GAPIC APIs

swcloud:


google-cloud-logging/lib/google/cloud/logging/v2/logging_service_v2_api.rb
```diff
@@ -299,13 +300,7 @@ def write_log_entries \
             )
             req.log_name = log_name unless log_name.nil?
             req.resource = resource unless resource.nil?
-            # Custom code here to work around setting a Hash where a Map is expected.
```
swcloud: Hmm. It looks like you added this work-around manually a week ago. Can you describe the background of this work around? @jmuk

jmuk: Oops, I just didn't notice this.  I don't think that's fixed -- I actually didn't notice you reported it :(

blowmage:
No problem. I had it on my list of things to talk about at standup... :smile:

blowmage:
Should the main GAPIC require (e.g. `google/cloud/datastore/v1.rb`) also be updated and have the additional requires removed as part of this PR?

swcloud:
I noticed that too. Yes. No need to call those 'require's with the fix in this PR. I will remove them in a later push for this PR.

google-cloud-logging/lib/google/cloud/logging/v2/logging_service_v2_api.rb
swcloud: As @jmuk suggested, this should be fixed in 'toolkit'. In the mean time, let me add your workaround back.

swcloud:
BTW, why for 'google-cloud-language', I don't see there is 'v1beta1.rb' from which the GAPIC code is loaded. Instead, GAPIC code is directly loaded from service.rb. This seems to be inconsistent to other APIs like 'vision'.

blowmage:
The inconsistency is an oversight. The Speech implementation was started over a month ago, and was mostly in parallel with the other services for that time. So it is slightly out of date.

Issue #980, opened 2016-10-18
Google::Cloud::Logging::Entry.extract_timestamp returns Time object with wrong precision


hxiong388:
Google::Cloud::Logging::Entry.extract_timestamp method calls 
```ruby
Time.at grpc.timestamp.seconds, grpc.timestamp.nanos/1000.0
```
The nano second division logic returns wrong time precision. See this snippet:
```ruby
require "google/cloud/logging"
timestamp = Time.parse "2014-10-02T15:01:23.045123456Z"
t_grpc = Google::Protobuf::Timestamp.new(seconds: timestamp.to_i, nanos: timestamp.nsec)
e_grpc = Google::Logging::V2::LogEntry.new(timestamp:t_grpc)
new_timestamp = Google::Cloud::Logging::Entry.extract_timestamp(e_grpc)

new_timestamp == timestamp # Returns false. Should be true.
```

hxiong388:
One solution to fix this precision problem is to change code to:
```ruby
Time.at grpc.event_time.seconds, Rational(grpc.event_time.nanos, 1000)
```
This should maintain correct time precision.

blowmage:
Awesome. Thanks so much for the report and the solution. Are you working on a PR for this or should I?

hxiong388:
I'm almost have the PR ready. :)

PR #981, opened 2016-10-18
Fix issue #980: extract_timestamp precision off

hxiong388:
Fix issue #980


blowmage:
Will merge then the build completes.

PR #1002, opened 2016-10-28
Rake jsondoc

quartzmo:
The goal of this PR is to fully automate the generation and deployment of JSON docs. It provides the following:
1. New `jsondoc:package` task that accepts accepts a tag with the format `<gem>/<version>`, generates the
   JSON docs for the given package and version, then assembles the complete `google-cloud` JSON docs from the previously built JSON docs for all gems, using the latest versions in `gh-pages`. Finally, it publishes the changes in the cloned `gh-pages` repo. This task also supports tags for the `google-cloud` package, which will be released after tags for other new gems.
2. Updated `jsondoc:master` task does the same, but generates fresh JSON docs for all gems.
3. Support for the new `stackdriver` package has been added, but is disabled pending release (see `TODO`s in `Rakefile`.)
4. Update `RELEASING.md` to support individual gems.

Not included in this PR (to be added in a future PR): 
1. Instead of assuming that the current branch (`master`) for the execution context holds the correct content for generating JSON docs for a tag, `git clone` a new repo and execute `jsondoc` in the correct tag (branch) passed to `jsondoc:package`.
2. Instead of resolving the dependencies of a `google-cloud` version with the latests versions of gems in `manifest.json`, better reproduce the dependency resolution of Rubygems.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
 require "bundler/setup"
 require "open3"
+require "json"
 
 task :bundleupdate do
   gems.each do |gem|
@@ -204,13 +205,10 @@ task :jsondoc, :bundleupdate do |t, args|
 end
 
 namespace :jsondoc do
-  # @param tag [String] A valid tag, e.g. "v0.10.0" (legacy), "google-cloud-datastore:v1.2.0", or "master".
   desc "Clones gh-pages branch to a temp dir"
-  task :init, :tag do |t, args|
-    tag = args[:tag]
-    fail "Missing required parameter 'tag'." if tag.nil?
-    gh_pages = Pathname.new(Dir.home) + "tmp/#{tag}-gh-pages"
-
+  task :clone_gh_pages, [:gh_pages_dir] do |t, args|
+    gh_pages_dir = extract_args args, :gh_pages_dir
+    gh_pages = gh_pages_path gh_pages_dir
     header "Cloning gh-pages branch to #{gh_pages}"
 
     FileUtils.remove_dir gh_pages if Dir.exists? gh_pages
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -594,6 +687,15 @@ def header str
   puts ""
 end
 
+# Returns [gem_name, gem_version]
+def split_tag str
+  return [nil, str] if str == "master" # Support use of "master" even without gem name
+  fail "'tag' must be in the format <gem>/<version> Actual: #{str}" unless str.include?("/")
+  parts = str.split("/")
```
blowmage: The tag format is actually `"{PACKAGE}/v{VERSION}"`. We had a regex for this before, why isn't that being used here?

Rakefile
```diff
@@ -203,13 +204,10 @@ task :jsondoc, :bundleupdate do |t, args|
...
 
 namespace :jsondoc do
-  # @param tag [String] A valid tag, e.g. "v0.10.0" (legacy), "google-cloud-datastore:v1.2.0", or "master".
   desc "Clones gh-pages branch to a temp dir"
-  task :init, :tag do |t, args|
-    tag = args[:tag]
-    fail "Missing required parameter 'tag'." if tag.nil?
-    gh_pages = Pathname.new(Dir.home) + "tmp/#{tag}-gh-pages"
-
+  task :clone_gh_pages, [:gh_pages_dir] do |t, args|
```
blowmage: It might be a good idea to move helper tasks like `jsondoc: clone_gh_pages` and `jsondoc: copy` and `jsondoc: too` out into a separate object that can be unit tested, and keeping tasks only for the actions that can be instigated by a user or a process like travis/appveyor.

But that could be a separate refactor after the desired behavior is achieved.

quartzmo: Sounds good!

Rakefile
```diff
@@ -227,100 +225,146 @@ namespace :jsondoc do
...
-      rm_rf gh_pages + "json/#{gem}/#{gem_version}", verbose: true
-      mkdir_p gh_pages + "json/#{gem}/"
-      cp_r "#{gem}/jsondoc", gh_pages + "json/#{gem}/#{gem_version}", verbose: true
+    header "Copying #{gem} jsondoc for '#{gem}/#{version}' to '#{gh_pages}'"
+
+    unless Dir.exist? gh_pages + "json/#{gem}"
+      mkdir_p gh_pages + "json/#{gem}", verbose: true
     end
+    rm_rf gh_pages + "json/#{gem}/#{version}", verbose: true
+    mkdir_p gh_pages + "json/#{gem}/"
```
hxiong388: Isn't this line redundant with line 236? I can remove this later in my stackdriver-json PR if you want.

Rakefile
```diff
@@ -227,100 +225,146 @@ namespace :jsondoc do
...
-    tag = args[:tag]
-    fail "Missing required parameter 'tag'." if tag.nil?
-    gh_pages = Pathname.new(Dir.home) + "tmp/#{tag}-gh-pages"
+  desc "Updates a gem's toc.json with correct tagName."
+  task :toc, [:gem, :version, :gh_pages_dir] => [:copy] do |t, args|
+    gem, version, gh_pages_dir = extract_args args, :gem, :version, :gh_pages_dir
+    gh_pages = gh_pages_path gh_pages_dir
+
+    header "Updating toc.json for '#{gem}/#{version}' in '#{gh_pages}'"
+    toc_path = gh_pages + "json/#{gem}/#{version}" + "toc.json"
```
hxiong388: Do you need another "/" at end of ".../#{version}" to build the correct path to toc.json?

quartzmo: No, `gh_pages` is a `Pathname` instance, and [`Pathname#+`](https://ruby-doc.org/stdlib-2.3.0/libdoc/pathname/rdoc/Pathname.html#method-i-2B) takes care of it.

Rakefile
quartzmo: Yes, sure looks like it! I decided to wrap in that conditional on 235 and I forgot to delete the original I think. Good catch.

quartzmo: I may need to rebase/update this PR after #1003, if so I'll fix it then.

Rakefile
hxiong388: Gotcha

Issue #1011, opened 2016-11-02
Error Reporting instrumentation error on Rails startup


hxiong388:
Right now when user tries to use Error Reporting instrumentation with Rails app, they get this error on Rails startup:
```
=> Booting Puma
=> Rails 5.0.0.1 application starting in production on http://0.0.0.0:3000
=> Run `rails server -h` for more startup options
Exiting
/usr/local/google/home/hexiong/rails_workspace/my_rails/gems/google-cloud-error_reporting/lib/google/cloud/error_reporting/rails.rb:62:in `block in <class:Railtie>': undefined local variable or method `error_reporting' for #<Google::Cloud::ErrorReporting::Railtie:0x0000000540c0e0> (NameError)
	from /usr/local/google/home/hexiong/.rvm/gems/ruby-2.3.1/gems/railties-5.0.0.1/lib/rails/initializable.rb:30:in `instance_exec'
	from /usr/local/google/home/hexiong/.rvm/gems/ruby-2.3.1/gems/railties-5.0.0.1/lib/rails/initializable.rb:30:in `run'
...
```

I think this bug is introduced during the refactoring commits in PR #961.

blowmage:
cc @hxiong388

hxiong388:
Closed by [PR#1013](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1013)

PR #961, opened 2016-10-07
Added google-cloud-error_reporting gem

hxiong388:
Created the google-cloud-error_reporting gem, which contains the client API library and instrumentation library for Stackdriver Error Reporting API.

This is the first revision of the API (version 0.1.0), which only contains support for the reporting API. More APIs will be added in future revisions.

The client API library is structured similarly to other google-cloud-* libraries in this repo. All the documentation, unit tests, acceptance tests, other rake tasks should be compatible with the rest of the repo.

The instrumentation logic/code is defined in two files:
lib/google/cloud/error_reporting/middleware.rb
lib/google/cloud/error_reporting/rails.rb
The middleware.rb can be loaded along with rest of client API code, by invoking "require google/cloud/error_reporting". But the rails.rb has to be explicitly loaded by "require google/cloud/error_reporting/rails" in order to enable the Railtie integration logic.

PR #1013, opened 2016-11-03
Fix issue #1011 by removing the undefined variable references.

hxiong388:
The error_reporting references were there when I had the Veneer Error Reporting library, and the code is to set default Error Reporting service_name and service_version. During the refactoring of PR #961, the Veneer library code was removed and the default values are now set in google/cloud/error_reporting/middleware.rb. So simply remove the left over references will suffice.

### DIFF
--- a/google-cloud-error_reporting/lib/google/cloud/error_reporting/rails.rb
+++ b/google-cloud-error_reporting/lib/google/cloud/error_reporting/rails.rb
@@ -58,10 +58,8 @@ module Google
             keyfile = er_config.keyfile || gcp_config.keyfile
 
             channel = self.class.grpc_channel keyfile
-            service_name = er_config.service_name ||
-              error_reporting.class.default_service_name
-            service_version = er_config.service_version ||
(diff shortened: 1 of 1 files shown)


blowmage:
Is it possible to add some test coverage that will make sure this won't happen again?

hxiong388:
Daniel and I were just talking about it. Since the effected code is actually inside around the Google::Cloud::ErrorReporting::Railtie initialization code, the best way to trigger that code would probably be to simulate Railtie initialization process. But instead of putting extra effort into making complex tests to cover these Railtie initializations, we think it's best to just cover this portion of the project with the internal CI project that will test using real Rails apps.

blowmage:
Makes sense.

blowmage:
Is this ready to merge? Did your internal builds pass?

hxiong388:
@blowmage Yeah, this is ready to merge. We don't have the automated internal testing in place yet, that's why the bug happened in the first place. All the manual testing I did yesterday passed.

Issue #986, opened 2016-10-19
Add GRPC documentation class for all GAPIC-enabled APIs


swcloud:
Currently there are no documentation info available in GRPC classes (generated from protobuf). We should deliver classes containing documentation for these GRPC objects so that they are part of a gem and hence the documentation will be available on the rubydoc.info site.

PR #1016, opened 2016-11-03
Add grpc dummy documentation class for datastore API

swcloud:
This was supposed to be part of #986, but documentation string for datastore API caused a bug in code generator to surface. Now that bug is fixed, so we need to add the grpc dummy documentation class for datastore API


swcloud:
@quartzmo Anything else needs to be done to merge this PR?

blowmage:
I was holding off on this until we got closer to a beta release and needed to include it. Once this is merged its harder to cut a point release without including this.

quartzmo:
@blowmage, Do you want to add a commit excluding these files from current jsondoc output?

blowmage:
Docs are already ignored in the .yardopts file.

PR #1019, opened 2016-11-04
Add streaming support for speech

swcloud:


blowmage:
What version of google-gax does this depend on?

swcloud:
There is no related change in google-gax for streaming support.

hxiong388:
I played around with existing speech streaming and the change form this PR (had to manually update service.rb to use the new GAPIC method). Since the new method is just a grpc stub, so I think it's functionally same as current speech streaming implementation.

blowmage:
I am the author of the new commit.

blowmage:
Did a bunch of edge case testing to see if I could get the method to misbehave, but it worked as expected. Errors returned by the API were sent to the enumerator and not handled/re-raised by google-gax.

quartzmo:
@blowmage Thanks for giving this an extra measure of manual testing!

PR #1020, opened 2016-11-04
Change product and API name to Stackdriver Error Reporting

steren:
As we can read on https://developers.google.com/apis-explorer/#p/clouderrorreporting/v1beta1/ the product and API name is "Stackdriver Error Reporting"

"Stackdriver Clouderrorreporting" has never been defined and should NOT be used.

### DIFF
--- a/google-cloud-error_reporting/README.md
+++ b/google-cloud-error_reporting/README.md
@@ -1,18 +1,18 @@
-Stackdriver Clouderrorreporting API for Ruby
+Stackdriver Error Reporting API for Ruby
 =================================================
 
 google-cloud-error_reporting uses [Google API extensions][google-gax] to provide an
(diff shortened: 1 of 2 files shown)


blowmage:
I believe this should also be changed in the gemspec file.

steren:
I'm not sure where I should change this, could you point me to an example?

blowmage:
The [summary on line 11](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-error_reporting/google-cloud-error_reporting.gemspec#L11).

steren:
Thanks, I performed the change

hxiong388:
I think all the current references to "clouderrorreporting" may be auto generated by VTK.

@swcloud, Steren has fixed the references manually in this PR. Can you please double check these references are also fixed in VTK? Thanks.

swcloud:
@hxiong388 That's right. We will take a look.

PR #1025, opened 2016-11-04
logger creates entries with timestamp

### LINKED ISSUE #1022: Logging should put timestamp on entries during creation
The async_writer has a performance feature of log entries batching that reduces number of gRPC requests. Since the Logging:Logger doesn't create entries with timestamp, the Stackdriver Logging service fills in the timestamp at time of receiving, which implies all the log entries in a batch to have same timestamp. The design is acceptable in most cases, but has couple potentially problems:
1. Without individual timestamp, logger now relies on async_writer and its queue to maintain log entries' ordering. Ideally, we want log entries' ordering to be only controlled by their true creation timestamp.
2. In extreme cases, log entries' time differences can be critical for diagnose problems. But this information is lost when a lot of log entries get batched together with same timestamp. Especially when volume is high or async_writer thread gets delayed for any reason.

I think we can solve these potential problems by filling in timestamp at Logging::Logger#write_entries, if users use Logging::Logger. For users manually creating log entries, it's their responsibility to provide timestamp information.

hxiong388:
Logging::Logger methods now creates Logging::Entry with timestamp at creation, which will provide a reliable chronological order of log entries getting send to Stackdriver Logging service.

close #1022

### DIFF
--- a/google-cloud-logging/lib/google/cloud/logging/logger.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/logger.rb
@@ -335,6 +335,7 @@ module Google
         # @private Write a log entry to the Stackdriver Logging service.
         def write_entry severity, message
           entry = Entry.new.tap do |e|
+            e.timestamp = Time.now
             e.severity = gcloud_severity(severity)
             e.payload = message
           end
--- a/google-cloud-logging/test/google/cloud/logging/logger/add_test.rb
+++ b/google-cloud-logging/test/google/cloud/logging/logger/add_test.rb
@@ -28,9 +28,15 @@ describe Google::Cloud::Logging::Logger, :add, :mock_logging do
   let(:logger) { Google::Cloud::Logging::Logger.new logging, log_name, resource, labels }
   let(:severity) { :DEBUG }
   let(:write_res) { Google::Logging::V2::WriteLogEntriesResponse.new }
+  let(:timestamp) { Time.parse "2016-10-02T15:01:23.045123456Z" }
 
   def write_req_args
-    entries = [Google::Logging::V2::LogEntry.new(text_payload: "Danger Will Robinson!", severity: severity)]
+    timestamp_grpc = Google::Protobuf::Timestamp.new seconds: timestamp.to_i,
+                                                     nanos: timestamp.nsec
+
+    entries = [Google::Logging::V2::LogEntry.new(text_payload: "Danger Will Robinson!",
+                                                 severity: severity,
+                                                 timestamp: timestamp_grpc)]
     [entries, log_name: "projects/test/logs/web_app_log", resource: resource.to_grpc, labels: labels, options: default_options]
   end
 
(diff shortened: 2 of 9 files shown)


quartzmo:
@hxiong388 I'm surprised if this change doesn't break some unit tests. What about the expected [`Google::Logging::V2::LogEntry`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-logging/test/google/cloud/logging/logger_test.rb#L32) value here for example?

hxiong388:
I didn't run the unit tests locally because I didn't expect this "small" change to break anything. And I just found out this line broke 84 of them...

Thanks for helping me catching this. Saved me hours to wait on Travis to do the same.

quartzmo:
The situation with Travis is certainly not good.

quartzmo:
I just had an idea, what about providing an option (or some mechanism) in case the user wants a server-assigned timestamp for some reason?

hxiong388:
Daniel and I were talking about that. Our rationale is that since logger in Ruby is a defined interface that doesn't take timestamp, then it makes sense for our logger implementation to fill in timestamp info. User who directly uses the underneath API, then they can manually create Logging::Entry objects by either providing or omit timestamp at will. This is why I made the change in Logger. 

Regarding weather or not provide a setting to control logger behavior, why would we every want log entries to share the batched timestamp? Isn't filled in creation timestamp always better?

blowmage:
We never gave that option before. The timestamp was set implicitly when writing to the logger, but now it needs to be set explicitly due to the behavior of the async writer.

quartzmo:
Set implicitly where, @blowmage? In the library? Or on the Logging service?

hxiong388:
@blowmage Does this mean you don't think we should explicitly set timestamps in logger or google-cloud-logging? Then logs on server side may suffer indistinguishable timestamp on server side, and future risk wrong ordering from async_writer bug. I think that would be a bigger evil compare with fix all the unit tests.

blowmage:
@quartzmo The timestamp was set implicitly when the entry was sent to the API, and the Logger called the API synchronously. Now that this isn't the case, and we are pooling entries in the async writer, we need to explicitly set the timestamp.

@hxiong388 No, I am agreeing with you. I have already approved the PR. Let's get a green build and I'll merge it.

quartzmo:
@blowmage So it was always set in the client (this library)? In that case ignore my suggestion. I was thinking it was previously set in the service.

blowmage:
No, if an entry is written to the API without a timestamp, the API will set the timestamp to when the entry gets written. That is why it was implicit.

quartzmo:
If this PR changes the behavior from setting the timestamp on the service ("the API") to setting it on the client for `synchronous` logging operations, then I think it might be nice to provide an option that prevents it from being set on the client, so that it can be set on the service as it is currently. Just a suggestion, I have no idea if anyone will actually want it.

blowmage:
I am not in favor of adding an option to control this behavior. The Logger should do the right thing and with this change it does.

quartzmo:
I agree that setting the timestamp on the client seems like the right thing to do. I'm not absolutely certain that it is *always* the right thing to do. Anyway, I also approve this PR, pending passing unit tests.

PR #1027, opened 2016-11-07
Update jbuilder version to be compatible with Rails 5 test app

hxiong388:
Update jbuilder version to 2.5, which is compatible with any Rails 5 apps.

Ran unit tests, all passed, with a lot warning statements. The warning messages don't see to be caused by this change.

### DIFF
--- a/gcloud-jsondoc.gemspec
+++ b/gcloud-jsondoc.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'yard', "~> 0.8"
   s.add_dependency 'kramdown', "~> 1.9"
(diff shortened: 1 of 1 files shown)


quartzmo:
Just to be on the safe side, I'll check out this branch in a second repo, and use it to build the docs on master.

blowmage:
I agree that would be wise. Can you also run all the acceptance tests as well?

quartzmo:
@blowmage sure

hxiong388:
I actually tested building master branch jsondoc with this version change last week when we were discussing it. I just didn't get a chance to submit the PR till today. But I agree, someone should double check it to be safe.

quartzmo:
I ran jsondoc, doctest, unit, and acceptance tasks. Everything passed. Merging.

PR #1030, opened 2016-11-08
Correctly pin activesupport 4.0

hxiong388:
The previous way of pinning activesupport didn't work when gem is used from other Gemfiles. So now by making activesupport 4.0 as an actually dependency of gcloud-jsondoc gem, it should work properly.

### DIFF
--- a/gcloud-jsondoc.gemspec
+++ b/gcloud-jsondoc.gemspec
@@ -28,8 +28,9 @@ Gem::Specification.new do |s|
   s.add_dependency 'yard', "~> 0.8"
   s.add_dependency 'kramdown', "~> 1.9"
   s.add_dependency 'jbuilder', "~> 2.5.0"
+  # Pin activesupport version. Remove this pin once we stop supporting Ruby 2.1
+  s.add_dependency "activesupport", "~> 4.0"
 
   s.add_development_dependency "minitest"
   s.add_development_dependency "minitest-autotest", "~> 1.0"
-  s.add_development_dependency "activesupport", "~> 4.0"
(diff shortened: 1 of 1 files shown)


blowmage:
I believe this should have been an actual dependency instead of a development dependency to begin with. But, wasn't this added in the first place so you can test against Rails 5? Doesn't this step block any future use of Rails 5.0 in the development environment of other gems? For example, you won't be able to create a Rails 5.0 app to test Logging because this dependency does not allow for it.

quartzmo:
@hxiong388 Ready to merge if this works for you (see @blowmage's comment above). Just let me know.

hxiong388:
Your right. I originally made PR #1027  because I figured I had to find a common set of gems that satisfies all the google-cloud-ruby development needs and the new Rails 5 test app. But then after #1027 and #1028 were merged last night, I figured out a way to separate out the google-cloud-ruby development gem set and the sample Rails 5 app's gem set, so they aren't interfering with each other anymore.

But since #1027 has been breaking master branch builds for Ruby 2.0 and 2.1 environments, I'm guessing it's better to pin activesupport 4.0 and stabilize the builds. Otherwise we can revert PR #1027, #1028, and this PR if you think it's better.

blowmage:
This PR is in conflict. Can you rebase your changes to remove the conflict?

blowmage:
Oh wait, never mind. It looks like this PR got some of the same commits intended for #1031, and my comment was intended for that PR. This issue is resolved now that #1031 is accepted. Closing.

PR #1031, opened 2016-11-08
Revert pr 1027

hxiong388:
PR #1027 and #1028 were submitted to make gcloud-json gem compatible with the soon to be added Rails 5 test app. But then I found an alternative solution to make a Rails 5 test app work without mixing the Rails 5 gem sets with google-cloud-ruby gems' gem sets.

After discussing with @blowmage, we decided to revert changes from PR #1027 and PR #1028.


hxiong388:
I dropped the extra commit. So only 0eb4b83 is in this PR for reverting.

blowmage:
Thanks!

PR #1050, opened 2016-11-12
Update Translate for V2

blowmage:
Update translate gem to use the new API endpoint. This requires that we replace the use of Google API Client with custom code. But this allows us to accomplish the following:

- [x] Use new URL
- [x] Support service accounts using OAuth2
- [x] Switch to POST for all API calls
- [x] Add `model` argument and attribute

### DIFF
--- a/google-cloud-translate/lib/google-cloud-translate.rb
+++ b/google-cloud-translate/lib/google-cloud-translate.rb
@@ -36,6 +36,14 @@ module Google
     # keys](https://cloud.google.com/translate/v2/using_rest#creating-server-api-keys).
     #
     # @param [String] key a public API access key (not an OAuth 2.0 token)
+    # @param [String, Array<String>] scope The OAuth 2.0 scopes controlling the
+    #   set of resources and operations that the connection can access. See
+    #   [Using OAuth 2.0 to Access Google
+    #   APIs](https://developers.google.com/identity/protocols/OAuth2).
+    #
+    #   The default scope is:
+    #
+    #   * `https://www.googleapis.com/auth/cloud-platform`
     # @param [Integer] retries Number of times to retry requests on server
     #   error. The default value is `3`. Optional.
     # @param [Integer] timeout Default timeout to use in requests. Optional.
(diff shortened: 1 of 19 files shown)


google-cloud-translate/lib/google-cloud-translate.rb
```diff
@@ -62,8 +70,10 @@ module Cloud
     #   translation = translate.translate "Hello world!", to: "la"
     #   translation.text #=> "Salve mundi!"
     #
-    def translate key = nil, retries: nil, timeout: nil
-      Google::Cloud.translate key, retries: (retries || @retries),
+    def translate key = nil, scope: nil, retries: nil, timeout: nil
```
quartzmo: Even though Translation now supports service accounts, both this method and the class method below have a different signature from other services. I think this is the correct choice since Translation continues to prioritize API key usage, and it is backward-compatible, but I just want to point this out.

google-cloud-translate/lib/google/cloud/translate/service.rb
```diff
@@ -77,10 +76,118 @@ def inspect
...
+        # Make a request and apply incremental backoff
         def execute
-          yield
-        rescue Google::Apis::Error => e
-          raise Google::Cloud::Error.from_error(e)
+          backoff = Backoff.new retries: retries
+          backoff.execute do
+            yield
+          end
+        rescue Faraday::ConnectionFailed
```
quartzmo: Does this duplicate the error handling in `post`, above?

google-cloud-translate/lib/google/cloud/translate/service.rb
```diff
@@ -77,10 +76,118 @@ def inspect
...
+                req.params = { key: @key }
+              else
+                @credentials.sign_http_request req
+              end
+            end
+          end
+
+          return JSON.parse(response.body)["data"] if response.success?
+
+          fail Google::Cloud::Error.gapi_error_class_for(response.status)
```
quartzmo: Can we extract error details from the response, and put them in the `Google::Cloud::Error`?

quartzmo:
Incredible work @blowmage! Can I add a commit to this PR making docs changes/additions?

google-cloud-translate/lib/google-cloud-translate.rb
blowmage: Yeah, this is an unfortunate situation. I have two reasons for leaving it this way:

1. Maintaining backwards compatibility. We currently direct users to create translate objects with API Key, and if we changed those arguments to both `project` and `keyfile`, as we have in other services, existing code will break.
2. The design guidelines we've been given state that API Key should have priority over service accounts, as you state.

google-cloud-translate/lib/google/cloud/translate/service.rb
blowmage: Yes, but it doesn't hurt anything to be duplicated.

quartzmo:
I confirm that I am the author of the commits.

blowmage:
Switching labels now that @quartzmo has confirmed.

blowmage:
Doc changes look good. :+1: from me. As soon as the builds pass we can merge.

PR #1052, opened 2016-11-15
Use latest rubies

blowmage:
Ruby 2.3.2 and 2.2.6 were recently relased.
Update Travis-CI config, which needs the versions to be explicit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,24 @@ language: ruby
 os:
 - linux
 - osx
-osx_image: xcode7.1
 rvm:
+- 2.3.2
 - 2.2.6
 - 2.1.9
 - 2.0.0-p648
 before_install:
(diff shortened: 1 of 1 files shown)


quartzmo:
@blowmage The build failed with:

```sh
ruby-2.3.2 is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.3.2 do rvm gemset create ' first, or append '--create'.``
```

blowmage:
We may have to sit on this PR for a little while until RVM and/or Travis-CI makes them available.

quartzmo:
Ruby 2.3.3 https://www.ruby-lang.org/en/news/2016/11/21/ruby-2-3-3-released/

blowmage:
Will revisit in the future, but closing for now.

PR #1058, opened 2016-11-16
New CI tasks

blowmage:
This PR adds `rake ci` and `rake ci:acceptance` tasks to the top-level project as well as every individual cloud service project. We have found it desirable to run the CI tasks locally before creating a PR. But as we add more tasks, such as `rubocop`, `jsondoc`, `doctest`, etc, we sometimes forget to run one. The `rake ci` task will run all the same tasks that CI runs. The `rake ci:acceptance` task will also run all the same tasks that CI runs, as well as the acceptance tests.

This PR also limits the coveralls report to only be run once per build, instead of running on every job in the build.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -23,6 +23,7 @@ task :each, :bundleupdate do |t, args|
         header "RUBOCOP, JSONDOC, TESTS FOR #{gem}"
         sh "bundle exec rake rubocop"
         sh "bundle exec rake jsondoc"
+        sh "bundle exec rake doctest"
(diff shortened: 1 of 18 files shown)


quartzmo:
@blowmage I like the `ci` name, very concise! This will be great to have.

@dazuma @hxiong388 PTAL at the changes/additions to Rakefiles you use.

blowmage:
I'm most excited to type `rake ci:a` and be able to switch to something else while it builds.

PR #1063, opened 2016-11-19
Fix empty protobuf as return type

swcloud:
Based on our usability study results, for those API returning google.protobuf.empty as the return type (Google::Protobuf::Empty), we should just return 'nil' to avoid exposing the low-level grpc details to the user.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/v1/subscriber_api.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/v1/subscriber_api.rb
@@ -452,6 +452,7 @@ module Google
               subscription: subscription
             }.delete_if { |_, v| v.nil? })
             @delete_subscription.call(req, options)
+            nil
           end
 
           # Modifies the ack deadline for a specific message. This method is useful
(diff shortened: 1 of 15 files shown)


blowmage:
I am so supportive of this! Can you rebase on master so this will run with the updated CircleCI configuration?

google-cloud-pubsub/lib/google/cloud/pubsub/v1/subscriber_api.rb
```diff
@@ -534,6 +536,7 @@ def acknowledge \
               ack_ids: ack_ids
             }.delete_if { |_, v| v.nil? })
             @acknowledge.call(req, options)
+            nil
```
swcloud: That doesn't matter I think. If you want to push your changes to this PR, please do that.

blowmage: Okay, I've pushed a commit to your branch that updates the tests.

blowmage:
I conform I am the author of the commit added to this PR.

PR #1069, opened 2016-11-21
Add Logging Project#entries resources argument

### LINKED ISSUE #971: Outdated usage of entries.list API request
Issue: In the logging/service.rb file, an deprecated fieldName (projectIds) of [entries.list](https://cloud.google.com/logging/docs/api/ref_v2beta1/rest/v2beta1/entries/list) API is being used. 

https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/2d6c92137152d5297e91e1b5c336ed94b5ac96e7/google-cloud-logging/lib/google/cloud/logging/service.rb#L95-L109

blowmage:
Effectively replaces the projects argument, which is deprecated and will eventually be removed.

Now that the generated GAPIC code has been updated and no longer raises we can implement this change.

[closes #971]

### DIFF
--- a/google-cloud-logging/lib/google/cloud/logging/entry/list.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/entry/list.rb
@@ -70,9 +70,9 @@ module Google
           def next
             return nil unless next?
             ensure_service!
-            grpc = @service.list_entries token: token, projects: @projects,
+            grpc = @service.list_entries token: token, resources: @resources,
                                          filter: @filter, order: @order,
-                                         max: @max
+                                         max: @max, projects: @projects
             self.class.from_grpc grpc, @service
           end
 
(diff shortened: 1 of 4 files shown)


quartzmo:
I'm wondering if at this early stage in google-cloud-logging adoption, and since after the library reaches 1.0 it will be more disruptive to make breaking changes, it might be best to remove the `projects` parameter from the method signature(s), rather than just mark it as deprecated?

blowmage:
I agree it would likely be better if the method parameter was removed by the time we hit 1.0. But I'm of the opinion we should have at least one release where it is marked deprecated before it is removed. That said, I will defer to @dazuma and @hxiong388 on how they want to handle this.

quartzmo:
@blowmage Makes sense to me.

quartzmo:
LGTM 

@dazuma, if you are OK with this PR, please merge when green.

quartzmo:
@dazuma @hxiong388 Good to merge?

PR #1081, opened 2016-11-22
Update logging API gapic code

swcloud:
This gapic code update for logging reflects changes in https://github.com/googleapis/googleapis/commit/60dde247a3af2f7178288b4018af1d65b57ef765.

The breaking change is in 'list_log_entries" where 'resource_names' becomes required while 'project_ids' becomes optional. @blowmage Please update veneer layer 'service.rb' and test code 'list_entries_test.rb'.

### DIFF
--- a/google-cloud-logging/lib/google/cloud/logging/service.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/service.rb
@@ -166,15 +166,18 @@ module Google
         end
 
         def create_sink name, destination, filter, version, start_time: nil,
-                        end_time: nil
+                        end_time: nil, unique_writer_identity: nil
           sink = Google::Logging::V2::LogSink.new({
             name: name, destination: destination, filter: filter,
             output_version_format: version,
             start_time: time_to_timestamp(start_time),
-            end_time: time_to_timestamp(end_time) }.delete_if { |_, v| v.nil? })
+            end_time: time_to_timestamp(end_time)
+          }.delete_if { |_, v| v.nil? })
 
           execute do
-            sinks.create_sink project_path, sink, options: default_options
+            sinks.create_sink project_path, sink,
+                              unique_writer_identity: unique_writer_identity,
+                              options: default_options
           end
         end
 
@@ -183,7 +186,7 @@ module Google
         end
 
         def update_sink name, destination, filter, version, start_time: nil,
-                        end_time: nil
+                        end_time: nil, unique_writer_identity: nil
           sink = Google::Logging::V2::LogSink.new({
             name: name, destination: destination, filter: filter,
             output_version_format: version,
@@ -191,7 +194,9 @@ module Google
             end_time: time_to_timestamp(end_time) }.delete_if { |_, v| v.nil? })
 
           execute do
-            sinks.update_sink sink_path(name), sink, options: default_options
+            sinks.update_sink sink_path(name), sink,
+                              unique_writer_identity: unique_writer_identity,
+                              options: default_options
           end
         end
 
--- a/google-cloud-logging/lib/google/cloud/logging/service.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/service.rb
@@ -165,10 +165,13 @@ module Google
           end
         end
 
-        def create_sink name, destination, filter, version
+        def create_sink name, destination, filter, version, start_time: nil,
+                        end_time: nil
           sink = Google::Logging::V2::LogSink.new({
             name: name, destination: destination, filter: filter,
-            output_version_format: version }.delete_if { |_, v| v.nil? })
+            output_version_format: version,
+            start_time: time_to_timestamp(start_time),
+            end_time: time_to_timestamp(end_time) }.delete_if { |_, v| v.nil? })
 
           execute do
             sinks.create_sink project_path, sink, options: default_options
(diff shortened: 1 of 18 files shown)


quartzmo:
The timing is incredible, we just merged #1069.

blowmage:
Do we need to expose `unique_writer_identity` in this PR as well?

blowmage:
@swcloud Do we need to expose unique_writer_identity in this PR as well?

blowmage:
Looks like V1::LogSink has added `start_time` and `end_time` as well. Should that also be added to the public API in this PR?

swcloud:
@blowmage I believe the answers to your questions should be all YES. These are added functionalities by the logging API and exposed by the proto and vkit, so they should be exposed by veneer as well.

blowmage:
I've been trying to figure out how we could treat `unique_writer_identity` as an attribute on Sink, but I don't think we can. That value is provided on create and update API calls, but never returned AFAICT. I don't understand why it was done this way, but it doesn't really give us many options for how to expose it.

blowmage:
I am the author of these new commits.

google-cloud-logging/lib/google/cloud/logging/service.rb
```diff
@@ -279,6 +285,24 @@ def default_options
           Google::Gax::CallOptions.new kwargs: default_headers
         end
 
+        ##
+        # @private Get a Google::Protobuf::Timestamp object from a Time object.
+        def time_to_timestamp time
+          return nil if time.nil?
+          # Make sure we have a Time object
+          return nil unless time.respond_to? :to_time
```
quartzmo: I wonder if raising an error is a better choice here. Since `start_time` and `end_time` are optional parameters, passing an unsupported type could lead to the user to believe the argument was successfully applied. It is not hard to imagine someone trying to pass a time value as Fixnum, for example.

google-cloud-logging/lib/google/cloud/logging/sink.rb
```diff
@@ -222,6 +264,24 @@ def self.resolve_version version
         def ensure_service!
           fail "Must have active connection to service" unless service
         end
+
+        ##
+        # @private Get a Google::Protobuf::Timestamp object from a Time object.
+        def time_to_timestamp time
+          return nil if time.nil?
+          # Make sure we have a Time object
+          return nil unless time.respond_to? :to_time
```
quartzmo: Same comment as `time_to_timestamp` in Service.

google-cloud-logging/lib/google/cloud/logging/service.rb
quartzmo: I wonder if raising an error is a better choice here. Since start_time and end_time are optional parameters, passing an unsupported type could lead to the user to believe the argument was successfully applied. It is not hard to imagine someone trying to pass a time value as Fixnum, for example.

google-cloud-logging/test/google/cloud/logging/project/list_entries_test.rb
```diff
@@ -224,7 +224,7 @@
     list_res = OpenStruct.new(page: OpenStruct.new(response: Google::Logging::V2::ListLogEntriesResponse.decode_json(list_entries_json(3, "next_page_token"))))
 
     mock = Minitest::Mock.new
-    mock.expect :list_log_entries, list_res, [["project1"], resource_names: nil, filter: nil, order_by: nil, page_size: nil, options: default_options]
+    mock.expect :list_log_entries, list_res, [[], project_ids: ["project1"], filter: nil, order_by: nil, page_size: nil, options: default_options]
```
blowmage: Oh wait, I see what you are asking. Before we only had one array, and we would make sure it had a default value. Now we can specify both resource_names _or_ project_ids. So if neither are proven fed we will default resource_names to the current project resource. But if projects are specified we don't apply that default.

swcloud: I think the doc says it's Required, so it should not be empty.

blowmage: I believe the behavior for this test is correct. The way Logging works is if you don't specify the request with a resource or project we will scope it to the project that the client is authenticated with. If the user specifies alternate projects ids to scope the request to then we should not add the current project as a resource.

The API has not changed, only the GAPIC code has. We are simply continuing the behavior that was in place since before resource names were added to the API.

google-cloud-logging/lib/google/cloud/logging/service.rb
blowmage: If we were to raise something like ArgumentError the place to do it is in the public method that accepts the value, not this private conversion method.

Do you want to see an error raised if a non-Time object is provided?

google-cloud-logging/test/google/cloud/logging/project/list_entries_test.rb
swcloud: Let me double check with the API producer next Monday to make sure this is the expected behavior.

blowmage: @swcloud Did you get confirmation?

quartzmo: In my opinion, we should follow the advice of the API owner and no longer use the deprecated `project_ids` field. In fact, I suspect the best choice at this point (pre-beta) would be a breaking change to our own method signature, removing `project_ids` there as well.

blowmage:
I'm gonna try to update this PR later tonight. But if you want you can accept this PR now and open a separate issue to make the changes in regards to project ids. This PR maintains compatibility, but that is a break in compatibility, so it might be wise to track that in a separate issue/PR.

quartzmo:
@blowmage I will update the PR now to no longer call `project_ids` in the API. Once it's done, you can either merge, or let me know if you want changes.

quartzmo:
Actually, since I'm not sure if I have access to @swcloud's repo, merging now. Will open a follow-up PR.

PR #1097, opened 2016-12-02
New spec to show encoding problem

dfournier:
Hi,

I'm having some issues to perform actions on files with strange filename. Files has been uploaded with [fog-google](https://github.com/fog/fog-google) without sanitizing the filename. Files can be listed but nothing can be done on them. Here is a sample:
```ruby
bucket = Google::Cloud::Storage.new.bucket 'my-bucket-name'
bucket.files.first.reload! # -> raises Google::Cloud::NotFoundError
bucket.files.first.delete # -> raises Google::Cloud::NotFoundError
```

I'm able to delete files via the [console](https://console.cloud.google.com/storage) and the problem might be an encoding problem. `#{[101, 769].pack("U*")}.png` is encoded `%C3%A9.png` [here](https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/http_command.rb#L146) but it should be encoded `e%CC%81.png`

```ruby
# url from the gem
`/storage/v1/b/gcloud-ruby-acceptance-2016-12-02t16-00-31z-06b8fa78/o/%C3%A9.png
# url from the google cloud storage console`
`/storage/v1_internal/b/gcloud-ruby-acceptance-2016-12-02t16-00-31z-06b8fa78/o/e%CC%81.png`
```

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -89,6 +89,17 @@ describe Google::Cloud::Storage::File, :storage do
     uploaded.delete
   end
 
+  it "should upload and delete a file with strange filename" do
+    original = File.new files[:logo][:path]
+    uploaded = bucket.create_file original, "#{[101, 769].pack("U*")}.png",
+      cache_control: "public, max-age=3600",
+      content_disposition: "attachment; filename=filename.ext",
+      content_language: "en",
+      content_type: "text/plain",
+      metadata: { player: "Alice", score: 101 }
+    uploaded.delete
+  end
+
   it "should upload and download a larger file" do
     original = File.new files[:big][:path]
     uploaded = bucket.create_file original, "BigLogo.png"


blowmage:
Thanks for the spec. I'll make sure this works.

blowmage:
This is a pretty interesting edge case in how different API clients handle URLs. The Google Cloud Storage API documentation presents the API endpoints using what looks like the _semantics_ of the [RFC 6570 URI Template](https://tools.ietf.org/html/rfc6570) standard. In fact, another API states this [pretty clearly](https://cloud.google.com/service-management/reference/rest/v1/services.configs#HttpRule.description.subsection). However, there is a difference between using the URI Template semantics and actually using the standard. This is because apparently RFC 6570 also states how UNICODE values are to be encoded when using a template, and this apparently differs from [RFC 3986 URI Generic Syntax](https://tools.ietf.org/html/rfc3986). This means that clients that use a URI Template library, like Google API Client, will handle UNICODE characters differently than clients that do not.


@sqrrrl @tbetbetbe Can you shed any light on this? Does Google have any guidance on whether API clients are supposed to use the URI Template semantics or adhere to the full standard?

Reproducing this outside of google-cloud-storage or google-api-client is pretty easy.

```ruby
require "addressable/template"
require "uri"

file = "#{[101, 769].pack("U*")}.png"
#=> "é.png"

template = Addressable::Template.new("http://example.com/bucket/{file}")
url1 = template.expand(file: file).to_str
#=> "http://example.com/bucket/%C3%A9.png"

url2 = "http://example.com/bucket/#{URI.encode file}"
#=> "http://example.com/bucket/e%CC%81.png"

URI.decode(url1) == URI.decode(url2)
#=> false
```

Google API Client creates and uses the URI Templating URL encoding, but the Storage Object's `self_link` URL is using the URI Generic Syntax encoding.

blowmage:
This online URI Template tool behaves the same way as the `addressable` gem, so I assume the behavior is per the RFC.

http://www.utilities-online.info/uritemplate/

sqrrrl:
AFAIK the intention was to adhere to the URI template behaviors.

blowmage:
Interesting. It looks like Storage does not adhere to the URI Template behavior then. Because the URL that Storage identifies the Object with is different than what two URI Template libraries that claim to adhere to the spec generate.

dfournier:
I signed it

sqrrrl:
Looks like it has to do with how the unicode normalization is happening. Looks like URI template (the RFC and libraries) use NFC form (composite character) for normalization while storage is using NFD (decomposed characters.)

They really should be considered equivalent though. They both represent the same thing (accented e in this case,) just different forms. The storage API *should* be normalizing the resource name to account for differences in user/client input.

blowmage:
The `é` in the original file string is comprised of two different characters. The first character, using the 101 codepoint, is an ASCII "e", and the second character uses the 769 code point and is a modifier and adds the slash to the previous character.  But when run through the URI Template, those two unicode code points (using 3 bytes) are replaced by a single code point for `é` that uses just two bytes. This is the source of the bug.

The [Character Encoding and Unicode Normalization](https://tools.ietf.org/html/rfc6570#section-1.6) section mentions equivalences between sequences of characters and normalizing character while encoding them. This is why I think the spec is the source of this behavior in the two URI Template libraries.

blowmage:
Ah dang, my browser didn't show me your reply until I posted my comment. Looks like we are on the same page though.

sqrrrl:
Filed a bug internally against cloud storage. It should accept both forms.

blowmage:
@sqrrrl Thank you sir! Since I don't have access to anything inside of Google, can you relay any activity on this issue here?

quartzmo:
Thank you @sqrrrl!

dfournier:
I encountered this problem with a migration script. I rewrote it with the [nodejs client](https://www.npmjs.com/package/google-cloud#google-cloud-storage) and the problem didn't occurred. So the bug might be in the ruby client.

blowmage:
The bug isn't in the ruby client, but its occurring because the ruby client is implemented differently than the node.js client. The bug is due to how unicode is being handled. In unicode the same character can be represented in several ways. You've specified one way when you named the file with `"#{[101, 769].pack("U*")}.png"`. The URI Template standard specifies a different way that those unicode characters are to be represented, and that is what the ruby client is doing that the node.js client is not.

The Cloud Storage API should accept both and treat them as equal, but it is not and you are seeing the Not Found response instead. So a bug has been opened against the Cloud Storage API to correct this.

At this point I don't expect to see any changes to the `google-cloud-storage` or the `google-api-client` gems to fix this. But once it is fixed we will accept this PR and the test to our acceptance tests so we will catch any API regressions.

blowmage:
@sqrrrl Can you give us an update on this? We can't see the internal bug, and are wondering if there is any movement on it. Thanks!

sqrrrl:
@blowmage No additional info at this time. Just pinged the bug again to get an update.

blowmage:
Thanks @sqrrrl! Yeah, I just checked this last week when we upgraded to Google API Client 0.10.x and the problem remains.

swcloud:
@blowmage Do you think this bug is release blocking?

blowmage:
@swcloud I don't know. The bug is with the storage service, and not the ruby lib. Can the ruby lib be blocked by upstream issues?

PR #1100, opened 2016-12-05
Add DATETIME, TIME, and BYTES query parameters

### LINKED ISSUE #1099: Add support for DATETIME, TIME, and BYTES query parameters
BigQuery has [DATETIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime-type), [TIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#time-type), and [BYTES](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#bytes-type) data types, but the current query parameter implementation does not yet support them. Make it so.

blowmage:
Add [DATETIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime-type), [TIME](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#time-type), and [BYTES](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#bytes-type) data types.

- [x] DATETIME
- [x] TIME
- [x] BYTES
- [x] Query Parameter Type Documentation

[closes #1099]


blowmage:
Here are how the BigQuery Query Parameter types map to Ruby types:

| BigQuery Query Parameter  | Ruby  | Notes  |
|---|---|---|
| `BOOL` | `true`/`false` | |
| `INT64` | `Integer` | |
| `FLOAT64` | `Float` | |
| `STRING` | `STRING` | |
| `DATETIME` | `DateTime` | DateTime objects's time zone is not sent, BigQuery DATETIME does not have time zone |
| `DATE` | `Date` | |
| `TIMESTAMP` | `Time` | |
| `TIME` | `BigQuery::Time` | There is a helper method to make creating bq time objects easier |
| `BYTES` | `File`/`StringIO` | We informally call these obects "IO-ish" |
| `ARRAY` | `Array` | Array must contain valid ruby types, but can't be a nested array |
| `STRUCT` | `Hash` | Hash can have String or Symbol keys, but the values must be valid ruby types. I think STRUCT is used as a JSON representation in BigQuery, but its hard to find docs for using STRUCT values in queries |

quartzmo:
@blowmage I will update this PR with additional documentation as you described above for data types. Thank you!

quartzmo:
I am okay with these commits being contributed to Google.

blowmage:
Yay!

PR #1101, opened 2016-12-06
Updates to speech using GAPIC LRO wrapper.

landrito:


blowmage:
I'll try to get this PR updated and passing tomorrow.

landrito:
Thanks @blowmage!

blowmage:
I get the following error when trying to use this code:

`LoadError: cannot load such file -- google/gax/operation`

That file path doesn't seem to exist. What version of `google-gax` will include it?

landrito:
Ah yes, it seems `google-gax` needs to be updated. I'll get it updated ASAP.

landrito:
Version `0.7.0` will include that file. We are holding off on publishing `0.7.0` because we are anticipating some changes to `google-gax` based on the feedback from this PR and expert review. Would you be able to make updates to this PR using the current master branch of `googleapis/gax-ruby`?

blowmage:
We are releasing tomorrow, so we won't be putting this in master until after the release.

landrito:
Going to hold off on this update until after some review in `googleapis/api-client-staging`

PR #1108, opened 2016-12-06
Fix post CI tasks

blowmage:
The post and coveralls were running on every node, but should only run once.


blowmage:
@quartzmo This can be merged. It fixes the build so the post task and coverage task are only run once.

PR #1109, opened 2016-12-07
Remove encryption_key_sha256

frankyn:
## Reasoning
@remi and I have been working on creating code samples to demonstrate encryption of Google Cloud Storage files using customer-supplied encryption keys. 

For consistency we reviewed other language samples currently shown in the Google Cloud Storage documentation ([storage/docs/encryption#sample_code](https://cloud.google.com/storage/docs/encryption#sample_code)) which demonstrate uploading files, providing an encryption key *only*.

Currently, the documentation for [Bucket#create_file](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.22.0/google/cloud/storage/bucket?method=create_file-instance) says that the `encryption_key_sha256` parameter *"Must be provided if encryption_key is provided"*.  If the SHA256 hash is not passed, however, the library currently [sets it by hashing the key](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/850fd5a5cb376d59ed379122ea18f23df937b8d0/google-cloud-storage/lib/google/cloud/storage/service.rb#L294).

To make using customer-supplied encryption keys easier and more consistent with other google-cloud libraries, this change removes the `encryption_key_sha256` parameter and updates the documentation. However, we did keep File#encryption_key_sha256 to get the SHA256 of an uploaded file. 

## Updates
Remove encryption_key_sha256 parameter from the following methods:

 - Bucket#file
 - Bucket#create_file
 - File#download
 - File#copy
 - File#get_file

The x-goog-encryption-key-sha256 header will be set to the SHA256 hash of the
encryption key provided.

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -33,7 +33,6 @@ describe Google::Cloud::Storage::File, :storage do
     cipher.encrypt
     cipher.random_key
   end
-  let(:encryption_key_sha256) { Digest::SHA256.digest encryption_key }
 
   before do
     # always create the bucket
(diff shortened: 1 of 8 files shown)


quartzmo:
@blowmage @omaray Is this proposed change `beta blocking`?

blowmage:
I've always been supportive of this change. If we do make this change I believe it would be best to launch beta with it rather than change this later.

PR #1111, opened 2016-12-07
Add doctest for BigQuery

blowmage:
Add doctest for BigQuery. A couple code examples proved difficult to cover using doctest, so they are skipped.

[refs #226]


quartzmo:
Incredible effort @blowmage! This was the biggest service to cover for #226, it will all be easier from here. Also, so great that you managed to catch and correct some incorrect examples.

PR #1129, opened 2016-12-12
Fix logging/middleware.rb doctest

hxiong388:
Fix doctest for logging/middleware.rb. This time, properly stub Google::Cloud::Core::Environment.get_metadata_attribute before each Middleware doctest, so no network connection is made from doctests.


quartzmo:
I tried out the changes to `doctest_helper.rb`, they fix the errors for me locally.

PR #1130, opened 2016-12-12
Update yard-doctest helpers, RELEASING.md, and CONTRIBUTING.md

quartzmo:
This PR stubs service factory methods and `Core` methods to raise errors if invoked. The purpose is to detect leaks in the doctest mocks that invoke actual backend resources.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contributing to google-cloud-ruby
+# Contributing to Google Cloud Ruby Client
 
 1. **Sign one of the contributor license agreements below.**
 2. Fork the repo, develop and test your code changes.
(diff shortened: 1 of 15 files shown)


quartzmo:
@swcloud @hxiong388 I made an attempt to document the [doctest testing](https://github.com/quartzmo/google-cloud-ruby/blob/doctest-raises/CONTRIBUTING.md#documentation-tests) for future contributors. Any feedback/ideas?

CONTRIBUTING.md
```diff
@@ -46,14 +46,38 @@ $ bundle exec rake console
...
+### Documentation Tests
+
+The project tests the code examples in the each gem's [YARD]()-based documentation.
+
+The example testing functions in a way that is very similar to unit testing, and in fact the library providing it, [yard-doctest](https://github.com/p0deje/yard-doctest), is based on the project's unit test library, [minitest](https://github.com/seattlerb/minitest).
+
+To run the documentation tests:
+
+``` sh
+$ rake jsondoc
```
hxiong388: Does doctest depend on jsondoc?

blowmage: No. It doesn't require any other task to be run before it. Not sure why `rake jsondoc` is listed here.

quartzmo: @hxiong388: It does in practice, but the dependency is not in rake, I think because `jsondoc` is pretty slow, and it's common to run `doctest` frequently while debugging the setup.

CONTRIBUTING.md
```diff
@@ -46,14 +46,38 @@ $ bundle exec rake console
 
 Tests are very important part of google-cloud-ruby. All contributions should include tests that ensure the contributed code behaves as expected.
 
+To run the unit tests, documentation tests, and code style checks together:
+
+``` sh
+$ rake ci
```
blowmage: Is it worth noting that users can run `rake ci:acceptance` (with handy alias of `rake ci:a`) to run all previous tasks plus acceptance tests?

CONTRIBUTING.md
quartzmo: On second thought, @blowmage, I can add the dependency in this PR if you like.

blowmage: I think adding the dependency would be best. You only have to add it to the services that actually implement doctest though! :smile:

blowmage:
```ruby
task doctest: :yard do
```

PR #1134, opened 2016-12-13
jsondoc: Raise error if return types is missing.

### LINKED ISSUE #1080: JSON docs generation should raise error if return type missing
The current gh-pages site is intolerant of a missing type declaration in `@return` tags. The JSON docs generator in `gcloud-jsondoc` branch should raise an error if missing, to avoid runtime errors that are hard to detect. For example, a missing return type currently causes the gh-pages site app to fail while rendering this view: https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.22.0/google/cloud/logging/logger

/cc @hxiong388

quartzmo:
I tested this change locally with all packages in `master`, `rake jsondoc` succeeded for all.

[closes #1080]

/cc @hxiong388

### DIFF
--- a/lib/gcloud/jsondoc/generator.rb
+++ b/lib/gcloud/jsondoc/generator.rb
@@ -58,5 +58,8 @@ module Gcloud
         end
       end
     end
+
+    class YardSyntaxError < RuntimeError
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


hxiong388:
Looks awesome! So much better than debugging the Angular app.

PR #1154, opened 2016-12-19
Refactor core protobuf conversion methods

blowmage:
Sort to match the type order of the Value protobufs.
Use the kind attribute that is used in the oneof definition.


blowmage:
I've updated this to add test coverage.

Most of these methods aren't used anywhere yet. Yet...

PR #1176, opened 2016-12-30
Exclude undocumented GAPIC Datastore and Logging code from YARD

### LINKED ISSUE #1146: Broken links to Protobuf types
Because YARD (our docs library) generates output that is scoped to Ruby classes and modules, Protobuf types that are exposed as **Ruby constants** are not currently published. So links in the gh-pages site to types like `Google::Datastore::V1::BeginTransactionResponse` (defined as a constant in lib/google/datastore/v1/datastore_pb.rb) are broken.

quartzmo:
[closes #1146]

### DIFF
--- a/google-cloud-datastore/.yardopts
+++ b/google-cloud-datastore/.yardopts
@@ -1,5 +1,6 @@
 --no-private
 --title=Google Cloud Datastore
+--exclude lib/google/datastore/v1
 --markup markdown
 
 ./lib/**/*.rb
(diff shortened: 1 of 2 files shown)


quartzmo:
Thanks @swcloud. Yes, because of (alphabetical?) processing order, the output from those files (xxx_pb.rb) was overwriting the output from the files in `/doc`, resulting in the broken links.

PR #1178, opened 2017-01-04
Update toc.json for GAPIC docs 

### LINKED ISSUE #1141: Low-level classes are displaying in the sidebar navigation of the documentation (Datastore::V1 and Logging::V2)
The sidebar navigation is showing new top-level headings for [`Datastore::V1`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.23.0/google/cloud/datastore/v1/datastoreclient) and [`Logging::V2`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.23.0/google/cloud/logging/v2/loggingservicev2client)

screen shot 2016-12-14 at 10 45 26 pm

If a user navigates to these pages (like I did), they will not find useful overviews of the client libraries for these services (like all of the other nav links show).

Instead, these display some of the internals of the gem that shouldn't be linked to from the top-level nav.

screen shot 2016-12-14 at 10 49 15 pm

screen shot 2016-12-14 at 10 49 35 pm

quartzmo:
This PR moves GAPIC items under Datastore and Logging services in the left nav of the gh-pages site. It enables a flag in `manifest.json` to use new focus support provided by GoogleCloudPlatform/gcloud-common#213. It also provides links in the pages shown for "Data Types" as discussed in GoogleCloudPlatform/gcloud-common#207.

[closes #1141]

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/v1/doc/google/datastore/v1/datastore.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/v1/doc/google/datastore/v1/datastore.rb
@@ -14,6 +14,43 @@
 
 module Google
   module Datastore
+    ##
+    # The `Google::Datastore::V1` module provides the following types:
+    #
+    # Class | Description
+    # ----- | -----------
+    # {Google::Datastore::V1::AllocateIdsRequest} | The request for Datastore::AllocateIds.
+    # {Google::Datastore::V1::AllocateIdsResponse} | The response for Datastore::AllocateIds.
+    # {Google::Datastore::V1::BeginTransactionRequest} | The request for Datastore::BeginTransaction.
+    # {Google::Datastore::V1::BeginTransactionResponse} | The response for Datastore::BeginTransaction.
+    # {Google::Datastore::V1::CommitRequest} | The request for Datastore::Commit.
+    # {Google::Datastore::V1::CommitResponse} | The response for Datastore::Commit.
+    # {Google::Datastore::V1::CompositeFilter} | A filter that merges multiple other filters.
+    # {Google::Datastore::V1::Entity} | A Datastore data object.
+    # {Google::Datastore::V1::EntityResult} | The result of fetching an entity from Datastore.
+    # {Google::Datastore::V1::Filter} | A holder for any type of filter.
+    # {Google::Datastore::V1::GqlQuery} | A query in the GQL grammar.
+    # {Google::Datastore::V1::GqlQueryParameter} | A binding parameter for a GQL query.
+    # {Google::Datastore::V1::Key} | A unique identifier for an entity.
+    # {Google::Datastore::V1::KindExpression} | A representation of a kind.
+    # {Google::Datastore::V1::LookupRequest} | The request for Datastore::Lookup.
+    # {Google::Datastore::V1::LookupResponse} | The response for Datastore::Lookup.
+    # {Google::Datastore::V1::Mutation} | A mutation to apply to an entity.
+    # {Google::Datastore::V1::MutationResult} | The result of applying a mutation.
+    # {Google::Datastore::V1::PartitionId} | A partition ID identifies a grouping of entities.
+    # {Google::Datastore::V1::Projection} | A representation of a property in a projection.
+    # {Google::Datastore::V1::PropertyFilter} | A filter on a specific property.
+    # {Google::Datastore::V1::PropertyOrder} | The desired order for a specific property.
+    # {Google::Datastore::V1::PropertyReference} | A property relative to the kind expressions.
+    # {Google::Datastore::V1::Query} | A query for entities.
+    # {Google::Datastore::V1::QueryResultBatch} | A batch of results produced by a query.
+    # {Google::Datastore::V1::ReadOptions} | The options shared by read requests.
+    # {Google::Datastore::V1::RollbackRequest} | The request for Datastore::Rollback.
+    # {Google::Datastore::V1::RollbackResponse} | The response for Datastore::Rollback.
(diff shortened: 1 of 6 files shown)


quartzmo:
@swcloud Should I add overview content for `google/cloud/datastore/v1` and `google/cloud/logging/v2` (currently undocumented and therefore blank)?

swcloud:
I guess what I meant was that since we don't have content for google/cloud/datastore/v1, when 'V1' is clicked, can we not change the content on the right panel? What it does just expands its two sub items. Is that doable?

google-cloud-datastore/lib/google/cloud/datastore/v1/doc/google/datastore/v1/datastore.rb
```diff
@@ -14,6 +14,43 @@
...
+    # {Google::Datastore::V1::PropertyReference} | A property relative to the kind expressions.
+    # {Google::Datastore::V1::Query} | A query for entities.
+    # {Google::Datastore::V1::QueryResultBatch} | A batch of results produced by a query.
+    # {Google::Datastore::V1::ReadOptions} | The options shared by read requests.
+    # {Google::Datastore::V1::RollbackRequest} | The request for Datastore::Rollback.
+    # {Google::Datastore::V1::RollbackResponse} | The response for Datastore::Rollback.
+    # {Google::Datastore::V1::RunQueryRequest} | The request for Datastore::RunQuery.
+    # {Google::Datastore::V1::RunQueryResponse} | The response for Datastore::RunQuery.
+    # {Google::Datastore::V1::Value} | Holds any of the supported value types and associated metadata.
+    #
```
swcloud: I'm worried that lib/google/cloud/datastore/v1/doc/google/datastore/v1/datastore.rb is auto-generated, now you have to manual-edit it.

google-cloud-logging/lib/google/cloud/logging/v2/doc/google/logging/v2/logging.rb
```diff
@@ -14,6 +14,35 @@
...
+    # {Google::Logging::V2::ListSinksResponse} | Result returned from ListSinks.
+    # {Google::Logging::V2::LogEntry} | An individual entry in a log.
+    # {Google::Logging::V2::LogEntryOperation} | Additional information about a potentially long-running operation with which a log entry is associated.
+    # {Google::Logging::V2::LogMetric} | Describes a logs-based metric.
+    # {Google::Logging::V2::LogSink} | Describes a sink used to export log entries outside of Stackdriver Logging.
+    # {Google::Logging::V2::UpdateLogMetricRequest} | The parameters to UpdateLogMetric.
+    # {Google::Logging::V2::UpdateSinkRequest} | The parameters to UpdateSink.
+    # {Google::Logging::V2::WriteLogEntriesRequest} | The parameters to WriteLogEntries.
+    # {Google::Logging::V2::WriteLogEntriesResponse} | Result returned from WriteLogEntries.
+    #
```
swcloud: I'm worried that lib/google/cloud/logging/v2/doc/google/logging/v2/logging.rb is auto-generated, now you have to manual-edit it.

quartzmo: @swcloud Good point. But I can't think of what else to do. I'm open to ideas.

quartzmo:
@swcloud I can map "V1" to `"google/cloud/datastore/v1/datastoreclient"` so it will place focus on the first child, "DatastoreClient". Is that better?

But I think it's really a huge omission not to have an overview of the GAPIC layer! Without it, there is nothing in the documentation that explains the nav item "V1".

quartzmo:
@swcloud To answer your question, I do not think it is appropriate to introduce special-case behavior to the navigation of the gh-pages site. I am also having second thoughts about my suggestion, above, to directing the "V1" nav item to the docs for `DatastoreClient `. Here is what would be presented:

> Each RPC normalizes the partition IDs of the keys in its input entities, and always returns entities with keys with normalized partition IDs. This applies to all keys and entities, including those in values, except keys with both an empty path and an empty or unset partition ID. Normalization of input keys sets the project ID (if not already set) to the project ID from the request.

I think anyone visiting the page expecting to find an explanation of "V1" will be hopelessly confused by this content.

The generated code contains docs for methods and attributes, should it not also contain an overview?

google-cloud-logging/lib/google/cloud/logging/v2/doc/google/logging/v2/logging.rb
swcloud: I'm not sure how the Angular app works. Can this be done by running some javascript from the Angular app?

quartzmo: We could add additional logic to the Angular app allowing us to map the nav item "V1" to a file with this content. That file would be separate from the generated code. But personally I think in practice it would be much easier to just manually update the new versions of the generated code. The separate file would also need to be versioned and updated, so it's not any easier, I think.

quartzmo:
@swcloud If you are OK with this PR, you can approve it and I will merge and update the docs with these changes. At least then the left nav will support the design from  GoogleCloudPlatform/gcloud-common#207, and we can discuss what to do about page content.

quartzmo:
@swcloud I'm going to merge this in about 30 minutes unless you have any objections.

swcloud:
OK. I think we need more discussion on "V1" content with @omaray  and how to generate the content for the table in some later PRs. I will file separate issues.

quartzmo:
Perfect, I agree.

PR #1179, opened 2017-01-05
Add Bucket#post_object method to facilitate multipart form uploads.

### LINKED ISSUE #1177: Storage POST objects with the xml-api using an html multipart form
Hi! 👋 

I'd like to upload objects to a storage bucket using signed post forms, like described in this guide:

https://cloud.google.com/storage/docs/xml-api/post-object

I thought everything I needed was already present in the [File::Signer class](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/file.rb#L728-L813), but I'm having a hard time generating signatures like described in the guide's example.

My idea was to add a method that would return an object with the URL to send the form to and a hash of fields to set in the input fields. I was thinking on allowing the api clients to pass a policy document that would be signed in the same fashion that the example shows, but I cannot make it work:

```python
from base64 import b64encode
from M2Crypto import EVP
def Sign(pem_key, message):
    assert pem_key
    key = EVP.load_key_string(pem_key)
    key.reset_context(md='sha256')
    key.sign_init()
    key.sign_update(str(message))
    return key.sign_final()

SecretKey = open('key.pem', 'r').read()
Policy = b64encode(open('policy.txt', 'r').read())
Signature = b64encode(Sign(SecretKey, Policy))
print Policy
print
print Signature
```

I'd really appreciate some advice.

calavera:
People can use this method to generate policy signatures for uploading
objects using html forms.

See: https://cloud.google.com/storage/docs/xml-api/post-object

Fixes #1177.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -676,6 +676,94 @@ module Google
         end
 
         ##
+        # Generate a PostObject that includes the fields and url to
+        # upload objects via html forms.
+        #
+        # Generating a PostObject requires service account credentials,
+        # either by connecting with a service account when calling
+        # {Google::Cloud.storage}, or by passing in the service account
+        # `issuer` and `signing_key` values. Although the private key can
+        # be passed as a string for convenience, creating and storing
+        # an instance of # `OpenSSL::PKey::RSA` is more efficient
+        # when making multiple calls to `post_object`.
+        #
+        # A {SignedUrlUnavailable} is raised if the service account credentials
+        # are missing. Service account credentials are acquired by following the
+        # steps in [Service Account Authentication](
+        # https://cloud.google.com/storage/docs/authentication#service_accounts).
+        #
+        # @see https://cloud.google.com/storage/docs/xml-api/post-object
+        #
+        # @param [String] path Path to of the file in Google Cloud Storage.
+        # @param [Hash] policy The security policy that describes what
+        #   can and cannot be uploaded in the form. When provided,
+        #   the PostObject fields will include a Signature based on the JSON
+        #   representation of this Hash and the same policy in Base64 format.
+        #   If you do not provide a security policy, requests are considered
+        #   to be anonymous and will only work with buckets that have granted
+        #   WRITE or FULL_CONTROL permission to anonymous users.
+        #   See [Policy Document](https://cloud.google.com/storage/docs/xml-api/post-object#policydocument)
+        #   for more information.
+        # @param [String] issuer Service Account's Client Email.
+        # @param [String] client_email Service Account's Client Email.
+        # @param [OpenSSL::PKey::RSA, String] signing_key Service Account's
+        #   Private Key.
+        # @param [OpenSSL::PKey::RSA, String] private_key Service Account's
+        #   Private Key.
+        #
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-todo-app"
+        #   post = bucket.post_object "avatars/heidi/400x400.png"
+        #
+        # @example Using a policy to define the upload authorization
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   policy = {
+        #     expiration: (Time.now + 3600).iso8601,
+        #     conditions: [
+        #       ["starts-with", "$key", ""],
+        #       {acl: "bucket-owner-read"},
+        #       {bucket: "travel-maps"},
+        #       {success_action_redirect: "http://example.com/success.html"},
+        #       ["eq", "$Content-Type", "image/jpeg"],
+        #       ["content-length-range", 0, 1000000]
+        #     ]
+        #   }
+        #
+        #   bucket = storage.bucket "my-todo-app"
+        #   post = bucket.post_object "avatars/heidi/400x400.png",
+        #                              policy: policy
+        #
+        # @example Using the issuer and signing_key options:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-todo-app"
+        #   key = OpenSSL::PKey::RSA.new "-----BEGIN PRIVATE KEY-----\n..."
+        #   post = bucket.post_object "avatars/heidi/400x400.png",
+        #                             issuer: "service-account@gcloud.com",
+        #                             signing_key: key
+        #
+        def post_object path, policy: nil, issuer: nil,
+                        client_email: nil, signing_key: nil,
+                        private_key: nil
+          ensure_service!
+          options = { issuer: issuer, client_email: client_email,
+                      signing_key: signing_key, private_key: private_key,
+                      policy: policy }
+
+          signer = File::Signer.from_bucket self, path
+          signer.post_object options
+        end
+
+        ##
         # The Bucket::Acl instance used to control access to the bucket.
         #
         # A bucket has owners, writers, and readers. Permissions can be granted
@@ -841,6 +929,14 @@ module Google
             @updates << attribute
             @updates.uniq!
           end
+        end
+      end
+
+      class PostObject
+        attr_reader :url, :fields
+        def initialize url, fields
+          @url = url
+          @fields = fields
         end
       end
     end
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -21,6 +21,8 @@ require "google/cloud/storage/file/verifier"
 module Google
   module Cloud
     module Storage
+      GOOGLEAPIS_URL = "https://storage.googleapis.com".freeze
+
       ##
       # # File
       #
@@ -749,7 +751,7 @@ module Google
           ##
           # The external url to the file.
           def ext_url
-            "https://storage.googleapis.com#{ext_path}"
+            "#{GOOGLEAPIS_URL}#{ext_path}"
           end
 
           def apply_option_defaults options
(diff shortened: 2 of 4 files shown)


quartzmo:
@calavera Thanks for this PR! (And for dealing with the CLA.) It looks like the build failed on rubocop style checks. Please see [CONTRIBUTING - Tests](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/CONTRIBUTING.md#tests) for details. (TLDR: `$ bundle exec rake ci` in `google-cloud-storage`.)

quartzmo:
Can you also add an acceptance test to [`acceptance/storage/bucket_test.rb`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/acceptance/storage/bucket_test.rb)? Assuming you've been developing this against the live service, it should be pretty easy to do. You already have a nice unit test from which to borrow.

quartzmo:
I want to hold off on merging this until early next week, to give @blowmage a chance to look at it. (He authored the Storage support in google-cloud-ruby, and is on vacation today and tomorrow.)

calavera:
no rush, I'm using my fork for now and I'm also waiting to get access to the HTTPS storage load balancing alpha before putting it in production.

I can make those changes in the meantime.

calavera:
@quartzmo I'm not sure what the best way to add an acceptance test.

This new method only generates the signature for people to use in multipart html forms. I guess the idea would be to create a multiparted object, send a post request to the storage api and see if it completed. Unfortunately, neither Ruby net/http nor httpclient(the library that google-api-client relies on) support multipart requests, as far as I know. I've used this other gem in the past for similar purposes, but I'm not sure if you're okay with me adding an extra dependency: https://github.com/nicksieger/multipart-post.

I'd really appreciate your feedback on how to proceed.

quartzmo:
@calavera Oh sorry--I didn't look closely enough at your use case! I think we can skip the acceptance test in this case, unless @blowmage is enthusiastic about adding the dependency. I don't feel like it's worth it.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -775,6 +777,35 @@ def determine_issuer options = {}
               @service.credentials.issuer
           end
 
+          def post_object options
+            options = apply_option_defaults options
+
+            fields = {
+              key: ext_path.sub("/", "")
```
blowmage: @calavera Why is the first `/` being removed here?

calavera: I'm following what the example in the docs says:

```md
Key: 
The name of the object that you are uploading.
If you specify the object name without the bucket name, 
you must append the bucket name to the URI (for example, storage.googleapis.com/<bucket>). 
If you specify the bucket name and object name (for example, bucket/object),
you must specify only the URI (for example, storage.googleapis.com).
You can also use the ${filename} variable if a user is providing a file name.
Also, if you are using a bucket name that is a CNAME redirect,
be sure that the bucket name is specified only once in your request.
```

This way, the URL is `https://storage.googleapis.com` and the KEY is `mybucket/myobject`.

To be honest, I didn't try to leave the ext path as it was, but I know this version works 😄

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -676,6 +676,88 @@ def signed_url path, method: nil, expires: nil, content_type: nil,
...
+        # are missing. Service account credentials are acquired by following the
+        # steps in [Service Account Authentication](
+        # https://cloud.google.com/storage/docs/authentication#service_accounts).
+        #
+        # @see https://cloud.google.com/storage/docs/xml-api/post-object
+        #
+        # @param [String] path Path to of the file in Google Cloud Storage.
+        # @param [Hash] policy When provider, the PostObject fields will
+        #   include a Signature based on the JSON representation of this Hash
+        #   and the same policy in Base64 format.
```
quartzmo: I'm thinking it's better to err on the side of too much description with this param. What about (changes in bold):

**The security policy that describes what can and cannot be uploaded in the form.** When **provided**, the PostObject fields will include a Signature based on the JSON representation of this Hash  and the same policy in Base64 format. **If you do not provide a security policy, requests are considered to be anonymous and will only work with buckets that have granted `WRITE` or `FULL_CONTROL` permission to anonymous users.**

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -676,6 +676,88 @@ def signed_url path, method: nil, expires: nil, content_type: nil,
...
+        #   }
+        #
+        #   bucket = storage.bucket "my-todo-app"
+        #   post = bucket.post_object "avatars/heidi/400x400.png",
+        #                              policy: policy
+        #
+        # @example Using the issuer and signing_key options:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud.storage
```
quartzmo: Use the new syntax instead: `storage = Google::Cloud::Storage.new`

calavera:
is there anything else you need from me to get this merged? I want to make sure I'm not missing anything.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -676,6 +676,88 @@ def signed_url path, method: nil, expires: nil, content_type: nil,
...
+        # https://cloud.google.com/storage/docs/authentication#service_accounts).
+        #
+        # @see https://cloud.google.com/storage/docs/xml-api/post-object
+        #
+        # @param [String] path Path to of the file in Google Cloud Storage.
+        # @param [Hash] policy When provider, the PostObject fields will
+        #   include a Signature based on the JSON representation of this Hash
+        #   and the same policy in Base64 format.
+        # @param [String] issuer Service Account's Client Email.
+        # @param [String] client_email Service Account's Client Email.
```
quartzmo: Do we need both `issuer` and `client_email` options? Also, downcase "Client Email" to "client email".

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -676,6 +676,88 @@ def signed_url path, method: nil, expires: nil, content_type: nil,
...
+        # @param [OpenSSL::PKey::RSA, String] signing_key Service Account's
+        #   Private Key.
+        # @param [OpenSSL::PKey::RSA, String] private_key Service Account's
+        #   Private Key.
```
quartzmo: Do we need both `signing_key ` and `private_key ` options? Also, downcase "Private Key" to "private key".

google-cloud-storage/lib/google/cloud/storage/bucket.rb
calavera: either option can be used to determine the issuer. I copied this from `signed_url`. What do you think about leaving it as it as for consistency?

See https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/file.rb#L773

google-cloud-storage/lib/google/cloud/storage/bucket.rb
calavera: either option can be used to determine the issuer. I copied this from `signed_url`. What do you think about leaving it as it as for consistency?

See https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/file.rb#L768

google-cloud-storage/lib/google/cloud/storage/bucket.rb
calavera: changed, thanks!

google-cloud-storage/lib/google/cloud/storage/bucket.rb
calavera: changed, thanks!

google-cloud-storage/lib/google/cloud/storage/bucket.rb
blowmage: This should remain consistent.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
blowmage: This should remain consistent.

quartzmo:
Sorry, the line with the link I suggested adding caused a rubocop error:

```
lib/google/cloud/storage/bucket.rb:705:108: C: Line is too long. [129/80]
        #   See [Policy Document](https://cloud.google.com/storage/docs/xml-api/post-object#policydocument) for more information.
                                                                                                           ^^^^^^^^^^^^^^^^^^^^^^
```

Just add a newline after the link.

calavera:
It's all 💚 now 🎉

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -676,6 +676,93 @@ def signed_url path, method: nil, expires: nil, content_type: nil,
...
+        # @param [Hash] policy The security policy that describes what
+        #   can and cannot be uploaded in the form. When provided,
+        #   the PostObject fields will include a Signature based on the JSON
+        #   representation of this Hash and the same policy in Base64 format.
+        #   If you do not provide a security policy, requests are considered
+        #   to be anonymous and will only work with buckets that have granted
+        #   WRITE or FULL_CONTROL permission to anonymous users.
+        #   See [Policy Document](https://cloud.google.com/storage/docs/xml-api/post-object#policydocument)
+        #   for more information.
+        # @param [String] issuer Service Account's Client Email.
```
quartzmo: Uh oh, it looks like `client_email` has disappeared from the docs.

calavera: @quartzmo what do you think if I remove it from everywhere in a next PR? I'm not sure if it makes sense to remove it from only one place when it should be removed from everywhere

quartzmo: @calavera I'm sorry, I didn't understand @blowmage's earlier comment "This should remain consistent", but what he wants is to keep both `issuer` and `client_email` options and both `signing_key` and `private_key` options.  (I didn't get that this was the existing design, sorry.) So can you please put it back as you had it before, with all four of these options? Then I promise to merge immediately!

calavera: OMG, I'm sorry, I totally misunderstood you. I just added the `client_email` back.

quartzmo: Awesome! I will merge when build completes. Thanks for your patience!

blowmage:
I'm confusing like that. :)

blowmage:
:+1:

calavera:
awesome, thanks! 🎊

PR #1183, opened 2017-01-05
Jsondoc tables

quartzmo:
Add the `table` class to markdown tables for correct style in Angular app.

@swcloud This fixes the table style issue in the "Data Types" page.


quartzmo:
Cancelled Circle CI, as it should not run for `gcloud-jsondoc` branch.

PR #1195, opened 2017-01-19
Add File#rotate

### LINKED ISSUE #730: Support rotating keys with File#rewrite
The following Storage API method is not currently supported by gcloud-ruby:
- [objects.rewrite](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite) - Rewrites a source object to a destination object. Optionally overrides metadata. Can be used to [rotate keys](https://cloud.google.com/storage/docs/encryption#customer-supplied) (see #722.)

quartzmo:
This PR provides a method for rotating customer-supplied encryption keys that is similar to [`rotate_encryption_key`](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/cloud-client/encryption.py#L96) in google-cloud-python.

It does *not* expose the full options list of [`objects.rewrite`](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite). I feel that should be discussed in a follow-up issue and pull request, since it may make sense to change `File#copy` to use `objects.rewrite` and expose the full power of `objects.rewrite`, including callbacks/controls for multiple requests.  

[closes #730]

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -27,9 +27,17 @@ describe Google::Cloud::Storage::File, :storage do
       big:  { path: "acceptance/data/three-mb-file.tif" } }
   end
 
-  let(:encryption_key) do
+  let(:cipher) do
     cipher = OpenSSL::Cipher.new "aes-256-cfb"
     cipher.encrypt
+    cipher
+  end
+
+  let(:encryption_key) do
+    cipher.random_key
+  end
+
+  let(:encryption_key_2) do
     cipher.random_key
   end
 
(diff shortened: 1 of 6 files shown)


google-cloud-storage/acceptance/storage/file_test.rb
```diff
@@ -288,6 +296,40 @@
     copied.delete
   end
 
+  it "should add, rotate, and remove customer-supplied encryption keys for an existing file" do
```
blowmage: I love this test. Great job!

quartzmo: 😻  Thanks!

PR #1211, opened 2017-01-27
Include GAPIC Natural Language code in docs

quartzmo:
Include GAPIC in YARD
Include GAPIC in google-cloud-language and google-cloud TOCs

### DIFF
--- a/google-cloud-language/.yardopts
+++ b/google-cloud-language/.yardopts
@@ -1,6 +1,5 @@
 --no-private
 --title=Google Cloud Language
---exclude lib/google/cloud/language/v1
 --exclude lib/google/cloud/language/v1.rb
 --markup markdown
 
--- a/google-cloud-language/docs/toc.json
+++ b/google-cloud-language/docs/toc.json
@@ -45,6 +45,17 @@
         {
           "title": "Annotation",
           "type": "google/cloud/language/annotation"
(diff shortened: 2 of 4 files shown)


google-cloud-language/.yardopts
```diff
@@ -1,6 +1,5 @@
 --no-private
 --title=Google Cloud Language
---exclude lib/google/cloud/language/v1
 --exclude lib/google/cloud/language/v1.rb
```
blowmage: Shouldn't this exclusion also be removed?

quartzmo: No, the namespaces are different in this package for some reason. There is no `lib/google/cloud/language/v1/doc/google/language/v1` path as in other packages, just `lib/google/cloud/language/v1/doc/google/cloud/language/v1`, so it needs to be included.

quartzmo: Sorry, misunderstood your question. The `lib/google/cloud/<package>/<version>.rb` always needs to be excluded so that the docs from types in the `doc` directory are not overwritten.

blowmage: Why does that one file need to be excluded?

blowmage: Gotcha.

PR #1220, opened 2017-02-07
Datastore.new accepts emulator_host params

rakusai:
For the local development or test purpose, I always use [Datastore Emulator](https://cloud.google.com/datastore/docs/tools/datastore-emulator).

The pain point to use it is we have to set the environmental variables like below before we start ruby server.

`ENV["DATASTORE_EMULATOR_HOST"]`

If the ruby (especially rails) development, those configuration should be included in the files like `settings.yaml`, etc.
We set project name params in such files but only the `emulator host` should in  environmental variables. 

Sometimes our team members start development without `ENV["DATASTORE_EMULATOR_HOST"]` and they write and read data remotely from the local computer. This is big accident.

So I believe we should accept `emulator_host` params just like `project` params in the `Datastore.new` method.

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore.rb
@@ -545,6 +545,8 @@ module Google
       # @param [Integer] timeout Default timeout to use in requests. Optional.
       # @param [Hash] client_config A hash of values to override the default
       #   behavior of the API client. See Google::Gax::CallSettings. Optional.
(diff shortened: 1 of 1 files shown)


rakusai:
I signed it!

blowmage:
Thanks for the PR @rakusai. This looks good to me. We will want to have tests for this, and add this parameter to Pub/Sub's initializer as well. I'm gonna wait to merge this until we get the other changes queued up and then we will merge both at the same time. Shouldn't be more than a day or two.

Issue #1215, opened 2017-01-31
Consider dropping GRPCUtils


blowmage:
Some of the internal shared code in google-cloud-core are some helper methods for dealing with common protobuf types such as `Google::Protobuf::Value`. It was assumed that all libraries would use these common types the same way, so these helper methods would be shared. However, we've learned that some libraries use these common types in slightly different ways, and the little reuse we were hoping for hasn't panned out.

I think it would be simpler to have all these helper methods local to the library. Yes, there will be some duplication of these helper methods, but in practice only a few methods are used by any service.

quartzmo:
@blowmage You want to leave this open until we remove `GRPCUtils`?

blowmage:
Not sure, I can see closing, and I can see keeping it open.

blowmage:
This change was made. All gems now own their own logic for dealing with core Protobuf types.

PR #1221, opened 2017-02-08
Reconsider GRPCUtils

blowmage:
`GRPCUtils` is a shared module, but a soon-to-be-released library has divergent logic and cannot use the shared logic. This situation will introduce confusion around when the shared module is to be used vs. not to be used. To avoid confusion, each library that needs some of the logic has added a local `Convert` module that will hold the conversion logic. The amount of logic duplicated across services is extremely minimal.

`GRPCUtils` should be removed when we release 1.0.

[refs #1215]

### DIFF
--- a/google-cloud-logging/test/google/cloud/logging/entry/to_grpc_test.rb
+++ b/google-cloud-logging/test/google/cloud/logging/entry/to_grpc_test.rb
@@ -27,7 +27,7 @@ describe Google::Cloud::Logging::Entry, :to_grpc, :mock_logging do
     grpc.severity.must_equal :DEFAULT
     grpc.timestamp.must_be :nil?
     grpc.insert_id.must_be :empty?
-    Google::Cloud::Core::GRPCUtils.map_to_hash(grpc.labels).must_be :empty?
+    Google::Cloud::Logging::Convert.map_to_hash(grpc.labels).must_be :empty?
     grpc.text_payload.must_be :empty?
     grpc.json_payload.must_be :nil?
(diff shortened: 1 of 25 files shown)


google-cloud-logging/test/google/cloud/logging/entry/to_grpc_test.rb
```diff
@@ -71,14 +71,14 @@
...
                                                                      "valueType"   => "STRING" })
 
     grpc.severity.must_equal :ERROR
     grpc.timestamp.must_equal Google::Protobuf::Timestamp.new(seconds: Time.parse("2016-01-02T03:04:05Z").to_i)
     grpc.insert_id.must_equal "insert123"
-    Google::Cloud::Core::GRPCUtils.map_to_hash(grpc.labels).must_equal("env" => "test", "fizz" => "buzz")
+    Google::Cloud::Logging::Convert.map_to_hash(grpc.labels).must_equal("env" => "test", "fizz" => "buzz")
     grpc.text_payload.must_equal "payload"
     grpc.json_payload.must_be :nil?
     grpc.proto_payload.must_be :nil?
```
quartzmo: Want to copy these `Convert.map_to_hash` tests over to Natural Language? I believe its implementation is lacking tests.

blowmage: There are no existing tests specifically for `map_to_hash`. I'm confident that the functionality is properly covered by the other unit and acceptance tests.

Issue #1180, opened 2017-01-05
Generate content automatically for the table shown on the 'Data Types' page.


swcloud:
Right now, the content is manually added to the generated dummy ruby classes. We need to either generate the content for the table or add Angular app logic to extract the data from the ruby classes in order to display the table.

quartzmo:
I agree.

quartzmo:
My thinking is that until other platforms using the Angular site app have the same problem, it will be better to implement this in the `gcloud-jsondoc` build tool. I think the best approach will be a flag in `toc.json` that will currently only be processed by the build tool, but which could be handled by the Angular app at a later date with no change needed in the configuration. For example, a `generateTocDescription` flag as shown near the bottom of this `toc.json` fragment:

```js
        {
          "title": "V1",
          "type": "google/cloud/datastore/v1",
          "patterns": ["\/datastore\/v1"],
          "nav": [
            {
              "title": "DatastoreClient",
              "type": "google/cloud/datastore/v1/datastoreclient"
            },
            {
              "title": "Data Types",
              "type": "google/datastore/v1",
              "generateTocDescription": true
            }
          ]
        }
```

When true, the description would be appended with a generated table linking to all children in the namespace in the `type` attribute.

blowmage:
We could build a YARD plugin to generate this, instead of putting it in gcloud-jsondoc.

quartzmo:
I'd rather keep all these hacks in one place ;)

blowmage:
I am not in favor of having `gcloud-jsondoc` implement this, since the documentation won't be added to the ruby files and be available to services like rubydoc.info that generate documentation from the files distributed in the gems. Would it be possible to have GAPIC generate the "Data Types" documentation?

quartzmo:
@blowmage I don't think access in rubydoc.info is a problem (see the left nav for [Google/Datastore/V1/AllocateIdsRequest](http://www.rubydoc.info/gems/google-cloud-datastore/Google/Datastore/V1/AllocateIdsRequest), for example.) The need for this table is purely in the gh-pages site, because of the lack of a complete "index" listing in the left nav, or am I mistaken?

blowmage:
Right now the gh-pages site needs a namespace (module or class) to navigate to , so my assumption here is that this documentation will be added to a module or class. Are you thinking about creating a non-existing namespace for the gh-pages app to use?

quartzmo:
[...]

PR #1224, opened 2017-02-09
Generate TOC pages in jsondoc

quartzmo:
See update to README, below, for description of this feature.

[refs #1180]

Updated screenshot of generated TOC for Datastore. Note that although multiple tables with different titles *can* be used (as below), all of the listed types could also be shown in a single table.

datatypes-2

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,10 +64,20 @@ require "gcloud/jsondoc"
 registry = YARD::Registry.load! ".yardoc"
 
 toc_config = {
-  types: [
+  documents: [
     {
+      type: "toc",
       title: "Google::Datastore::V1::DataTypes",
-      toc: { package: "Google::Datastore::V1", include: "google/datastore/v1/" }
+      modules: [
+        {
+          title: "Google::Protobuf",
+          include: ["google/protobuf"]
+        },
+        {
+          title: "Google::Datastore::V1",
+          include: ["google/datastore/v1"]
+        }
+      ]
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -34,7 +34,9 @@ generator = Gcloud::Jsondoc::Generator.new registry
 generator.write_to "jsondoc"
 ```
 
-Note: If the source code lives in a sub-directory of the main repo, as it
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -49,6 +51,31 @@ generator.write_to "jsondoc"
...
+```ruby
+require "gcloud/jsondoc"
+
+registry = YARD::Registry.load! ".yardoc"
+
+toc_config = {
+  types: [
+    {
+      title: "Google::Datastore::V1::DataTypes",
+      toc: { package: "Google::Datastore::V1", include: "google/datastore/v1/" }
```
blowmage: This works well for Datastore, because `Google::Cloud::Datastore::V1` and `Google::Datastore::V1` are separate namespaces, but for other services the GAPIC code and Protobuf code share the same namespace. How will Jsondoc differentiate between the two?

blowmage: Also, I believe some services have types that are defined in multiple namespaces. Datastore has types in `Google::Datastore::V1` and `Google::Protobuf`.

quartzmo: Ugh, thanks for reminding me, originally `include` accepted an array. I meant to reinstate that but just got impatient due to how much time this seemingly "easy" feature has already taken.

quartzmo: Adding `exclude` [Array] as well, just in case it's needed.

test/generated_toc_doc_test.rb
```diff
@@ -0,0 +1,106 @@
...
+            {
+              title: "IncludedModule2",
+              include: ["includedmodule2/"],
+              exclude: ["includedmodule2/nested"]
+            }
+          ]
+        }
+      ]
+    }
+  end
```
quartzmo: @blowmage Let me know if you think this configuration design is sufficient.

quartzmo: The first `exclude` property is a copy/paste error, but it has no effect.

blowmage: Yeah, I think that is a great design.

PR #1225, opened 2017-02-09
Support for Per-Object Storage Class

blowmage:
Support Per-Object Storage Class (POSC) on File creation, and File update.

Files are created with a custom POSC using the `storage_class` method argument:

```ruby

bucket.storage_class #=> "STANDARD"

bucket.create_file "/var/todo-app/avatars/heidi/400x400.png",
                   "avatars/heidi/400x400.png",
                   storage_class: :regional
```

Existing files can modify their POSC using the `storage_class=` method:

```ruby

file = bucket.file "avatars/heidi/400x400.png"

file.storage_class #=> "REGIONAL"

# Update POSC
file.storage_class = "NEARLINE"
```

/cc @jskeet @omaray @danoscarmike

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -247,6 +247,18 @@ describe Google::Cloud::Storage::File, :storage do
     uploaded.metadata["title"].must_equal meta[:metadata][:title]
   end
 
+  it "should create and update storage_class" do
+    uploaded = bucket.create_file files[:logo][:path], "CloudLogo-storage_class.png", storage_class: :nearline
+
+    uploaded.storage_class.must_equal "NEARLINE"
+    uploaded.storage_class = :dra
+    uploaded.storage_class.must_equal "DURABLE_REDUCED_AVAILABILITY"
+
+    retrieved1 = bucket.file "CloudLogo-storage_class.png"
+
+    retrieved1.storage_class.must_equal "DURABLE_REDUCED_AVAILABILITY"
+  end
+
   it "should copy an existing file" do
     uploaded = bucket.create_file files[:logo][:path], "CloudLogo"
     copied = try_with_backoff "copying existing file" do
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -514,6 +514,16 @@ module Google
         # @param [Hash] metadata A hash of custom, user-provided web-safe keys
         #   and arbitrary string values that will returned with requests for the
         #   file as "x-goog-meta-" response headers.
+        # @param [Symbol, String] storage_class Storage class of the file.
+        #   Determines how the file is stored and determines the SLA and the
+        #   cost of storage. Values include `:multi_regional`, `:regional`,
+        #   `:nearline`, `:coldline`, `:standard`, and `:dra` (Durable Reduced
+        #   Availability), as well as the strings returned by
+        #   {Bucket#storage_class}. For more information, see [Storage
+        #   Classes](https://cloud.google.com/storage/docs/storage-classes) and
+        #   [Per-Object Storage
+        #   Class](https://cloud.google.com/storage/docs/per-object-storage-class).
+        #   The default value is the default storage class for the bucket.
         # @param [String] encryption_key Optional. A customer-supplied, AES-256
         #   encryption key that will be used to encrypt the file.
         #
@@ -561,14 +571,14 @@ module Google
                         content_disposition: nil, content_encoding: nil,
                         content_language: nil, content_type: nil,
                         crc32c: nil, md5: nil, metadata: nil,
-                        encryption_key: nil
+                        storage_class: nil, encryption_key: nil
           ensure_service!
           options = { acl: File::Acl.predefined_rule_for(acl), md5: md5,
                       cache_control: cache_control, content_type: content_type,
                       content_disposition: content_disposition, crc32c: crc32c,
-                      content_encoding: content_encoding,
-                      content_language: content_language, metadata: metadata,
-                      key: encryption_key }
+                      content_encoding: content_encoding, metadata: metadata,
+                      content_language: content_language, key: encryption_key,
+                      storage_class: storage_class_for(storage_class) }
           ensure_file_exists! file
           # TODO: Handle file as an IO and path is missing more gracefully
           path ||= Pathname(file).to_path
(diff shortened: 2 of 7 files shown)


google-cloud-storage/acceptance/storage/file_test.rb
```diff
@@ -247,6 +247,18 @@
     uploaded.metadata["title"].must_equal meta[:metadata][:title]
   end
 
+  it "should create and update storage_class" do
+    uploaded = bucket.create_file files[:logo][:path], "CloudLogo-storage_class.png", storage_class: :nearline
+
+    uploaded.storage_class.must_equal "NEARLINE"
+    uploaded.storage_class = :dra
+    uploaded.storage_class.must_equal "DURABLE_REDUCED_AVAILABILITY"
```
quartzmo: What do you think about changing the target storage class? I'm just afraid that every once in while, people look to tests for best practices, and `DURABLE_REDUCED_AVAILABILITY` is one that it is recommended to [move *away* from](https://cloud.google.com/storage/docs/storage-classes#dra).

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -268,6 +268,29 @@ def encryption_key_sha256
...
+          @gapi.storage_class
+        end
+
+        ##
+        # Updates how the file is stored and determines the SLA and the cost of
+        # storage. Values include `:multi_regional`, `:regional`, `:nearline`,
+        # `:coldline`, `:standard`, and `:dra` (Durable Reduced Availability),
+        # as well as the strings returned by {File#storage_class} or
+        # {Bucket#storage_class}. For more information, see [Per-Object Storage
+        # Class](https://cloud.google.com/storage/docs/per-object-storage-class).
```
quartzmo: If only one official docs page is linked, I think [docs/storage-classes](https://cloud.google.com/storage/docs/storage-classes) is the better one to clarify the choices. The `docs/per-object-storage-class` page linked here is also valuable, especially for the region vs multi-region section, but less relevant.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -903,9 +938,14 @@ def format_extension_headers headers
         ##
         # Yielded to a block to accumulate changes for a patch request.
         class Updater < File
+          # @private Do not allow storage_class to be set in an update call.
```
quartzmo: This is because `patch` doesn't support it? What if `patch` does support it in the future, would this be removed?

quartzmo: I suppose changing the storage class must always be a `rewrite` op.

quartzmo: Since I had to investigate, might be good to note in this comment.

google-cloud-storage/test/google/cloud/storage/file_update_test.rb
```diff
@@ -106,6 +106,24 @@
...
+    patched_file_gapi.storage_class = "DURABLE_REDUCED_AVAILABILITY"
+    rewrite_resp = OpenStruct.new done: true, resource: patched_file_gapi
+    mock.expect :rewrite_object, rewrite_resp,
+      [bucket_name, file.name, bucket_name, file.name, patch_file_gapi]
+
+    file.service.mocked_service = mock
+
+    file.storage_class.must_equal "STANDARD"
+    file.storage_class = :dra
+    file.storage_class.must_equal "DURABLE_REDUCED_AVAILABILITY"
```
quartzmo: Again, DRA is probably not the best choice for example target.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -514,6 +514,15 @@ def file path, generation: nil, encryption_key: nil
         # @param [Hash] metadata A hash of custom, user-provided web-safe keys
         #   and arbitrary string values that will returned with requests for the
         #   file as "x-goog-meta-" response headers.
+        # @param [Symbol, String] storage_class Storage class of the file.
+        #   Determines how the file is stored and determines the SLA and the
+        #   cost of storage. Values include `:multi_regional`, `:regional`,
+        #   `:nearline`, `:coldline`, `:standard`, and `:dra` (Durable Reduced
```
quartzmo: The API does support changing to `STANDARD`, correct? I only ask because standard isn't a "real" storage class, so it might be worth testing...

> Cloud Storage offers four storage classes: Multi-Regional Storage, Regional Storage, Nearline Storage, and Coldline Storage.

google-cloud-storage/test/google/cloud/storage/file_update_test.rb
blowmage: The reason I chose it was because of the automatic mapping we do for `storage_class`. The values `:dra` and `"DURABLE_REDUCED_AVAILABILITY"` have the biggest difference, so I wanted the tests to exercise show that happens.

google-cloud-storage/lib/google/cloud/storage/file.rb
blowmage: I think this should be pretty apparent by the implementation of `storage_class=`.

google-cloud-storage/lib/google/cloud/storage/file.rb
blowmage: I'll update the File documentation to have both.

google-cloud-storage/test/google/cloud/storage/file_update_test.rb
quartzmo: The point is that no one should be changing to this storage class.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
blowmage: Yes, the API supports that. I am able to call both `bucket.create_file path, storage_class: :standard` and `file.storage_class = :standard` and it works.

quartzmo: 👍  cool, thanks for checking!

blowmage: AFAIK, `STANDARD` is the same as `REGIONAL` or `MULTI_REGIONAL`, depending on the bucket's location. Meaning, the behavior is different depending if it is in the US or Asia.

quartzmo: Yes, it's like a virtual class for conditionally setting one of those.

google-cloud-storage/test/google/cloud/storage/file_update_test.rb
blowmage: That's fine. But my point this is written for testing and not for best practices. I want to exercise more of the code.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -268,6 +268,35 @@ def encryption_key_sha256
...
+        # `:coldline`, `:standard`, and `:dra` (Durable Reduced Availability),
+        # as well as the strings returned by {File#storage_class} or
+        # {Bucket#storage_class}. For more information, see [Storage
+        # Classes](https://cloud.google.com/storage/docs/storage-classes) and
+        # [Per-Object Storage
+        # Class](https://cloud.google.com/storage/docs/per-object-storage-class).
+        # The  default value is the default storage class for the bucket. See
+        # {Bucket#storage_class}.
+        # @param [Symbol, String] storage_class Storage class of the file.
+        def storage_class= storage_class
```
quartzmo: Should we have an example?

blowmage: I didn't add one because the other setter methods don't have examples, or return types. I didn't want this method to stand out from the rest. But yes, they should have them.

blowmage:
@jskeet Do you have any input on this PR?

PR #1247, opened 2017-02-18
Fix Spanner acceptance test use of Map

blowmage:
The Protobuf Map object does not directly implement to_h. Use to_hash instead.
The to_h method comes from the Enumerable mixin, but in Ruby 2.0 it does not
have the to_h method.


quartzmo:
I'll merge once circle ci is green.

blowmage:
The PR's build won't run the acceptance tests, so I'll merge now.

PR #1249, opened 2017-02-20
Fix Generator#generate_docs

quartzmo:
Sort types in TOC alphabetically.

### DIFF
--- a/lib/gcloud/jsondoc/generator.rb
+++ b/lib/gcloud/jsondoc/generator.rb
@@ -87,7 +87,7 @@ module Gcloud
                 )
               end
             end
-            OpenStruct.new title: m[:title], types: matched_types.values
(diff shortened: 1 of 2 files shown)


lib/gcloud/jsondoc/generator.rb
```diff
@@ -87,7 +87,8 @@ def generate_docs
                 )
               end
             end
-            OpenStruct.new title: m[:title], types: matched_types.values
+            matched_types = types_hash.values.sort {|a,b| a.name <=> b.name }
+            OpenStruct.new title: m[:title], types: matched_types
```
blowmage: What about:

```ruby
OpenStruct.new title: m[:title], types: matched_types.values.sort_by(&:name)
```

quartzmo: good idea, will do

PR #1254, opened 2017-02-21
Change Document#url? to Document#gcs_url?

landrito:
Got some internal feedback from the language team to change this method name because url? is misleading.

### DIFF
--- a/google-cloud-language/lib/google/cloud/language/document.rb
+++ b/google-cloud-language/lib/google/cloud/language/document.rb
@@ -74,12 +74,6 @@ module Google
           @grpc.source == :gcs_content_uri
         end
 
-        def url?
-          puts "The method Document#url? is deprecated. " \
-          "Use Document#gcs_url? instead."
-          gcs_url?
-        end
-        
         ##
         # @private The source of the document's content.
         #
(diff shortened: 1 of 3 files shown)


google-cloud-language/lib/google/cloud/language/document.rb
```diff
@@ -70,10 +70,16 @@ def content?
         ##
         # @private Whether the document source is a Google Cloud Storage URI.
         #
-        def url?
+        def gcs_url?
           @grpc.source == :gcs_content_uri
         end
 
+        def url?
```
blowmage: The previous `url?` method is marked private, and not part of the public API. It does not show up in the documentation, and end users should not be aware of it. Why is this new `url?` method and warning added, which are _not_ longer marked private?

We should be free to make internal implementation changes without this step. Right?

quartzmo: +1, it's a "package-private" method *by convention* in Ruby

landrito: Ah Yes. I was being a bit overly protective, trying to not break users doing some bad OOP things (calling private instance methods through #send). I removed it.

Issue #1255, opened 2017-02-21
Natural Language: Remove encoding argument


landrito:
Based on internal feedback, we should default the encoding based on the fact that the user is using Ruby v2.0+.

landrito:
@danoscarmike Can you add this to the ML project.

swcloud:
@blowmage Can you make this priority since we want to fix this before March 3.

blowmage:
Yep.

blowmage:
AFAICT this already happens. Actions that do not specify an encoding generate requests to the API that include `:UTF8`.

I submitted #1267 to update the documentation to make this more clear. Can this issue be closed?

blowmage:
Closing as this is already implemented.

landrito:
The encoding type is actually in regards to the encoding type the user expects from the API. Since Ruby's default encoding is UTF-8 for versions 2.0+ we should not even expose this parameter.

Heres the bit of documentation in regards to this: https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/language_service.proto#L927

blowmage:
@landrito Should the Ruby client always use `UTF8`? Or should it check if `Encoding.default_internal` is `Encoding::UTF_16` or `Encoding::UTF_32` and use that encoding?

landrito:
Since the api supports utf16 and utf32 I think it would be best to use utf8 if `Encoding::UTF_8` is found, utf16 if `Encoding ::UTF_16` is found and, utf32 if `Encoding::UTF_32` is found. 

`Encoding.default_internal` can find other encodings besides utf8, utf16, or utf32. It's up in the air what we should do for that scenario. I'm leaning towards defaulting to utf8 if it is not utf16 or utf32. Thoughts?

blowmage:
I think it would be best to use UTF8 unless the system is set to UTF-16 or UTF-32. So if the system is set to ISO-8859-1 the API call will be made with UTF8. Does that work for you?

landrito:
Works for me.

PR #1267, opened 2017-02-24
Update encoding documentation

blowmage:
Add Natural Language documentation for the acceptable values and the default value of the encoding method parameter. [refs #1255]

### DIFF
--- a/google-cloud-language/lib/google/cloud/language/document.rb
+++ b/google-cloud-language/lib/google/cloud/language/document.rb
@@ -179,8 +179,9 @@ module Google
         # @param [Boolean] syntax Whether to perform syntactic analysis.
         #   Optional. The default is `false`. If every feature option is
         #   `false`, **all** features will be performed.
-        # @param [String] encoding The encoding type used by the API to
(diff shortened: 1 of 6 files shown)


google-cloud-language/lib/google/cloud/language/document.rb
```diff
@@ -180,7 +180,8 @@ def language= new_language
         #   Optional. The default is `false`. If every feature option is
         #   `false`, **all** features will be performed.
         # @param [String] encoding The encoding type used by the API to
```
quartzmo: The param type is `[String]`, but the acceptable values listed are symbols. Want to add `Symbol` to the types lists, or remove the colon from the acceptable values to make them strings? (I vote for the strings, since they are upper case).

blowmage: Yeah, I'll update the docs.

blowmage: The format and language arguments have the param type of `[String]`, but accept symbols as well. I'll update their docs as well.

PR #1269, opened 2017-02-24
Regenerate Datastore Gapic to use new HTTP headers.

landrito:
Updates: https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1268

cc @swcloud @bjwatson

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/service.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/service.rb
@@ -58,8 +58,8 @@ module Google
             channel: channel,
             timeout: timeout,
             client_config: client_config,
-            app_name: "gcloud-ruby",
-            app_version: Google::Cloud::Datastore::VERSION)
+            lib_name: "gccl",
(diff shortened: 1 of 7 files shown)


google-cloud-datastore/lib/google/cloud/datastore/version.rb
```diff
@@ -16,7 +16,7 @@
 module Google
   module Cloud
     module Datastore
-      VERSION = "0.24.0"
+      VERSION = "0.24.1"
```
blowmage: Please don't adjust version numbers. We do this as part of the release process.

landrito: Ah Gotcha. I'll do revert this the rest.

blowmage: Can you please remove this commit from the PR.

blowmage: Thanks!

Issue #1268, opened 2017-02-24
Update Gapic Clients to use new http headers.


landrito:
This is a tracking meta-issue for updating the gapic clients to use the new http headers.

Each PR should:

  * The `lib_name` and `lib_version` keyword arguments to the GAPIC client for the clients that are wrapped by a hand-written layer.

Completion List:
  * [x] Datastore (#1269)
  * [x] Error Reporting (#1270)
  * [x] Language (#1271)
  * [x] Logging (#1272)
  * [x] Monitoring (#1273)
  * [x] PubSub (#1274)
  * [x] Spanner (#1275)
  * [x] Speech (#1276)
  * [x] Trace (#1277)
  * [x] Vision (#1278)

landrito:
Hopefully doing it like this will reduce your cognitive load @blowmage!

blowmage:
Closing as all these pull requests are now accepted.

bjwatson:
@blowmage @landrito Can you let me know when all of these updates packages have been published?

blowmage:
Should be later this week or early next week.

bjwatson:
@blowmage Please make every effort to deploy this week, especially for Vision and Logging. Every day we wait, we're losing usage metrics.

blowmage:
@bjwatson I can release them all today, except for Natural Language. Is that okay?

bjwatson:
Sounds great! Thanks @blowmage.

blowmage:
@bjwatson All these changes are released, except for Natural Language, which has some breaking changes and will be released with the next batch of major releases.

bjwatson:
Thanks Mike!

PR #1273, opened 2017-02-24
Regenerate Monitoring Gapic to use new HTTP headers.

landrito:
Updates #1268

cc @bjwatson @swcloud


bjwatson:
LGTM, but @blowmage has the final sign-off on these since I'm not as familiar with the Ruby code.

PR #1275, opened 2017-02-24
Regenerate Spanner Gapic to use new HTTP headers.

landrito:
Updates #1268

cc @bjwatson @swcloud


google-cloud-spanner/lib/google/spanner/v1/spanner_pb.rb
```diff
@@ -10,7 +10,6 @@
 require 'google/protobuf/timestamp_pb'
 require 'google/spanner/v1/keys_pb'
 require 'google/spanner/v1/mutation_pb'
-require 'google/spanner/v1/query_plan_pb'
```
blowmage: Why was this require removed?

landrito: This was directly from the proto compiler. I'll do some digging when I get back from lunch.

blowmage: this file is also required in `result_set_pb`, which is the next file required. This is probably okay.

blowmage:
Ah, I forgot that spanner also creates an instance of OperationsClient, and like Speech this PR depends on google-gax 0.8.1.

PR #1276, opened 2017-02-24
Regenerate Speech Gapic to use new HTTP headers.

landrito:
Updates #1268

cc @bjwatson @swcloud

### DIFF
--- a/google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
@@ -26,6 +26,9 @@ require "json"
 require "pathname"
 
 require "google/gax"
+require "google/gax/operation"
+require "google/longrunning/operations_client"
+
 require "google/cloud/speech/v1beta1/cloud_speech_pb"
 
 module Google
@@ -36,7 +39,7 @@ module Google
         #
         # @!attribute [r] speech_stub
         #   @return [Google::Cloud::Speech::V1beta1::Speech::Stub]
-        class SpeechApi
+        class SpeechClient
           attr_reader :speech_stub
 
           # The default address of the service.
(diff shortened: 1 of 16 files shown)


google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
```diff
@@ -82,17 +79,40 @@ def initialize \
...
+              service_path: service_path,
+              port: port,
+              channel: channel,
+              chan_creds: chan_creds,
+              scopes: scopes,
+              client_config: client_config,
+              timeout: timeout,
+              app_name: app_name,
+              app_version: app_version,
+              lib_name: lib_name,
```
blowmage: It looks like this change probably requires google-gax 0.8.1, right? Because these method arguments aren't added until that point release.

google-cloud-speech/lib/google/cloud/speech/version.rb
```diff
@@ -16,7 +16,7 @@
 module Google
   module Cloud
     module Speech
-      VERSION = "0.22.0"
+      VERSION = "0.22.1"
```
blowmage: Can you remove this commit, please?

landrito: Ah missed those.

landrito: Fixed.

google-cloud-speech/lib/google/cloud/speech/v1beta1/speech_client.rb
landrito: Correct, this will require 0.8.1. IIUC That is covered by the pessimistic dependency (`"google-gax", "~> 0.8.0"`) correct?

blowmage: Yes, it is covered by the dependency, but we need to bump the minimum to 0.8.1, since 0.8.0 will error.

blowmage:
I confirm I am okay with their commits being contributed to this project.

PR #1290, opened 2017-02-25
Natural Language Acceptance Test Fix

blowmage:
It looks like Natural Language is rolling out an update, and these checks are failing.
The behavior we are starting to see is that this Sentiment is returned.
Sometimes the Sentiment is empty, sometimes it is not.
Disable the sentiment nil checks so the build will pass.


blowmage:
Not sure if it will be forever, but we do have unit tests that are ensuring we are sending the correct request to the API. Not sure what else we can do about changing API behavior.

PR #1292, opened 2017-02-26
Update to Google API Client 0.10.x

### LINKED ISSUE #1256: Update to google-api-client 0.10.x
The Google API Client library has cut its first 0.10.x release. We should stay on the technology treadmill and upgrade to the latest version.

blowmage:
Update Google API Client to the recently released 0.10.x version.
	
[closes #1256]

### DIFF
--- a/google-cloud-core/Gemfile
+++ b/google-cloud-core/Gemfile
@@ -3,10 +3,8 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake", "~> 11.0"
-gem "google-api-client", "~> 0.9.11"
-gem "grpc", "~> 1.0"
-gem "google-protobuf", "~> 3.0"
-gem "googleapis-common-protos", "~> 1.3"
+gem "google-api-client", "~> 0.10.0"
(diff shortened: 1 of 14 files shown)


google-cloud-core/Gemfile
```diff
@@ -3,10 +3,8 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake", "~> 11.0"
-gem "google-api-client", "~> 0.9.11"
-gem "grpc", "~> 1.0"
-gem "google-protobuf", "~> 3.0"
-gem "googleapis-common-protos", "~> 1.3"
+gem "google-api-client", "~> 0.10.0"
+gem "google-gax", "~> 0.8.0"
```
quartzmo: Just curious, why is this PR adding `google-gax`?

blowmage: Its replacing the `grpc` and `google-protobuf` and `googleapis-common-protos` dependencies, the same way that those direct dependencies have been removed from gemspec files. The `google-gax` gem declares all those dependencies, and we should use the versions that it specifies.

Issue #1238, opened 2017-02-16
Write more integration tests for BigQuery


omaray:
Please see email I sent about the details and test cases from @tswast . 

Also, C# has [some examples](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/ExhaustiveTypesTest.cs) that might be worth looking at.

/cc @jskeet @jba @tswast @danoscarmike

blowmage:
The process of adding additional acceptance tests has uncovered several incomplete or missing features. New issues have been created for them.

PR #1307, opened 2017-02-27
Update handling of BigQuery types

blowmage:
Fix several bugs with the handling of types in BigQuery. Make sure that the format of the value objects returned from BigQuery match the format used for the query parameters. Time, DateTime, Date, etc.

This is part of a larger set of changes discovered by adding additional acceptance tests (#1238).

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
@@ -73,6 +73,8 @@ module Google
                 StringIO.new Base64.decode64 value
               elsif type == "TIMESTAMP"
                 ::Time.at Float(value)
+              elsif type == "TIME"
+                Bigquery::Time.new value
               else
                 value
               end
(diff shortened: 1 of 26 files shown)


quartzmo:
I don't see any additions to acceptance testing in this PR, wouldn't it be good to expand that to cover these changes?

blowmage:
Yes, I'm working on new acceptance tests to fulfill #1238, but they don't come close to passing without these changes. There are more changes coming that will be needed to get the new acceptance tests to pass. It uncovered many problems with the existing implementation. The new acceptance tests will be added in a future PR. Right now, these changes pass all existing tests.

quartzmo:
Oh yeah, that's what prompted this work in the first place, isn't it? ;) SGTM.

blowmage:
These changes also bring BigQuery in line with how Spanner will handle returned types. Exciting changes ahead!

blowmage:
Found and resolved an issue with formatting TIMESTAMP values. This is ready for review.

PR #1308, opened 2017-02-28
Remove Natural Language encoding argument

### LINKED ISSUE #1255: Natural Language: Remove encoding argument

blowmage:
Switch to detecting the system encoding and using that instead.
Ruby's default encoding is UTF-8, and the default is API encoding is :UTF8.
Only use :UTF16 or :UTF32 if the system is explicitly set to those encodings.

[closes #1255]

### DIFF
--- a/google-cloud-language/lib/google/cloud/language.rb
+++ b/google-cloud-language/lib/google/cloud/language.rb
@@ -53,14 +53,11 @@ module Google
     #
     # ## Creating documents
     #
-    # Cloud Natural Language API supports UTF-8, UTF-16, and UTF-32 encodings.
-    # (Ruby uses UTF-8 natively, which is the default sent to the API, so unless
-    # you're working with text processed in different platform, you should not
-    # need to set the encoding type.) Be aware that only English, Spanish, and
-    # Japanese language content are supported.
-    #
     # Use {Language::Project#document} to create documents for the Cloud Natural
-    # Language service. You can provide text or HTML content as a string:
+    # Language service. (The Cloud Natural Language API currently supports
+    # English, Spanish, and Japanese.)
+    #
+    # You can provide text or HTML content as a string:
     #
     # ```ruby
     # require "google/cloud/language"
--- a/google-cloud-language/lib/google/cloud/language/annotation.rb
+++ b/google-cloud-language/lib/google/cloud/language/annotation.rb
@@ -240,9 +240,17 @@ module Google
         # Represents a piece of text including relative location.
         #
         # @attr_reader [String] text The content of the output text.
-        # @attr_reader [Integer] offset The API calculates the beginning offset
-        #   of the content in the original document according to the `encoding`
-        #   specified in the API request.
+        # @attr_reader [Integer] offset The beginning offset
+        #   of the content in the original document.
+        #
+        #   The API calculates the beginning offset according to the client
+        #   system's default encoding. In Ruby this defaults to UTF-8. To change
+        #   the offset calculation you will need to change Ruby's default
+        #   encoding. This is commonly done by setting
+        #   `Encoding.default_internal` to `Encoding::UTF_16` or
+        #   `Encoding::UTF_32`. If the system is configured to use an encoding
+        #   other than UTF-16 or UTF-32 the offset will be calculated using
+        #   UTF-8.
         #
         # @example
         #   require "google/cloud/language"
@@ -384,9 +392,17 @@ module Google
           alias_method :content, :text
 
           ##
-          # The API calculates the beginning offset of the content in the
-          # original document according to the `encoding` specified in the
-          # API request. See {TextSpan#offset}.
+          # The beginning offset of the content in the original document. See
+          # {TextSpan#offset}.
+          #
+          # The API calculates the beginning offset according to the client
+          # system's default encoding. In Ruby this defaults to UTF-8. To change
+          # the offset calculation you will need to change Ruby's default
+          # encoding. This is commonly done by setting
+          # `Encoding.default_internal` to `Encoding::UTF_16` or
+          # `Encoding::UTF_32`. If the system is configured to use an encoding
+          # other than UTF-16 or UTF-32 the offset will be calculated using
+          # UTF-8.
           #
           # @return [Integer]
           #
@@ -535,11 +551,31 @@ module Google
             @lemma            = lemma
           end
 
+          ##
+          # The content of the output text. See {TextSpan#text}.
+          #
+          # @return [String]
+          #
           def text
             @text_span.text
           end
           alias_method :content, :text
 
+          ##
+          # The beginning offset of the content in the original document. See
+          # {TextSpan#offset}.
+          #
+          # The API calculates the beginning offset according to the client
+          # system's default encoding. In Ruby this defaults to UTF-8. To change
+          # the offset calculation you will need to change Ruby's default
+          # encoding. This is commonly done by setting
+          # `Encoding.default_internal` to `Encoding::UTF_16` or
+          # `Encoding::UTF_32`. If the system is configured to use an encoding
+          # other than UTF-16 or UTF-32 the offset will be calculated using
+          # UTF-8.
+          #
+          # @return [Integer]
+          #
           def offset
             @text_span.offset
           end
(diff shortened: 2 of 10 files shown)


google-cloud-language/lib/google/cloud/language.rb
```diff
@@ -54,10 +54,10 @@ module Cloud
     # ## Creating documents
     #
     # Cloud Natural Language API supports UTF-8, UTF-16, and UTF-32 encodings.
-    # (Ruby uses UTF-8 natively, which is the default sent to the API, so unless
-    # you're working with text processed in different platform, you should not
-    # need to set the encoding type.) Be aware that only English, Spanish, and
-    # Japanese language content are supported.
+    # (Ruby uses UTF-8 natively, which is the default sent to the API. To change
```
landrito: s/`which is the default sent to the API`/`which is the default received from the API`

google-cloud-language/lib/google/cloud/language.rb
```diff
@@ -54,10 +54,10 @@ module Cloud
     # ## Creating documents
     #
     # Cloud Natural Language API supports UTF-8, UTF-16, and UTF-32 encodings.
```
landrito: This line is ambiguous whether it means requests support these encodings or responses. To be clear, request can be anything since the proto converter will convert it to utf8 before sending. The responses can be any utf8 utf16 or utf32 based on default internal encoding.

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      language.service.mocked_service = mock
+      annotation = language.annotate text_content
+      mock.verify
+
+      annotation.must_be_kind_of Google::Cloud::Language::Annotation
+    end
+  end
+
+  it "uses UTF16 when the internal encoding is UTF-8" do
```
landrito: UTF8

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      language.service.mocked_service = mock
+      annotation = language.annotate text_content
+      mock.verify
+
+      annotation.must_be_kind_of Google::Cloud::Language::Annotation
+    end
+  end
+
+  it "uses UTF16 when the internal encoding is UTF-32" do
```
landrito: UTF32

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
```diff
@@ -0,0 +1,149 @@
...
+    mock.expect :annotate_text, grpc_resp, [grpc_doc, features, :UTF8, options: default_options]
+
+    language.service.mocked_service = mock
+    annotation = language.annotate text_content
+    mock.verify
+
+    annotation.must_be_kind_of Google::Cloud::Language::Annotation
+  end
+
+  it "uses UTF16 when the internal encoding is nil" do
```
landrito: UTF8

google-cloud-language/lib/google/cloud/language/annotation.rb
```diff
@@ -241,8 +241,8 @@ def self.from_grpc grpc
         #
         # @attr_reader [String] text The content of the output text.
         # @attr_reader [Integer] offset The API calculates the beginning offset
-        #   of the content in the original document according to the `encoding`
-        #   specified in the API request.
+        #   of the content in the original document according to the default
+        #   encoding of the system. Ruby defaults to UTF-8.
```
quartzmo: Do you want to add:

> To change this you will need to set `Encoding.default_internal` to `Encoding ::UTF_16` or `Encoding ::UTF_32`.

Otherwise, you have to hope that someone interested in the offset is going to stumble upon this in the guide or Document docs.

google-cloud-language/lib/google/cloud/language/service.rb
```diff
@@ -134,6 +118,23 @@ def default_options
           Google::Gax::CallOptions.new kwargs: default_headers
         end
 
+        def default_encoding
+          # memoize the result so the check is only made once
```
landrito: IIUC (although rare and probably terrible), Users can reassign `Encoding.default_internal`. Memoizing results will cause encodings to fail if on the (very) low chance of someone changing the default encoding directly. That being said, do you think the performance improvements warrant memoizing the results?

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
```diff
@@ -0,0 +1,149 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "helper"
+
+describe Google::Cloud::Language, :default_encoding, :mock_language do
+  it "detects the default encoding" do
```
landrito: If the user has set the default encoding to UTF16 or UTF32 in their environment, this test will fail correct?

blowmage: Yes, but nobody ft does that. :)

google-cloud-language/lib/google/cloud/language.rb
blowmage: @quartzmo Can you think of a better description for this?

google-cloud-language/lib/google/cloud/language/annotation.rb
blowmage: We have that language in the guide documentation. I felt like it was distracting on the class documentation. My preference would be to remove this entire paragraph on the class documentation than pull it even more off-topic.

google-cloud-language/lib/google/cloud/language.rb
quartzmo: Change:

```
    # Cloud Natural Language API supports UTF-8, UTF-16, and UTF-32 encodings.
    # (Ruby uses UTF-8 natively, which is the default sent to the API, so unless
    # you're working with text processed in different platform, you should not
    # need to set the encoding type.) Be aware that only English, Spanish, and
    # Japanese language content are supported.
    #
    # Use {Language::Project#document} to create documents for the Cloud Natural
    # Language service. You can provide text or HTML content as a string:
```
to
```
    # Use {Language::Project#document} to create documents for the Cloud Natural
    # Language service. (The Cloud Natural Language API currently supports 
    # English, Spanish, and Japanese.)
    #
    # You can provide text or HTML content as a string:
```

google-cloud-language/lib/google/cloud/language/annotation.rb
blowmage: Oops, this is the `TextSpan#offset` documentation. I thought it was the `Document` documentation. I don't have a problem with how this is written on this attribute.

quartzmo: I think all `offset` attrs/methods need a full discussion of encoding. (Here's my logic: the encoding only affects one thing, the offset. So you will only care about changing encoding when you have issues with offset. And when you have issues with offset, you will look at the method docs for offset.)

google-cloud-language/lib/google/cloud/language.rb
blowmage: Changed.

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
blowmage: Fixed.

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
blowmage: Fixed.

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
blowmage: Fixed.

google-cloud-language/lib/google/cloud/language/service.rb
blowmage: Updated. Memoization is removed.

google-cloud-language/lib/google/cloud/language.rb
blowmage: This has been changed.

@landrito, @quartzmo and I paired on this and revised all the documentation. We removed mention of encoding from everywhere except for the `offset` values. Can you take a look and tell us if this new verbiage works for you.

google-cloud-language/test/google/cloud/language/default_encoding_test.rb
blowmage: The CI servers are configured to use the Ruby defaults, so I am not worried about this failing. I left this test un-stubbed because if it ever does fail I would like to know.

blowmage:
@landrito, @quartzmo and I paired on this and revised all mentions of encoding in documentation. We removed mentions of encoding from everywhere except the documentation of the offset values, since those values are the only ones that will be affected by the system configuration. Can you take a look and let us know what you think?

PR #1312, opened 2017-03-02
Update Vision to V1.1

blowmage:


blowmage:
Both @quartzmo and @swcloud have contributed these changes, but are unfortunately out for the next several days. So I'm submitting this PR for them.

blowmage:
> What is causing the circle CI to be yellow? It looks to pass to me.

Not sure. Maybe an upstream dependency on S3? Too soon? :trollface:

PR #1313, opened 2017-03-02
Switch BigQuery to Standard SQL by default

blowmage:
Standard SQL has many more features than Legacy SQL. It makes sense for Standard SQL to be the default.

This change allows nested values to be returned by default, and legacy support is switched to an option flag.

```ruby

db = bigquery.dataset "mine"

# standard sql is default and returns properly formatted values
# no standard_sql argument needed!
db.query "SELECT * FROM stuff"
rows == [{"name"=>"mike", "foo"=>[{"bar"=>"hey", "baz"=>{"bif"=>1}}, {"bar"=>"world", "baz"=>{"bif"=>2}}]}] #=> true

# legacy sql can be supported and will flattened nested and repeated records
db.query "SELECT * FROM stuff", legacy_sql: true
rows == [{"name"=>"mike", "foo_bar"=>"hey", "foo_baz_bif"=>1}, {"name"=>"mike", "foo_bar"=>"world", "foo_baz_bif"=>2}] #=> true
```

This is part of a larger set of changes for BigQuery.

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb
@@ -15,7 +15,7 @@
 require "bigquery_helper"
 
 describe Google::Cloud::Bigquery, :bigquery do
-  let(:publicdata_query) { "SELECT url FROM [publicdata:samples.github_nested] LIMIT 100" }
+  let(:publicdata_query) { "SELECT url FROM publicdata.samples.github_nested LIMIT 100" }
   let(:dataset_id) { "#{prefix}_dataset" }
   let(:dataset) do
     d = bigquery.dataset dataset_id
--- a/google-cloud-bigquery/lib/google/cloud/bigquery.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery.rb
@@ -95,33 +95,12 @@ module Google
     # explained in [Querying
     # Data](https://cloud.google.com/bigquery/querying-data).
     #
-    # ### Legacy SQL (formerly BigQuery SQL)
-    #
-    # Before version 2.0, BigQuery executed queries using a non-standard SQL
-    # dialect known as BigQuery SQL. This variant is still the default, and will
-    # be used unless you pass the flag `standard_sql: true` with your query.
-    # (If you get an SQL syntax error with a query that may be written in
-    # standard SQL, be sure that you are passing this option.)
-    #
-    # ```ruby
-    # require "google/cloud/bigquery"
-    #
-    # bigquery = Google::Cloud::Bigquery.new
-    #
-    # sql = "SELECT TOP(word, 50) as word, COUNT(*) as count " \
-    #       "FROM [publicdata:samples.shakespeare]"
-    # data = bigquery.query sql
-    # ```
-    #
-    # Notice that in legacy SQL, a fully-qualified table name uses the following
-    # format: `[my-dashed-project:dataset1.tableName]`.
-    #
     # ### Standard SQL
     #
     # Standard SQL is the preferred SQL dialect for querying data stored in
     # BigQuery. It is compliant with the SQL 2011 standard, and has extensions
-    # that support querying nested and repeated data. It has several advantages
-    # over legacy SQL, including:
+    # that support querying nested and repeated data. This is the default
+    # syntax. It has several advantages over Legacy SQL, including:
     #
     # * Composability using `WITH` clauses and SQL functions
     # * Subqueries in the `SELECT` list and `WHERE` clause
@@ -136,8 +115,7 @@ module Google
     # For examples that demonstrate some of these features, see [Standard SQL
     # highlights](https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql#standard_sql_highlights).
     #
-    # Legacy SQL is still the default. To use standard SQL instead, pass the
-    # option `standard_sql: true` with your query.
+    # Standard SQL is the default.
     #
     # ```ruby
     # require "google/cloud/bigquery"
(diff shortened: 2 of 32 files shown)


google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb
```diff
@@ -15,7 +15,7 @@
 require "bigquery_helper"
 
 describe Google::Cloud::Bigquery, :bigquery do
-  let(:publicdata_query) { "SELECT url FROM [publicdata:samples.github_nested] LIMIT 100" }
+  let(:publicdata_query) { "SELECT url FROM publicdata.samples.github_nested LIMIT 100" }
```
landrito: Missing ticks.

google-cloud-bigquery/lib/google/cloud/bigquery.rb
```diff
@@ -147,12 +125,35 @@ module Cloud
...
+    # data = bigquery.query sql
+    # ```
+    #
+    # Notice that in standard SQL, a fully-qualified table name uses the
+    # following format: <code>`my-dashed-project.dataset1.tableName`</code>.
+    #
+    # ### Legacy SQL (formerly BigQuery SQL)
+    #
+    # Before version 2.0, BigQuery executed queries using a non-standard SQL
+    # dialect known as BigQuery SQL. This variant is still the default, and will
```
landrito: This variant is no longer the default.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -416,28 +424,30 @@ def create_table table_id, name: nil, description: nil, fields: nil
...
         #   dataset = bigquery.dataset "my_dataset"
         #   view = dataset.create_view "my_view",
-        #             "SELECT name, age FROM [proj:dataset.users]",
+        #             "SELECT name, age FROM proj.dataset.users",
         #             name: "My View", description: "This is my view"
         #
         # @!group Table
         #
-        def create_view table_id, query, name: nil, description: nil
+        def create_view table_id, query, name: nil, description: nil,
```
landrito: We can make this more error proof by making sure that standard_sql and legacy_sql are mutually exclusive parameters.

Calling this function with both fields set is incorrect and we should protect against that.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -759,7 +760,16 @@ def query_job query, params: nil, priority: "INTERACTIVE", cache: true,
...
         #   is true and `flatten` is false. Optional. The default value is
-        #   false.
+        #   true.
+        # @param [Boolean] legacy_sql Specifies whether to use BigQuery's
+        #   [legacy
+        #   SQL](https://cloud.google.com/bigquery/docs/reference/legacy-sql)
+        #   dialect for this query. If set to false, the query will use
+        #   BigQuery's [standard
+        #   SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/)
+        #   When set to false, the values of `large_results` and `flatten` are
```
landrito: Can we warn users when these values are going to be ignored?

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -252,8 +252,8 @@ def self.default_project
         #
         def query_job query, params: nil, priority: "INTERACTIVE", cache: true,
                       table: nil, create: nil, write: nil, large_results: nil,
-                      flatten: nil, dataset: nil, legacy_sql: nil,
-                      standard_sql: nil
+                      flatten: nil, dataset: nil, standard_sql: nil,
```
landrito: Is there a specific reason to rearrange these parameters?

google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
```diff
@@ -473,6 +473,7 @@ def query_config query, options = {}
         def resolve_legacy_sql legacy_sql, standard_sql
```
landrito: Is there a better place for this function? It is repeated across multiple files.

google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb
blowmage: Yes, but since the project doesn't have any `-` characters we don't need to ticks. Its good to have coverage both ways. :)

google-cloud-bigquery/lib/google/cloud/bigquery.rb
blowmage: Oops. Fixed

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
blowmage: I don't know if that is needed. The nature of the BigQuery API is that some options override others. We felt having a positive assertion (`standard_sql: true`) was better than a negative assertion (`legacy_sql: false`)

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
blowmage: That text is taken from the [API documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.query.useLegacySql). Every time we get fancy with stuff like that we live to regret it. I'd prefer to leave the warnings to the documentation.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
blowmage: To give `standard_sql` prominence over `legacy_sql`.

google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
blowmage: I agree it borders on too much duplication, but this is a private method and can be refactored in the future without affecting the public API. I'd rather feel the pain of the duplication before DRYing it, and right now its not causing me any pain.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
landrito: Gotcha. Can we at the make resolve_legacy_sql place standard_sql at a high precedence than legacy_sql since standard_sql has prominence over legacy_sql.

blowmage: Great catch! I meant to do that before but I had too many balls in the air. Fixed.

google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
blowmage: Well I finally felt pain when fixing the `resolve_legacy_sql` precedence. These methods have been moved to the Convert module.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
```diff
@@ -239,6 +239,12 @@ def self.to_json_value value
           end
         end
 
+        def self.resolve_legacy_sql legacy_sql, standard_sql
```
landrito: `standard_sql` should be first for consistency.

google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb
landrito: nit: Might be good to add a comment noting this.

PR #1316, opened 2017-03-03
BigQuery Schema Changes

blowmage:
This PR refactors the internals of Schema to make maintenance easier. Since we are in the midst of other breaking changes, some changes have been made here. The biggest ones are:

* Nested fields no longer yield a Schema object, but the Field object
* Field has been updated to act like Schema, and can add nested fields
* Removal of `Schema#fields=`, users should use the API to modify fields
* Removal of `fields` argument on `Dataset#create_table`, users should use the yielded `table` object to define the schema
* Remove `Data#raw` because we now have much higher confidence in the format of the data
* Additional convenience methods
* Updated documentation

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -566,31 +566,28 @@ module Google
         #   * `append` - BigQuery appends the data to the table.
         #   * `empty` - A 'duplicate' error is returned in the job result if the
         #     table exists and contains data.
-        # @param [Boolean] large_results If `true`, allows the query to produce
-        #   arbitrarily large result tables at a slight cost in performance.
-        #   Requires `table` parameter to be set.
-        # @param [Boolean] flatten Flattens all nested and repeated fields in
-        #   the query results. The default value is `true`. `large_results`
-        #   parameter must be `true` if this is set to `false`.
         # @param [Boolean] standard_sql Specifies whether to use BigQuery's
         #   [standard
         #   SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/)
         #   dialect for this query. If set to true, the query will use standard
         #   SQL rather than the [legacy
         #   SQL](https://cloud.google.com/bigquery/docs/reference/legacy-sql)
-        #   dialect. When set to true, the values of `large_results` and
-        #   `flatten` are ignored; the query will be run as if `large_results`
-        #   is true and `flatten` is false. Optional. The default value is
-        #   true.
+        #   dialect. Optional. The default value is true.
         # @param [Boolean] legacy_sql Specifies whether to use BigQuery's
         #   [legacy
         #   SQL](https://cloud.google.com/bigquery/docs/reference/legacy-sql)
         #   dialect for this query. If set to false, the query will use
         #   BigQuery's [standard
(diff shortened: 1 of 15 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/schema.rb
```diff
@@ -52,48 +54,27 @@ def fields
...
         end
 
+        ##
+        # The names of the fields as symbols.
         def headers
           fields.map(&:name).map(&:to_sym)
         end
 
+        ##
+        # Retreive a fields by name.
```
landrito: s/fields/field

google-cloud-bigquery/lib/google/cloud/bigquery/schema.rb
```diff
@@ -272,6 +253,33 @@ def record name, description: nil, mode: nil
           nested_field
         end
 
+        # @private
+        def changed?
+          return false if frozen?
```
landrito: Can't you change it before you freeze it?

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -379,7 +356,7 @@ def delete force: nil
         #
         # @!group Table
         #
-        def create_table table_id, name: nil, description: nil, fields: nil
+        def create_table table_id, name: nil, description: nil
```
landrito: Why remove fields here?

blowmage: The approach of accepting Field objects pre-dates the yielding of the table updater object in this method. The yielded table object allows a user to configure the schema using schema API, which is a much easier way to define the schema and matches the API used by Rails migrations.

Accepting fields on a Schema object also adds lots of complexity to a frozen/immutable logic. Since we removed it from Schema I felt it was appropriate to remove it from this call as well.

google-cloud-bigquery/lib/google/cloud/bigquery/schema.rb
```diff
@@ -276,141 +244,86 @@ def date name, description: nil, mode: :nullable
         #   end
         #
         def record name, description: nil, mode: nil
+          # TODO: do we need to fail if no block was given?
```
landrito: Based on the definition set here: https://cloud.google.com/bigquery/data-types we might want to even ensure there is at least on field in the record.

blowmage: Right, the BigQuery API will enforce that a record field has nested fields and return an appropriate error if they do not. But passing a block is no guarantee that a field will be added.

```ruby

table = dataset.create_table "things" do |schema|
  schema.record "properties" do |properties|
    # not doing anything here but passing a block
    # no nested records being added
  end
end
```

I don't think its needed anymore, but this is something I'd like to discuss with @quartzmo when he gets back. If we did remove the error, users _could_ decide to use code like this:

```ruby

table = dataset.create_table "things" do |schema|
  properties = schema.record "properties"
  properties.string "name"
  properties.integer "size"
end
```

landrito: FWIW: If we are going to leave the block check then we might as well ensure the record has a nested field. I am fine with leaving as is until we discuss with @quartzmo.

PR #1318, opened 2017-03-07
Correct encoding of BYTES parameters

yugui:
This change allows users to send BYTES fields in parameter lists,
which are necessary in `google.spanner.v1.Write.values` or
`google.spanner.v1.KeySet.keys` for example.

Such fields used to be rejected, saying

> TypeError: Invalid type Array to assign to submessage field.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/convert.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/convert.rb
@@ -133,8 +133,7 @@ module Google
               obj.rewind
               content = obj.read.force_encoding("ASCII-8BIT")
               encoded_content = Base64.strict_encode64(content)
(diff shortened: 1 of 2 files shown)


blowmage:
I'll look at this. We should finally be able to get work into Spanner this week!

yugui:
ping?

blowmage:
I was hoping to pull the sin as part of a larger change that is being worked on, but I'll accept it now to unblock you. Just know that the ruby API for spanner is going to change significantly in the future.

yugui:
@blowmage Thank you.

> Just know that the ruby API for spanner is going to change significantly in the future.

Now I'm trying to add transaction support in another PR. Then, should I wait for your larger API change?

PR #1335, opened 2017-03-15
Upload to Storage with IO

blowmage:
Allow IO-ish objects to be uploaded to Storage to create new files. This looks to have been allowed since the Google Cloud API 0.10.0 release.


blowmage:
/hattip to @ernie for letting us know this was working with Google API Client. 😄

ernie:
@blowmage

 I'm useful!

PR #1341, opened 2017-03-18
Language: Round the sentiment values to nearest tenth.

landrito:
DO NOT MERGE.

The language service should only be returning a floating value that is rounded to the nearest tenth. Somewhere along the pipeline, precision is lost making the values slightly off. I'm going to be doing some investigating to see where the precision is lost. This PR is a workaround in the case that I don't find a solution.


blowmage:
The extra precision is added by google-protobuf. We give the proto object `0.1`, and it returns `0.10000000149011612`

quartzmo:
@landrito This has been a nagging annoyance. Much thanks.

Increased precision has also been an issue in other packages. In this commit for Vision on [Feb 14](https://github.com/GoogleCloudPlatform/google-cloud-ruby/commit/5a8491744b962c65fca21691dd84d5d231f15205#diff-859576205537e1c9900362638447eb22), you can see that I updated the example output in order to get passing documentation tests (unit tests):

```sh
-        #   landmark.score #=> 0.91912264
+        #   landmark.score #=> 0.9191226363182068
```

blowmage:
@landrito The more I look into this, the more I am convinced that google-protobuf is doing something wrong. Ruby floats are IEEE 754 double-precision 64-bit floats, so I don't see why the values would be changing. For example, here is the code for packing `0.1` into little-endian bytes:

```ruby
[0.1].pack("E") #=> "\x9A\x99\x99\x99\x99\x99\xB9?"
```

Here are those same little-endian bytes being unpacked into a float:

```ruby
"\x9A\x99\x99\x99\x99\x99\xB9?".unpack("E").first #=> 0.1
```

But, when setting the same value on a protobuf object it is changed:

```ruby
proto = Google::Cloud::Language::V1::Sentiment.new
proto.score = 0.1
proto.score #=> 0.10000000149011612
```

Originally I figured google-protobuf needed to change the values because of some requirement with the [Protobuf spec](https://developers.google.com/protocol-buffers/docs/encoding#non-varint-numbers). But AFAICT we should be able to store the ruby floats without changing the value.

quartzmo:
What @blowmage has shown above is exactly what we have seen with unit testing. Although we have tried to use values with less precision in fixtures, the actual values end up with more precision.

Thanks for digging in, @blowmage.

blowmage:
I looked into this further, and it turns out the [Language proto spec](https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/language_service.proto#L194) for Sentiment uses `float`, which means it would be a 32 bit float, not a 64 bit float.

Packing and unpacking a 32 bit floats uses "e", not "E". When we pack the float to 32 bits we her the following little-endian bytes:

```ruby
[0.1].pack("e") #=> "\xCD\xCC\xCC="
```

When we unpack those little-endian bytes back to a float we get the following:

```ruby
"\xCD\xCC\xCC=".unpack("e").first #=> 0.10000000149011612
```

So this conversion appears to be the source of the problem.

blowmage:
A workaround for the 32-bit to 64-bit float conversion could be to round the value to the nearest 8 positions:

```ruby
0.10000000149011612.round 8 #=> 0.1
```

I think this should be done in the google-protobuf library, not in google-cloud-ruby. Thoughts?

blowmage:
I think the [Protobuf ruby test coverage](optional_float) is incomplete. They are testing `0.5`, which packs and unpacks just fine:

```ruby
[0.5].pack("e").unpack("e").first #=> 0.5
```

It would be much better if they tested something like `0.1`:

```ruby
[0.1].pack("e").unpack("e").first #=> 0.10000000149011612
```

blowmage:
@quartzmo wrote:

> In this commit for Vision on Feb 14, you can see that I updated the example output in order to get passing documentation tests (unit tests):

That change was made because of the [google-protobuf 3.2.0 release](https://rubygems.org/gems/google-protobuf/versions/3.2.0). So something in that release changed how floats are dealt with. It looks like 32-bit floats were rounded to 8 positions before that release. @landrito, can you confirm that?

blowmage:
Rounding to 7 i likely better than rounding to 8:

```ruby
[0.4].pack("e").unpack("e").first.round 8 #=> 0.40000001
[0.4].pack("e").unpack("e").first.round 7 #=> 0.4

[0.9].pack("e").unpack("e").first.round 8 #=> 0.89999998
[0.9].pack("e").unpack("e").first.round 7 #=> 0.9
```

landrito:
Thanks a lot for your work looking into this! Going forward I think I shall do the following.

1. (landrito) File a bug with `google-protobuf` outlining this bug and holes in their testing.
2. (landrito) Change this PR to round to 7 as a work around until a fix is made in `google-protobuf`

Does this sound good?

blowmage:
1. Sounds good
2. If we are going to round all protobuf `float` values, we should make this change everywhere, not just in Natural Language. IMO. :wink:

landrito:
Hm. I agree that we should make the change everywhere. Is there a way to make the change in one place in `google-cloud-ruby` rather than adding it to every reader of floating point grpcs? Worst case, `google-protobuf` takes many months to fix this issue, remembering to add `.round` to every new instance of a float field wont scale well since it will introduce a point of failure.

blowmage:
AFAICT the conversion happens in google-protobuf's C code, so there is no mechanism we have to hook into a method and alter the return value. This looks like something we just add in our Ruby code along with a comment that references the issue.

landrito:
I'm still working out what do do with `Sentiment` values specifically since the api is sending (the nearest approximated 32bit float of) one decimal values. After some thought I think we shouldn't be doing a global `.round` on float32 protobuf fields. I think this will introduce precision loss rather than preserve it as well as unnecessary cognitive overhead when introducing new apis that use float32s.

In regards to the tests that fail would we be able hook make a function `float_32_rep(val)` which would return `[val].pack('e').unpack('e').first` and use that for assertions involving float32 fields (although I doubt that this will work for documentation tests).

quartzmo:
@landrito The documentation/examples/snippets tests also run on minitest, like the unit tests, only they are loaded via yard-doctest.

landrito:
Ah so then I suppose introducing `float_32_rep(val)` will be an adequate solution. Thoughts?

blowmage:
For the tests, I think `assert_in_delta` and `must_be_close_to` are sufficient. I don't think we need a `float_32_rep`.

landrito:
Closing. Leaving as is is fine.

PR #1350, opened 2017-03-21
GAE Rails integration tests compatible with gcloud SDK v147

hxiong388:
gcloud SDK starts to prepand `exec` in docker image entrypoint in version 147.0.0. This change has broken the GAE Rails integration test workflows. We can wrap the Rails test applications' entrypoints in `bash -c` to work around this change.

To test this change, upgrade gcloud SDK to version 147.0.0 or later, then run `rake integration:gae` and verify this task complete successfully.

I also noticed GAE Flex has fixed the load balancer unavailable right after deployment bug. So we can safely remove that `sleep` statement.

### DIFF
--- a/integration/deploy.rb
+++ b/integration/deploy.rb
@@ -31,12 +31,6 @@ def deploy_gae_flex app_dir
     last_gae_version = get_gae_versions.last
     sh "gcloud app deploy -q" do |ok, res|
       if ok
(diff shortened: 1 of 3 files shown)


quartzmo:
Thanks @hxiong388!!!! LGTM, but I will test tomorrow before merging.

PR #1364, opened 2017-03-27
Add integration test for file name normalization

swcloud:
This test ensures the client does not perform file name normalization. But it looks like the current ruby client (google-api-client?) does that. So the test fails.

### DIFF
--- a/google-cloud-storage/acceptance/storage/filename_test.rb
+++ b/google-cloud-storage/acceptance/storage/filename_test.rb
@@ -1,0 +1,40 @@
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "storage_helper"
+
+describe Google::Cloud::Storage::File, :storage do
+  let(:bucket_name) { "storage-library-test-bucket" }
+  let :bucket do
+    storage.bucket(bucket_name)
+  end
+
+  # Normalization Form C: a single character for e-acute U+00e9.
+  # URL should end with Cafe%CC%81
+  # Normalization Form D: an ASCII e followed by U+0301 combining character
+  # URL should end with Caf%C3%A9
+  let(:filenames) { ["Caf\u00e9", "Cafe\u0301"] }
+  let(:filecontent) { ["Normalization Form C", "Normalization Form D"] }
+  
+  it "does not perform any file name normalization" do
+    filenames.each_with_index do |name, i|
+      file = bucket.file name
+      file.name.must_equal name
+      Tempfile.open ["cafe"] do |tmpfile|
+        downloaded = file.download tmpfile
+        File.read(downloaded.path).must_equal filecontent[i]
+      end
+    end
+  end
+end


blowmage:
@swcloud This PR looks to be the opposite of #1097, which attempts to show that the upload should match regardless of unicode normalization. @sqrrrl says that an internal GCS bug has been created for the file name normalization. Is that bug still open? What can you tell us about this?

blowmage:
@swcloud I doubt that `google-api-client` will remove its dependency on the `addressable` gem, which implements the [RFC 6570 URI Template](https://tools.ietf.org/html/rfc6570) spec faithfully. So the only way to accept this PR is to remove the dependency on `google-api-client`, similar to how we have done that for `google-cloud-translate`. Is that what you are asking with this PR?

blowmage:
For more about the gems and the spec, see [this comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1097#issuecomment-264565019).

swcloud:
The internal bug is still open. The service side fix is tricky. Currently
we want to make sure the client does not do normalization.

swcloud:
My understanding is that different API had their different way to define
resource name semantics either byte semantics or text semantics, GCS is
byte semantics, so there should not be any normalization.

On Mon, Mar 27, 2017, 9:20 AM Song Wang <songwang@google.com> wrote:


The internal bug is still open. The service side fix is tricky. Currently
we want to make sure the client does not do normalization.

On Mon, Mar 27, 2017, 8:55 AM Mike Moore <notifications@github.com> wrote:

@swcloud <https://github.com/swcloud> This PR looks to be the opposite of
#1097 <https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1097>,
which attempts to show that the upload should match regardless of unicode
normalization. @sqrrrl <https://github.com/sqrrrl> says that an internal
GCS bug has been created for the file name normalization. Is that bug still
open? What can you tell us about this?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1364#issuecomment-289497699>,
or mute the thread

blowmage:
@swcloud Would you prefer we push a fix to google-api-client or would you prefer we remove google-api-client?

swcloud:
Sounds like you already have code to replace google-api-client
functionality? I would prefer to remove google-api-client to be consistent
with Translation.

blowmage:
> Sounds like you already have code to replace google-api-client functionality?

Not really.

> I would prefer to remove google-api-client to be consistent with Translation.

Translation is a very simple library using a single API call with HTTP GET, so it was much easier to replace. Storage has dozens of API calls, both downloading and uploading files. This will be a significant effort to implement and verify this change.

swcloud:
I see. What about another option? Maybe you can add a configable parameter
to enable/disable normalization?

blowmage:
> I see. What about another option? Maybe you can add a configable parameter to enable/disable normalization?

I've been looking into that as well. By reading the closed issues for the `addressable` gem, I doubt that any configuration to change the normalization behavior will be accepted. I see the following options:

1. Fork the `addressable` gem, add configuration to control normalization behavior, make `google-api-client` switch to use this newly forked gem.
2. Fork the relevant parts of the google-api-client gem, replace the dependency on `addressable`, switch `google-cloud-storage` to use this newly forked gem.
3. Move `google-cloud-storage` off of `google-api-client` in the manner of `google-cloud-translate`.

The _least_ complicated option seems to be option 3. If byte semantics for GCS API calls is a hard requirement, and we cannot wait for GCS to support text semantics, then I think option 3 is the way to go. Since all changes would be under our control this would also be the quickest to complete.

But that doesn't mean this is an easy or quick change.

swcloud:
Agreed with option 3. Do you have ETA?

blowmage:
> Do you have ETA?

Maybe the end of the week? Depending on other priorities?

sqrrrl:
FWIW, I'm working on submitting a PR for addressable to make unicode normalization optional. Should have that done shortly.

blowmage:
@sqrrrl Wonderful news!

@swcloud Does this change our plans? We unfortunately don't have control over if the change is accepted, and when it would be released.

sporkmonger:
I am fully onboard w/ unicode normalization being optional for URI templates. The default should be whatever the RFC says, however, on principle, I don't like library code that thinks it knows better than the developer just because the "RFC said so".

That said, while the RFC didn't say you **MUST** normalize, it does say you **SHOULD** normalize. Anybody that's trying to push raw bytes in a URI should probably seriously consider url-safe base64 encoding as a first pass for representing raw byte data that's not subject to the vagaries of unicode, NFKC, and percent-encoding. That goes double for URI templates, where the data is guaranteed to be messed with by any conformant processor. Even after Addressable makes it possible to disable normalization, it seems to me that you're still firmly in "this is unwise" territory here. I would recommend deprecating raw bytes as values expanded within URI templates and suggest phasing them out entirely. It's simply a bad idea and you shouldn't be doing it.

sporkmonger:
As for release schedule, adding an extra parameter shouldn't be a breaking change, so it can be a patch release. All the commits between now and the last release have been changes to the tests, so there's no obstacle to releasing immediately after merge.

sporkmonger:
2.5.1 released. You should be all set to update dependency and use the new option. My commentary above still applies but at least you have better options now.

sqrrrl:
Thanks Bob. FWIW, agree with the general sentiment -- not great that resource names are sensitive here.

I'll update google-api-client in the next day or two. Version will be bumped to 0.11.0

blowmage:
@sqrrrl How long do you think it will take to see a release that makes use of [addressable 2.5.1](https://rubygems.org/gems/addressable/versions/2.5.1)?

sqrrrl:
Aiming for end of week, earlier if I can. There are some other changes that need to be dusted off and merged in, but shouldn't take too long to do.

blowmage:
Man, I missed your earlier comment again. That all sounds great. Please let me know how I can help. My email is mike@blowmage.com

swcloud:
@sqrrrl @sporkmonger Thanks a lot for pushing the changes in Addressable so quickly! @blowmage  Yes, it makes sense to change the plan and stick with the google-api-client once @sqrrrl released a new version. This path is less risky.

sqrrrl:
@blowmage @swcloud -- ruby client 0.11 just pushed, please give it a try. LMK if you run in to any issues.

PR #1371, opened 2017-03-29
Add support for BigQuery job cancellation

leklund:
The `google-api-ruby-client` library supports BigQuery job cancellation:  https://github.com/google/google-api-ruby-client/blob/master/generated/google/apis/bigquery_v2/service.rb#L340

This PR adds support issuing a cancel request on a job.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
@@ -206,6 +206,13 @@ module Google
         end
 
         ##
+        # Cancel the job and return a gapi cancel job response.
+        def cancel
+          ensure_service!
+          service.cancel_job job_id
+        end
+
+        ##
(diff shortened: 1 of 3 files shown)


blowmage:
Hi @leklund, thank you for this pull request. We are happy to add this functionality. However, I have no idea how this was accepted. I know GitHub is saying I merged it, but this is the first I'm seeing this PR. But the good news is that this is great and I would have accepted and merged it anyway.

I'm going to make some small changes to the code that you contributed, to bring it inline with other services that have similar functionality. I would have requested these changes in the PR, if I didn't somehow already merge it...

¯\\\_(ツ)\_/¯ 

Thanks again!

leklund:
@blowmage Thanks! I was a little surprised how quickly it was merged :)

I see the updates you made in #1373 and it helped make sense of some of the confusion I had around what to return from the method. Thanks for the merge and the updates! Much appreciated.

Issue #1333, opened 2017-03-14
Test top 4 APIs on GAE Flex, GKE, and GCP


omaray:
Test BigQuery, Storage, Datastore, and Logging client libraries on all the GCP platforms (don't include GAE Standard).

Close this issue once we confirm it all works.

blowmage:
@swcloud Any input on how to do this? Should this be automated?

quartzmo:
I think the [Logging integration tests](https://github.com/GoogleCloudPlatform/google-cloud-ruby/tree/master/google-cloud-logging/integration) may serve as a foundation.

thagomizer:
If it doesn't need to be automated I can help. I used logging on 2 flex apps (one in production). I used storage via fog but can switch to our storage library in both. And deploying that code to GCE and GKE is straightforward. I'm only testing the happy path but it is working great.

quartzmo:
Thanks @thagomizer!

I will start down the automation path today, but manual tested may very well be needed depending on the schedule. 

I also propose creating separate issues for each package and each GCP platform.

omaray:
Hey guys - if automating is too much work let's hold off and chat at the next meeting. The idea was indeed to sanity check first to make sure we're not hitting anything obvious here that our customers would hit simply by deploying and calling our libraries.

But if we can easily automate, then yes that would also be awesome.

quartzmo:
@omaray I'll do a bit more research into test automation, so that we can have a better-informed discussion at the next meeting. I will also try to manually or semi-manually execute some tests as part of the exploration.

quartzmo:
@hxiong388 I believe I have successfully executed your amazing (and amazingly involved) automated GAE and GKE testing for Error Reporting and Logging, and I am hoping to use expand it to run tests for the packages in this issue. These tests will be as simple as possible, which I believe still involves deploying them within a Sinatra web app. At this point, I'm wondering: What is your take on testing GCE?

hxiong388:
GAE and GKE tests can be automated relatively easier, thanks to gcloud SDK and kubectl. While GCE may be a bit harder due to its IaaS nature. I think if we can get the tests automated on GAE and GKE, it may be sufficient to imply the coverage for GCE, because:
1. GAE and GKE are currently hosted from GCE vms. 
2. The GCE environment's two main services, metadata service accessibility and application default authentication, should both be covered by the GAE and GKE automation tests as well.

[...]

PR #1375, opened 2017-03-30
Integration test support

quartzmo:
This PR does the following in support of #1333:

* Unit tests
  * Fix stub signature for Translate Service.
  * Prevent return of actual project in any environment including GCE.
* Acceptance tests
  * Add rake task `acceptance:unsafe` to support using default credentials for acceptance testing.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -134,6 +134,21 @@ namespace :acceptance do
     end
   end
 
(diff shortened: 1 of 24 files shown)


blowmage:
Great catch on the translate stubs. As soon as this passes CI I'll merge.

Issue #1388, opened 2017-04-03
Language: Support for AnalyzeEntitySentiment.


landrito:
[`AnalyzeEntitySentiment`](https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1beta2/language_service.proto#L47) method was introduced. Support for this method would be great.

blowmage:
@landrito Are you going to submit a PR adding the `v1beta2` GAPIC files?

landrito:
Closing for now.

The method responses from v1beta2 are not of GA quality yet so we do not want to add them to the main surface. We do, however, want to give the user the ability to use the v1beta2 endpoint. This can be done by including the v1beta2 gapic in the package.

remi:
@landrito How can I analyze entity sentiment today using google-cloud-language?

landrito:
Sorry, The issue should read. **Language: Support for AnalyzeEntitySentiment in manual layer**.

After the next release cycle of NL. The way to analyze entity sentiment would be through the v1beta gapic.

PR #1390, opened 2017-04-04
Generate gapic client for language v1beta2

landrito:


quartzmo:
@landrito Looks like you will need to exclude `v1beta2` from [Rubocop checks](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-language/.rubocop.yml#L7-L8).

landrito:
Fixed!

landrito:
Updates #1388

blowmage:
The dependency on google-gax will need to be updated to 0.8.2 when this is released.

PR #1392, opened 2017-04-05
Update Storage File#copy to call rewrite_object

### LINKED ISSUE #1196: Expose objects.rewrite in File
As a follow-up to #1195, which adds `File#rotate` based on `objects.rewrite`, I would like to expose the full power of `objects.rewrite` in `File`. It may make the most sense to replace the implementation of `File#copy` with something that uses `objects.rewrite` instead of `objects.copy`, since on the server `objects.copy` seems to just be a single-request version of `objects.rewrite` under the hood. If so, `File#copy` could be expanded to accept a block to monitor and even control the multiple-request behavior used now in `File#rotate`. It could also accept options for rotating keys, just like `File#rotate`.

Comments/discussion welcome.

### LINKED ISSUE #1369: Objects.rewrite operations should support multiple requests
Currently, `File#storage_class=` and `File#rotate` use `Objects.rewrite`, but do not provide for multiple requests in case of an incomplete operation (incomplete rewrites should return a `rewriteToken`).

Similarly, a solution for #1196 and #1328 based on `Objects.rewrite` should support multiple requests using `rewriteToken`.

From [Objects.rewrite](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite):

> This method copies data using multiple requests so large objects can be copied with a normal length timeout per request rather than one very long timeout for a single request. 

and

> **rewriteToken string** Include this field (from the previous rewrite response) on each rewrite request after the first one, until the rewrite response 'done' flag is true. Calls that provide a rewriteToken can omit all other request fields, but if included those fields must match the values provided in the first rewrite request.

blowmage:
Copying large files may cause an API timeout. A better solution is
to switch to calling rewrite_object instead of calling copy_object.
Change the implementation, and other changes around rewrite_object.

[closes #1196, closes #1369]

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/service.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/service.rb
@@ -209,13 +209,16 @@ module Google
         def copy_file source_bucket_name, source_file_path,
                       destination_bucket_name, destination_file_path,
                       options = {}
+          key_options = rewrite_key_options options[:key],
+                                            options[:key]
           execute do
-            service.copy_object \
+            service.rewrite_object \
               source_bucket_name, source_file_path,
               destination_bucket_name, destination_file_path,
               destination_predefined_acl: options[:acl],
               source_generation: options[:generation],
-              options: key_options(options[:key])
+              rewrite_token: options[:token],
+              options: key_options
           end
         end
 
@@ -224,8 +227,8 @@ module Google
         def rewrite_file source_bucket_name, source_file_path,
                          destination_bucket_name, destination_file_path,
                          options = {}
-          options = rewrite_key_options options[:source_key],
-                                        options[:destination_key]
+          key_options = rewrite_key_options options[:source_key],
+                                            options[:destination_key]
           execute do
             service.rewrite_object \
               source_bucket_name, source_file_path,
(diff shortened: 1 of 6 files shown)


quartzmo:
Sorry for this late-breaking idea, but what about adding an option to control the number of API requests? We have this in BigQuery:

> @param [Integer] request_limit The upper limit of API requests to make to load all data. Default is no limit.

google-cloud-storage/lib/google/cloud/storage/service.rb
```diff
@@ -224,27 +226,29 @@ def copy_file source_bucket_name, source_file_path,
         def rewrite_file source_bucket_name, source_file_path,
```
quartzmo: I might be missing something, but isn't the implementation of this method exactly the same as `copy_file`, above? Sorry I just can't see the difference.

quartzmo:
Have you tested this with a file large enough to require multiple requests, and does the request indeed return immediately? If instead it blocks until a 30s timeout (my guess), we do not need the `sleep 1`.

quartzmo:
On second thought, I think you must be right about it returning immediately.

blowmage:
> but what about adding an option to control the number of API requests?

I didn't want to change the existing API, and `rotate!` already had the an implementa without any config. I'm open to discussing it, but I think that should be a separate issue.

google-cloud-storage/lib/google/cloud/storage/service.rb
blowmage: The only difference is the presence of the `rewriteToken` argument. There was a lot of churn to get this functionality, and I didn't want to add even more to combine all the calls to these two methods. We could do this as part of a refactor, but it doesn't need to be in this PR either. (I have another PR queued up once this is merged.)

blowmage: The other difference is in how the encryption key is handled. `copy_file` take just `key`, but rewrite_file takes `source_key` and `destination_key`.

quartzmo:
I realized that I've been confusing this operation with one that moves data between the local environment and the service. Instead, the long-running work is performed only on the service, so really we're just requesting the status of that work (as you pointed out once; sorry it failed to register). So please disregard the comments above about `request_limit` and `sleep`.

PR #1394, opened 2017-04-05
Update Storage File copy

blowmage:
Allow `File#copy` to update a file's attributes when, including `storage_class`.


quartzmo:
Is there unit test coverage for this?

blowmage:
Yes. Unit test coverage exists for updating `storage_class` in `update`. Just added unit tests for calling `copy` with a block.

PR #1397, opened 2017-04-05
Add IAM methods to Bucket

### LINKED ISSUE #1383: Support IAM methods in File
The IAM methods have been added to Storage Bucket and FIle/Object. These methods are considered Alpha, and should be added to the Ruby Storage client.

quartzmo:
This PR adds a Policy class as well as the IAM methods for Bucket to Storage.

It does not add the IAM methods for File (Object).

Acceptance tests are passing for me locally now after whitelisting projects.

[refs #1383]

### DIFF
--- a/google-cloud-storage/acceptance/storage/bucket_test.rb
+++ b/google-cloud-storage/acceptance/storage/bucket_test.rb
@@ -110,4 +110,33 @@ describe Google::Cloud::Storage::Bucket, :storage do
     random_bucket = storage.bucket "#{bucket_name}_does_not_exist"
     random_bucket.must_be :nil?
   end
+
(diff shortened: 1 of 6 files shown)


blowmage:
@omaray Do you want this released while the API is returning an error? Or should we hold off on this until the API behaves correctly?

quartzmo:
The PR is now complete and ready to merge except for the issue of the `InvalidArgumentError`.

omaray:
That's a good question. Let's discuss this at the next standup.

blowmage:
We've gotten confirmation that the Bucket IAM methods should be supported separately from the File/Object IAM methods. So this PR will only include changes for Bucket, and a future PR will include changes for File.

PR #1402, opened 2017-04-06
Language version docs

quartzmo:


landrito:
Why are we removing the beta warning?

quartzmo:
It's no longer on the cloud.google.com docs.

landrito:
Ah gotcha, consistency is good. I just wanted to verify that the SLA is still at a beta level. If so LGTM.

quartzmo:
I searched the official docs for any mention of "beta" and found only [this usage](https://cloud.google.com/natural-language/pricing):

> As of December 1, 2016,the Beta pricing discount of 50% for the Cloud Natural Language API is no longer in effect. This page reflects current prices.

But if you are sure that the SLA is still beta, then maybe I should remove that commit. Happy to do it.

blowmage:
When do we consider the Natural Language ruby gem to be Beta? currently the README lists it [as Alpha](https://github.com/GoogleCloudPlatform/google-cloud-ruby#cloud-natural-language-api-alpha).

landrito:
Woops sorry NL should be alpha right now.

landrito:
I think the confusion is that the API is v1 whereas the client is alpha. I think removing the beta documentation is fine since that was in reference to the API. The gem we will change to beta at a later date. Keep it alpha in the README.

PR #1411, opened 2017-04-07
Debugger::Railtie doesn't stop Rails from starting up 

### LINKED ISSUE #1410: Debugger Railtie raises exception if project id missing
The current `Google::Cloud::Debugger::Railtie` initialization procedure raises an exception if a GCP project id isn't provided through environment variable or Rails configuration while running from non-GCP environments.

Ideally, the `Google::Cloud::Debugger::Railtie` should behave similarly to the other Stackdriver Railties. It should log an warning message explaining the Debugger Railtie initialization failed due to missing project id, and not interrupt the Rails application from starting.

hxiong388:
Correctly wrap the `Google::Cloud::Debugger::Project` instantiating code inside the condition check block.

[closes #1410]

### DIFF
--- a/google-cloud-debugger/lib/google/cloud/debugger/rails.rb
+++ b/google-cloud-debugger/lib/google/cloud/debugger/rails.rb
@@ -58,19 +58,20 @@ module Google
         end
 
         initializer "Stackdriver.Debugger" do |app|
-          debugger_config = Railtie.parse_rails_config config
-
-          project_id = debugger_config[:project_id]
-          keyfile = debugger_config[:keyfile]
-          module_name = debugger_config[:module_name]
-          module_version = debugger_config[:module_version]
-
-          debugger = Google::Cloud::Debugger.new project: project_id,
-                                                 keyfile: keyfile,
-                                                 module_name: module_name,
-                                                 module_version: module_version
-
           if self.class.use_debugger? app.config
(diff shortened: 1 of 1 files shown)


google-cloud-debugger/lib/google/cloud/debugger/rails.rb
```diff
@@ -58,19 +58,20 @@ class Railtie < ::Rails::Railtie
         end
 
         initializer "Stackdriver.Debugger" do |app|
-          debugger_config = Railtie.parse_rails_config config
+          if self.class.use_debugger? app.config
+            debugger_config = Railtie.parse_rails_config config
```
blowmage: Where is `config` coming from? Shouldn this be `app.config`?

hxiong388: This `config` is defined in `railties` gem: https://github.com/rails/rails/blob/master/railties/lib/rails/railtie.rb#L122

The `config` and `app.config` are interchangeable in this initializer code. They both reference the same object. I actually thought about changing it here from `config` to `app.config`, but then I noticed the same code is referenced in the other three Railties we have. (This code is probably inherited from the previous `appengine` gem)

blowmage: kk, just checking. :+1:

PR #1415, opened 2017-04-10
Update IAM Policy docs

quartzmo:
Demonstrate IAM updates using block instead of setter.


quartzmo:
@blowmage Let me know if this change looks OK, and I'll add similar commits for Resource Manager, Spanner, and Storage. Please note that I removed some `# Api call` comments from the examples to improve readability, since this is adequately covered in `#policy` docs.

blowmage:
@quartzmo That looks great!

PR #1424, opened 2017-04-13
Add Spanner Client API

blowmage:
This PR implements the Client API, which is what users will use to connect to Cloud Spanner in their applications. A client is created from the main Spanner object:

```ruby
require "google/cloud/spanner"

spanner = Google::Cloud::Spanner.new
db = spanner.client "my-instance", "my-database"
```

The Client object offers the same functionality as Session, but also allows the creation of a Read/Write transaction:

```ruby
require "google/cloud/spanner"

spanner = Google::Cloud::Spanner.new
db = spanner.client "my-instance", "my-database"

db.transaction do |tx|
  results = tx.execute "SELECT * FROM users"

  results.rows.each do |row|
    puts "User #{row[:id]} is #{row[:name]}""
  end
end
```

As well as a Read/Only snapshot:

```ruby
require "google/cloud/spanner"

spanner = Google::Cloud::Spanner.new
db = spanner.client "my-instance", "my-database"

db.snapshot do |snp|
  results = snp.execute "SELECT * FROM users"

  results.rows.each do |row|
    puts "User #{row[:id]} is #{row[:name]}""
  end
end
```

This PR completes the major pieces of the API that users will use, but does not complete the Spanner implementation. There are still several missing features, such as the session pool, retrying aborted transactions, and resuming broken streams. But those features will not see much change to the public API as defined in this PR.


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -0,0 +1,429 @@
...
+        #
+        #   db = spanner.client "my-instance", "my-database"
+        #
+        #   results = db.read "users", ["id, "name"], streaming: false
+        #
+        #   results.rows.each do |row|
+        #     puts "User #{row[:id]} is #{row[:name]}""
+        #   end
+        #
+        def read table, columns, id: nil, limit: nil, streaming: true
```
quartzmo: I think you explained this to me once, but why not use a `Table` abstraction for all these methods that take `table` as the first argument?

quartzmo:
> The Client object offers the same functionality as Session, but

Is `Session` now gone for good?

blowmage:
Yes, Session is no longer part of the public API, but it will remain in the international API and will see additional features when session pooling is added.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -0,0 +1,608 @@
...
+        #
+        #   db.transaction do |tx|
+        #     results = tx.execute "SELECT * FROM users"
+        #
+        #     results.rows.each do |row|
+        #       puts "User #{row[:id]} is #{row[:name]}""
+        #     end
+        #   end
+        #
+        def transaction
```
quartzmo: Should an option for snapshot read-only transactions be available here?

Here are the advantages of their use, from the GAPIC docs: 

 > Snapshot read-only. This transaction type provides guaranteed consistency across several reads, but does not allow writes. Snapshot read-only transactions can be configured to read at timestamps in the past. Snapshot read-only transactions do not need to be committed.

> For transactions that only read, snapshot read-only transactions provide simpler semantics and are almost always faster. In particular, read-only transactions do not take locks, so they do not conflict with read-write transactions. As a consequence of not taking locks, they also do not abort, so retry loops are not needed.

quartzmo: I see, maybe this is covered by `#snapshot`? In that case, maybe adding a reference to `#snapshot` from the `#transaction` docs will help users understand.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
```diff
@@ -330,19 +333,42 @@ def streaming_read_table session_path, table_name, columns, id: nil,
...
+        end
+
+        def create_snapshot session_name, strong: nil, timestamp: nil,
+                            staleness: nil
+          tx_opts = Google::Spanner::V1::TransactionOptions.new(read_only:
+            Google::Spanner::V1::TransactionOptions::ReadOnly.new({
+              strong: strong,
+              read_timestamp: Convert.time_to_timestamp(timestamp),
+              exact_staleness: Convert.number_to_duration(staleness),
+              return_read_timestamp: true
```
quartzmo: Is there any support for bounded staleness modes? I would expect to see it here, using `ReadOnly#max_staleness` and `ReadOnly#min_read_timestamp`.

blowmage: No, those options are for single-use transactions, and as such are available on `Client#read` and not used on `Snapshot#read`.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
blowmage: Probably, we will need to take a comprehensive pass on the docs in a future PR.

blowmage:
Gonna merge now so we can open the follow-up PRs.

PR #1428, opened 2017-04-14
Speech use LINEAR16

### LINKED ISSUE #1426: Update Speech Client parameter from :raw to :LINEAR16
Hi,

We received a comment about updating samples to use :LINEAR16 instead of the file type :raw.

> Both RAW and WAV can encapsulate both LINEAR16 and Mulaw.

It may be a good idea to update snippets to use `:LINEAR16` instead of `:raw`. 

[1] https://github.com/GoogleCloudPlatform/ruby-docs-samples/issues/166
[2] https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/168

blowmage:
Replace the value :raw with :linear16, at the request of the Speech team. Update documentation and code examples to use :linear16.

[closes #1426]

### DIFF
--- a/google-cloud-speech/README.md
+++ b/google-cloud-speech/README.md
@@ -26,7 +26,7 @@ require "google/cloud/speech"
 speech = Google::Cloud::Speech.new
 
 audio = speech.audio "path/to/audio.raw",
-                     encoding: :raw,
+                     encoding: :linear16,
                      language: "en-US",
                      sample_rate: 16000
 
--- a/google-cloud-speech/acceptance/speech/async_test.rb
+++ b/google-cloud-speech/acceptance/speech/async_test.rb
@@ -22,7 +22,7 @@ describe "Asynchonous Recognition", :speech do
   let(:gcs_url)  { gcs_file.to_gs_url }
 
   it "recognizes audio from local file path" do
-    op = speech.process filepath, encoding: :raw, language: "en-US", sample_rate: 16000
+    op = speech.process filepath, encoding: :linear16, language: "en-US", sample_rate: 16000
 
     op.must_be_kind_of Google::Cloud::Speech::Operation
     op.wont_be :done?
@@ -37,7 +37,7 @@ describe "Asynchonous Recognition", :speech do
   end
 
   it "recognizes audio from local file object" do
-    op = speech.process File.open(filepath, "rb"), encoding: :raw, language: "en-US", sample_rate: 16000
+    op = speech.process File.open(filepath, "rb"), encoding: :linear16, language: "en-US", sample_rate: 16000
 
     op.must_be_kind_of Google::Cloud::Speech::Operation
     op.wont_be :done?
@@ -52,7 +52,7 @@ describe "Asynchonous Recognition", :speech do
   end
 
   it "recognizes audio from GCS URL" do
-    op = speech.process gcs_url, encoding: :raw, language: "en-US", sample_rate: 16000
+    op = speech.process gcs_url, encoding: :linear16, language: "en-US", sample_rate: 16000
 
     op.must_be_kind_of Google::Cloud::Speech::Operation
     op.wont_be :done?
@@ -67,7 +67,7 @@ describe "Asynchonous Recognition", :speech do
   end
 
   it "recognizes audio from Storage File URL" do
-    op = speech.process gcs_file, encoding: :raw, language: "en-US", sample_rate: 16000
+    op = speech.process gcs_file, encoding: :linear16, language: "en-US", sample_rate: 16000
 
     op.must_be_kind_of Google::Cloud::Speech::Operation
     op.wont_be :done?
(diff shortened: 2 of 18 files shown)


remi:
Any plan to remove `:raw`, which appears to have caused this confusion?

> *"Both RAW and WAV can encapsulate both LINEAR16 and Mulaw."* - issue [ruby-docs-samples/166](https://github.com/GoogleCloudPlatform/ruby-docs-samples/issues/166) 

As for a doc and sample change, I'm a big 👍  based on user feedback 😄

frankyn:
Is it possible to list the available symbols found [here](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/52b512b09ac5070a6c038f058793ecaa1f1267fe/google-cloud-speech/lib/google/cloud/speech/project.rb#L625-L630) as well?

remi:
+1 to @frankyn ~ it would be really helpful to document these!!

remi:
If `:raw` is to be removed, it is a time sensitive decision.  As that is backwards-incompatible, it should be decided on and done before any upcoming releases.

frankyn:
@blowmage The Speech team has asked to have :raw removed. As raw can be multiple different formats (e.g., mulaw, linear16, amr).

Thanks Mike!

blowmage:
> it would be really helpful to document these!!

They are documented. See [`Audio#encoding`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-speech/v0.24.0/google/cloud/speech/audio?method=encoding-instance) and all instances of the `encoding` method argument, such as on [`Project#recognize`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-speech/v0.24.0/google/cloud/speech/project?method=recognize-instance)

blowmage:
> The Speech team has asked to have :raw removed.

This PR removes all mention of `:raw` from the documentation and code examples. I would prefer to keep the undocumented alias in the implementation, since removing it may break existing code.

google-cloud-speech/README.md
```diff
@@ -26,7 +26,7 @@ require "google/cloud/speech"
 speech = Google::Cloud::Speech.new
 
 audio = speech.audio "path/to/audio.raw",
```
quartzmo: Should the file extension in the examples also be changed?

remi:
@blowmage 

> They are documented. See Audio#encoding and all instances of the encoding method argument, such as on Project#recognize

These mention the acceptable values but I don't mention that :linear and :linear16 are available options.  I had to go into the code to see that we support these useful aliases 😄

google-cloud-speech/README.md
remi: raw and wav are both file formats, but a raw file can include various types of encoded audio.

So I think raw is correct.

That said... using the Speech API introduced me to the .raw format, which I didn't previously know.  .wav, however, is something that "everybody knows"...

quartzmo:
@remi The PR removes `raw` and adds `linear16` (for example, in [audio.rb](https://github.com/blowmage/gcloud-ruby/blob/188818f873384eb4466d2c44b422ae5c57f1661c/google-cloud-speech/lib/google/cloud/speech/audio.rb#L62) to the list. It doesn't add `linear` afaik.

remi:
> I would prefer to keep the undocumented alias in the implementation, since removing it may break existing code.

The feedback we got was that :raw isn't technically *correct* because .raw doesn't necessarily mean LINEAR16, it could also be Mulaw encoded audio, for example.  This makes me thing it is really a bug.  So if we don't remove :raw now, is there a deprecation path that we could take to eventually remove it?  Could it be removed before something like a major version release?

quartzmo:
@blowmage Based on [this comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1426#issuecomment-294242805):

> raw: :LINEAR16 is not correct. raw can be multiple different formats (e.g., mulaw, linear16, amr)
so mapping raw to linear16 is not really correct

I feel the better choice is to remove the alias, and potentially break existing code.

blowmage:
> Could it be removed before something like a major version release?

Yes, I think it would be appropriate to remove it at the 1.0.0 release.

remi:
@quartzmo See [Project#convert_encoding](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/52b512b09ac5070a6c038f058793ecaa1f1267fe/google-cloud-speech/lib/google/cloud/speech/project.rb#L625-L630)

```ruby
>> Google::Cloud::Speech.new.send :convert_encoding, :linear
=> :LINEAR16
```

remi:
Also +1 on not using .raw

blowmage:
> I would also suggest changing the file extensions in the examples from .raw

Those files and that extension comes from the examples for node and python. I think the extension is appropriate, since those files are just the raw bytes, and not a WAV file with a WAV envelope. If it were a WAV file, it would make sense to use the .wav extension.

frankyn:
The issue is not the extension but the inferred format from using `:raw`. I agree with @blowmage about keeping the `.raw` extension.

PR #1431, opened 2017-04-18
Add Subscription retain_acked and retention

### LINKED ISSUE #1286: Add Pub/Sub Subscription retain_acked/retention_duration
#1274 adds the retain_acked/retention_duration values to subscriptions. Add these values to the model, create method, and update method.

quartzmo:
[closes #1286]

### DIFF
--- a/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
+++ b/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
@@ -174,10 +174,11 @@ describe Google::Cloud::Pubsub, :pubsub do
     end
 
     it "should allow creation of a subscription with options" do
-      subscription = topic.subscribe "#{$topic_prefix}-sub3", retain_acked: true
+      subscription = topic.subscribe "#{$topic_prefix}-sub3", retain_acked: true, retention: 600
       subscription.wont_be :nil?
       subscription.must_be_kind_of Google::Cloud::Pubsub::Subscription
       assert subscription.retain_acked
+      subscription.retention.must_equal 600
       subscription.delete
     end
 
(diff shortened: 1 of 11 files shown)


google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
```diff
@@ -173,10 +173,12 @@ def retrieve_snapshot project, subscription, snapshot_name
       next_subs.token.must_be :nil?
     end
 
-    it "should allow creation of a subscription" do
-      subscription = topic.subscribe "#{$topic_prefix}-sub3"
+    it "should allow creation of a subscription with options" do
```
blowmage: Can we have a test that creates a subscription with options and one that creates a subscription without options?

quartzmo: Sure. The reason I didn't add a new test is because it's acceptance testing, I thought it wasn't worth the cost.

blowmage: After looking at the implementation more I don't think this is as important as I thought at first. The potential issue is like to avoid isn't very likely. Sometimes creating protobuf objects with nil values causes issues, but we aren't doing that here with Subscription. Up to you to cover that with acceptance tests or not.

quartzmo: That sounds worth covering, I'll add it, just a moment.

quartzmo: @blowmage Actually, I just noticed that [other tests](https://github.com/quartzmo/google-cloud-ruby/blob/49d37fe5321c7d79a5b02ab45ee9ea8e7ad9b915/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb#L191-L214) create subscriptions without options, so it's covered.

blowmage: kk 👍

Issue #1430, opened 2017-04-17
[Cloud Storage] bucket-level IAM change request


frankyn:
Hi @blowmage and @quartzmo!

I recently worked on samples using the new bucket-level IAM API surface. ([PR](https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/167#discussion_r111654037)). 

I want to request that the API surface for reloading bucket-level IAM policy settings be changed from `bucket.policy force: true` to `bucket.policy.reload!`. It's more idiomatic and self-descriptive. 

I welcome your feedback, thanks!

quartzmo:
@frankyn Thanks for this suggestion, and please keep them coming!

Since google-cloud-storage is now GA/1.x, I don't think we'll want to remove the `force` option. What do you think about adding `bucket.policy.reload!` even though we keep `force`?

frankyn:
Is the rational for not removing `force:` because it would cause a breaking change? 
Could this be an exception being that Bucket-level IAM is still in Alpha? 

If not possible, I prefer not documenting the named parameter `force:` and only promoting the use of  `bucket.policy.reload!`.  

Thanks @quartzmo!

quartzmo:
> Could this be an exception being that Bucket-level IAM is still in Alpha?

Good question! I don't know if an exception can be made in this case. @blowmage what's your opinion on this?

blowmage:
AFAIK diversion and policy is for the gem, not services the gem covers

blowmage:
Guidance should be to update the IAM policy with the block. If this is the only way the Policy values are changed I don't see a compelling need for `force: true` or `Policy#reload!` in the samples.

```ruby
require "google/cloud/storage"

storage = Google::Cloud::Storage.new

bucket = storage.bucket "my-todo-app"

bucket.policy do |p|
  p.remove "roles/storage.admin", "user:owner@example.com"
  p.add "roles/storage.admin", "user:newowner@example.com"
  p.roles["roles/storage.objectViewer"] = ["allUsers"]
end
```

frankyn:
@blowmage that would make sense when updating the policy, but what about when testing changes similar to [this](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/72390d943e7e47b6ff25d4d8592382b02d18db05/storage/spec/iam_spec.rb#L46-L65)?

quartzmo:
[...]

PR #1433, opened 2017-04-20
Remove Bucket#policy force example.

quartzmo:
[refs #1430]

Although I am also working on an implementation of `Policy#reload!`, this PR can be merged earlier if desired, since I think everyone agrees on the change.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -920,17 +920,7 @@ module Google
         #   policy = bucket.policy # API call
         #   policy_2 = bucket.policy # No API call
         #
-        # @example Use `force` to retrieve the latest policy from the service:
-        #   require "google/cloud/storage"
-        #
-        #   storage = Google::Cloud::Storage.new
-        #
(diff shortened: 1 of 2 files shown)


google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -920,17 +920,7 @@ def default_acl
         #   policy = bucket.policy # API call
         #   policy_2 = bucket.policy # No API call
```
frankyn: This example raises the question of how to "force" a reload. Although, it does show the client does cache values to reduce the number of API calls.

frankyn:
Follow-up question, I think we should remove the [documentation](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.28.0/google/cloud/storage/bucket?method=policy-instance) of the `force` parameter as well once we have an alternative solution.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: I think we'd better keep this, since memoization is not typical in google-cloud-ruby.

PR #1435, opened 2017-04-21
Add Storage Policy#reload!

### LINKED ISSUE #1430: [Cloud Storage] bucket-level IAM change request

quartzmo:
[closes #1430]

This PR provides an implementation of `Policy#reload!` so that discussion can continue regarding the change proposed in #1430.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -898,10 +898,6 @@ module Google
         # @see https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy
         #   Buckets: setIamPolicy
         #
-        # @param [Boolean] force Force load the latest policy when `true`.
-        #   Otherwise the policy will be memoized to reduce the number of API
-        #   calls made. The default is `false`.
-        #
         # @yield [policy] A block for updating the policy. The latest policy
         #   will be read from the service and passed to the block. After the
         #   block completes, the modified policy will be written to the service.
@@ -920,6 +916,16 @@ module Google
         #   policy = bucket.policy # API call
         #   policy_2 = bucket.policy # No API call
         #
+        # @example Retrieve the latest policy from the service:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-todo-app"
+        #
+        #   policy = bucket.policy # API call
+        #   policy_2 = policy.reload! # API call
+        #
         # @example Retrieve the latest policy and update it in a block:
         #   require "google/cloud/storage"
         #
@@ -936,7 +942,7 @@ module Google
           @policy ||= begin
             ensure_service!
             gapi = service.get_bucket_policy name
-            Policy.from_gapi gapi
+            Policy.from_gapi gapi, service
           end
           return @policy unless block_given?
           p = @policy.deep_dup
@@ -977,7 +983,7 @@ module Google
         def policy= new_policy
           ensure_service!
           gapi = service.set_bucket_policy name, new_policy.to_gapi
-          @policy = Policy.from_gapi gapi
+          @policy = Policy.from_gapi gapi, service
         end
 
         ##
--- a/google-cloud-storage/lib/google/cloud/storage/policy.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/policy.rb
@@ -70,13 +70,31 @@ module Google
       #   end
       #
       class Policy
-        attr_reader :etag, :roles
+        ##
+        # @private The Service object.
+        attr_accessor :service
+
+        ##
+        # @private The Google API Client object.
+        attr_accessor :gapi
 
         ##
         # @private Creates a Policy object.
-        def initialize etag, roles
-          @etag = etag
-          @roles = roles
+        def initialize
+          @service = nil
+          @gapi = Google::Apis::StorageV1::Policy.new
+        end
+
+        ##
+        # The name of the bucket.
+        def resource_id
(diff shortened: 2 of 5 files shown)


google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -898,10 +898,6 @@ def default_acl
         # @see https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy
         #   Buckets: setIamPolicy
         #
-        # @param [Boolean] force Force load the latest policy when `true`.
```
blowmage: I do not want this parameter removed from the documentation. There are valid use cases for using it.

google-cloud-storage/lib/google/cloud/storage/policy.rb
```diff
@@ -162,43 +186,76 @@ def role role_name
         end
 
         ##
-        # Returns a deep copy of the policy.
+        # Returns a deep copy of the policy. The service object referenced by
```
blowmage: This note about the service object is going to be confusing. The fact that the Policy object holds a reference to a Service object is outside of the public API. I think this change to the documentation should be reverted.

google-cloud-storage/lib/google/cloud/storage/policy.rb
```diff
@@ -162,43 +186,76 @@ def role role_name
...
+          @gapi
         end
 
         ##
         # @private New Policy from a
         # Google::Apis::StorageV1::Policy object.
-        def self.from_gapi gapi
-          roles = gapi.bindings.each_with_object({}) do |binding, memo|
-            memo[binding.role] = binding.members.to_a
+        def self.from_gapi gapi, conn
```
blowmage: The name of the second argument should probably be `service` instead of `conn`. I realize that this was missed in other code when migrating from Google API Client 0.8 to 0.9, but we shouldn't propagate that mistake.

quartzmo: 👍

google-cloud-storage/lib/google/cloud/storage/policy.rb
quartzmo: Good point, agreed.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: I can see your point. @frankyn?

frankyn: In response to the question raised in [issue 1430](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1430). I think we should leave the parameter in the documentation, because it's able to refresh the `Bucket` memoized `Policy` that `Policy#reload!` doesn't. I'm not sure if there's another use-case that I'm not aware of, but that for sure is important. 

@quartzmo  I'm not sure how to proceed while `Policy#reload!` doesn't update the `Policy` instance of `Bucket`. @remi and @dazuma, what are your thoughts on having `Policy#reload` returning an updated `Policy` but not updating the `Policy` instance in `Bucket`? Example shown below. 

> Quoting @quartzmo 
```ruby

policy = bucket.policy # API call
policy_2 = bucket.policy.reload! # API call
policy_3 = bucket.policy # No API call. Stale reference, same instance as first line.
```

remi: Do we use `force: true` for these semantics in other Google Cloud Client Libraries today?

For a bit of background, this conversation started with a code review I was doing for Frank for Bucket IAM samples.

I saw this line of code: `@bucket.policy force: true` and [commented](https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/167#discussion_r111654037) that I couldn't, for the life of me, guess what it did.  Like *"What does it mean to 'force' a policy?  Or... what!?  I can't even..."*

I like `policy.reload!` because of its idiomatic nature (read: Rails).  If a param is also available, I would much prefer `bucket.policy reload: true` because it has semantic meaning and I can guess/infer from the language what the code does.

remi: #fwiw Sorry we didn't point this out with the initial API!  Of the GA interfaces, this was one of the last added and we *just* wrote samples with it.

Due to GA, `force` should be backwards compatible.  But I would be very pleased to see an additive `reload` named param in addition to a `#reload!` method

If we add `#reload!` but the parameter to reload when calling `Bucket#policy` is `force`, then that becomes an inconsistency.

Vote: Deprecate `force:` in favor of `reload:`.  Implementation something like `fetch_from_api! if reload || force`

frankyn: @quartzmo, how do you feel about not introducing #reload! given the stale state shown in the example usage and instead provide an alias for `force:` being `reload:` as mentioned by remi?

quartzmo: @frankyn Sounds great! Can you post this idea to the original issue?

frankyn:
@quartzmo, what is the status of this PR? Will you create another PR or continue with this one?

quartzmo:
Closing this PR because we have decided that removing memoization is a better solution for #1430.

frankyn:
Thanks, @quartzmo!

Issue #1237, opened 2017-02-15
Use google-cloud-resource-prefix Header


blowmage:
Google Cloud Spanner should send the project name, instance name, or database name in the `google-cloud-resource-prefix` header. Currently only project is being sent, but we should send the most specific value that is available. All data requests made by the client should provide the database name in the header.

vkedia:
To clarify this is only needed for the data api and not for the admin apis.

PR #1439, opened 2017-04-27
Spanner retry for broken streams

### LINKED ISSUE #1236: Resume broken streams
The PartialResultSet object can contain a resume token which can be used to resume a broken stream. The resume token is not guaranteed to be in every message. Results should be buffered until a resume token is present before yielding.

If a stream is broken, a new API request should be made with the token, and results continued to yielded from the new streaming response. If the broken stream cannot be resumed a failure should be raised.

blowmage:
This PR adds retry to streams responses that receive an ABORTED error. Responses are buffered while waiting for a `resume_token`. Values will be returned when the buffer has reached its upper bound.

closes #1236

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/results.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/results.rb
@@ -14,6 +14,7 @@
 
 
 require "google/cloud/spanner/convert"
+require "google/cloud/errors"
 
 module Google
   module Cloud
@@ -85,6 +86,7 @@ module Google
 
           fields = @metadata.row_type.fields
           values = []
+          cached_responses = []
           chunked_value = nil
           resume_token = nil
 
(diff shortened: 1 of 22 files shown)


google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -85,6 +86,7 @@ def rows
 
           fields = @metadata.row_type.fields
           values = []
+          cached_responses = []
```
quartzmo: Is `cached_responses` bounded somewhere? My understanding is that the buffer should be bounded, and once full, data returned regardless of `resume_token`.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -482,12 +482,19 @@ def delete table, *id
...
+        def transaction &block
           ensure_service!
           tx_session = session
           tx_grpc = @project.service.begin_transaction tx_session.path
           tx = Transaction.from_grpc(tx_grpc, tx_session)
-          yield tx
+          begin
+            block.call tx
+          rescue Google::Cloud::AbortedError
+            # Retry the entire transaction
```
quartzmo: I believe this should be performed with a backoff delay, using a value returned by the service?

google-cloud-spanner/lib/google/cloud/spanner/service.rb
```diff
@@ -376,12 +389,10 @@ def inspect
...
         protected
 
-        def default_headers
-          { "google-cloud-resource-prefix" => "projects/#{@project}" }
-        end
-
-        def default_options
-          Google::Gax::CallOptions.new kwargs: default_headers
+        def default_options_from_session session_name
+          default_prefix = session_name.split("/sessions/").first
```
quartzmo: The value of `default_prefix` will be the most specific of database name, instance name, or project name, correct?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
blowmage: No, the buffer is not bounded. The requirements document doesn't state how large the buffer should be, or how to measure the buffer size. The node implementation buffers up to 10 responses before responding. I don't see where the python implementation buffers responses. So I left it out for now.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
blowmage: I also left this out for now. I haven't been able to observe an ABORTED response from the API, so I don't know for sure how the value will be returned. Also, the grpc client is currently fixing some issues with exposing the trailing metadata.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
blowmage: For Data API calls it will always be the database name. In #1237 we are instructed that those are the only calls that require the header.

blowmage:
@quartzmo I have updated this PR with the bounds check matching the node implementation. I have removed the transaction retry and header changes and will re-submit them in separate PRs. Can you please review again?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -85,6 +86,8 @@ def rows
 
           fields = @metadata.row_type.fields
           values = []
+          buffered_responses = []
+          buffer_upper_bound = 10
```
quartzmo: Is `10` the best possible value for this bound? I have seen no suggested value in any documentation. I assume you got this from the google-cloud-node implementation since you mention it in the PR description. Is it just a matter of guessing for each platform?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -98,24 +101,71 @@ def rows
...
-                grpc.values.unshift merge(chunked_value, grpc.values.shift)
-                chunked_value = nil
-              end
-              to_iterate = values + grpc.values
-              chunked_value = to_iterate.pop if grpc.chunked_value
+              buffered_responses << grpc
+
+              if (grpc.resume_token && grpc.resume_token != "") ||
+                buffered_responses.size >= buffer_upper_bound
+                # This can set the resume_token to nil
```
quartzmo: This is (understandably) involved, and either comments or refactoring to smaller methods with meaningful names will help make it easier to follow. Just adding some comments will be fine for now.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
blowmage: This is simply the method used by node. I also haven't seen any direction in the requirements document about how to configure the buffer, or how to determine if its full. Is it the number of requests? Is it the total size of data in the buffer? I don't know.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
blowmage: In this specific case I felt a longer, unfactored method was better than implementing the wrong abstraction. It allows for easier changes.

We can't simply break the look up into smaller methods, since it all needs to execute within the same method context because of the Enumerator. It can probably be better, but I'd rather have it working and unfactored now than fully factored sometime in the future.

PR #1440, opened 2017-04-27
Add Image#annotate

### LINKED ISSUE #1437: Request: a way to run multiple types of Vision detection on a single Image
I have a program that runs multiple types of detection on an image.  I make a `vision.image` and then call `image.text` and `image.landmarks` and `image.labels` and more.  But this is slow *and* effects my Vision API quota.

So I looked for a method that I could call on my Vision::Image to detect multiple things in one API call, eg. `image.detect landmarks: true, text: true`.

Being very familiar with the Vision API, I happen to know that these use the `annotate` rpc method and I had previously scrolled past an `annotate` method in the docs for Vision.

But today, when I had a use-case for it, I was surprised to see that Image has no method that will run multiple types of detection (one annotate call) AFAICT!

I see that Project#annotate exists and accepts multiple images.

My ask is for a new method Image#annotate which returns the same Annotation object but for only that image.  I found it confusing that the Image didn't have this method available.

quartzmo:
[closes #1437]

### DIFF
--- a/google-cloud-vision/lib/google/cloud/vision/image.rb
+++ b/google-cloud-vision/lib/google/cloud/vision/image.rb
@@ -434,6 +434,96 @@ module Google
           annotation.web
         end
 
+        ##
+        # Performs detection of Cloud Vision
+        # [features](https://cloud.google.com/vision/reference/rest/v1/images/annotate#Feature)
+        # on the image. If no options for features are provided, **all** image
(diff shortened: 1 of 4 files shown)


quartzmo:
@remi Can you review this?

quartzmo:
@remi Ping!

PR #1443, opened 2017-04-28
Spanner Session Pool

### LINKED ISSUE #1235: Session Pool
Use a pool to manage sessions. Session creation is expensive and the number of sessions that can be created is limited. Sessions and the pool will be hidden from end users, but the pool behavior can be configured by the following settings:

1. `min`: Minimum number of sessions that the pool will maintain at any point in time.
2. `max`: Maximum number of active sessions that the pool will have at any point in time.
3. `keepalive`: The frequency that the session pool checks for idle sessions and keep them alive. If the number of idle sessions is more than `min` then they can be closed and culled from the pool.

Creating transactions with `BeginTransaction` is an expensive call, so some fraction of the pool may preallocate transactions. If this is done, end users should be able to control what percentage of the pool will do this. However, we should investigate if using `TransactionSelector.begin` adds the same amount of overhead that `BeginTransaction` does.

The pool should follow the behavior of other ruby connection pools and raise if there are no sessions available, so end users can handle the error and adjust the pool size if needed/wanted.

blowmage:
Implements a session pool with the following features

* Configure minimum sessions
* Configure maximum sessions
* Configure keep alive setting

The session pool does not pre-create transactions. And trying to use more sessions than the pool has configured will raise an error. There is no configuration to have the pool block until a session becomes available.

(closes #1235)

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/session.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/session.rb
@@ -494,6 +494,22 @@ module Google
         end
 
         ##
+        # @private
+        # Keeps the session alive by calling SELECT 1
+        def keepalive!
+          ensure_service!
+          execute "SELECT 1"
+          return true
+        rescue Google::Cloud::NotFoundError
+          @grpc = service.create_session \
+            Admin::Database::V1::DatabaseAdminClient.database_path(
+              project_id, instance_id, database_id)
+          self
+          execute "SELECT 1"
+          return false
+        end
+
+        ##
         # @private Creates a new Session instance from a
         # Google::Spanner::V1::Session.
         def self.from_grpc grpc, service
--- a/google-cloud-spanner/lib/google/cloud/spanner/session.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/session.rb
@@ -303,10 +303,10 @@ module Google
         #     c.insert "users", [{ id: 2, name: "Harvey",  active: true }]
         #   end
         #
-        def commit
+        def commit transaction_id: nil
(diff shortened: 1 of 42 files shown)


google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -487,10 +487,26 @@ def replace table, *rows
...
-          service.commit path, commit.mutations
+          service.commit path, commit.mutations, transaction_id: transaction_id
+        end
+
+        ##
+        # @private
+        # Keeps the session alive by calling SELECT 1
+        def keepalive!
+          ensure_service!
+          execute "SELECT 1"
```
quartzmo: This is a single-read query that does not involve a transaction, correct?

blowmage: Correct.

quartzmo:
I compared the background thread management introduced here to [`TimerTask`](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/timer_task.rb) and its related [`RubyExecutorService`](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/executor/ruby_executor_service.rb) in [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby). The only major difference I noticed was the synchronization on the stop/close/kill and the "task" operation, but I guess with the task performed here (the keepalive query), we don't really need to worry if it completes or not when the client is closed.

blowmage:
Right, that's my expectation. We don't have any resources held up in the thread, so it should be enough to simply kill it.

PR #1447, opened 2017-05-01
Remove non-streaming API calls

### LINKED ISSUE #1446: Remove non-streaming mode from Spanner
The client library "should not use the non streaming requests."

The `streaming` option currently present in various execute and read methods should be removed, along with all related support for these operations.

blowmage:
[closes #1446]


google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -198,29 +190,10 @@ def delete_session
...
+        def execute sql, params: nil, transaction: nil
           ensure_service!
-          if streaming
-            Results.execute service, path, sql,
-                            params: params, transaction: transaction
-          else
-            Results.from_grpc service.execute_sql \
-              path, sql, params: params, transaction: transaction
-          end
+          Results.execute service, path, sql,
```
quartzmo: Where is `Results.execute` defined? Sorry, just can't find it.

quartzmo: Never mind, i was on wrong branch. Found it.

google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -198,29 +190,10 @@ def delete_session
...
-        #   puts "User #{user_row[:id]} is #{user_row[:name]}""
-        #
-        def execute sql, params: nil, transaction: nil, streaming: true
+        def execute sql, params: nil, transaction: nil
           ensure_service!
-          if streaming
-            Results.execute service, path, sql,
-                            params: params, transaction: transaction
-          else
-            Results.from_grpc service.execute_sql \
```
quartzmo: Can we remove `Service#execute_sql` and `Service#read` if they are no longer needed?

blowmage: I wanted to keep them so I can still call the non-streaming API. Its buried deep in the private API so I saw no harm in keeping them on Service.

quartzmo: SGTM

PR #1452, opened 2017-05-04
Spanner Transaction Improvements

blowmage:
This PR makes the following changes to Transaction:

- [x] Add `Transaction#rollback`
- [x] Retry aborted transactions as many times as needed.
- [x] Add `deadline` to control how long a transaction can take to complete.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -446,6 +446,9 @@ module Google
         # a single logical point in time across columns, rows, and tables in a
         # database.
         #
+        # @param [Numeric] deadline The total amount of time in seconds the
+        #   transaction has to succeed.
+        #
         # @yield [transaction] The block for reading and writing data.
         # @yieldparam [Google::Cloud::Spanner::Transaction] transaction The
         #   Transaction object.
@@ -464,17 +467,24 @@ module Google
         #     end
         #   end
         #
-        def transaction &block
+        def transaction deadline: 120, &block
           ensure_service!
+
+          deadline = validate_deadline deadline
+          backoff = 1.0
+          start_time = current_time
+
           @pool.with_session do |session|
-            tx_grpc = @project.service.begin_transaction session.path
-            tx = Transaction.from_grpc(tx_grpc, session)
+            tx = create_transaction! session
             begin
               block.call tx
             rescue Google::Cloud::AbortedError => err
-              sleep delay_from_aborted(err)
-              tx_grpc = @project.service.begin_transaction session.path
-              tx = Transaction.from_grpc(tx_grpc, session)
+              # Re-raise if deadline has passed
+              raise err if current_time - start_time > deadline
+              # Sleep the amount from RetryDelay, or incremental backoff
+              sleep(delay_from_aborted(err) || backoff *= 1.3)
+              # Create new transaction and retry the block
+              tx = create_transaction! session
               retry
             end
           end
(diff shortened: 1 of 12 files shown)


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -599,6 +598,25 @@ def validate_snapshot_args! strong: nil, timestamp: nil, staleness: nil
...
+            seconds = err.metadata["retryDelay"]["seconds"].to_i
+            nanos = err.metadata["retryDelay"]["nanos"].to_i
+            return seconds if nanos.zero?
+            return seconds + (nanos / 1000000000.0)
+          end
+          # No metadata? Try the inner error
+          delay_from_aborted(err.cause)
+        rescue
+          # Any error gets the default delay, which is 0
+          return 0
```
vkedia: Can we default to exponential backoff.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -471,12 +471,11 @@ def transaction &block
...
-              # TODO: retrieve delay from ABORTED error
-              # Retry the entire transaction
-              tx2_grpc = @project.service.begin_transaction session.path
-              tx2 = Transaction.from_grpc(tx2_grpc, session)
-              block.call tx2
+            rescue Google::Cloud::AbortedError => err
+              sleep delay_from_aborted(err)
+              tx_grpc = @project.service.begin_transaction session.path
+              tx = Transaction.from_grpc(tx_grpc, session)
+              retry
```
vkedia: So this will just keep on retrying till the txn succeds or fails with some other error? I can imagine some pathological cases where transaction keeps on aborting many times. Why not have a deadline?

blowmage: Can you tell us how the deadline is supposed to behave? What values should be used? Is it configurable?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
blowmage: How should this behave? Does it increment the last known delay from the previous UNAVAILABLE? Or should we maintain a separate delay value? If we maintain a separate delay value, does it increment when we the UNAVAILABLE response has a delay?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
vkedia: It should be user configurable with some default value. Foe eg if user sets it to 2 minutes, you keep on trying till then. Also for the outgoing commit rpc, you should set the timeout to be whatever is remaining. So lets say 1 minute out of those 2 mintues have elapsed, then you set the timeout for the outgoing commit rpc to 1 minute.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
vkedia: Isn't unavailable retried at a lower level? I think this can be a separate delay which is increased each time you get an aborted.

blowmage: No, GAPIC does not handle errors during streaming responses. The GRPC errors are passed through the Enumerator object that responses are pulled from, and the GAPIC layer isn't involved. The GAPIC layer will only handles errors on the call that instigates the streaming request and creates the Enumerator object.

vkedia: Ok. This can just be a separate delay.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
quartzmo: Python implementation: https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/spanner/google/cloud/spanner/session.py#L280-L288

blowmage:
This PR has been updated. PTAL.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
blowmage: A deadline has been added.

quartzmo:
Reviewing now.

Issue #1449, opened 2017-05-04
How to handle missing/duplicate field names in Spanner?


blowmage:
In spanner the returned data can have missing/empty field names.

```ruby
db.execute("SELECT 1").rows.to_a #=> [{:""=>1}]
```

The first question is how to handle multiple missing/empty field names. The results are returned in Ruby hashes, and Hash does not allow duplicate field names, so the empty string symbol (`:""`) used for both columns is overridden.

```ruby
db.execute("SELECT 1, 2").rows.to_a #=> [{:""=>2}]
```

The second question is a variation on the first. How do we handle duplicate field names. For example:

```ruby
db.execute("SELECT 1 AS count, 2 AS count").rows.to_a #=> [{:count=>2}]
```

/cc @vkedia

blowmage:
My first impulse to the first question is to replace missing/empty field names with a default name that includes the column position. For example:

```ruby
db.execute("SELECT 1, 2").rows.to_a #=> [{:_col_0=>1, :_col_1=>2}]
```

But, this doesn't address the second question, since only the empty field names would be changed.

One options is to find all duplicates, and append a count to them:

```ruby
db.execute("SELECT 1, 2, 3 AS count, 4 AS count").rows.to_a
#=> [{:""=>2, :_1=>2, :count=>3, :count_1=>4}]
```

A variation on this is to modify all filed names that have a duplicate, not just the duplicate fields:

```ruby
db.execute("SELECT 1, 2, 3 AS count, 4 AS count").rows.to_a
#=> [{:_0=>2, :_1=>2, :count_0=>3, :count_1=>4}]
```

The returned field names would also be modified, so the names would match up. Currently, we have:

```ruby
db.execute("SELECT 1, 'hello'").types #=> {:""=>:STRING}
```

So we need some way to differentiate the field names.

```ruby
db.execute("SELECT 1, 'hello'").types #=> {:_0=>:INT64, :_1=>:STRING}
```

blowmage:
One possible solution is return a Hash that has values by name _and_ position. The duplicate names would be squashed just like it is now:

````ruby
db.execute("SELECT 1, 2, 3, 4").rows.to_a
#=> [{:""=>4, 0=>1, 1=>2, 2=>3, 3=>4}]

db.execute("SELECT 1 AS count, 2 AS count, 3 AS count, 4 AS count").rows.to_a
[...]

PR #1453, opened 2017-05-05
Add pairs argument to Spanner Results

blowmage:
The pair method argument allows rows and types to be returned as a nested Array of pairs rather than a Hash. This will be useful for when users have rows that have duplicate names. The values can be
represented in a data structure that allows for duplicate keys, while being familiar to Rubyists.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/results.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/results.rb
@@ -26,7 +26,12 @@ module Google
         ##
         # Indicates the field names and types for the rows in the returned data.
         #
-        # @return [Hash] The types of the returned data.
+        # @param [Boolean] pairs Allow the types to be represented as a nested
+        #   Array of pairs rather than a Hash. This is useful when results have
+        #   duplicate names. The default is `false`.
+        #
+        # @return [Hash, Array] The types of the returned data. The default is a
+        #   Hash. Is a nested Array of Arrays when `pairs` is specified.
         #
         # @example
         #   require "google/cloud/spanner"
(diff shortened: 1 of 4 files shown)


blowmage:
@quartzmo @vkedia This is an attempt at handling #1449. What do you think?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -26,7 +26,12 @@ class Results
         ##
         # Indicates the field names and types for the rows in the returned data.
         #
-        # @return [Hash] The types of the returned data.
+        # @param [Boolean] pairs Allow the types to be represented as a nested
+        #   Array of pairs rather than a Hash. This is useful when results have
+        #   duplicate names. The default is `false`.
```
vkedia: We should error out if this if false and we encounter duplicate names or multiple fields with no names.

blowmage: I don't think raising here is the best approach. A Hash is the most common data structure for querying a database, and I think most Rubyists will expect duplicate names to be squashed. As long as we give them a way to work with duplicate names I think we should be fine.

I have another change coming that will improve how missing names are handled. The duplicate names will still be an issue.

blowmage:
@quartzmo @vkedia What do you think about separating these structures into separate methods? I'm thinking `Results#rows` could yield a Hash, and `Results#row_pairs` could yield an array of array pairs. Similarly, `Results#types` could return a Hash, and `Results#row_pairs` could return an array of array pairs. Thoughts?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
vkedia: If someone has duplicate columns in his query, then I guess he would want to know the value of both the columns, otherwise why did he ask for both the columns in the first place. In that case why would he want to get back a Hash which removes the values of one of the columns. I would think in most cases, asking for the Hash when you have duplicate columns is a bug and throwin an error would make it easier to realise that.

vkedia:
yeah that sounds reasonable to me. But I would still prefer if the rows method threw an error if there are duplicates

quartzmo:
I like the separate methods ` Results#rows` and `Results#row_pairs`.

PR #1455, opened 2017-05-08
Spanner Snapshot Improvements

blowmage:
This PR adds the `single_use` argument to `Client#read` and `Client#execute` for specifying the single-use read-only transaction type. It also adds `Results#timestamp`.

### DIFF
--- a/google-cloud-spanner/acceptance/spanner/client/execute_test.rb
+++ b/google-cloud-spanner/acceptance/spanner/client/execute_test.rb
@@ -33,8 +33,24 @@ describe "Spanner Client", :execute, :spanner do
     row[:num].must_equal 42
   end
 
+  it "runs a simple query using a single-use strong option" do
+    results = db.execute "SELECT 42 AS num", single_use: { strong: true }
+    results.must_be_kind_of Google::Cloud::Spanner::Results
+
+    results.types.must_be_kind_of Hash
+    results.types.keys.count.must_equal 1
+    results.types[:num].must_equal :INT64
+
+    rows = results.rows.to_a # grab all from the enumerator
+    rows.count.must_equal 1
+    row = rows.first
+    row.must_be_kind_of Hash
+    row.keys.must_equal [:num]
+    row[:num].must_equal 42
+  end
+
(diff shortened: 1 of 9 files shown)


quartzmo:
While the coverage of read-only options in [the official docs](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionOptions.ReadOnly) is certainly a bit tricky to consume, I'm not sure that changing the underlying API option names from `read_timestamp`, `exact_staleness`, `min_read_timestamp`, and `max_staleness` to `timestamp`, `staleness`, `bounded_timestamp`, and `bounded_staleness` helps me better keep track of how this library manages its interactions with the service. I'm going to look into this more deeply.

blowmage:
I found [these docs](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.TransactionOptions.description.subsection_1.subsection_1) to be way more illuminating for how the Snapshots/Read-only options work. There are three main groups, **strong**, **exact**, and **bounded**. Single-use snapshots can use all three groups, but longer-running snapshots can only use **strong** and **exact**. Therefore I felt it was better to use those terms than to use the API names, which I feel are not as clear.

vkedia:
I would also prefer if we used the same terms as in the api definition otherwise it will create more confusion.

blowmage:
I've updated the PR and added support for the API names.

FWIW, I agree that there is confusion around these concepts, but I think the cause are the names the API has chosen. Even the [GAPIC docs](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/v1/doc/google/spanner/v1/transaction.rb#L143-L147) use strong/exact/bounded to explain the concepts. I tried to lessen the confusion by simplifying the names and bringing them more inline with the concepts.

quartzmo:
@blowmage The addition of the API names alongside the new names resolves my objection to the new names, since it establishes clear links to the API docs. I am concerned that client application code using only the new names might be problematic for maintainers of such code, but I'll leave it to users to make the decision of which names to pick.

PR #1463, opened 2017-05-10
Fix 'stack level too deep'

shahoxo:
maybe typo `@project`

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -53,7 +53,7 @@ module Google
         # The Spanner project connected to.
         # @return [Project]
         def project
-          project
+          @project
         end
 
         # The unique identifier for the project.
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -50,12 +50,6 @@ module Google
           @pool = Pool.new self, min: min, max: max, keepalive: keepalive
         end
 
-        # The Spanner project connected to.
-        # @return [Project]
-        def project
-          @project
-        end
-
         # The unique identifier for the project.
         # @return [String]
         def project_id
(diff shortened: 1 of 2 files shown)


quartzmo:
@shahoxo Thank you for finding this bug, and submitting this pull request! Much appreciated!

I think however that since this class also provides `#project_id` that returns the same value, a better fix would be to either:

1. Remove the `#project` method if it is not needed.

or 

2. Replace the body of `#project` with the body of `#project_id` (which appears to be OK), delete `#project_id`, and then alias `project_id` to `project`, similar to [how this is done in `project.rb`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/project.rb#L65-L68).

@blowmage Thoughts?

quartzmo:
I would also note that across the library, the presence of both `#project` and `#project_id` to return the project identifier is a little confusing.

shahoxo:
This is my intuition impression without knowing technical background.
#project is expected to return Project object.
#project_id is expected to return Project identifier. (ex. project.identifier)

blowmage:
I am okay with the commits being contributed to the project.

blowmage:
Apologies for not weighing in sooner, I was out yesterday with the flu.

First, this fix is correct. Second, Spanner is different than the other libraries, and the behavior to `#project` is different. Spanner is really 3 different APIs, two Admin APIs for Instances and Databases that is represented with the Project object, and one Data API that is represented with the Client object. It is most likely that users will create a shared Client object, meaning it may be difficult to reach back to the Admin API. That is why Client holds on to a reference to the Project object.

I've added test coverage for these methods to ensure that they work properly, and so that if we ever have this bug again it will be caught. I've also added calls to the Client's instance and database object, making it easier for users to move from the Data API to the Admin APIs.

I've also rebased this commit on to the latest master, so that the build will (hopefully) succeed.

quartzmo:
Thanks again for this PR, @shahoxo!

PR #1468, opened 2017-05-15
Session Pool Updates

blowmage:
This PR adds the remaining Session Pool features:

-[x] write transaction frequency
-[x] block/fail

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -52,12 +52,12 @@ module Google
         ##
         # @private Creates a new Spanner Client instance.
         def initialize project, instance_id, database_id, min: 2, max: 10,
-                       keepalive: 1500, write_ratio: 0.5
+                       keepalive: 1500, write_ratio: 0.5, fail: false
           @project = project
           @instance_id = instance_id
           @database_id = database_id
           @pool = Pool.new self, min: min, max: max, keepalive: keepalive,
-                                 write_ratio: write_ratio
+                                 write_ratio: write_ratio, fail: fail
         end
 
         # The unique identifier for the project.
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -52,7 +52,7 @@ module Google
         ##
         # @private Creates a new Spanner Client instance.
         def initialize project, instance_id, database_id, min: 2, max: 10,
-                       keepalive: 1500, write_ratio: 0.5, fail: false
+                       keepalive: 1500, write_ratio: 0.5, fail: true
           @project = project
           @instance_id = instance_id
           @database_id = database_id
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -52,11 +52,12 @@ module Google
         ##
         # @private Creates a new Spanner Client instance.
         def initialize project, instance_id, database_id, min: 2, max: 10,
-                       keepalive: 1500
+                       keepalive: 1500, write_ratio: 0.5
           @project = project
           @instance_id = instance_id
           @database_id = database_id
-          @pool = Pool.new self, min: min, max: max, keepalive: keepalive
+          @pool = Pool.new self, min: min, max: max, keepalive: keepalive,
+                                 write_ratio: write_ratio
         end
 
         # The unique identifier for the project.
@@ -602,8 +603,7 @@ module Google
           backoff = 1.0
           start_time = current_time
 
-          @pool.with_session do |session|
-            tx = create_transaction! session
+          @pool.with_transaction do |tx|
             begin
               block.call tx
             rescue Google::Cloud::AbortedError => err
@@ -611,15 +611,15 @@ module Google
               raise err if current_time - start_time > deadline
               # Sleep the amount from RetryDelay, or incremental backoff
               sleep(delay_from_aborted(err) || backoff *= 1.3)
-              # Create new transaction and retry the block
-              tx = create_transaction! session
+              # Create new transaction on the session and retry the block
+              tx = tx.session.create_transaction
               retry
             rescue RollbackError
               # Transaction block was interrupted, allow to continue
             rescue => err
               # Rollback transaction when handling unexpeced error
               # Don't use Transaction#rollback since it raises
-              session.rollback tx.send(:transaction_id)
+              tx.safe_rollback
               raise err
             end
           end
@@ -816,11 +816,6 @@ module Google
         # available.
         def ensure_service!
           fail "Must have active connection to service" unless @project.service
-        end
-
-        def create_transaction! session
-          tx_grpc = @project.service.begin_transaction session.path
-          Transaction.from_grpc(tx_grpc, session)
         end
 
         ##
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -27,39 +27,75 @@ module Google
       # {Google::Cloud::Spanner::Session} instances.
       #
       class Pool
-        attr_accessor :min, :max, :pool, :queue
+        attr_accessor :all_sessions, :session_queue, :transaction_queue
 
-        def initialize client, min: 2, max: 10, keepalive: 1500
+        def initialize client, min: 2, max: 10, keepalive: 1500,
+                       write_ratio: 0.5
           @client = client
           @min = min
           @max = max
           @keepalive = keepalive
-          @pool = []
-          @queue = []
+          @write_ratio = write_ratio
+          @write_ratio = 0 if write_ratio < 0
+          @write_ratio = 1 if write_ratio > 1
 
           # initialize pool and availability queue
           init
         end
 
         def with_session
-          session = checkout
-          yield session
-          checkin session
-        end
-
-        def checkout
-          if queue.empty?
-            fail "No available sessions" if pool.size >= @max
-            new_session!
+          session = checkout_session
+          begin
+            yield session
+          ensure
+            checkin_session session
           end
-          queue.shift
         end
 
-        def checkin session
-          fail "Cannot checkin session" unless pool.include? session
+        def checkout_session
+          read_session = session_queue.shift
+          return read_session if read_session
+          write_transaction = transaction_queue.shift
+          return write_transaction.session if write_transaction
+
+          fail "No available sessions" if all_sessions.size >= @max
+          new_session!
+        end
+
+        def checkin_session session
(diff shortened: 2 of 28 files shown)


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -52,11 +52,12 @@ class Client
         ##
         # @private Creates a new Spanner Client instance.
         def initialize project, instance_id, database_id, min: 2, max: 10,
-                       keepalive: 1500
+                       keepalive: 1500, write_ratio: 0.5, fail: false
```
quartzmo: I think it would be helpful to somehow group the following options, which are all for session pool configuration: `min`, `max`, `keepalive`, `write_ratio`, and `fail`. Perhaps with a `pool_` prefix or else by grouping them in a hash under a `pool_options` or `pool_config` option?

blowmage: I agree. Similar to how we use a `single_use` hash. I'll make the change and add documentation for it.

This method won't change, as it is marked private. I will only make this change on the public `Project#client` method.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
           @client = client
           @min = min
           @max = max
           @keepalive = keepalive
-          @pool = []
-          @queue = []
+          @write_ratio = write_ratio
+          @write_ratio = 0 if write_ratio < 0
+          @write_ratio = 1 if write_ratio > 1
+          @fail = fail
```
vkedia: This does not seem to have clean up of idle sessions?

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
       # {Google::Cloud::Spanner::Session} instances.
       #
       class Pool
-        attr_accessor :min, :max, :pool, :queue
+        attr_accessor :all_sessions, :session_queue, :transaction_queue
 
-        def initialize client, min: 2, max: 10, keepalive: 1500
+        def initialize client, min: 2, max: 10, keepalive: 1500,
```
vkedia: I think we can have the max to be higher. We can make it 100

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
 
-        def checkin session
-          fail "Cannot checkin session" unless pool.include? session
+        def checkout_transaction
+          @mutex.synchronize do
+            loop do
+              write_transaction = transaction_queue.shift
+              return write_transaction if write_transaction
+              read_session = session_queue.shift
+              return read_session.create_transaction if read_session
```
vkedia: I guess create_transaction makes an rpc? In that case you should not hold mutex for all that time.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
+
+        def checkout_session
+          @mutex.synchronize do
+            loop do
+              read_session = session_queue.shift
+              return read_session if read_session
+              write_transaction = transaction_queue.shift
+              return write_transaction.session if write_transaction
+
+              return new_session! if can_allocate_more_sessions?
```
vkedia: you should not hold mutext while calling new_session since that can take a while.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
 
         def keepalive!
           ensure_keepalive_thread!
 
-          # Call keep alive only on the queue, not all sessions.
-          # If a session is checked out, we can assume its being used.
-          queue.each(&:keepalive!)
+          # Call keep alive only on the queue, not all sessions or transactions.
+          # If a session or transaction is checked out assume its being used.
+          session_queue.each(&:keepalive!)
```
vkedia: How do you ensure that a session on which you are calling keep alive is not handed out to the user? There is no synchronization here?

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -88,29 +154,43 @@ def reset
...
         def init
-          @pool = []
-          @queue = []
+          @all_sessions = []
+          @session_queue = []
+          @transaction_queue = []
           ensure_keepalive_thread!
-          @min.times { new_session! }
+          # init session queue
+          @min.times { session_queue << new_session! }
```
vkedia: This and the the line below will make init block for a long time if min is high. Can this be done in a separate thread so as to not block init.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -88,29 +154,43 @@ def reset
 
         def close
           @thread.kill if @thread
-          @pool.each(&:delete_session)
-          @pool = []
-          @queue = []
+          @all_sessions.each(&:delete_session)
```
vkedia: what happens if delete_session raises an error for some session? Will it still be called for all the other sessions? If not then, you should catch that error and make sure that other sessions are still deleted.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
         def keepalive!
           ensure_keepalive_thread!
 
-          # Call keep alive only on the queue, not all sessions.
-          # If a session is checked out, we can assume its being used.
-          queue.each(&:keepalive!)
+          # Call keep alive only on the queue, not all sessions or transactions.
+          # If a session or transaction is checked out assume its being used.
+          session_queue.each(&:keepalive!)
+          transaction_queue.each(&:keepalive!)
```
vkedia: calling keepalive would invalidate the transaction that was previously begun. So either you should create a new transaction or take the session out of transaction_queue and put it in session_queue

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
vkedia: I dont think it makes sense to do keepalive on all the sessions in the queue. Instead you should keep track of when the session was last used and only if the last use time was more than 30 minutes in the past should keep alive be called.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
+              read_session = session_queue.shift
+              return read_session if read_session
+              write_transaction = transaction_queue.shift
+              return write_transaction.session if write_transaction
+
+              return new_session! if can_allocate_more_sessions?
+
+              fail SessionLimitError if @fail
+
+              @resource.wait @mutex
```
vkedia: what is close() on the pool is called while some thread is waiting here? I think that thread will remain blocked forever?

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -27,49 +29,113 @@ module Spanner
...
+
+        def checkin_session session
+          unless all_sessions.include? session
+            fail ArgumentError, "Cannot checkin session"
+          end
+
+          @mutex.synchronize do
+            session_queue.push session
+
+            @resource.broadcast
```
vkedia: why not signal instead of broadcast?

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll change `max` to 100.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: There is no culling of sessions.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll change this.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll change this.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: Agreed. I'll fix it.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll change it.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: Sessions checked out of the pool are not in the queue. I'll add synchronization.

blowmage: Okay.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: The keep alive is being called on the transaction. It includes the transaction_id in the API call.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll add error handling.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: I'll put this in a separate thread.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
vkedia: Yes I was talking about the case where while keepalive is ongoing on the session, it is checked out and the user then does some other operation. Or the other way round, where just before you are about to call keep alive, the session is checked out and user does some operation on the session.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
vkedia: I see. But that is not ideal since for one it will be a much expensive call then just doing it in a separate snapshot transaction. Also that will start the clock ticking for the transaction and it might actually take some locks (even though that seems surprising). So you should instead do it in a new single use readonly transaction. Another option here can be to just call begintransaction again instead of select 1 for the transaction ready sessions.

blowmage: I'll change `Transaction#keepalive` to call `BeginTransaction` again.

PR #1469, opened 2017-05-15
Update Spanner Results docs

quartzmo:


google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -64,8 +64,8 @@ def timestamp
         #
         #   results = db.execute "SELECT * FROM users"
         #
-        #   results.fields.each do |name, type|
-        #     puts "Column #{name} is type {type}"
+        #   results.fields.pairs.each do |field|
```
blowmage: Pretty sure we could say the following:

```ruby
results.fields.pairs.each do |key, value|
```

quartzmo: that's better for sure, thanks

PR #1478, opened 2017-05-16
Spanner docs updates and fixes

quartzmo:
This PR adds spanner to the docs manifest, updates the TOC listing, and makes various updates and fixes. It also adds the `true` return value to some `Session` methods.

### DIFF
--- a/google-cloud-spanner/docs/toc.json
+++ b/google-cloud-spanner/docs/toc.json
@@ -64,6 +64,14 @@
               "type": "google/cloud/spanner/v1/spannerclient"
             },
             {
+              "title": "InstanceAdminClient",
+              "type": "google/cloud/spanner/admin/instance/v1/instanceadminclient"
+            },
+            {
+              "title": "DatabaseAdminClient",
+              "type": "google/cloud/spanner/admin/database/v1/databaseadminclient"
+            },
+            {
               "title": "Data Types",
               "type": "google/spanner/v1/datatypes"
             }
--- a/google-cloud-spanner/docs/toc.json
+++ b/google-cloud-spanner/docs/toc.json
@@ -32,24 +32,36 @@
     {
       "title": "Spanner",
       "type": "google/cloud/spanner",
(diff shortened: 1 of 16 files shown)


google-cloud-spanner/docs/toc.json
```diff
@@ -32,26 +32,38 @@
...
           "type": "google/cloud/spanner/v1",
-          "patterns": ["\/spanner\/v1", "\/spanner\/admin\/instance\/v1", "\/spanner\/admin\/database\/v1", "google\/iam", "google\/protobuf", "google\/rpc"],
+          "patterns": ["\/spanner\/v1", "google\/iam", "google\/protobuf", "google\/rpc"],
           "nav": [
             {
               "title": "SpannerClient",
               "type": "google/cloud/spanner/v1/spannerclient"
             },
             {
-              "title": "InstanceAdminClient",
```
blowmage: Why were these removed? Spanner has 3 APIs, I think it would be good to show all 3.

google-cloud-spanner/lib/google/cloud/spanner/commit.rb
```diff
@@ -127,6 +129,8 @@ def upsert table, *rows
         #   See [Data
         #   types](https://cloud.google.com/spanner/docs/data-definition-language#data_types).
         #
+        # @return [Array<Hash>] The array of rows.
```
blowmage: Why is `upsert` returning the rows? I don't think it should be returning anything.

quartzmo: I don't know why, but that is what the method returns.

google-cloud-spanner/docs/toc.json
quartzmo: Right, I'll restore those other GAPIC clients, thanks.

google-cloud-spanner/lib/google/cloud/spanner/commit.rb
blowmage: I think it would be better to leave that behavior undocumented. It's not part of the API design.

quartzmo: Done, removed.

Issue #1451, opened 2017-05-04
Test Spanner code examples


quartzmo:
Also known as: "snippets testing". Provide automated testing in the Rake `doctest` task for the Spanner documentation examples (YARD's `@example` tag) using the [yard-doctest](https://github.com/p0deje/yard-doctest) library, as discussed in #226.

PR #1489, opened 2017-05-17
Spanner guide

quartzmo:
I adapted this guide from [Getting Started with Cloud Spanner in Python](https://cloud.google.com/spanner/docs/getting-started/python/). The examples remain untested, pending #1451.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner.rb
@@ -195,7 +195,7 @@ module Google
     #
     # db = spanner.client "my-instance", "my-database"
     #
-    # results = db.read "users", ["id", "name"], keys: 1..5
+    # results = db.read "users", [:id, :name], keys: 1..5
     #
     # results.rows.each do |row|
     #   puts "User #{row[:id]} is #{row[:name]}"
@@ -286,7 +286,7 @@ module Google
     #   # Perform another read using the `read` method. Even if the data
     #   # is updated in-between the reads, the snapshot ensures that both
     #   # return the same data.
-    #   results_2 = db.read "users", ["id", "name"]
+    #   results_2 = db.read "users", [:id, :name]
     #   results_2.rows.each do |row|
     #     puts "User #{row[:id]} is #{row[:name]}"
     #   end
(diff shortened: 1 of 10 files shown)


google-cloud-spanner/lib/google/cloud/spanner.rb
```diff
@@ -19,7 +19,310 @@
...
+    # lists and ranges:
+    #
+    # ```ruby
+    # require "google/cloud/spanner"
+    #
+    # spanner = Google::Cloud::Spanner.new
+    #
+    # db = spanner.client "my-instance", "my-database"
+    #
+    # results = db.read "users", ["id", "name"], keys: 1..5
```
blowmage: Even thought both String and Symbol are supported, I think it would be better to specify columns names with symbols, since that is what is being used to access the row values later on in the code example:

```ruby
results = db.read "users", [:id, name], keys: 1..5
```

quartzmo: Sure, I will change this example in a new commit, since it now is used across several files (12 usages total).

quartzmo: Actually, the docs for `#read` only specify String: `@param [Array<String>] columns`. I suggest changing this in a separate PR.

blowmage: We should change it to `Array<String, Symbol>` then.

quartzmo:
Are there tests for symbol arguments? I recommend making this change in a separate PR.

google-cloud-spanner/lib/google/cloud/spanner.rb
quartzmo: Are there tests for symbol arguments? I recommend making this change in a separate PR.

blowmage: Yes. Tests use symbols, and the protobuf only accepts strings. It's fine to update the documention with the code it as-is.

quartzmo: OK, thanks. SGTM, will update.

PR #1490, opened 2017-05-18
Add video intelligence GAPIC

swcloud:


quartzmo:
@swcloud I am reviewing this now.

quartzmo:
@swcloud The build appears to be failing on Rubocop errors. Can you either remove Rubocop from the `Rakefile`, or else add a `.rubocop.yml` file excluding the source directory? To my knowledge, we do not run Rubocop on GAPIC files in other packages.

quartzmo:
@swcloud Should I update the `Rakefile` and/or `.rubocop.yml` to match those in `google-cloud-monitoring`?

quartzmo:
I am OK with my commits being contributed to this project.

quartzmo:
The build is now failing with 3 unit test errors. The cause is:

```
RuntimeError: Could not load the default credentials.
```

which makes me think that the unit tests do not adequately mock all requests to the live service.

blowmage:
Yep, makes sense.

google-cloud-video_intelligence/.yardopts
```diff
@@ -0,0 +1,8 @@
+--no-private
+--title=Google Cloud Video Intelligence API
+--exclude lib/google/cloud/video_intelligence/v1beta1
```
blowmage: Doesn't excluding this directory hide the GAPIC docs from the YARD output? Are we sure we want this? @quartzmo, can you confirm the docs are still correct?

quartzmo: @blowmage I'm looking into it now.

quartzmo: The exclusions in .yardopts appear incorrect. If it is like other packages, it should only include GAPIC and protobuf docs from `doc` dirs, in this case: `google-cloud-video_intelligence/lib/google/cloud/video_intelligence/v1beta1/doc`. I will update the file. 

I will also add `video_intelligence` to `gcloud-ruby/docs/manifest.json`, in order for the docs to be loaded and accessible via top navigation in the gh-pages site. This will **not** add the docs to the `google-cloud` meta-package.

blowmage: Sounds good. I already added the entry to the main manifest file in the gem setup commit.

blowmage:
I've rebased this branch to master to pick up a fix to the CircleCI configuration. It appears RVM is no longer installing the bundler gem by default now.

blowmage:
This PR has seen a lot of churn because of the directive to get this released quickly. I suggest we squash the PR into one commit when merging.

blowmage:
@swcloud Is it your intention that this gem is added as a dependency on the `google-cloud` gem?

google-cloud-video_intelligence/.yardopts
quartzmo: I have locally fixed and tested the `.yardopts` exclusions, and fixed incorrect paths in `toc.json`. I am now looking into problems with the "Data Types" generation.

quartzmo:
@blowmage There is a whitelist in the top-level Rakefile's `google_cloud` task that needs to be updated.

PR #1495, opened 2017-05-22
Spanner Nested Merge

blowmage:
The merge logic needed to be updated to support deeply nested lists. @vkedia, can you verify this behavior is correct?

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/results.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/results.rb
@@ -256,7 +256,8 @@ module Google
           elsif left.kind == :list_value
             left_val = left.list_value.values.pop
             right_val = right.list_value.values.shift
-            if left_val.kind == :string_value && right_val.kind == :string_value
+            if (left_val.kind == right_val.kind) &&
+               (left_val.kind == :list_value || left_val.kind == :string_value)
               left.list_value.values << merge(left_val, right_val)
             else
               left.list_value.values << left_val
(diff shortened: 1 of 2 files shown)


blowmage:
The new test shows how deeply nested lists should be merged. The current [merge test](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/test/google/cloud/spanner/results/merge_test.rb) already tests merging lists with mixed value types, e.g. `numberValue`, `stringValue`, nullValue`, etc...

vkedia:
Looks perfect. Thanks for the comprehensive test. BTW I could not find read_test.json file in this repo which has a test case that could have caught this:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/test/resources/com/google/cloud/spanner/read_tests.json#L150

blowmage:
When we first wrote these tests that JSON wouldn't load in the Protobuf objects, so we had to diverge our tests.

vkedia:
I see. What error were you getting?

vkedia:
Also the chunks field in that JSON would load into protocol buffers of PartialResultSet. Everything else is not a protocol buffer.

blowmage:
The PartialResultSet data defined in the JSON file wouldn't load in the protobuf objects. I don't remember what the problem was, but the data structure wasn't correct. So I recreated the test scenarios defined in the JSON file in test fixtures. This was in late 2016, IIRC.

PR #1499, opened 2017-05-24
Expose Commit Timestamp

blowmage:
This PR adds timestamp as a return value for methods that will modify data. Previously, these methods returned `true`, but will now return a `Time` value returned from Spanner in `Google::Spanner::V1::CommitResponse.commit_timestamp`.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -372,7 +372,7 @@ module Google
         # @yield [commit] The block for mutating the data.
         # @yieldparam [Google::Cloud::Spanner::Commit] commit The Commit object.
         #
-        # @return [Boolean] Returns `true` if the operation succeeded.
+        # @return [Time] The timestamp at which the operation committed.
         #
         # @example
         #   require "google/cloud/spanner"
@@ -419,7 +419,7 @@ module Google
         #   See [Data
         #   types](https://cloud.google.com/spanner/docs/data-definition-language#data_types).
         #
-        # @return [Boolean] Returns `true` if the operation succeeded.
+        # @return [Time] The timestamp at which the operation committed.
         #
         # @example
         #   require "google/cloud/spanner"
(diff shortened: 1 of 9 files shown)


google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -340,10 +341,9 @@ def commit transaction_id: nil
         #                       { id: 2, name: "Harvey",  active: true }]
         #
         def upsert table, *rows, transaction_id: nil
-          commit = Commit.new
-          commit.upsert table, rows
-          service.commit path, commit.mutations, transaction_id: transaction_id
-          true
+          commit transaction_id: transaction_id do |c|
```
quartzmo: Nice work refactoring these methods to use `#commit` internally!

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -372,7 +372,7 @@ def read table, columns, keys: nil, index: nil, limit: nil,
         # @yield [commit] The block for mutating the data.
         # @yieldparam [Google::Cloud::Spanner::Commit] commit The Commit object.
         #
-        # @return [Boolean] Returns `true` if the operation succeeded.
+        # @return [Time] The timestamp at which the operation committed.
```
quartzmo: I would change all of these that read "the operation committed" to "the transaction committed". Or perhaps make it even clearer with the exact API doc from `CommitResponse`: "The Cloud Spanner timestamp at which the transaction committed."

blowmage: I made these different because when users call `client.insert row` it isn't obvious to them that a single-use transaction is being used. Since the client handles all that for the users I didn't think it made sense to reference a transaction in the return value description.

quartzmo: I think it's best to use a precise definition, and if it signals that there is a bit more to the story (a transaction is at work), so much the better. But your call.

PR #1501, opened 2017-05-25
Add acceptance tests for concurrent transactions

### LINKED ISSUE #1484: Concurrent Transaction Coverage
Add test coverage for multiple concurrent transactions operating on the same data.

quartzmo:
Add test coverage for multiple concurrent transactions.

These tests add a total of 4 seconds of sleep time. If that is unacceptable, let me know and I'll remove it.

[closes #1484]

### DIFF
--- a/google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
+++ b/google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 require "spanner_helper"
+require "concurrent"
 
 describe "Spanner Client", :transaction, :spanner do
   let(:db) { spanner_client }
   let(:columns) { [:account_id, :username, :friends, :active, :reputation, :avatar] }
   let(:fields_hash) { { account_id: :INT64, username: :STRING, friends: [:INT64], active: :BOOL, reputation: :FLOAT64, avatar: :BYTES } }
   let(:additional_account) { { account_id: 4, username: "swcloud", reputation: 99.894, active: true, friends: [1,2] } }
+  let(:query_reputation) { "SELECT reputation FROM accounts WHERE account_id = 1 LIMIT 1" }
 
   before do
     db.commit do |c|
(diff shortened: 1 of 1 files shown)


blowmage:
@vkedia Can you confirm these are the tests you are looking for?

vkedia:
Overall looks ok
Can you add a toggle to keep track whether it is the first attempt of a txn and only sleep on the first attempt. 
You can also use something like this to prevent sleeping:
http://www.rubydoc.info/github/ruby-concurrency/concurrent-ruby/Concurrent/CountDownLatch

We use something similar in java tests:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java#L106

quartzmo:
@vkedia Yes, I will try to add a mechanism like `commitBarrier` in the Java tests. Thanks for the tip.

quartzmo:
@vkedia I locally removed the `sleep` statements and added a `CountDownLatch` (and some debug print statements.) I have run the test repeatedly, and the output has always been:

```sh
# Running:

read 2
read 1
write 2
commit 2
read 1
write 1
commit 1
.

Finished in 2.299352s, 0.4349 runs/s, 1.3047 assertions/s.

1 runs, 3 assertions, 0 failures, 0 errors, 0 skips
```

Does this look correct to you? I am assuming that the initial transaction for thread `1` is being aborted and retried.

vkedia:
I am not sure where you put the debug statement. But shouldn't the "write
1" log come twice?

quartzmo:
@vkedia I updated the PR with my local changes. The debug statements are *after* the operations.

vkedia:
Looking at that shouldn't the "write 1" log also come twice? Txn would only abort when commit() is called on it. Can you also print the value that is read by both the threads.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -126,6 +128,83 @@
       assert_accounts_equal expected, actual
     end
   end
+  focus
```
blowmage: `focus`

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -126,6 +128,83 @@
...
+      latch = Concurrent::CountDownLatch.new 1
+      thr_1 = Thread.new do
+        db.transaction do |tx|
+          tx_results = tx.read "accounts", [:reputation], keys: 1, limit: 1
+          puts "read 1"
+          tx_val = tx_results.rows.first[:reputation]
+          new_val = tx_val + 1
+          tx.update "accounts", [{ account_id: 1, reputation: new_val }]
+          puts "write 1"
+          latch.wait
```
blowmage: Don't you want to wait before you do the update?

quartzmo:
@vkedia I added the read result as follows: 

```ruby
        db.transaction do |tx|
          tx_results = tx.read "accounts", [:reputation], keys: 1, limit: 1
          tx_val = tx_results.rows.first[:reputation]
          puts "read 1: #{tx_val}"
```


```sh
# Running:

read 2: 63.5
read 1: 63.5
write 2
commit 2
read 1: 64.5
write 1
commit 1
.
```

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
quartzmo: I copied the [Java test](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/it/ITTransactionTest.java#L108), which appears to me to be waiting after the update.

vkedia: In Java, the buffer method just buffers the mutations in memory and does not make the commit call. Commit is made automatically by the library when the callable returns. So it is waiting before the commit rpc is made. But here I think the update call also makes the rpc, so you are waiting after the rpc. @blowmage is correct in that you should wait before the commit. Then only you can guarantee that both the threads have read the value before they try to commit.

vkedia:
This looks ok. I had not realized that `tx.update` actually makes the commit rpc. How would I update multiple rows, possible in different tables, in the transaction?

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
quartzmo: >  @blowmage is correct in that you should wait before the commit. 

The current position of the wait is just before the end of the block, and therefore just before the commit. @blowmage asked:

> Don't you want to wait before you do the update?

vkedia: Looking at the code here, it seems that update calls commit:
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/session.rb#L433

quartzmo:
I was also thinking that rpc commit would not be called until the completion of the block passed to `db.transaction`. This is despite having my having reviewed all of this several times over.

vkedia:
In that case we need to fix this API for couple of reasons. Firstly I think Its behaviour is non intutive.  In all the other languages, update just buffers the mutations in memory. Then either user has to explicitly call commit or the library automatically calls commit when the callback returns. More importantly the way API is right now, it does not allow multiple mutations to be done in the same mutation. That definitely needs to be fixed.

quartzmo:
@vkedia Mike (@blowmage) is out today, but I spoke with him on the phone this morning and he agrees with changing the behavior of `Transaction` to collect mutations until an explicit commit is called or the block returns. I plan to work on it today, and submit as part of this PR.

vkedia:
Thanks!

quartzmo:
I have rebased this PR on master to get both the correct transaction behavior from #1516 and the addition of concurrent-ruby from #1502. I have also updated it as follows: The test now lets thread 1 read first, then makes thread 1 wait until thread 2 has read, written, and committed. The debug output now looks like this:

```sh
# Running:

read 1: 63.5
read 2: 63.5
write 2
commit 2
write 1
read 1: 64.5
write 1
commit 1
.

Finished in 18.937448s, 0.0528 runs/s, 0.1584 assertions/s.

1 runs, 3 assertions, 0 failures, 0 errors, 0 skips
```

or sometimes like this:

```sh
# Running:

read 1: 63.5
read 2: 63.5
write 2
read 2: 63.5
write 2
commit 2
write 1
read 1: 64.5
write 1
commit 1
.

Finished in 19.395449s, 0.0516 runs/s, 0.1547 assertions/s.

1 runs, 3 assertions, 0 failures, 0 errors, 0 skips
```

Issue #1485, opened 2017-05-17
Spanner Pool Changes


blowmage:
The following changes have been requested:

- [x] Change max to 100
- [x] Create new sessions/transactions outside of mutex
- [x] Ensure Pool#close does not block waiting threads
- [x] Use signal instead of broadcast
- [x] Synchronize when calling keepalive
- [x] Track last used time on session, use for more intelligent keepalive
- [x] Handle errors when calling Session#delete_session
- [x] Add threading to Pool#init, so it won't block
- [x] Cull idle sessions
- [ ] ~Change Transaction#keepalive to call BeginTransaction~

vkedia:
Can you also add cleanup of idle sessions to this list?

blowmage:
The Transaction#keepalive item has been removed from the list.

blowmage:
Delivered in #1502

PR #1502, opened 2017-05-25
Spanner pool updates

blowmage:
This PR addresses the following items from #1485:

- [x] Change max to 100
- [x] Create new sessions/transactions outside of mutex
- [x] Ensure Pool#close does not block waiting threads
- [x] Use signal instead of broadcast
- [x] Synchronize when calling keepalive
- [x] Track last used time on session, use for more intelligent keepalive
- [x] Handle errors when calling Session#delete_session
- [x] Add threading to Pool#init, so it won't block
- [x] Cull idle sessions
- [x] Change Transaction#keepalive to call BeginTransaction

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -32,7 +32,8 @@ module Google
         attr_accessor :all_sessions, :session_queue, :transaction_queue
 
         def initialize client, min: 10, max: 100, keepalive: 1800,
-                       write_ratio: 0.3, fail: true, block_on_init: nil
+                       write_ratio: 0.3, fail: true, block_on_init: nil,
+                       skip_background_thread: nil
           @client = client
           @min = min
           @max = max
@@ -46,7 +47,8 @@ module Google
           @resource = ConditionVariable.new
 
           # initialize pool and availability queue
-          init block_on_init: block_on_init
+          init block_on_init: block_on_init,
+               skip_background_thread: skip_background_thread
         end
 
         def with_session
@@ -166,7 +168,7 @@ module Google
           @mutex.synchronize do
             @all_sessions.map do |s|
               Thread.new do
-                s.delete_session rescue nil
+                s.release!
               end
             end.map(&:join)
             @all_sessions = []
@@ -177,13 +179,43 @@ module Google
           true
         end
 
+        def keepalive_or_release!
+          to_keepalive = []
+          to_release = []
+
+          @mutex.synchronize do
+            available_count = session_queue.count + transaction_queue.count
+            release_count = @min - available_count
+            release_count = 0 if release_count < 0
+
+            to_keepalive += (session_queue + transaction_queue).select do |x|
+              x.idle_since? @keepalive
+            end
+
+            # Remove a random portion of the sessions and transactions
+            to_release = to_keepalive.sample release_count
+            to_keepalive -= to_release
+
+            # Remove those to be released from circulation
+            @all_sessions -= to_release.map(&:session)
+            @session_queue -= to_release
+            @transaction_queue -= to_release
+          end
+
+          to_release.map! { |x| Thread.new { x.release! } }
+          to_keepalive.map! { |x| Thread.new { x.keepalive! } }
+
+          # join all the threads before returning
+          (to_release + to_keepalive).map(&:join)
+        end
+
         private
 
-        def init block_on_init: nil
+        def init block_on_init: nil, skip_background_thread: nil
           @all_sessions = []
           @session_queue = []
           @transaction_queue = []
-          ensure_valid_thread!
+          ensure_background_thread! unless skip_background_thread
           # init session queue
           num_transactions = (@min * @write_ratio).round
           num_sessions = @min.to_i - num_transactions
@@ -226,29 +258,11 @@ module Google
           all_sessions.size < @max
         end
 
-        def ensure_valid!
-          ensure_valid_threads = []
-          @mutex.synchronize do
-            ensure_valid_threads += session_queue.map do |s|
-              Thread.new do
-                s.ensure_valid! since: @keepalive
-              end
-            end
-            ensure_valid_threads += transaction_queue.map do |tx|
-              Thread.new do
-                tx.session.ensure_valid! since: @keepalive
-              end
-            end
-          end
-          # join all the threads before returning
-          ensure_valid_threads.map(&:join)
-        end
-
-        def ensure_valid_thread!
+        def ensure_background_thread!
           @thread ||= Thread.new do
             # before calling keepalive
             loop do
-              ensure_valid!
+              keepalive_or_release!
               sleep 300
             end
           end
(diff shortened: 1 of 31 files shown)


blowmage:
The remaining items for #1485 are:

-[ ] Cull idle sessions
-[ ] Change Transaction#keepalive to call BeginTransaction

I'll be opening separate PRs for those items.

blowmage:
Since @vkedia hasn't reviewed yet, I added a couple more commits that should complete issue #1485.

We continue to see some intermittent `GRPC::Aborted: 10:Aborted due to transient fault` errors during the acceptance tests. Not sure why this is yet. Most have been on this branch, but we've had a couple CI builds with the errors as well, so it may not be related to changes in this PR.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -59,31 +61,40 @@ def with_session
...
             loop do
+              fail ClientClosedError if @closed
+
               read_session = session_queue.shift
               return read_session if read_session
               write_transaction = transaction_queue.shift
               return write_transaction.session if write_transaction
 
-              return new_session! if can_allocate_more_sessions?
+              if can_allocate_more_sessions?
```
vkedia: this can cause you to go over limit. Consider if max sessions is 1 and currently there is no session. First thread enters this block and decides it can create a session. While it is creating, second thread can enter this block and also decide it can create one since there still are 0 sessions. So it will also go and create 1 more session. To avoid this you need to keep track of how many sessions are under creation right now and also factor this while calculating if you can allocate more sessions.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -99,52 +110,50 @@ def with_transaction
...
+          end
+          return new_transaction! if action == :new
         end
 
         def checkin_transaction tx
           unless all_sessions.include? tx.session
             fail ArgumentError, "Cannot checkin session"
           end
 
+          tx = tx.session.reload!.create_transaction
```
vkedia: calling create_transaction here defeats the purpose of pre preparing transactions. The reason we want to pre prepare txn is to avoid the latency of BeginTransaction call. But this is just moving that latency here rather than in the beginning of the txn. That is why this needs to be done in a background thread.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -59,31 +61,40 @@ def with_session
...
+
+          return new_session! if action == :new
         end
 
         def checkin_session session
           unless all_sessions.include? session
             fail ArgumentError, "Cannot checkin session"
           end
 
+          session.reload!
```
vkedia: this will make rpcs, right? Why do we need to do this everytime a session is checked in?

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -154,32 +163,91 @@ def reset
...
-          @all_sessions = []
-          @session_queue = []
-          @transaction_queue = []
+          # Unblock all waiting threads
+          @closed = true
+          @resource.broadcast
+          # Delete all sessions
+          @mutex.synchronize do
+            @all_sessions.map do |s|
+              Thread.new do
```
vkedia: if there are 1000 sessions, this will create a 1000 threads. can you instead use some sort of a threadpool?
https://github.com/ruby-concurrency/concurrent-ruby/blob/master/doc/thread_pools.md

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -154,32 +163,91 @@ def reset
...
-          @all_sessions.each(&:delete_session)
-          @all_sessions = []
-          @session_queue = []
-          @transaction_queue = []
+          # Unblock all waiting threads
+          @closed = true
+          @resource.broadcast
+          # Delete all sessions
+          @mutex.synchronize do
+            @all_sessions.map do |s|
```
vkedia: There is a chance of session leak here. What if there are multiple threads creating sessions at this point. Those sessions would not yet be in all_sessions so would never get deleted.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -154,32 +163,91 @@ def reset
...
+            to_keepalive -= to_release
+
+            # Remove those to be released from circulation
+            @all_sessions -= to_release.map(&:session)
+            @session_queue -= to_release
+            @transaction_queue -= to_release
+          end
+
+          to_release.map! { |x| Thread.new { x.release! } }
+          to_keepalive.map! { |x| Thread.new { x.keepalive! } }
```
vkedia: As above, can you use a common threadpool for all these operations.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -154,32 +163,91 @@ def reset
...
 
           true
         end
 
+        def keepalive_or_release!
+          to_keepalive = []
+          to_release = []
+
+          @mutex.synchronize do
+            available_count = session_queue.count + transaction_queue.count
```
vkedia: can you explain the overall logic here? I dont quite follow it.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: We've been doing our best to avoid unnecessary external dependencies. It can add compatibility conflicts for users, as we've experienced in other Google Cloud libraries. But if its a hard requirement then we can't avoid it. I'll give it some thought over the weekend.

vkedia: It will be good to have some sort of bounds on how many threads we use. If you cannot use an external library for this, maybe you can do something like create 10 threads and have each thread close num_sessions/10 sessions sequentially in a loop.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: If there are more idle sessions than the minimum number of sessions configured, then the difference will be released instead of kept alive. So if there are 15 idle sessions (haven't been used in 30 minutes), and the minimum number of sessions is 10, 10 will be kept alive and 5 will be released.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: Added.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: Thread pool added.

google-cloud-spanner/lib/google/cloud/spanner/pool.rb
blowmage: An in process count has been added to track this.

quartzmo:
@vkedia Can you please review the addition of `Concurrent::FixedThreadPool` in this PR?

PR #1506, opened 2017-05-25
Remove memoization of Policy

### LINKED ISSUE #1486: Remove IAM Policy Memoization
Remove the memoization for IAM Policy. Deprecate or remove the `force` argument, and issue a warning when used while deprecated.

refs #1430

quartzmo:
This PR updates `#policy` and `#policy=` methods to remove the memoization of the IAM `Policy`, deprecates the `force` param in `#policy`, and prints a warning if `force` is passed to `#policy`.

Question: Should this PR also deprecate public method `Policy#deep_dup`, or does that method provide value for users?

[closes #1486]

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
@@ -513,11 +513,6 @@ module Google
         # @see https://cloud.google.com/pubsub/docs/reference/rpc/google.iam.v1#iampolicy
         #   google.iam.v1.IAMPolicy
         #
-        # @param [Boolean] force Force the latest policy to be retrieved from
-        #   the Pub/Sub service when `true`. Otherwise the policy will be
-        #   memoized to reduce the number of API calls made to the Pub/Sub
-        #   service. The default is `false`.
-        #
         # @yield [policy] A block for updating the policy. The latest policy
         #   will be read from the Pub/Sub service and passed to the block. After
         #   the block completes, the modified policy will be written to the
@@ -527,23 +522,13 @@ module Google
         #
         # @return [Policy] the current Cloud IAM Policy for this subscription
         #
-        # @example Policy values are memoized to reduce the number of API calls:
+        # @example
         #   require "google/cloud/pubsub"
         #
         #   pubsub = Google::Cloud::Pubsub.new
         #   sub = pubsub.subscription "my-subscription"
         #
-        #   policy = sub.policy # API call
-        #   policy_2 = sub.policy # No API call
-        #
-        # @example Use `force` to retrieve the latest policy from the service:
-        #   require "google/cloud/pubsub"
-        #
-        #   pubsub = Google::Cloud::Pubsub.new
-        #   sub = pubsub.subscription "my-subscription"
-        #
-        #   policy = sub.policy force: true # API call
-        #   policy_2 = sub.policy force: true # API call
+        #   policy = sub.policy
         #
         # @example Update the policy by passing a block:
         #   require "google/cloud/pubsub"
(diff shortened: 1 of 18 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -513,10 +513,9 @@ def seek snapshot
         # @see https://cloud.google.com/pubsub/docs/reference/rpc/google.iam.v1#iampolicy
         #   google.iam.v1.IAMPolicy
         #
-        # @param [Boolean] force Force the latest policy to be retrieved from
-        #   the Pub/Sub service when `true`. Otherwise the policy will be
-        #   memoized to reduce the number of API calls made to the Pub/Sub
-        #   service. The default is `false`.
+        # @param [Boolean] force [Deprecated] Force the latest policy to be
```
blowmage: Since Pub/Sub isn't 1.0 yet, I think we should remove the option. We can still make breaking changes before we hit 1.0.

I would use these deprecation descriptions for Storage though.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -553,19 +542,16 @@ def seek snapshot
...
-          @policy ||= begin
-            ensure_service!
-            grpc = service.get_subscription_policy name
-            Policy.from_grpc grpc
-          end
-          return @policy unless block_given?
-          p = @policy.deep_dup
-          yield p
-          self.policy = p
+          puts "DEPRECATED: 'force' in Subscription#policy" unless force.nil?
```
blowmage: If `force` is removed we don't need this. But for Storage we should use `warn` instead of `puts`.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
quartzmo: Sounds good. Should we also remove `Policy#deep_dup`? It is no longer used internally after the removal of memoization.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
quartzmo: Sounds good, will change to `warn` in >= 1.0 packages.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: Yes, I meant to mention that in the PR as well. I would remove it, but keep it with a `warn` for Storage.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: :+1:

quartzmo:
@blowmage I believe this PR is now ready to merge.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -898,9 +898,9 @@ def default_acl
         # @see https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy
         #   Buckets: setIamPolicy
         #
-        # @param [Boolean] force Force load the latest policy when `true`.
-        #   Otherwise the policy will be memoized to reduce the number of API
-        #   calls made. The default is `false`.
+        # @param [Boolean] force [Deprecated] Force the latest policy to be
+        #   retrieved from the Storage service when `true`. Deprecated because
+        #   the latest policy is now always retrieved. The default is `false`.
```
blowmage: Wouldn't it be:

> The default is `nil`.

quartzmo:
Thanks, I'll fix the doc for the default! Good catch.

PR #1510, opened 2017-05-26
Add Storage Bucket#labels

blowmage:
Adds `Bucket#labels` and `Bucket#labels=`, allows labels to be set on `Project#create_bucket` when using the yielded bucket object.


quartzmo:
Can you add test coverage for `Project#create_bucket`?

Also, I think the `labels` and `labels=` docs are missing `@return` and `@param`.

blowmage:
I added test coverage for `Project#create_bucket` specifying labels.

I didn't add `@return` or `@param` because _all_ the Bucket and File attribute methods are missing them. `File#metadata` is the most direct comparison. It might be a good idea to do a pass of the Storage docs, but I think that should be a separate issue/PR.

PR #1513, opened 2017-05-26
Add Pub/Sub GRPC channel args

### LINKED ISSUE #1467: Received message larger than max: PubSub error and message loop
### Reproduce Steps:

- Publish message > 4MB to a topic
- Retrieve message through subscription: `Error: Received message larger than max (xxxxx vs. 4194304)`

### References

- https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2190
- https://github.com/mia-0032/fluent-plugin-gcloud-pubsub-custom/pull/37

### Impact

I believe gRPC introduced a request limit of 4MB. But **PubSub quotas** clearly says, maximum message size can be 10MB. So many people already utilize this quota as a guidance, and publish messages upto 10MB. [Example](https://github.com/mia-0032/fluent-plugin-gcloud-pubsub-custom/blob/master/lib/fluent/plugin/out_gcloud_pubsub.rb#L27).

image

Because the quota is broken out-of-the-box by PubSub ruby client (which uses gRPC internally in one of the latest releases I think), messages don't get acked, and circulate in the system for 1 week (PubSub has fixed 1 week retention policy). So put all together, the backpressure is really bad. We had to drop all our subscriptions twice now in the course of debugging this error.

blowmage:
Add configuration needed to pull messages larger than 4MB. Resolves #1467.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -41,7 +41,12 @@ module Google
 
         def channel
           require "grpc"
-          GRPC::Core::Channel.new host, nil, chan_creds
+          GRPC::Core::Channel.new host, chan_args, chan_creds
+        end
+
+        def chan_args
+          { "grpc.max_send_message_length"    => -1,
+            "grpc.max_receive_message_length" => -1 }
         end
 
         def chan_creds


blowmage:
> I am assuming we do not want to add an acceptance test for > 4MB due to cost/speed.

Yep. It was tested manually.

PR #1516, opened 2017-05-31
Improved Transaction Behavior

blowmage:
This PR updates the Transaction behavior in the following ways:

- [x] All changes in the transaction block are held in memory and are committed when the block is finished executing.
- [x] Remove `Transaction#commit` and `Transaction#rollback` from public API.
- [x] `Client#transaction` now returns the timestamp of the commit.

These changes were made pair programming with @quartzmo.


google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -78,7 +80,7 @@
   end
 
   it "can rollback a transaction" do
```
vkedia: can you also add a test case where you are doing multiple mutations

google-cloud-spanner/lib/google/cloud/spanner/errors.rb
```diff
@@ -20,6 +20,16 @@ module Cloud
     module Spanner
       ##
       # @private
+      # # CommitError
+      #
+      # Signal that the Transaction has been committed.
```
vkedia: why do we want to allow user to explicitly call commit, if we are automatically doing that at the end of block.

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
```diff
@@ -45,6 +45,10 @@ class Transaction
         # @private The Session object.
         attr_accessor :session
 
+        def initialize
+          @commit = Commit.new
```
vkedia: Where is this Commit class defined?

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
```diff
@@ -302,7 +305,7 @@ def upsert table, *rows
         #
         def insert table, *rows
```
vkedia: can you update the documentation to make it clear that this buffers the mutations in memory.

google-cloud-spanner/lib/google/cloud/spanner/errors.rb
quartzmo: The preferred usage for simple scenarios should be to not call `Transaction#commit`, but just to let the block complete.  My understanding is that `Transaction#commit` should only be called in case of some conditional logic or similar scenario.

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
quartzmo: Yes, definitely.

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
quartzmo: [lib/google/cloud/spanner/commit.rb](https://github.com/blowmage/gcloud-ruby/blob/84a610bbe653a8a805c3584938d90d10ac903d67/google-cloud-spanner/lib/google/cloud/spanner/commit.rb)

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
quartzmo: Yes, we can do something like [this unit test](https://github.com/blowmage/gcloud-ruby/blob/84a610bbe653a8a805c3584938d90d10ac903d67/google-cloud-spanner/test/google/cloud/spanner/client/transaction_test.rb#L134-L153).

blowmage: Added.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -92,6 +94,8 @@
 
       tx.insert "accounts", additional_account
```
blowmage: No, its only there to illustrate that it won't be called (and hence won't raise an error) because the transaction has already been rolled back. I'll update it to raise an error with a more descriptive message.

google-cloud-spanner/lib/google/cloud/spanner/errors.rb
blowmage: Calling `commit` would break out of the transaction block and return immediately. Similar to how calling `rollback` breaks out of the block. Also similar to calling `break` in a loop, or `return` in a method. Here is a use-case I have in mind:

```ruby
db.transaction do |tx|
  rows = tx.execute sql_query
  rows.each do |row|
    modified_row = modify_row row
    tx.upsert modified_row
    tx.commit if threshold_met(modified_row)    
  end

  # since the threshold isn't met, modify more...
  more_rows = tx.execute another_sql_query
  more_rows.each do |row|
    modified_row = modify_row row
    tx.upsert modified_row
  end
end
```

vkedia: I am not sure adding this extra control flow mechanism is a good idea. Is this common practice in Ruby. I would rather people used the standard mechanisms. Is there not some standard mechanism to do this i.e break out of a loop early.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
vkedia: Ah, I had not realized that call rolback breaks out of the block.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
vkedia: Thanks. Can you add a similar test for the normal commit case as well where multiple mutations are done and txn is committed instead of rollback as in this test.

google-cloud-spanner/lib/google/cloud/spanner/errors.rb
quartzmo: What about copying Rails' ActiveRecord Transaction flow control? In ActiveRecord (afaik), unless an exception is thrown, any return (early return, break, returning a falsy value, whatever) will commit the transaction. Only exceptions result in rollback. A special [`ActiveRecord::Rollback`](http://api.rubyonrails.org/classes/ActiveRecord/Rollback.html) error type is provided to users for triggering rollback. It is caught by the framework and does not result in an application error. (See link above for example code.)

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
quartzmo: See my comment below about rollback in ActiveRecord.

google-cloud-spanner/lib/google/cloud/spanner/errors.rb
yugui: I also doubt if the extract control flow mechanism for `commit` is a good idea.
And application developers can use `Kernel#catch` and `Kernel#throw` when they really need to globally escape from a deep call stack.

Even if it is good, again, I recommend `Kernel#throw` for the implementation of the mechanism because of the following consideration.

ActiveRecord consistently rolls back the current transaction on exception. That's why `ActiveRecord::Rollback` is an exception.
On the other hand, this PR is trying to introduce another rule -- sometimes rollback, sometimes commit.
Since you look to need a generic way of global escape but not a consistency to application errors, I recommend `Kernel#throw`, which is a canonical way of global escape in Ruby.

blowmage: Agreed. @quartzmo and I discussed this last night and agreed to remove both `#commit` and `#rollback` from Transaction.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
blowmage: Added.

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
blowmage: Docs have been updated. Please take a look.

blowmage:
This PR has been updated to remove `Transaction#commit` and `Transaction#rollback` to match ActiveRecord's API.

quartzmo:
@blowmage Do we need to document how users can manually rollback the transaction from within a block passed to `Client#transaction`?

vkedia:
+1. I agree we should document thst.

quartzmo:
@vkedia We are also adding the new error for user-controlled rollback, similar to [`ActiveRecord::Rollback`](http://api.rubyonrails.org/classes/ActiveRecord/Rollback.html).

blowmage:
Okay, updated behavior of the Rollback exception and added additional documentation and code examples describing the behavior. PTAL.

Issue #1519, opened 2017-06-06
Cloud Spanner Client should prevent nested transactions


vkedia:
Cloud Spanner does not support [nested transactions](https://en.wikipedia.org/wiki/Nested_transaction). But the client library does allow users to call `database.transaction` inside of the callback that is provided to an outer `database.transaction`. This is misleading since users might believe that this will behave like a nested transaction but in reality these will be two independent transactions. This is confusing and a source of bugs. Specifically the inner transaction might succeed while the outer might ABORT, in which case the callback will be rerun thus rerunning the inner transaction. 
We should prevent this from happening by detecting and raising an error if someone calls `database.transaction` inside the callback provided to `database.transaction`.  Note that this needs to be done before Beta since this is a known breaking change.
Also note that if we come across a use case where we do want to allow calling a nested `database.transaction`, we should be able to enable that in future without making a breaking change. 

cc @jgeewax @bjwatson

blowmage:
How do other languages detect and prevent this?

blowmage:
I expect we can put the transaction_id on the current Thread when a transaction starts, and check for the value when other transactions start, but that may block users from creating and using concurrent transactions with Fibers.

I suppose we could walk the closure's AST and detect if it opens another transaction, but I don't know if that approach is performant enough for production use. This is an interesting problem...

vkedia:
Does Ruby have Fibers? Where can I read up more on them? Ideally as you said we would put something on Thread local but presence of FIbers make this interesting. 
In Go, which has goroutines which are in some sense similar to Fibers, there is a Context object being passed around. So we use that to record the fact that there is already an ongoing transaction.

blowmage:
Yes, Ruby does have a [`Fiber`](https://ruby-doc.org/core-2.0.0/Fiber.html) class.

https://ruby-doc.org/core-2.0.0/Fiber.html

vkedia:
I guess we can use Fiber local variables
https://ruby-doc.org/core-2.2.2/Thread.html#class-Thread-label-Fiber-local+vs.+Thread-local

blowmage:
Yeah, it looks like we get/set values using `Thread.current[]` using either threads or fibers and the behavior will be as expected.

That is, unless users create a nested thread or fiber in the transaction block:

```ruby
client.transaction do |tx|
  tx.insert "users", [{ id: 2, name: "Harvey",  active: true }]

  Thread.new do
[...]

PR #1522, opened 2017-06-06
Prevent nested transactions

blowmage:
This is an attempt to satisfy #1519 by not allowing nested transaction blocks. It does so by setting a value on `Thread.current`. This would allow only one transaction on a thread at a time. Looking for feedback.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -663,6 +663,9 @@ module Google
           end
         end
 
+        # rubocop:disable Metrics/AbcSize
+        # rubocop:disable Metrics/MethodLength
+
         ##
         # Creates a transaction for reads and writes that execute atomically at
         # a single logical point in time across columns, rows, and tables in a
@@ -722,6 +725,9 @@ module Google
         #
         def transaction deadline: 120, &block
           ensure_service!
+          unless Thread.current[:transaction_id].nil?
+            fail "Nested transactions are not allowed"
+          end
 
           deadline = validate_deadline deadline
           backoff = 1.0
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -728,7 +731,11 @@ def transaction deadline: 120, &block
           start_time = current_time
 
           @pool.with_transaction do |tx|
+            unless Thread.current[:transaction_id].nil?
+              fail "Nested transactions are not allowed"
+            end
             begin
+              Thread.current[:transaction_id] = tx.transaction_id
```
quartzmo: Should this assignment be made (and cleared) in a new, outer `begin`/`ensure` block to avoid the gap it appears to leave open on retry?

blowmage: I don't think that is necessary. We only care if a value is set or not. The `begin` block can be retried with a new transaction value, so it is cleaner to have the begin block set the value. When a value is present an error is raised _before_ the begin block begins.

quartzmo: OK, sounds good.

blowmage:
@vkedia Do you want this change to be included in the first release?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
vkedia: This will still allow nested read only transactions, right? Can we prevent that as well?

blowmage:
It currently will allow nested snapshots. I can add the same logic there as well.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
blowmage: It currently will allow nested snapshots. I can add the same logic there as well.

blowmage: The check for snapshots has been added.

quartzmo:
🎉

PR #1530, opened 2017-06-09
Add Pub/Sub Publisher

blowmage:
This PR adds a Publisher object that has the following:

- [x] Asynchronously publish messages in batches
- [x] Will push batches in intervals (every 0.25 seconds)
- [x] Will push batches when message count hits a limit (2000)
- [x] Will push batches when RPC payload reaches a size (10MB)

This functionality is expressed as `Topic#publish_async` and `Project#publish_async`.

This change is to be merged into the `pubsub` branch, which will be used as the target for the upcoming Pub/Sub changes.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -19,6 +19,7 @@ require "google/cloud/pubsub/service"
 require "google/cloud/pubsub/credentials"
 require "google/cloud/pubsub/topic"
 require "google/cloud/pubsub/snapshot"
+require "google/cloud/pubsub/publisher"
 
 module Google
   module Cloud
@@ -217,7 +218,8 @@ module Google
         #
         #   pubsub = Google::Cloud::Pubsub.new
         #
-        #   msg = pubsub.publish "my-topic", File.open("message.txt")
+        #   file = File.open "message.txt", mode: "rb"
+        #   msg = pubsub.publish "my-topic", file
         #
         # @example Additionally, a message can be published with attributes:
         #   require "google/cloud/pubsub"
@@ -249,6 +251,47 @@ module Google
           yield publisher if block_given?
           return nil if publisher.messages.count.zero?
           publish_batch_messages topic_name, publisher
+        end
+
+        ##
+        # Publishes a message to the given topic asynchonously.
+        #
+        # @param [String] topic_name Name of a topic.
+        # @param [String, File] data The message data.
+        # @param [Hash] attributes Optional attributes for the message.
+        # @yield [result] the callback for when the message has been published
+        # @yieldparam [PublishResult] result the result of the asynchonous
+        #   publish
+        #
+        # @example
+        #   require "google/cloud/pubsub"
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #
+        #   pubsub.publish_async "my-topic", "task completed" do |result|
+        #     puts result.msg_id if result.succeeded?
+        #   end
+        #
+        # @example A message can be published using a File object:
+        #   require "google/cloud/pubsub"
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #
+        #   file = File.open "message.txt", mode: "rb"
+        #   pubsub.publish_async "my-topic", file
+        #
+        # @example Additionally, a message can be published with attributes:
+        #   require "google/cloud/pubsub"
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #
+        #   pubsub.publish_async "my-topic",
+        #                        "task completed",
+        #                        foo: :bar, this: :that
+        #
+        def publish_async topic_name, data = nil, attributes = {}, &block
+          ensure_service!
+          service.publish_async topic_name, data, attributes, &block
         end
 
         ##
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/publish_result.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/publish_result.rb
@@ -1,0 +1,89 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+module Google
+  module Cloud
+    module Pubsub
+      ##
+      # # PublishResult
+      #
+      class PublishResult
+        ##
+        # @private Create an PublishResult object.
+        def initialize message, error = nil
+          @message = message
+          @error = error
+        end
+
+        ##
+        # The message.
+        def message
+          @message
+        end
+        alias_method :msg, :message
+
+        ##
+        # The message's data.
+        def data
+          message.data
+        end
+
+        ##
+        # The message's attributes.
+        def attributes
+          message.attributes
+        end
+
+        ##
+        # The ID of the message, assigned by the server at publication
+        # time. Guaranteed to be unique within the topic.
+        def message_id
+          message.message_id
+        end
+        alias_method :msg_id, :message_id
+
+        ##
+        # The error that was raised when published, if any.
+        def error
+          @error
+        end
+
+        ##
+        # Whether the publish request was successful.
+        def succeeded?
+          error.nil?
+        end
+
+        # Whether the publish request failed.
+        def failed?
+          !succeeded?
+        end
+
+        ##
+        # @private Create an PublishResult object from a message protobuf.
+        def self.from_grpc msg_grpc
+          new Message.from_grpc(msg_grpc)
+        end
+
+        ##
+        # @private Create an PublishResult object from a message protobuf and an
+        # error.
+        def self.from_error msg_grpc, error
+          new Message.from_grpc(msg_grpc), error
+        end
+      end
+    end
+  end
+end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
@@ -1,0 +1,202 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+require "monitor"
+require "concurrent"
+require "google/cloud/pubsub/publish_result"
+require "google/cloud/pubsub/service"
+
+module Google
+  module Cloud
+    module Pubsub
+      ##
+      # @private
+      # # Publisher
+      #
+      class Publisher
+        include MonitorMixin
+
+        MAX_BYTES = 5*1024*1024
+        MAX_MESSAGES = 1000
+        TIMER_DELAY = 0.25
+
+        attr_reader :service, :queue
+        attr_reader :max_bytes, :max_messages, :timer_delay
+        attr_reader :thread_pool, :timer_task
+
+        def initialize service, max_bytes: MAX_BYTES,
+                       max_messages: MAX_MESSAGES, timer_delay: TIMER_DELAY
+          @service = service
+          @max_bytes = max_bytes
+          @max_messages = max_messages
+          @timer_delay = timer_delay
+
+          @queue = {}
+          @thread_pool = Concurrent::ThreadPoolExecutor.new
+          @timer_task = Concurrent::TimerTask.new(
+            execution_interval: @timer_delay) { flush }
+
+          # init MonitorMixin
+          super()
+        end
+
+        def publish topic_name, data = nil, attributes = {}, &block
+          topic_name = service.topic_path topic_name
+          msg = create_pubsub_message data, attributes
+
+          synchronize do
+            batch = @queue[topic_name]
+            if batch.nil?
+              # Create a new batch.
+              batch = QueuedBatch.new msg, block
+            elsif !batch.try_add(msg, block, max_messages: max_messages,
+                                             max_bytes: max_bytes)
+              # We hit a threshold, publish existing batch.
+              publish_batch_async topic_name, batch
+              # Create a new batch.
+              batch = QueuedBatch.new msg, block
+            end
+            @queue[topic_name] = batch
+          end
+          nil
+        end
+
+        def start
+          @timer_task.execute
+        end
+
+        def stop
+          flush
+          @timer_task.shutdown
+        end
+
+        def started?
+          @timer_task.running?
+        end
+
+        def stopping?
+          @timer_task.shuttingdown?
+        end
+
+        def stopped?
+          @timer_task.shutdown?
+        end
+
+        def wait timeout = nil
+          @timer_task.wait_for_termination timeout
+        end
+
+        def flush
+          synchronize do
+            @queue.each do |topic_name, batch|
+              publish_batch_async topic_name, batch
+            end
+            @queue = {}
+          end
+        end
+
+        protected
+
+        def create_pubsub_message data, attributes
+          attributes ||= {}
+          if data.is_a?(::Hash) && attributes.empty?
+            attributes = data
+            data = nil
+          end
+
+          # Convert IO-ish objects to strings
+          if data.respond_to?(:read) && data.respond_to?(:rewind)
+            data.rewind
+            data = data.read
+          end
+          # Convert data to encoded byte array to match the protobuf defn
+          data = String(data).force_encoding("ASCII-8BIT")
+
+          # Convert attributes to strings to match the protobuf definition
+          attributes = Hash[attributes.map { |k, v| [String(k), String(v)] }]
+
+          Google::Pubsub::V1::PubsubMessage.new data: data,
+                                                attributes: attributes
+        end
+
+        def publish_batch topic_name, batch
+          grpc = service.publish_messages topic_name, batch.messages
+          batch.items.zip(Array(grpc.message_ids)) do |item, id|
+            item.msg.message_id = id
+            if item.callback
+              item.callback.call PublishResult.from_grpc(item.msg)
+            end
+          end
+        rescue => e
+          batch.items.each do |item|
+            if item.callback
+              item.callback.call PublishResult.from_error(item.msg, e)
+            end
+          end
+        end
+
+        def publish_batch_async topic_name, batch
+          Concurrent::Future.new(executor: @thread_pool) do
+            publish_batch topic_name, batch
+          end.execute
+        end
+
+        class QueuedBatch
+          attr_reader :messages, :callbacks
+
+          def initialize msg, callback
+            @messages = [msg]
+            @callbacks = [callback]
+          end
+
+          def add msg, callback
+            @messages << msg
+            @callbacks << callback
+          end
+
+          def try_add msg, callback,
+                      max_messages: MAX_MESSAGES, max_bytes: MAX_BYTES
+            return false if total_message_count + 1 > max_messages
+            return false if total_message_size + msg.to_proto.size >= max_bytes
+            add msg, callback
+            true
+          end
+
+          def total_message_count
+            @messages.count
+          end
+
+          def total_message_size
+            @messages.map(&:to_proto).map(&:size).inject(0, :+)
+          end
+
+          def items
+            @messages.zip(@callbacks).map do |msg, callback|
+              Item.new msg, callback
+            end
+          end
+
+          class Item
+            attr_accessor :msg, :callback
+            def initialize msg, callback
+              @msg = msg
+              @callback = callback
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -18,6 +18,7 @@ require "google/cloud/pubsub/credentials"
 require "google/cloud/pubsub/version"
 require "google/cloud/pubsub/v1"
 require "google/gax/errors"
+require "google/cloud/pubsub/publisher"
 
 module Google
   module Cloud
@@ -87,6 +88,12 @@ module Google
           credentials == :this_channel_is_insecure
         end
 
+        def async_publisher
+          return mocked_async_publisher if mocked_async_publisher
+          @async_publisher ||= Publisher.new(self).tap(&:start)
+        end
+        attr_accessor :mocked_async_publisher
+
         ##
         # Gets the configuration of a topic.
         # Since the topic only has the name attribute,
@@ -150,10 +157,23 @@ module Google
               data: data, attributes: attributes)
           end
 
+          publish_messages topic, messages
+        end
+
+        ##
+        # Unlike publish, publish_messages takes an array of message grpc
+        # objects instead of an array of data/attribute arrays.
+        def publish_messages topic, messages
           execute do
             publisher.publish topic_path(topic), messages,
                               options: default_options
           end
+        end
+
+        ##
+        # Publish to the async publisher
+        def publish_async topic_name, data, attributes, &block
+          async_publisher.publish topic_name, data, attributes, &block
         end
 
         ##
--- a/google-cloud-pubsub/support/doctest_helper.rb
+++ b/google-cloud-pubsub/support/doctest_helper.rb
@@ -15,7 +15,7 @@
 require "google/cloud/pubsub"
 
 class File
-  def self.open f
+  def self.open *args
     "task completed"
   end
 end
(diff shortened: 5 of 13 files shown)


blowmage:
Does anyone know Thomas Kneeland's github username?

geigerj:
Thomas is @tbkneeland.

blowmage:
@tbkneeland Can you comment on this PR so I can add you as a reviewer?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,189 @@
...
+          msg = create_pubsub_message data, attributes
+
+          synchronize do
+            batch = @queue[topic_name]
+            if batch.nil?
+              # Create a new batch.
+              batch = QueuedBatch.new msg, block
+            elsif !batch.try_add(msg, block)
+              # We hit a threshold, publish existing batch.
+              publish_batch_async topic_name, batch
```
quartzmo: Can any more of this logic be moved into `QueuedBatch`? I feel like this could be a little more OO.

google-cloud-pubsub/support/doctest_helper.rb
```diff
@@ -43,8 +43,9 @@ def mock_pubsub
 
     pubsub.service.mocked_publisher = Minitest::Mock.new
     pubsub.service.mocked_subscriber = Minitest::Mock.new
+    pubsub.service.instance_variable_set :@async_publisher, Minitest::Mock.new
```
quartzmo: Maybe add `Service#mocked_async_publisher` for consistency?

google-cloud-pubsub/support/doctest_helper.rb
```diff
@@ -199,6 +200,19 @@ def mock_pubsub
     end
   end
 
+  doctest.before "Google::Cloud::Pubsub::Project#publish_async" do
+    mock_pubsub do |mock_publisher, mock_subscriber, async_publisher|
```
quartzmo: Should `async_publisher `  be `mock_async_publisher` for consistency? It's a mouthful, but...

blowmage:
> Are there unit tests for the batch limits MAX_COUNT and MAX_SIZE? Might be good.

Agreed. They got lost during the churn of development. I have added them.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -283,6 +284,46 @@ def publish topic_name, data = nil, attributes = {}
...
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #
+        #   pubsub.publish_async "my-topic",
+        #                        "task completed",
+        #                        foo: :bar, this: :that
+        #
+        def publish_async topic_name, data = nil, attributes = {}, &block
+          ensure_service!
+          service.async_publisher.publish topic_name, data, attributes, &block
```
blowmage: Good idea. Added.

google-cloud-pubsub/support/doctest_helper.rb
blowmage: Agreed. Added.

google-cloud-pubsub/support/doctest_helper.rb
blowmage: Changed.

lukesneeringer:
> Will push batches when RPC payload reaches a size (10MB)

Can we change the default on this to 5 MB?

My rationale is: The Pub/Sub _service_ has a 10 MB limit. With the maximum size (unlike messages or latency), it is likely that you will leapfrog the limit slightly. For example, if you have 9.9 MB of messages and then publish a 200 KB message, you now have 10.1 MB. Pub/Sub _will give you an error_ when you try to publish these, and the client library will either error out or (worse) thrash the batch with its retry logic and then error out.

I assert that the right thing for most users is to just have a 5 MB default limit. The user can always change it if they have a reason to want something else.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -283,6 +284,46 @@ def publish topic_name, data = nil, attributes = {}
...
+        #   pubsub.publish_async "my-topic", "task completed" do |result|
+        #     puts result.msg_id if result.succeeded?
+        #   end
+        #
+        # @example A message can be published using a File object:
+        #   require "google/cloud/pubsub"
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #
+        #   pubsub.publish_async "my-topic", File.open("message.txt")
```
lukesneeringer:
```ruby
File.open("message.txt", mode: "rb")
```

blowmage:
We can change the limits, no problem. But the check makes sure that the total message payload size is less than the limit. So if you have a queued batch with 9.9MB of messages, and want to add a 200KB message, it does the following:

* Publish the 9.9MB batch of messages
* Create a new queued batch with the 200KB message

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+            data = nil
+          end
+
+          # Convert IO-ish objects to strings
+          if data.respond_to?(:read) && data.respond_to?(:rewind)
+            data.rewind
+            data = data.read
+          end
+          # Convert data to encoded byte array to match the protobuf defn
+          data = String(data).force_encoding("ASCII-8BIT")
```
lukesneeringer: Summoning @remi. :-) So, disclaimer: I am a Python guy so I have Python biases.

That said, I think this should be an error rather than an implicit conversion. Same for line 128 below.
This is a hard spot, though, because I know that being implicit and Postel's Law are important in Ruby.

Here is why I think this might be important enough to break Ruby convention: Bear in mind that with Pub/Sub, you do not know anything about the client on the other side. They may not (in fact, likely will not) be using this library, or Ruby at all. You can not assume _anything_ about the folks on the consuming side of this. Well, that is not _quite_ true: You can basically assume that the client will base64 decode the message, and nothing else.

I am doing a quick bit of research on string encoding in Ruby (since I do not know much about the ins and outs of this topic in Ruby; note that I _am_ fluent on the topic more generally -- an entire chapter of my book was devoted to this issue), and as best as I can tell, Ruby [basically has the same problems that everyone else has][1]. Basically, strings have an explicit `encoding`, but it might be wrong, especially if you read in from a file.

This merits discussion. But I _think_ that users need to send the right thing, and we should actually complain loudly rather than coerce for them, because the fundamental awareness here is so critical.

  [1]: http://www.justinweiss.com/articles/3-steps-to-fix-encoding-problems-in-ruby/

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+                                                attributes: attributes
+        end
+
+        def publish_batch topic_name, batch
+          grpc = service.publish_messages topic_name, batch.messages
+          batch.items.zip(Array(grpc.message_ids)) do |item, id|
+            item.msg.message_id = id
+            item.block.call PublishResult.from_grpc(item.msg) if item.block
+          end
+        rescue => e
```
lukesneeringer: Ideally it would be nice to detect retryable errors and retry the publish in these cases. (Note that if you were watching the Python implementation at all, I have not yet implemented that, and my code looks...a lot like this actually. :-))

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+          end
+
+          def add msg, block
+            @messages << msg
+            @blocks << block
+          end
+
+          def try_add msg, block, max_count: MAX_COUNT, max_size: MAX_SIZE
+            return false if total_message_count + 1 > max_count
+            return false if total_message_size + msg.to_proto.size >= max_size
```
lukesneeringer: Oh, I see where you came up with the 10 MB limit now. You are actually _failing the add_ when the next message goes over, which (I saw earlier) is what triggers the new batch.

I still think it would be better to have a smaller limit, just in case the client and server do not exactly agree on what the size is (e.g. if `msg.to_proto.size` has a bug due to a character encoding or some such -- it will totally happen). However, I recognize now that you accounted for the case I posited earlier.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+          end
+        end
+
+        def publish_batch_async topic_name, batch
+          Concurrent::Future.new(executor: @thread_pool) do
+            publish_batch topic_name, batch
+          end.execute
+        end
+
+        class QueuedBatch
```
lukesneeringer: I do not see anything here about the batch being committed if enough time elapses. That is probably the single most important of the three limits.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
```diff
@@ -87,6 +88,12 @@ def insecure?
           credentials == :this_channel_is_insecure
         end
 
+        def async_publisher
+          return mocked_async_publisher if mocked_async_publisher
```
lukesneeringer: Huh?

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
```diff
@@ -150,13 +157,26 @@ def publish topic, messages
               data: data, attributes: attributes)
           end
 
+          publish_messages topic, messages
+        end
+
+        ##
+        # Unlike publish, publish_messages takes an array of message grpc
```
lukesneeringer: Why?

lukesneeringer:
> We can change the limits, no problem. But the check makes sure that the total message payload size is less than the limit. So if you have a queued batch with 9.9MB of messages, and want to add a 200KB message, it does the following:
> 
> * Publish the 9.9MB batch of messages
> * Create a new queued batch with the 200KB message

I discovered this when I did a more thorough review just now. I agree that your implementation is not as naive as I assumed, but still believe a lower limit would be wise. I covered this in my review, but essentially my rationale is that I do not really trust the protobuf byte-size check to be sufficiently precise, and see no risk or downside to a 5 MB limit.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
lukesneeringer: Also, if we are going to encode strings for users (see above), we _*definitely*_ should not be using `force_encoding`. We should be using `encode!`.

According to the Ruby documentation, [encode][1] actually transcodes the string to the destination encoding, while [force_encoding][2] simply changes what the string says its encoding is. In other words, if the former is like translating from English to Greek, the latter is like _claiming that_ the text is Greek.

The fact that our own (excellent) developers got this wrong pushes me even further toward wanting to avoid implicit conversion of encodings, though. Our users _will_ send data and wonder why they are getting something else on the other end.

  [1]: https://ruby-doc.org/core-2.4.0/String.html#method-i-encode-21
  [2]: https://ruby-doc.org/core-2.4.0/String.html#method-i-force_encoding

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: Changed.

google-cloud-pubsub/lib/google/cloud/pubsub/publish_result.rb
```diff
@@ -0,0 +1,89 @@
...
+# limitations under the License.
+
+
+module Google
+  module Cloud
+    module Pubsub
+      ##
+      # # PublishResult
+      #
+      class PublishResult
```
blowmage: We feel the encapsulation has value and is what we have done everywhere. It protects users from changes to the Low Level API, which has experienced more changes than you may expect.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+module Google
+  module Cloud
+    module Pubsub
+      ##
+      # # Publisher
+      #
+      class Publisher
+        include MonitorMixin
+
+        MAX_COUNT = 2000
```
blowmage: No reason. Changed.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+        MAX_SIZE = 10000000
```
blowmage: Changed.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+
```
blowmage: It was a magic value. Added the `TIMER_DELAY` constant for it.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+        attr_reader :service, :queue
+        attr_reader :max_size, :max_count
+        attr_reader :thread_pool, :publish_task
+
+        def initialize service, max_size: MAX_SIZE, max_count: MAX_COUNT
```
blowmage: Changed.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+              publish_batch_async topic_name, batch
+              # Create a new batch.
+              batch = QueuedBatch.new msg, block
+            end
+            @queue[topic_name] = batch
+          end
+          nil
+        end
+
+        def start
```
blowmage: FWIW, I don't understand why `#start` and `#stop` are needed either. We only added these methods because they were in the spec. It does add some confusion though, since there are two constructs that can be stopped, the `TimerTask`, and the `ThreadPool`. I'm fine with removing these methods if that is allowed per the spec.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+            end
+            @queue = {}
+          end
+        end
+
+        protected
+
+        def create_pubsub_message data, attributes
+          attributes ||= {}
+          if data.is_a?(::Hash) && attributes.empty?
```
blowmage: Right now users are allowed to call the `publish` method without providing a `data` argument. If users are allowed to _not_ pass in a message `data`, then this check is not needed. I'm fine making the `data` argument required if we can get some confirmation on this.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: FWIW, we understand the difference between `#force_encoding` and `#encode!`, and we chose `#force_encoding` to try to preserve the data that the user presents. Before we make any changes to how the method encodes the data, can we get consensus on whether we should continue to be liberal in allowing non-binary data and encoding it to binary, or should we be conservative and force users to encode the data before calling `#publish`?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: The GAPIC layer already provides retry, which is why we aren't handling it explicitly.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: That is handled by the `TimerTask` object in `Publisher`, and not in `QueuedBatch`.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: Changed to 5MB.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
blowmage: This is a hook used by our tests. This object is not part of the public API exposed to the users.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
blowmage: Publisher breaks an existing design layer in the code because it needs to calculate the size of the messages over the wire (`msg.to_proto.size`). Because of this, we needed to add a separate publish method to our internal service layer. Its possible this implementation detail is changed in a future refactoring, since this object is not part of the public API exposed to the users.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: It occurs to me that another reason we chose `force_encoding` was to match the behavior of the library pre-GRPC. When using the REST API the `data` was Base64 encoded, and that encoding/decoding was more permissive and used `force_encoding` as opposed to `encode!`.

But again, not opposed to changing this behavior if the requirements specify something else.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: At standup we agreed to create a separate issue to discuss whether a message can be published without `data`. So let's move this discussion to #1535.

Any changes made to `data` being required will be made in a separate PR. So please review this PR with that in mind.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: At standup we agreed to create a separate issue to discuss the encoding decisions. So let's move this discussion to #1536.

Any changes made for the encoding issue will be made in a separate PR. So please review this PR with that in mind.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
lukesneeringer: Waiting on @remi for a confirmation, but it is fine by me.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
lukesneeringer: I am fine with a separate tracking issue and will not block on that question. Thanks!

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
lukesneeringer: Got it. Thanks!

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
lukesneeringer: I must have missed it. Thanks. (I will re-read later but I have no real concerns.)

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
lukesneeringer: That makes sense. Thank you!

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
lukesneeringer: Moving discussion to #1536. Thanks so much! :-)

google-cloud-pubsub/lib/google/cloud/pubsub/publish_result.rb
lukesneeringer: That is kind of my concern. After all, it also "protects" users from _additions_, which on other APIs has been an issue.

That said, Pub/Sub is super stable by comparison and it is hard to see things being added here, so no need to block on this. I am not sure it is a wise pattern (in other languages it has caused more problems than it solved in my assessment), but I agree that it is widely used in our libraries and this is not a problem case.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,195 @@
...
+          )
+          @publish_task = Concurrent::TimerTask.new(execution_interval: 0.25) do
+            flush
+          end
+
+          # init MonitorMixin
+          super()
+        end
+
+        def publish topic_name, data = nil, attributes = {}, &block
```
lukesneeringer: Superceded by #1536.

blowmage:
I would like to merge this so we can move on to the Subscriber. We have a couple open issues for the questions raised in so far in this PR's review process, and we will address them in future PRs. Can I get some approvals for this PR?

blowmage:
@lukesneeringer @remi Do you have any other concerns about this PR? (other than the ones that have separate issues created for them?) If not, can you please approve?

@tbkneeland We haven't heard from you on this at all yet. Can you please look this over?

tbkneeland:
Apologies for the extensive delays in getting to this. No-op comment so I can be added as a reviewer.

blowmage:
@tbkneeland Thanks! I can't add you yet. Can you start a review?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,205 @@
...
+        rescue => e
+          batch.items.each do |item|
+            if item.callback
+              item.callback.call PublishResult.from_error(item.msg, e)
+            end
+          end
+        end
+
+        def publish_batch_async topic_name, batch
+          Concurrent::Future.new(executor: @thread_pool) do
```
tbkneeland: You probably want an unbounded threadpool here -- each publish will consume an entire thread for the duration. gRPC's Ruby wrapper gives up the GVL before blocking, so we can make forward progress, as long as we have threads on which to perform work.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,205 @@
...
+                       max_messages: MAX_MESSAGES, timer_delay: TIMER_DELAY
+          @service = service
+          @max_bytes = max_bytes
+          @max_messages = max_messages
+          @timer_delay = timer_delay
+
+          @queue = {}
+          @thread_pool = Concurrent::FixedThreadPool.new(
+            [2, Concurrent.processor_count * 2].max,
+            fallback_policy: :caller_runs
```
tbkneeland: This `fallback_policy` seems dangerous -- it means that a function that advertises itself as async can occasionally become blocking.

Why not leave `max_queue` unbounded or set `fallback_policy` to `:abort`?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,205 @@
...
+
+          # Convert attributes to strings to match the protobuf definition
+          attributes = Hash[attributes.map { |k, v| [String(k), String(v)] }]
+
+          Google::Pubsub::V1::PubsubMessage.new data: data,
+                                                attributes: attributes
+        end
+
+        def publish_batch topic_name, batch
+          grpc = service.publish_messages topic_name, batch.messages
```
tbkneeland: Does `service.publish_messages` include retry logic, or does that need to be added here?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -0,0 +1,205 @@
...
+      ##
+      # @private
+      # # Publisher
+      #
+      class Publisher
+        include MonitorMixin
+
+        MAX_BYTES = 5*1024*1024
+        MAX_MESSAGES = 1000
+        TIMER_DELAY = 0.25
```
tbkneeland: The batching logic is conflating two different durations here.

- The maximum duration the user wants us to buffer a batch while waiting for more messages. This may be arbitrarily long, and is set by the user.
- How frequently we check to see if any duration-based limits have elapsed; this is an implementation detail ("Often enough that no one notices" -- instead of polling with a fixed cadence, you might consider adjusting the execution interval to the next potential expiration). 

By extension: We should not be publishing all batches at the same time. 

More precisely, a batch of messages will be published to a given topic when any of the following conditions is satisfied:

- Number of messages exceeds max_messages
- Number of bytes exceeds max_bytes
- The time elapsed since the first publish call in this batch exceeds max_delay
- User calls publisher.stop/flush/shutdown

blowmage: The spec states:

> a Publisher is instantiated which runs on its own Thread publishing enqueued message on a fast cadence (eg. every 0.25 seconds)

To me that indicated that the Publisher manages the duration that messages are published, not the batches of messages. Should this be discussed on the specification document?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: AFAIK we can't set `max_queue` on a FixedThreadPool. The spec documentation seems to be implying that the thread pool is a fixed size.

Personally I'd favor using a CachedThreadPool without specifying the number of threads, or a ThreadPoolExecutor with a min and a max number of threads (and where we can set `max_queue`).

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: The retry logic is part of the GAPIC layer, which is called within `service.publish_messages`.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: See my other comment. Is this something we should discuss on the specification document?

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: > or set `fallback_policy` to `:abort`?

I will make this change. I believe `:abort` is the default, actually. It was useful to fallback to `:caller_runs` during development, but these PRs are stable enough now that this should have been changed.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: Oh, my mistake. The Subscriber has configuration that indicates a fixed thread pool. The Publisher doesn't indicate anything about the thread pool. I'll change the implementation to be unbounded.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: The thread pool has been changed to a `Concurrent::ThreadPoolExecutor`, which is unbounded by default.

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
blowmage: The thread pool has been changed to a `Concurrent::ThreadPoolExecutor`, which is unbounded by default.

blowmage:
Closing as we decided to go with #1561 instead.

PR #1532, opened 2017-06-09
Remove Pub/Sub autocreate

blowmage:
This PR removes the `autocreate` feature in Pub/Sub. This feature was implemented early on in the initial design of the Pub/Sub library, along with `autoack`. Since `autoack` is being removed, I think it makes sense to reconsider `autocreate` as well.

The `autocreate` option simply allows for the creation of a topic when pulling it. So, instead of this:

```ruby
require "google/cloud/pubsub"

topic = pubsub.topic("existing-topic") ||
        pubsub.create_topic("my-topic")
```

Users could instead do this:

```ruby
require "google/cloud/pubsub"

topic = pubsub.topic "existing-topic", autocreate: true
```

But this feature has been problematic at times, especially when attempted to be combined with the `skip_lookup` option. Since we are about to break backwards compatibility by removing `autoack` and adding new features, I think it might be wise to remove `autocreate` as well. Thoughts?

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -83,13 +83,7 @@ module Google
         ##
         # Retrieves topic by name.
         #
-        # The topic will be created if the topic does not exist and the
-        # `autocreate` option is set to true.
-        #
         # @param [String] topic_name Name of a topic.
-        # @param [Boolean] autocreate Flag to control whether the requested
-        #   topic will be created if it does not exist. Ignored if `skip_lookup`
-        #   is `true`. The default value is `false`.
         # @param [String] project If the topic belongs to a project other than
         #   the one currently connected to, the alternate project ID can be
         #   specified here.
@@ -99,9 +93,7 @@ module Google
         #   does not exist. Default is `false`.
         #
         # @return [Google::Cloud::Pubsub::Topic, nil] Returns `nil` if topic
-        #   does not exist. Will return a newly created{
-        #   Google::Cloud::Pubsub::Topic} if the topic does not exist and
-        #   `autocreate` is set to `true`.
+        #   does not exist.
         #
         # @example
         #   require "google/cloud/pubsub"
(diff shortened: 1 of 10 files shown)


blowmage:
@tbkneeland.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -133,14 +119,13 @@ def self.default_project
...
         #   topic = pubsub.topic "another-topic", skip_lookup: true
         #
-        def topic topic_name, autocreate: nil, project: nil, skip_lookup: nil
+        def topic topic_name, project: nil, skip_lookup: nil
           ensure_service!
           options = { project: project }
           return Topic.new_lazy(topic_name, service, options) if skip_lookup
           grpc = service.get_topic topic_name
           Topic.from_grpc grpc, service
         rescue Google::Cloud::NotFoundError
```
lukesneeringer: This now seems to amount to silently swallowing the exception. `NotFoundError` should be actually raised here, yes?

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -331,32 +296,14 @@ def publish topic_name, data = nil, attributes = {}
...
-                      retain_acked: false, retention: nil, endpoint: nil,
-                      autocreate: nil
+                      retain_acked: false, retention: nil, endpoint: nil
           ensure_service!
           options = { deadline: deadline, retain_acked: retain_acked,
                       retention: retention, endpoint: endpoint }
           grpc = service.create_subscription topic_name,
                                              subscription_name, options
           Subscription.from_grpc grpc, service
-        rescue Google::Cloud::NotFoundError => e
```
lukesneeringer: Here you seem to have removed the `rescue` block completely, which is what I expected above.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: The documentation states that if a Topic resource does not exist `nil` is returned.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: The documentation for this method does not state that `nil` will be returned when the resource does not exist. This is because `NOT_FOUND` can be raised for a missing Topic resource or a missing Subscription resource. The users have to check the error to see for sure.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: Returning `nil` for a missing resource instead of raising is a common idiom in most Ruby libraries. Most other Google Cloud libraries behave similarly in their lookup methods.

lukesneeringer: Okay, and you state below that this is idiomatic. Sold.

PR #1533, opened 2017-06-09
Add Dataset#load to BigQuery

### LINKED ISSUE #1528: Create, configure and load table in one call
Regarding to issue https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/939 I'm opening this issue to allow the creation of `Dataset#load`.

@blowmage as you said me in this comment https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/939#issuecomment-284005734 

> I can see an argument for a Dataset#load method that will allow the creation of a table, configuring of a schema, and loading of data in a single BigQuery API call. If that is what is wanted I would respectfully ask that a new issue be opened for that.

I would like to know if this operation can be supported, we have this use case: we create the table or truncate it with the attached schema and loading data in only one call. I think this can be helpful for more people, you can do it with `bq` command, why not with ruby? In addition, I'd like to get rid of my fork and use the official gem.

quartzmo:
Support creating, configuring and loading a table in one API call.

[closes #1528]

### DIFF
--- a/google-cloud-bigquery/.rubocop.yml
+++ b/google-cloud-bigquery/.rubocop.yml
@@ -30,7 +30,7 @@ Metrics/PerceivedComplexity:
 Metrics/AbcSize:
   Max: 25
 Metrics/MethodLength:
-  Max: 20
+  Max: 25
 Metrics/ParameterLists:
   Enabled: false
 Style/RescueModifier:
--- a/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
@@ -41,6 +41,7 @@ describe Google::Cloud::Bigquery::Dataset, :bigquery do
     end
     t
   end
+  let(:local_file) { "acceptance/data/kitten-test-data.json" }
 
   before do
     table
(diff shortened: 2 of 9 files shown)


quartzmo:
@AlfonsoUceda Please take a look, and let me know if you have any feedback or suggestions, thanks!

AlfonsoUceda:
@quartzmo great PR thanks!, just a question, is there a change to pass the schema as an option instead a block? for example the following snippet:
```ruby
bigquery.dataset("my_dataset").load "my_table", file, write: "truncate", schema: schema
```

We create schema objects `#from_gapi(nil)` because then we inject them in the `table#load` using our fork. I know `#from_gapi` method is private but is there a way to create schema objects from a public interface? have you though in that?

blowmage:
> is there a change to pass the schema as an option instead a block?

No. We don't allow a schema object to be defined outside of the block syntax. For instance, there is no `schema` argument to `Dataset#create_table`.

Calling `Schema#from_gapi(nil)` is the private API and subject to change. You should not use it. If you want to be able to create a schema object outside of the block syntax for creating a table then that should be requested separately as it would affect more methods than `Dataset#load`.

quartzmo:
@AlfonsoUceda Can you explain why you prefer to create the schema object outside of a block and pass it as an argument? This will help us justify adding a schema option as well as a (documented) way to create a new schema object.

AlfonsoUceda:
@quartzmo we have our custom reporter component, I mean an ETL process. We have schema separated by files, every schema its class, so we can inject them in our ETL class, every ETL has a file and we define a extractor, transformer and loader (csv). When the CSV file is created then with this file and a generated schema object we inject into our bigquery loaders class to create or truncate the table.

```ruby
bigquery_loader.load(file, schema: schema)

# inside `#load`
table.load file, write: 'truncate', schema: schema # this is because of our fork
```

This is because we can't test our schema in a separate test or because we share schemas by ETL.

quartzmo:
@AlfonsoUceda Thank you. I will merge this PR now, and create a new issue for the schema option and schema creation.

AlfonsoUceda:
@quartzmo thank you for your work and considering the schema option.

quartzmo:
@AlfonsoUceda No problem! Have patience, we'll get you off your fork soon. ;)

PR #1542, opened 2017-06-14
Add Pub/Sub Subcriber

blowmage:
This PR adds a Subscriber object that has the following:

- [x] Uses StreamingPull to pull, acknowledge, and delay received messages
- [x] Calls a user-defined callback that can be used to acknowledge or delay received messages
- [x] Uses a ThreadPool to call callbacks, so streaming requests and responses are handled in a non-blocking manner
- [x] Maintain lease on received messages until they are asked or nacked by users.

This functionality is expressed as `Subscription#listen`.

This change is to be merged into the pubsub branch, which will be used as the target for the upcoming Pub/Sub changes.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
@@ -34,16 +34,19 @@ module Google
 
         ##
         # Subscriber attributes.
-        attr_reader :callback, :subscription_name, :deadline, :threads
+        attr_reader :callback, :subscription_name, :deadline, :inventory,
+                    :threads
 
         ##
         # @private Create an empty {Subscriber} object.
-        def initialize callback, subscription_name, deadline, threads, service
+        def initialize callback, subscription_name, deadline, inventory,
+                       threads, service
           @request_queue = nil
           @output_enum = nil
           @callback = callback
           @subscription_name = subscription_name
-          @deadline = deadline
+          @deadline = deadline || 60
+          @inventory = Inventory.new self, (inventory || 100)
           @threads = threads || [2, Concurrent.processor_count * 2].max
           @service = service
 
@@ -66,20 +69,28 @@ module Google
           synchronize do
             return if @request_queue.nil?
             @request_queue.push self
+            @inventory.stop
+            @stopped = true
           end
 
           true
         end
 
+        def stopped?
+          synchronize { @stopped }
+        end
+
+        def paused?
+          synchronize { @paused }
+        end
+
         def wait!
           synchronize do
-            return if @background_thread.nil?
-            @background_thread.join
-          end
+            @background_thread.join if @background_thread
 
-          return true if @thread_pool.shutdown?
-          @thread_pool.shutdown
-          @thread_pool.wait_for_termination
+            @thread_pool.shutdown
+            @thread_pool.wait_for_termination
+          end
 
           true
         end
@@ -95,6 +106,8 @@ module Google
 
           synchronize do
             @request_queue.push ack_request
+            @inventory.remove ack_ids
+            unpause_streaming!
           end
 
           true
@@ -113,14 +126,47 @@ module Google
 
           synchronize do
             @request_queue.push deadline_ack_request
+            @inventory.remove deadline_ack_ids
+            unpause_streaming!
           end
 
           true
         end
 
+        def inventory
+          synchronize { @inventory.count }
+        end
+
+        ##
+        # @private
+        def delay_inventory!
+          synchronize do
+            return true if @inventory.empty?
+
+            inv_mod_ack_deadline = @inventory.ack_ids.map { deadline }
+            inv_mod_ack_request = Google::Pubsub::V1::StreamingPullRequest.new
+            inv_mod_ack_request.modify_deadline_ack_ids += @inventory.ack_ids
+            inv_mod_ack_request.modify_deadline_seconds += inv_mod_ack_deadline
+
+            @request_queue.push inv_mod_ack_request
+          end
+
+          true
+        end
+
+        ##
+        # @private
+        def clear_inventory!
+          synchronize do
+            @inventory.clear
+            unpause_streaming!
+          end
+        end
+
         # @private
         def to_s
-          format "(subscription: %s)", subscription_name
+          format "(subscription: %s, inventory: %i)", subscription_name,
+                 inventory
         end
 
         # @private
@@ -130,29 +176,45 @@ module Google
 
         protected
 
+        # rubocop:disable all
+
         def background_run
-          @output_enum.each do |response|
-            response.received_messages.each do |rec_msg_grpc|
-              perform_callback_async @callback, rec_msg_grpc
+          until synchronize { @stopped }
+            Thread.current.kill if synchronize { @paused }
+
+            begin
+              # Cannot syncronize the enumerator, causes deadlock
+              response = @output_enum.next
+              response.received_messages.each do |rec_msg_grpc|
+                rec_msg = ReceivedMessage.from_grpc(rec_msg_grpc, self)
+                synchronize do
+                  @inventory.add rec_msg.ack_id
+
+                  perform_callback_async rec_msg
+                end
+              end
+              synchronize { pause_streaming! }
+            rescue StopIteration
+              break
             end
           end
         rescue GRPC::DeadlineExceeded
           # The GRPC client will raise when stream is opened longer than the
           # timeout value it is configured for. When this happends, restart the
           # stream stealthly.
-          synchronize do
-            start_streaming!
-          end
         rescue => e
-          @request_queue.unshift Google::Cloud::Error.from_error(e)
-          Thread.current.kill
+          synchronize do
+            @request_queue.unshift Google::Cloud::Error.from_error(e)
+          end
         ensure
           Thread.pass
         end
 
-        def perform_callback_async callback, rec_msg_grpc
+        # rubocop:enable all
+
+        def perform_callback_async rec_msg
           Concurrent::Future.new(executor: @thread_pool) do
-            callback.call ReceivedMessage.from_grpc(rec_msg_grpc, self)
+            @callback.call rec_msg
           end.execute
         end
 
@@ -160,7 +222,28 @@ module Google
           @request_queue.unshift initial_input_request
           @output_enum = service.streaming_pull @request_queue.each_item
 
+          @background_thread.kill if @background_thread
           @background_thread = Thread.new { background_run }
+          @stopped = nil
+        end
+
+        def pause_streaming!
+          return if @paused
+
+          @paused = true if inventory_full?
+        end
+
+        def unpause_streaming!
+          return if @paused.nil? || inventory_full?
+
+          @paused = nil
+          # Need to recreate the enums otherwise we get the error:
+          # "fiber called across threads"
+          start_streaming!
+        end
+
+        def inventory_full?
+          @inventory.full?
         end
 
         def initial_input_request
@@ -180,6 +263,92 @@ module Google
         end
 
         # @private
+        class Inventory
+          include MonitorMixin
+
+          attr_reader :subscriber, :limit
+
+          def initialize subscriber, limit
+            @subscriber = subscriber
+            @limit = limit
+            @_ack_ids = []
+            @wait_cond = new_cond
+
+            super()
+          end
+
+          def ack_ids
+            @_ack_ids
+          end
+
+          def add *ack_ids
+            ack_ids = Array(ack_ids).flatten
+            synchronize do
+              @_ack_ids += ack_ids
+              @start_time ||= Time.now
+              @background_thread ||= Thread.new { background_run }
+            end
+          end
+
+          def remove *ack_ids
+            ack_ids = Array(ack_ids).flatten
+            synchronize do
+              @_ack_ids -= ack_ids
+              clear if @_ack_ids.empty?
+            end
+          end
+
+          def clear
+            synchronize do
+              @_ack_ids = []
+              if @background_thread
+                @background_thread.kill
+                @background_thread = nil
+              end
+              @start_time = nil
+            end
+          end
+
+          def count
+            synchronize do
+              @_ack_ids.count
+            end
+          end
+
+          def empty?
+            synchronize do
+              @_ack_ids.empty?
+            end
+          end
+
+          def stop
+            synchronize do
+              @stopped = true
+            end
+          end
+
+          def full?
+            count >= limit
+          end
+
+          protected
+
+          def background_run
+            synchronize do
+              until @stopped
+                @wait_cond.wait calc_delay
+
+                @subscriber.delay_inventory!
+              end
+            end
+          end
+
+          def calc_delay
+            (@subscriber.deadline - 3) * rand(0.8..0.9)
+          end
+        end
+
+        # @private
         class EnumeratorQueue
           extend Forwardable
           def_delegators :@q, :push
@@ -193,7 +362,10 @@ module Google
           def unshift request
             new_queue = Queue.new
             new_queue.push request
-            new_queue.push @q.pop while @q.size > 0
+            while @q.size > 0
+              r = @q.pop
+              new_queue.push r unless r.equal? @sentinel
+            end
             @q = new_queue
           end
 
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
@@ -15,9 +15,7 @@
 
 require "google/cloud/pubsub/service"
 require "google/cloud/pubsub/subscriber/stream"
-require "google/cloud/pubsub/subscriber/enumerator_queue"
 require "monitor"
-require "concurrent"
 
 module Google
   module Cloud
@@ -46,17 +44,13 @@ module Google
           @deadline = deadline || 60
           @streams = streams || 1
           @inventory = inventory || 100
-          @threads = threads || [2, Concurrent.processor_count * 2].max
-          @thread_pool = Concurrent::FixedThreadPool.new @threads
+          @threads = threads || 4
           @service = service
 
-          @stream_pool = @streams.times.map do
-            Thread.new do
-              Thread.current[:stream] = Stream.new self
-            end
+          stream_pool = @streams.times.map do
+            Thread.new { Stream.new self }
           end
-          @stream_pool.map!(&:join)
-          @stream_pool.map! { |t| t[:stream] }
+          @stream_pool = stream_pool.map(&:value)
 
           super() # to init MonitorMixin
         end
@@ -90,11 +84,6 @@ module Google
             end
           end
           wait_pool.join
-
-          synchronize do
-            @thread_pool.shutdown
-            @thread_pool.wait_for_termination
-          end
 
           self
         end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
@@ -14,9 +14,9 @@
 
 
 require "google/cloud/pubsub/service"
-require "google/cloud/errors"
+require "google/cloud/pubsub/subscriber/stream"
+require "google/cloud/pubsub/subscriber/enumerator_queue"
 require "monitor"
-require "forwardable"
 require "concurrent"
 
 module Google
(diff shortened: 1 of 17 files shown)


blowmage:
This PR is independent of #1530, but there will be some merge conflicts that will have to be fixed once either this or #1530 are merged.

remi:
`subscriber.wait!` is exiting for me immediately.  Trying to use `#listen`.  Am I doing something wrong || can you provide a working snippet for me to test with?

```ruby
$stdout.sync = true

require "google/cloud/pubsub"

pubsub       = Google::Cloud::Pubsub.new
topic        = pubsub.topic "subscriber-test"
subscription = topic.subscription "test-listen"
subscriber   = subscription.listen do |message|
  puts "[#{Thread.current.object_id}] message #{message.message_id} => #{messsage.data.inspect}"
end

subscriber.start
puts "Listening..."
subscriber.wait!
``` 

^--- exits immediately.

Also, #listen's callback isn't ever triggering for me if I call `subscriber.start` playing in IRB, incase that should work.

blowmage:
@remi, you need to call `#stop` before calling `#wait`

blowmage:
.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,217 @@
...
+              threads,
+              fallback_policy: :caller_runs
+            )
+            start_streaming!
+          end
+
+          true
+        end
+
+        def stop
```
remi: I get this error if I #stop from a SIGINT trap:

```ruby
# ...
trap "INT" do
  puts "Exiting..."
  subscription.stop
  subscription.wait!
  exit
end
# ...
```

```
be ruby subscriber_test.rb 
Listening...
^CExiting...
/Users/remily/.rbenv/versions/2.3.1/lib/ruby/2.3.0/monitor.rb:187:in `lock': can't be called from trap context (ThreadError)
	from /Users/remily/.rbenv/versions/2.3.1/lib/ruby/2.3.0/monitor.rb:187:in `mon_enter'
	from /Users/remily/.rbenv/versions/2.3.1/lib/ruby/2.3.0/monitor.rb:212:in `mon_synchronize'
	from /Users/remi/Desktop/pubsub_review/gcloud-ruby/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb:69:in `stop'
	from subscriber_test.rb:19:in `block in <main>'
	from subscriber_test.rb:25:in `sleep'
	from subscriber_test.rb:25:in `block in <main>'
	from subscriber_test.rb:24:in `loop'
	from subscriber_test.rb:24:in `<main>'
```

I expect stopping from a trapped signal to be somewhat common (it's how I frequently use Pub/Sub, personally).

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,217 @@
...
+          synchronize do
+            return if @request_queue.nil?
+            @request_queue.push self
+          end
+
+          true
+        end
+
+        def wait!
```
remi: #idea Should all of the behavior of #wait! be part of #stop? And remove #wait!

So far, I *always* do this in my code using Subscriber:

```ruby
subscriber.stop
subscriber.wait!
```

I do not understand the use-case where I would only call stop but not wait, as wait is currently implemented.

FWIW When I tried this for the first time, I guessed that `wait!` would block my script indefinitely while the pull ran (*until #stop was called, eg. from capturing a signal*).

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: Hmmm, I don't know how we can reliably stop without being able to call `lock`...

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: We haven't committed time to the documentation of these methods since we aren't sure they are correct. But I will do my best to add some dev document them here:

Subscriber#stop only halts the pulling of new messages. It does not ensure that all messages have been processed. It updates the state of the Subscriber object and will return quickly.

Subscriber#wait! will block until all the current messages in the queue are processed. Because it blocks it has the bang to indicate that it could take a while. This is similar to BigQuery's `Job#wait_until_done!`, Spanner's `Job#wait_until_done!`, Speech's `Operation#wait_until_done!`, Speech's `Stream#wait_until_complete!`, and DNS's `Change#wait_until_done!`. In all of these other cases these object will complete eventually, but in Pub/Sub the Subscription is intended to operate open ended. There is no logical end for how long it will run, it will only be stopped eventually.

If it were me, I'd name `wait!` something like `wait_until_empty!`.

blowmage: Another analogy I'd draw is concurrent-ruby's API. In it you stop using [`#shutdown`](https://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html#shutdown-instance_method) and you wait using [`#wait_for_termination`](https://ruby-concurrency.github.io/concurrent-ruby/Concurrent/ThreadPoolExecutor.html#wait_for_termination-instance_method). Shutting down a thread pool means new jobs can't be added to it, but it doesn't mean that all existing jobs are completed.

(Then analog for start would be [`#execute`](https://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Future.html#execute-instance_method), although that call isn't present on thread pool objects.)

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: So it appears that any code that can potentially cause a deadlock isn't allowed in a trap. How about something like this instead?

```ruby
begin
  puts "Starting..."
  subscription.start
  sleep
rescue Interrupt
  puts "Received SIGINT..."
ensure
  puts "Stopping..."
  subscription.stop
  subscription.wait!
end
```

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
lukesneeringer: > `Subscriber#stop` only halts the pulling of new messages. It does not ensure that all messages have been processed. It updates the state of the Subscriber object and will return quickly.

This sounds correct to me.

> `Subscriber#wait!` will block until all the current messages in the queue are processed. Because it blocks it has the bang to indicate that it could take a while. This is similar to BigQuery's `Job#wait_until_done!`, Spanner's `Job#wait_until_done!`, Speech's `Operation#wait_until_done!`, Speech's `Stream#wait_until_complete!`, and DNS's `Change#wait_until_done!`. In all of these other cases these object will complete eventually, but in Pub/Sub the Subscription is intended to operate open ended. There is no logical end for how long it will run, it will only be stopped eventually.
> 
> If it were me, I'd name `wait!` something like `wait_until_empty!`.

There is also a use case where I _want_ both to keep receiving messages and the thread to block indefinitely while messages are coming in. (Essentially, if I was writing a process that I intended to run as a daemon.) I would expect that the way to do this is to call `#wait!` but _not_ `#stop`. Is that what happens here?

I am not picky on the name of the method (@remi gets the final call on that) but I would hope that the `#wait!` method does not terminate if the subscription is still running, even if the queue is empty.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,214 @@
...
+          synchronize do
+            @request_queue.push ack_request
+          end
+
+          true
+        end
+
+        ##
+        # @private
+        def delay new_deadline, *messages
```
lukesneeringer: I think I would prefer this be named `modify_ack_deadline` to match the API (and the method name in other languages). It actually took me a moment to figure out what this method did.

I would also note that the method is _not_ a delay. You can call this method to make the deadline _shorter_. For example, if you call:

```ruby
delay 20 message
delay 1 message
```

The new deadline will be 1 second from when the latter call went in, not 21 seconds.
(Gotcha: Unless those calls end up getting processed out of order, but I am not too worried about this as this is a private method and calling it in such a way is an obvious anti-pattern.)

The verb `delay` sounds to me like "add N seconds to this timeout", when the behavior is actually "set the timeout to N seconds".

google-cloud-pubsub/support/doctest_helper.rb
```diff
@@ -562,10 +562,10 @@ def rec_message_json message, id = rand(1000000)
   }.to_json
 end
 
-def rec_messages_json message
+def rec_messages_json message, id = nil
```
lukesneeringer: Do we receive messages in JSON in GRPC?

blowmage: No, Pub/Sub is not using JSON. It is using GRPC and all data is received using the binary encoded protobuf objects.

These methods are here for convenience in instantiating the protobuf objects used in the documentation's code example tests.

blowmage:
@tbkneeland @remi @lukesneeringer This PR has been updated to add inventory management to the Subscriber. PTAL.

blowmage:
The inventory management additions are written assuming that received messages are not pulled from the stream until a request is made to the Ruby Enumerator. Basically, we stop polling messages from the response enumerator in an attempt to tell the GRPC client that we want to stop receiving messages on the stream while still being able to add messages to the request enumerator to send messages on the stream. However, in our manual testing it appears that messages may be queued in the Enumerator. Messages may be sent from the API but never pulled in to the Ruby objects. So it may be possible that we could have some stale objects in the enumerator, or orphaned responses that are never handled by the ruby code.

@apolcyn @murgatroid99 Do either of you have any input on controlling streaming requests/responses like this?

murgatroid99:
When you say that you want to "stop receiving messages", do you mean that you are done with the stream and do not intend to ever read any more messages from it, or do you mean that you want to apply backpressure to the stream so that the remote end temporarily stops sending until you want to read more messages?

In the first case, ending a stream on the reading end such that the remote end knows to stop sending is not part of the gRPC protocol. That information should be communicated to the remote end in whatever messages you are sending. The client can cancel a request, but that terminates the entire call with an error, which does not seem to be what you want.

In the second case, it is in general possible to apply backpressure to a stream to temporarily pause it until you want to start reading again, though I am not sure if the Ruby implementation does this. However, this should only be used for resource management, such as controlling usage of bandwidth and memory. Due to propagation delay and buffering at both ends in multiple layers, you should not expect that the contents of any particular set of messages will be affected by applying this backpressure, or even that the server application code will be made aware that this is happening.

blowmage:
In this case we want the latter. We want to apply back pressure to pause (and unpause) responses from the server. We do not want to cancel the stream if possible. Who can inform us on how/if we can accomplish this using the Ruby client?

apolcyn:
So what's wanted here is specifically for back pressure to be felt <i>by the server</i> when the <i>client</i> stops reading messages from a stream? 

If this is the case, then this should be happening in the ruby client. But back pressure should kick in based only on the receive-side http2 flow control. (e.g., the server will be able to continue to send messages until the total amount outstanding/unread  bytes exceeds the flow control limit, which will happen eventually if the client isn't reading messages). Flow control window size is tunable in the c-core but it doesn't look like that api is exposed in ruby. Is this sufficient?

blowmage:
What is wanted is for the client to apply back pressure on the server, however that is possible. I would love for the client to continue to accept and read responses from the server as long as the client could apply the back pressure to have the server pause responses eventually.

Right now, it looks like the client doesn't have any other way to apply that back pressure, other than to stop accepting responses by not pulling them from the Ruby Enumerator. Is this what we should be doing?

apolcyn:
If preventing an OOM on the client caused by the server's messages is the goal, then flow control should take care of preventing this (I think default config per-stream allowed-outstanding window should be roughly around 64K). The act of reading or not reading from the client-side enumerator will determine whether flow control starts to eventually push back to the server or allow more. Not sure if that works for this?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -66,20 +69,28 @@ def stop
           synchronize do
             return if @request_queue.nil?
             @request_queue.push self
+            @inventory.clear
```
quartzmo: Won't this delete messages that need to be processed? (Or is the idea that this occurs just locally, and they will be retrieved again later?)

quartzmo: You [commented below](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1542/commits/b6e1096b5ac9b009900d2632dbba8712d4ac0ef0#r123351117) to @remi: 

> Subscriber#stop only halts the pulling of new messages. It does not ensure that all messages have been processed. It updates the state of the Subscriber object and will return quickly. Subscriber#wait! will block until all the current messages in the queue are processed. Because it blocks it has the bang to indicate that it could take a while. 

So I had the idea that user calls `#stop` and `#wait!` to stop retrieving, but then process the inventory.

blowmage: Messages removed from inventory can still be processed. What we want is when the Subscriber stops to stop the Subscriber's background thread, _and_ the Inventory's background thread. so I'm using `Inventory#clear` as a way to stop the Inventory's background thread.

But, you bring up a good point, messages can still be processed by the Subscriber's thread pool and that will add the messages to Inventory. I suppose what we need is to add a `Inventory#stop` method that can stop the background thread. `Inventory#clear` is too much of a cheat. :)

I'll update the PR. Thanks for pointing this out!

quartzmo: Oh, I get it now. Thanks for explaining. Sounds like a good change.

blowmage:
@apolcyn Okay, that is what we are doing now. We simply stop reading responses from the client-side response enumerator. Can you confirm that we can still push requests to the client-side request enumerator even though we've stopped reading responses from the response enumerator? The GRPC client's flow control halts streaming responses while allowing streaming requests?

murgatroid99:
Yes, flow control in each direction is totally independent.

blowmage:
@murgatroid99 Thanks of the confirmation!

blowmage:
@tbkneeland @remi @lukesneeringer It looks like the only mechanism available to the Pub/Sub Ruby client for flow control is to stop pulling responses and let them queue up in the GRPC client. This is what this PR is doing. This also explains why some messages are not immediately available when the Subscriber starts pulling from the stream again once the inventory is no longer at capacity.

blowmage:
@mdietz94 Since you are taking over for @tbkneeland can you look at this PR as well?

blowmage:
This PR has been updated now that AsyncPublisher has been accepted and all merge conflicts have been resolved. PTAL.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
mdietz94: +1

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,386 @@
...
+
+        def perform_callback_async rec_msg
+          Concurrent::Future.new(executor: @thread_pool) do
+            @callback.call rec_msg
+          end.execute
+        end
+
+        def start_streaming!
+          @request_queue.unshift initial_input_request
+          @output_enum = service.streaming_pull @request_queue.each_item
```
mdietz94: You may want to consider adding any ack ids from inventory on this initial request. We will process ack ids that are part of the initial request, but may miss some that are sent shortly after the stream is started.

This is an implementation detail, but it would be good to hide this from end users.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,386 @@
...
+        end
+
+        ##
+        # @private
+        def delay_inventory!
+          synchronize do
+            return true if @inventory.empty?
+
+            ack_ids_deadline = @inventory.ack_ids.map { deadline }
+            inventory_ack_request = Google::Pubsub::V1::StreamingPullRequest.new
```
mdietz94: I would make sure that inventory does not exceed 10 MB. This should be unlikely, but if it does you want to split it into multiple StreamingPullRequests.

mdietz94: I would also call this mod_ack_request, since you are not acking, but modifying the deadlines.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -0,0 +1,386 @@
...
+            @thread_pool.shutdown
+            @thread_pool.wait_for_termination
+          end
+
+          true
+        end
+
+        ##
+        # @private
+        def acknowledge *messages
```
mdietz94: I would batch this so that messages are marked as 'acked' but then all sent together as one request. This isn't incorrect, but I think it would save CPU if you just had a background thread polling (could be 20x a second or something like that) instead of constructing a request for every message processed.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: I did not know we can do that. What if the size of the ack_ids in inventory is larger than 10MB?

blowmage: > This is an implementation detail, but it would be good to hide this from end users.

This method is not part of the public API and is shown to users in the documentation.

mdietz94: I meant "adding to the initial request being more successful" is an implementation detail, not this function.

If it's larger than 10 MB then we can only guarantee the first 10 MB of acks will be sent successfully. The others should be sent on a small delay (0.5s) after which the stream should have completed connecting.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: I've updated the variable name to be more clear.

blowmage: Agreed. I'll work on this as part of this PR.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: I've considered this as well. Similar to the AsyncPublisher used to publish messages in batches. I think this is a good candidate for a future enhancement in a separate PR. Once this PR is accepted I will create an issue to track this.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: Should this 0.5s delay be applied to all messages sent after the stream is created? Or just the inventory mod_ack messages?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: The method alias `modify_ack_deadline` has been added in a separate PR.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
mdietz94: I think maybe set the stream on a delay for sending requests until either:

1. 0.5s has passed from its creation
2. We have received a message from the stream

It's also okay to not wait the 0.5s if it makes the code simpler, acks are best effort, and it's possible they will still be processed, and possible we will end up buffering them in the future and that it's not worth optimizing too much to fit today's implementation.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
mdietz94: SGTM

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: K, I'll open an issue to add a 0.5s delay after initializing a stream and we can discuss in more detail in a separate PR.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -0,0 +1,354 @@
...
+
+          ##
+          # @private Create an empty Subscriber::Stream object.
+          def initialize subscriber
+            @subscriber = subscriber
+
+            @request_queue = nil
+            @output_enum = nil
+
+            @inventory = Inventory.new self, subscriber.inventory
```
blowmage: I have a question on how each Stream object's Inventory should be handled. Consider the following code:

```ruby
subscriber = subscription.listen streams: 4, inventory 1000 do |msg|
  puts msg.msg_id
end
```

Does each Stream object have an inventory of 1000, with a total inventory of 4000? Or does each Stream object have an inventory of 250, with a total inventory of 1000? The code is implementing the former, but I can see an argument to be made for the latter. Thoughts?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: Now that we are adding the option for multiple streams in this PR I think it would be best to track this in a separate issue/PR.

blowmage:
This PR has been updated to allow a Subscriber to contain multiple Streams. Users can configure this with the `streams` optional argument on `Subscription#listen`:

```ruby
subscriber = subscription.listen streams: 4 do |msg|
  puts msg.msg_id
  msg.ack!
end
```

The default value for `streams` is 1.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -0,0 +1,348 @@
...
+          # Makes sure the values are the `ack_id`. If given several
+          # {ReceivedMessage} objects extract the `ack_id` values.
+          def coerce_ack_ids messages
+            Array(messages).flatten.map do |msg|
+              msg.respond_to?(:ack_id) ? msg.ack_id : msg.to_s
+            end
+          end
+
+          # @private
+          class Inventory
```
mdietz94: You may want to consider sharing the limit across all created streams, at the Subscriber level. It may be better to say that you want 100 messages across all streams rather than 10 messages per stream, especially if for one reason or another one stream or more becomes throttled or has poorer performance on the server side.

blowmage: Each stream is going to manage its own inventory. The more streams you configure, the more total inventory held.

mdietz94: I understand that, but I think it's more confusing for users, that don't want to think about streams, just their total memory limits, and it's different than we've done this in Java/C# where the number of streams is invisible to the user. I'm happy for the number to be user configurable, but I think that flow control limits should be per subscriber.

PR #1543, opened 2017-06-14
Add requester pays to Storage

### LINKED ISSUE #1540: Add requester pays to Storage Bucket
Requester Pays enables the owner of a bucket to indicate that a user accessing the bucket or an object it contains should assume the transit costs related to the access. The feature is whitelist only at present.

- [ ] Add a metadata flag on a bucket to indicate that it is a requester pays bucket.
- [ ] Add documentation of whitelist-only availability to API docs and guide.

### LINKED ISSUE #1541: Add userProject to support requester pays in Bucket and File operations
Requester Pays enables the owner of a bucket to indicate that a user accessing the bucket or an object it contains should assume the transit costs related to the access. The availability of this feature is whitelist-only at present.

- [ ] Add a`user_pays` flag or `user_project` string to the relevant `Bucket` and `File` methods.
- [ ] When param above is true or given, send a project ID in the `userProject` header/query parameter for each relevant operation (see list below.)
- [ ] Add unit test coverage for the new header/query parameter for each relevant operation (see list below.)
- [ ] Add documentation of whitelist-only availability to API docs and guide.

API methods appearing in [`googleapis.com/discovery/v1`](https://www.googleapis.com/discovery/v1/apis/storage/v1/rest) that accept `userProject`: (Note: It may be that not all methods listed below are in google-cloud-ruby.)

- [x] `storage.bucketAccessControls.delete`
- [ ] `storage.bucketAccessControls.get`
- [x] `storage.bucketAccessControls.insert`
- [x] `storage.bucketAccessControls.list`
- [ ] `storage.bucketAccessControls.patch`
- [ ] `storage.bucketAccessControls.update`
- [ ] `storage.buckets.delete`
- [x] `storage.buckets.get`
- [x] `storage.buckets.getIamPolicy`
- [x] `storage.buckets.patch`
- [x] `storage.buckets.setIamPolicy`
- [x] `storage.buckets.testIamPermissions`
- [ ] `storage.buckets.update`
- [x] `storage.defaultObjectAccessControls.delete`
- [ ] `storage.defaultObjectAccessControls.get`
- [x] `storage.defaultObjectAccessControls.insert`
- [x] `storage.defaultObjectAccessControls.list`
- [ ] `storage.defaultObjectAccessControls.patch`
- [ ] `storage.defaultObjectAccessControls.update`
- [ ] `storage.notifications.delete`
- [ ] `storage.notifications.get`
- [ ] `storage.notifications.insert`
- [ ] `storage.notifications.list`
- [x] `storage.objectAccessControls. delete `
- [ ] `storage.objectAccessControls.get`
- [x] `storage.objectAccessControls.insert`
- [x] `storage.objectAccessControls.list`
- [ ] `storage.objectAccessControls.patch`
- [ ] `storage.objectAccessControls.update`
- [ ] `storage.objects.compose`
- [ ] `storage.objects.copy`
- [x] `storage.objects. delete `
- [x] `storage.objects.get`
- [ ] `storage.objects. getIamPolicy`
- [x] `storage.objects.insert`
- [x] `storage.objects.list`
- [x] `storage.objects.patch`
- [x] `storage.objects.rewrite`
- [ ] `storage.objects.setIamPolicy`
- [ ] `storage.objects.testIamPermissions`
- [ ] `storage.objects.update`
- [ ] `storage.objects. watchAll`

quartzmo:
This PR has two parts:

- [x] Add `requester_pays` to `Bucket` (#1540)
- [x] Add `userProject` request param to support requester pays in all relevant `Bucket` and `File` operations (#1541)

[closes #1540, closes #1541]

```ruby

# Set `user_pays` property once when retrieving bucket.
# All relevant RPCs made for this bucket and its objects
# will include `userProject` set to the currently authenticated
# project.

bucket = storage.bucket "my-todo-app", user_pays: true # Send `userProject`

file = bucket.file "avatars/heidi/400x400.png" # Send `userProject`

file.acl.add_reader "group-authors@example.net" # Send `userProject`
```

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -43,6 +43,11 @@ module Google
         attr_accessor :service
 
         ##
+        # @private The user_pays flag for sending `userProject` query param for
+        # operations on requester_pays buckets and their files.
+        attr_accessor :user_pays
+
+        ##
         # @private The Google API Client object.
         attr_accessor :gapi
 
@@ -51,6 +56,7 @@ module Google
         def initialize
           @service = nil
           @gapi = Google::Apis::StorageV1::Bucket.new
+          @user_pays = nil
         end
 
         ##
@@ -393,7 +399,7 @@ module Google
         #
         def delete
           ensure_service!
-          service.delete_bucket name
+          service.delete_bucket name, user_pays: user_pays
           true
         end
 
@@ -449,9 +455,10 @@ module Google
           ensure_service!
           gapi = service.list_files name, prefix: prefix, delimiter: delimiter,
                                           token: token, max: max,
-                                          versions: versions
+                                          versions: versions,
+                                          user_pays: user_pays
           File::List.from_gapi gapi, service, name, prefix, delimiter, max,
-                               versions
+                               versions, user_pays: user_pays
         end
         alias_method :find_files, :files
 
@@ -487,8 +494,9 @@ module Google
         def file path, generation: nil, encryption_key: nil
           ensure_service!
           gapi = service.get_file name, path, generation: generation,
-                                              key: encryption_key
-          File.from_gapi gapi, service
+                                              key: encryption_key,
+                                              user_pays: user_pays
+          File.from_gapi gapi, service, user_pays: user_pays
         rescue Google::Cloud::NotFoundError
           nil
         end
@@ -629,7 +637,8 @@ module Google
                       content_disposition: content_disposition, crc32c: crc32c,
                       content_encoding: content_encoding, metadata: metadata,
                       content_language: content_language, key: encryption_key,
-                      storage_class: storage_class_for(storage_class) }
+                      storage_class: storage_class_for(storage_class),
+                      user_pays: user_pays }
           ensure_io_or_file_exists! file
           path ||= file.path if file.respond_to? :path
           path ||= file if file.is_a? String
@@ -977,7 +986,7 @@ module Google
         def policy force: nil
           warn "DEPRECATED: 'force' in Bucket#policy" unless force.nil?
           ensure_service!
-          gapi = service.get_bucket_policy name
+          gapi = service.get_bucket_policy name, user_pays: user_pays
           policy = Policy.from_gapi gapi
           return policy unless block_given?
           yield policy
@@ -1018,7 +1027,8 @@ module Google
         #
         def policy= new_policy
           ensure_service!
-          gapi = service.set_bucket_policy name, new_policy.to_gapi
+          gapi = service.set_bucket_policy name, new_policy.to_gapi,
+                                           user_pays: user_pays
           Policy.from_gapi gapi
         end
 
@@ -1051,7 +1061,8 @@ module Google
         def test_permissions *permissions
           permissions = Array(permissions).flatten
           ensure_service!
-          gapi = service.test_bucket_permissions name, permissions
+          gapi = service.test_bucket_permissions name, permissions,
+                                                 user_pays: user_pays
           gapi.permissions
         end
 
@@ -1065,10 +1076,11 @@ module Google
 
         ##
         # @private New Bucket from a Google API Client object.
-        def self.from_gapi gapi, conn
+        def self.from_gapi gapi, service, user_pays: nil
           new.tap do |f|
             f.gapi = gapi
-            f.service = conn
+            f.service = service
+            f.user_pays = user_pays
           end
         end
 
@@ -1088,7 +1100,7 @@ module Google
             [attr, @gapi.send(attr)]
           end]
           patch_gapi = Google::Apis::StorageV1::Bucket.new patch_args
-          @gapi = service.patch_bucket name, patch_gapi
+          @gapi = service.patch_bucket name, patch_gapi, user_pays: user_pays
         end
 
         ##
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -43,20 +43,56 @@ module Google
         attr_accessor :service
 
         ##
-        # @private The user_pays flag for sending `userProject` query param for
-        # operations on requester_pays buckets and their files.
-        attr_accessor :user_pays
-
-        ##
         # @private The Google API Client object.
         attr_accessor :gapi
+
+        ##
+        # A `true`/`false` value or a project ID string to be sent as the
+        # `userProject` query param for operations on requester pays buckets and
+        # their files. The default is `nil`. For convenience, this attribute
+        # should be set when first retrieving the bucket, by providing the
+        # `user_project` option to {Project#bucket}.
+        #
+        # If the `requester_pays` flag is enabled for the bucket, and if this
+        # attribute is set to `true`, transit costs for operations on the bucket
+        # will be billed to the current project for this client. (See
+        # {Project#project} for the ID of the current project.) If this
+        # attribute is set to a project ID other than the current project, and
+        # that project is authorized for the currently authenticated service
+        # account, transit costs will be billed to the given project.
+        #
+        # The requester pays feature is currently available only to whitelisted
+        # projects.
+        #
+        # See also {#requester_pays=} and {#requester_pays}.
+        #
+        # @example With `user_project` set to pay for a requester pays bucket:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "other-project-bucket", user_project: true
+        #   files = bucket.files # Billed to current project
+        #   bucket.user_project #=> true
+        #
+        # @example With `user_project` set to a project other than the default:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "other-project-bucket",
+        #                           user_project: "my-other-project"
+        #   files = bucket.files # Billed to "my-other-project"
+        #   bucket.user_project #=> "my-other-project"
+        #
+        attr_accessor :user_project
 
         ##
         # @private Create an empty Bucket object.
         def initialize
           @service = nil
           @gapi = Google::Apis::StorageV1::Bucket.new
-          @user_pays = nil
+          @user_project = nil
         end
 
         ##
@@ -302,7 +338,8 @@ module Google
         ##
         # Indicates that a client accessing the bucket or a file it contains
         # must assume the transit costs related to the access. The requester
-        # must indicate the project to which the access costs should be billed.
+        # must pass the `user_project` option to {Project#bucket} to indicate
+        # the project to which the access costs should be billed.
         #
         # This feature is currently available only to whitelisted projects.
         #
@@ -317,12 +354,24 @@ module Google
         ##
         # Indicates that a client accessing the bucket or a file it contains
         # must assume the transit costs related to the access. The requester
-        # must indicate the project to which the access costs should be billed.
+        # must pass the `user_project` option to {Project#bucket} to indicate
+        # the project to which the access costs should be billed.
         #
         # This feature is currently available only to whitelisted projects.
         #
         # @param [Boolean] new_requester_pays When set to `true`, the bucket is
         #   requester pays.
+        #
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-bucket"
+        #
+        #   bucket.requester_pays = true # API call
+        #   # Other projects must now provide `user_project` option when calling
+        #   # Project#bucket to access this bucket.
         #
         def requester_pays= new_requester_pays
           @gapi.billing ||= Google::Apis::StorageV1::Bucket::Billing.new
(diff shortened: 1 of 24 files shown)


google-cloud-storage/lib/google/cloud/storage/project.rb
```diff
@@ -249,6 +249,12 @@ def bucket bucket_name
         #   does not exist. For more information, see [How to Host a Static
         #   Website
         #   ](https://cloud.google.com/storage/docs/website-configuration#step4).
+        # @param [Boolean] requester_pays Indicates that a client accessing the
```
blowmage: This method already has too many optional arguments IMO.  Would it be acceptable to set this value on the yielded bucket instead of as a method argument ?

blowmage: Acceptable to set this value _only_ on the yielded the bucket, I mean. :)

quartzmo: It's acceptable to me. We can always add it later if there is sufficient demand.

blowmage: :+1:

jskeet:
I'm really not a suitable person to review this, I'm afraid - I wouldn't know what's idiomatic and what's not, for a start.

quartzmo:
@jskeet No problem, thanks anyway!

frankyn:
@quartzmo I can review usage of this feature!

quartzmo:
@frankyn Thank you!

quartzmo:
@frankyn Updated PR description with example.

quartzmo:
> what do you think about having user_pays: true automatically be set if the bucket is a requester pays bucket?

I don't think we should do this given the "opt-in" intention of setting a value for `userProject`. However, your suggestion to clarify the relationship between `user_pays` and the feature as a whole is great. I will update the PR shortly.

quartzmo:
I'm thinking we should change `user_pays` to `user_project` in order to better associate the option name with the error `Google::Cloud::InvalidArgumentError: required: Bucket is requester pays bucket but no user project provided.`

The option could also accept a string project ID as well as a boolean. If `true` is given, use the current project, if string project ID is given, store and use it instead.

@blowmage @frankyn Thoughts?

blowmage:
@quartzmo If you can show that users are able to provide a project ID other than the one they are authenticating with then I can see changing the argument name/value. But if that is not the case then I'm not in favor of changing it.

jskeet:
A single service account may have sufficient permissions to be billable for multiple different projects.

quartzmo:
@jskeet Thanks for the clarification. @blowmage Sound good?

quartzmo:
Confirmed offline with @blowmage. I'm going to make this change.

frankyn:
Sorry I came in late to the conversation. I agree with the change, but the parameter name `user_project` could be ambiguous. What about `billed_project`?

blowmage:
^ @jskeet ?

quartzmo:
@frankyn You are certainly right that `user_project` does not communicate the whole story of "The project to be billed for this request, for Requester Pays buckets." (From the discovery document.) But for continuity with broader documentation, SO issues, and other sources covering its usage, I think sticking to the same name as the API and other libraries such as [google-cloud-python's impl](https://github.com/GoogleCloudPlatform/google-cloud-python/blob/master/storage/google/cloud/storage/bucket.py#L89-L91) is the best route.

frankyn:
Thanks @quartzmo, I understand the idea of consistency.

quartzmo:
@frankyn Leveraging the "docs ecosystem." ;)

frankyn:
I meant to say that I agree with the information you provided me. It would be jarring for this parameter to have different names per-language in the documentation. :)

quartzmo:
@frankyn Updated, PTAL.

geigerj:
Disclaimer: I'm very new to Ruby. But it feels off to me to have `user_project` accept a boolean. What does it mean for `user_project` to be false? (Should it accept a symbol instead?)

quartzmo:
@geigerj Our goal is to make it very easy to send the client's current project ID. Our take is that Rubyists are *very* comfortable with type-agnostic truthy/falsy values. The default is `nil` (falsy) and passing `false` is equivalent. Accepting  `true` is a convenience: We aren't going to force you to look up your current project ID since we already have it. This is what we expect most users to go with. Accepting a string as well is a hedge, just in case someone has multiple projects for their service account, as @jskeet pointed out above.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -47,10 +47,52 @@ class Bucket
         attr_accessor :gapi
 
         ##
+        # A `true`/`false` value or a project ID string to be sent as the
```
frankyn: Could you define what it means for `true` to be set? (Use project id defined by default to use `project`)

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -47,10 +47,52 @@ class Bucket
...
+        # will be billed to the current project for this client. (See
+        # {Project#project} for the ID of the current project.) If this
+        # attribute is set to a project ID other than the current project, and
+        # that project is authorized for the currently authenticated service
+        # account, transit costs will be billed to the given project.
+        #
+        # The requester pays feature is currently available only to whitelisted
+        # projects.
+        #
+        # See also {#requester_pays=} and {#requester_pays}.
```
frankyn: Maybe change to: See also {#requester_pays=} and {#requester_pays} to turn on requester pays for a bucket.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -47,10 +47,52 @@ class Bucket
...
+        #
+        # @example With `user_project` set to pay for a requester pays bucket:
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "other-project-bucket", user_project: true
+        #   files = bucket.files # Billed to current project
+        #   bucket.user_project #=> true
```
frankyn: I don't understand why a user would want to print the value of bucket.user_project after requesting associated files using requester pays.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -294,12 +336,58 @@ def website_404= website_404
...
+        # @return [Boolean, nil] Returns `true` if requester pays is enabled for
+        #   the bucket.
+        #
+        def requester_pays
+          @gapi.billing.requester_pays if @gapi.billing
+        end
+        alias_method :requester_pays?, :requester_pays
+
+        ##
+        # Indicates that a client accessing the bucket or a file it contains
```
frankyn: Instead of indicates, explain that this operation enables requester pays?

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -294,12 +336,58 @@ def website_404= website_404
...
+        # must assume the transit costs related to the access. The requester
+        # must pass the `user_project` option to {Project#bucket} to indicate
+        # the project to which the access costs should be billed.
+        #
+        # This feature is currently available only to whitelisted projects.
+        #
+        # @param [Boolean] new_requester_pays When set to `true`, the bucket is
```
frankyn: Change "the bucket is requester pays" to "requester pays is enabled for the bucket".

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: 👍

google-cloud-storage/lib/google/cloud/storage/bucket.rb
frankyn: FWIW, I understand it's described in the next paragraph, but they are introduced here.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: 👍

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: Hah, that's what I deserve for copying official docs!

google-cloud-storage/lib/google/cloud/storage/bucket.rb
frankyn: Whoops, "turn on" -> "enable"

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: On line 56:

> if this attribute is set to `true`, transit costs for operations on the bucket will be billed to the current project for this client.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: Hmmm, should I change this to a more complicated example, in which the value of `user_project` is changed for an operation, then restored?

```ruby
bucket = storage.bucket "other-project-bucket", user_project: true
files = bucket.files # Billed to current project
bucket.user_project = "my-other-project"
files = bucket.files # Billed to "my-other-project"
bucket.user_project = true
```

frankyn: +1 this would provide a multiple projects workflow example. However, I feel that `bucket.user_project = true` isn't necessary in this snippet. 

My original comment was to remove `bucket.user_project #=> true` from the snippet and only keep. 
```
bucket = storage.bucket "other-project-bucket", user_project: true
files = bucket.files # Billed to current project
```

google-cloud-storage/lib/google/cloud/storage/bucket.rb
frankyn: I misunderstood its context when displayed in documentations. One last nit but not required. 

Why `true`/`false` and not boolean?

quartzmo:
@frankyn Updated, thanks for the edits!

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -47,10 +47,42 @@ class Bucket
...
+        # In general, this attribute should be set when first retrieving the
+        # bucket by providing the `user_project` option to {Project#bucket}.
+        #
+        # The requester pays feature is currently available only to whitelisted
+        # projects.
+        #
+        # See also {#requester_pays=} and {#requester_pays} to enable requester
+        # pays for a bucket.
+        #
+        # @example Setting to a project other than the default:
```
frankyn: Setting a non-default project:

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -47,10 +47,42 @@ class Bucket
...
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "other-project-bucket", user_project: true
+        #   files = bucket.files # Billed to current project
```
frankyn: Suggestion: Billed to "other-project-bucket" (I like explicit).

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -294,12 +326,59 @@ def website_404= website_404
...
+        # to the access. The requester must pass the `user_project` option to
+        # {Project#bucket} to indicate the project to which the access costs
+        # should be billed.
+        #
+        # This feature is currently available only to whitelisted projects.
+        #
+        # @param [Boolean] new_requester_pays When set to `true`, requester pays
+        #   is enabled for the bucket.
+        #
+        # @example
```
frankyn: Add description for this example for it to be consistent with other example. Suggestion: Enable requester pays for a bucket.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
quartzmo: But `user_project: true` means costs for accessing "other-project-bucket" and its files will be billed to the current (or default) project.

quartzmo: Unfortunately the example does not include the name of the current project, it is obtained from an environment variable.

frankyn: Whoops, my mistake. Understood.

quartzmo:
@frankyn Updated, PTAL.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -55,43 +55,34 @@ class File
...
         #
         # The requester pays feature is currently available only to whitelisted
         # projects.
         #
-        # See also {Bucket#requester_pays=} and {Bucket#requester_pays}.
+        # See also {Bucket#requester_pays=} and {Bucket#requester_pays} to
+        # enable requester pays for a bucket.
         #
-        # @example With `user_project` set to pay for a requester pays bucket:
+        # @example Setting to a project other than the default:
```
frankyn: Could this be consistent with "Setting a non-default project"?

frankyn:
Last comment thanks for your patience @quartzmo. The documentation looks great!

quartzmo:
@frankyn Can you approve?

google-cloud-storage/lib/google/cloud/storage/file.rb
quartzmo: Done.

frankyn:
Thanks @quartzmo this looks great!

quartzmo:
@blowmage and @frankyn, thank you both for your help with this feature!

PR #1547, opened 2017-06-15
Use latest generation of google-cloud-debugger GAPIC

arhamahmed:
google-cloud-debugger GAPIC: https://github.com/googleapis/api-client-staging/tree/master/generated/ruby/google-cloud-ruby/google-cloud-debugger


hxiong388:
The PR was merged preemptively, but the build should pass once the DNS test is fixed in master branch.

PR #1559, opened 2017-06-23
Spanner unbounded thread pool

blowmage:
This PR replaces the fixed thread pool with an unbounded thread pool that will expand and contract the number of threads as needed.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -215,9 +215,7 @@ module Google
 
         def init
           # init the thread pool
-          @thread_pool = Concurrent::FixedThreadPool.new(
-            [2, Concurrent.processor_count*2].max
-          )
+          @thread_pool = Concurrent::ThreadPoolExecutor.new
           # init the queues
           @new_sessions_in_process = @min.to_i
           @all_sessions = []


blowmage:
@vkedia What do you think about this change?

vkedia:
Why do we need to make this change? Have we observed any issues with the fixed thread pool?
Also can we allow user to specify a threapool with some default?

blowmage:
One reason I can see for this change is memory overhead for less heavily used client instances. Also, currently there is no way for users to specify the size (or type) of the thread pool. That's sort of why I'm asking. :) Is configuration wanted?

@remi, since we are actively working on something similar for Pub/Sub, and I know you are working on Spanner code examples, do you have any thoughts on configuration for Spanner's Session Pool's Thread Pool?

vkedia:
What is the current size of the thread pool. I don't think we should change to unbounded since we do want to throttle session creations by default. So can we keep the default to be a bounded thread pool and have users specify there own thread pool if needed?

blowmage:
> What is the current size of the thread pool.

The current fixed thread pool size is double the number of CPUs:

```ruby
@thread_pool = Concurrent::FixedThreadPool.new(
  [2, Concurrent.processor_count*2].max
)
```

blowmage:
I'll create a separate PR for the thread pool size configuration.

blowmage:
Closing as this is not the direction we are going.

PR #1561, opened 2017-06-25
Add Pub/Sub AsyncPublisher

blowmage:
This PR adds a AsyncPublisher object to Topic that performs the following:

- [x] Asynchronously publish messages in batches
- [x] Will push batches in intervals _after_ first publish (every 0.25 seconds)
- [x] Will push batches when message count hits a limit (1000)
- [x] Will push batches when RPC payload reaches a size (5MB)

This functionality is expressed as `Topic#publish_async`. The publisher object can be accessed as `Topic#async_publisher`.

This change is to be merged into the pubsub branch, which will be used as the target for the upcoming Pub/Sub changes.

This PR is an alternative implementation to #1530, based on feedback provided.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -197,9 +197,10 @@ module Google
         # @param [String] topic_name Name of a topic.
         # @param [String, File] data The message data.
         # @param [Hash] attributes Optional attributes for the message.
-        # @yield [publisher] a block for publishing multiple messages in one
+        # @yield [batch] a block for publishing multiple messages in one
         #   request
-        # @yieldparam [Topic::Publisher] publisher the topic publisher object
+        # @yieldparam [Topic::BatchPublisher] batch the topic batch publisher
+        #   object
         #
         # @return [Message, Array<Message>] Returns the published message when
         #   called without a block, or an array of messages when called with a
@@ -245,7 +246,7 @@ module Google
             data = nil
           end
           ensure_service!
-          publisher = Topic::Publisher.new data, attributes
+          publisher = Topic::BatchPublisher.new data, attributes
           yield publisher if block_given?
           return nil if publisher.messages.count.zero?
           publish_batch_messages topic_name, publisher
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -119,12 +119,12 @@ module Google
         #   pubsub = Google::Cloud::Pubsub.new
         #   topic = pubsub.topic "another-topic", skip_lookup: true
         #
-        def topic topic_name, project: nil, skip_lookup: nil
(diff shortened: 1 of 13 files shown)


blowmage:
@tbkneeland I don't have permissions to add you as a reviewer. Can you start a review anyway? Please and thank you.

blowmage:
@dazuma I've added you to the reviewers for this PR since it shares some similarity to Logger's AsyncWriter. Can you look this over an provide any feedback?

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -144,10 +144,10 @@ def topic topic_name, project: nil, skip_lookup: nil
         #   pubsub = Google::Cloud::Pubsub.new
         #   topic = pubsub.create_topic "my-topic"
         #
-        def create_topic topic_name
+        def create_topic topic_name, async: nil
```
remi: Project#topic doesn't accept async options, only create_topic does. Which means I can't do the following:

```ruby
topic = pubsub.topic "my-topic", max_messages: 50
```

😢

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -0,0 +1,230 @@
...
+
+          def run_background
+            synchronize do
+              until @stopped
+                if @batch.nil?
+                  @cond.wait
+                  next
+                end
+
+                time_since_last_publish = Time.now - @time
```
remi: This is really `time_since_first_publish`. The variable name threw me off.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: Oops, that's a mistake! I'll fix it.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: Agreed. i'll update the name.

google-cloud-pubsub/support/doctest_helper.rb
```diff
@@ -416,6 +416,20 @@ def mock_pubsub
     end
   end
 
+  doctest.skip "Google::Cloud::Pubsub::Topic#publish_async" do
```
remi: OOC why is publish_async skipped? Subscription#listen has a useful code comment describing why it's being skipped.

blowmage: The implementation has changed so much between the previous PR and this one, that the previous doc test configuration was no longer applicable. @quartzmo and I will need to create a strategy for the AsyncPublisher doc tests, but we will do that in a later PR, along with the final documentation, after the behavior added in this PR is accepted.

remi: Sgtm! Thanks 😄

PR #1564, opened 2017-06-27
Pub/Sub Message data encoding

blowmage:
Update documentation to detail how message data is processed when sending to Pub/Sub API, and how message data is encoded when received from Pub/Sub API.

The Google Cloud Pub/Sub Ruby library will follow how Google Protobuf [handles string data for bytes fields](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#fields), namely:

> The protobuf library will duplicate the string, convert it to ASCII-8BIT encoding, and freeze it.

The Google Cloud Pub/Sub Ruby library should not provide less convenience than Google Protobuf. In fact, it provides more convenience for `data` values, as it will also accept an IO-ish object, such as File, and convert it to a string/bytes.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message.rb
@@ -61,13 +61,14 @@ module Google
         end
 
         ##
-        # The received data.
+        # The message payload. This data is a list of bytes encoded as
+        # ASCII-8BIT.
         def data
           @grpc.data
         end
 
         ##
-        # The received attributes.
+        # Optional attributes for the message.
         def attributes
           return @grpc.attributes.to_h if @grpc.attributes.respond_to? :to_h
           # Enumerable doesn't have to_h on Ruby 2.0, so fallback to this
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -194,8 +194,12 @@ module Google
         ##
         # Publishes one or more messages to the given topic.
         #
+        # The message payload must not be empty; it must contain either a
+        # non-empty data field, or at least one attribute.
+        #
         # @param [String] topic_name Name of a topic.
-        # @param [String, File] data The message data.
+        # @param [String, File] data The message payload. This will be converted
+        #   to bytes encoded as ASCII-8BIT.
         # @param [Hash] attributes Optional attributes for the message.
         # @yield [batch] a block for publishing multiple messages in one
         #   request
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/received_message.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/received_message.rb
@@ -66,13 +66,14 @@ module Google
         alias_method :msg, :message
 
         ##
-        # The received message's data.
+        # The received message payload. This data is a list of bytes encoded as
+        # ASCII-8BIT.
         def data
           message.data
         end
 
         ##
-        # The received message's attributes.
+        # Optional attributes for the received message.
         def attributes
           message.attributes
         end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
@@ -240,7 +240,11 @@ module Google
         ##
         # Publishes one or more messages to the topic.
         #
-        # @param [String, File] data The message data.
+        # The message payload must not be empty; it must contain either a
+        # non-empty data field, or at least one attribute.
+        #
+        # @param [String, File] data The message payload. This will be converted
+        #   to bytes encoded as ASCII-8BIT.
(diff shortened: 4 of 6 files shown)


blowmage:
ATTN @tbkneeland

geigerj:
Would appreciate explicit approval from the Pub/Sub team on this particular issue. @mdietz94 will take over for @tbkneeland.

mdietz94:
We would like to be consistent and require UTF8 encoding on all languages. We would like to potentially be more restrictive than the Protocol Buffer API in order to ensure compatibility across languages.

It's important if a user publishes "\xce\xb8" from any language, the subscriber will always see "θ". Failure to do this can lead to inconsistency, and it seems like a huge mistake to not use UTF8 when choosing an encoding.

blowmage:
What if users want to publish binary data, like an image file?

mdietz94:
They can encode that as UTF8, decode as UTF8, and write it back out on the
other side. It's important we don't make it difficult for people to use
CP/S and achieve good localization.

Also that is a poor use case for CP/S. Users should be writing the image to
disk somewhere and just publishing the fact that it happened. Our average
message size is quite small.

remi:
When I last met in person with Thomas a few weeks ago, we looked at this issue for over an hour. Unless he has since stated otherwise, we came to the conclusion that, for Ruby, following the behavior of Google Protobuf is correct.

FWIW we spent quite some time digging into the Ruby code and C bindings for the Ruby protobuf library to verify the [documented behavior for bytes fields](https://developers.google.com/protocol-buffers/docs/reference/ruby-generated#fields) that @blowmage commented earlier *(as it's different from other langs afaik)*

> "The protobuf library will duplicate the string, convert it to ASCII-8BIT encoding, and freeze it."

So the answer is pretty clear.

The [PubsubMessage `data` field](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#google.pubsub.v1.PubsubMessage) is a `bytes` field, not a `string` field.  So ASCII-8BIT will be used, not UTF-8.  If the Pub/Sub API intended for strings, then the proto field should've been a string instead of bytes.

mdietz94:
If we publish a UTF8 string, would ASCII-8BIT throw an exception or drop some characters? Or would it just publish the bytes.

I agree as long as we're just publishing a byte array, but I definitely don't want to stop people from publishing UTF8 strings.

In Python, for example, we're supporting both Unicode strings as UTF8 and bytes()

mdietz94:
See @tbkneeland's comments on https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3463

blowmage:
> If we publish a UTF8 string, would ASCII-8BIT throw an exception or drop some characters? Or would it just publish the bytes.

It publishes the bytes of the string, sans the encoding. We take great care not to modify the bytes provided.

blowmage:
> It's important if a user publishes "\xce\xb8" from any language, the subscriber will always see "θ". Failure to do this can lead to inconsistency, and it seems like a huge mistake to not use UTF8 when choosing an encoding.

I think there is real danger of taking the bytes from the data payload and always representing them as UTF-8. Take the following single byte payload:

```ruby
[255].pack("C") #=> "\xFF"
[255].pack("C").encoding #=> ASCII-8BIT
[255].pack("C").encode("utf-8") #=> Encoding::UndefinedConversionError: "\xFF" from ASCII-8BIT to UTF-8
```

There are byte sequences that are illegal in UTF-8. The ruby library can't guarantee what client published the message, and since PubsubMessage doesn't have any encoding information on it I think it is dangerous to make assumptions about the payload's original encoding. It seems to me that the safest option is to assume bytes unless you have evidence for encoding otherwise.

blowmage:
> It's important we don't make it difficult for people to use CP/S and achieve good localization.

What is "CP/S"?

quartzmo:
CP/S is Cloud Pub/Sub, I'm guessing.

blowmage:
> It's important if a user publishes "\xce\xb8" from any language, the subscriber will always see "θ". Failure to do this can lead to inconsistency, and it seems like a huge mistake to not use UTF8 when choosing an encoding.

I'm looking at @tbkneeland's [comment](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/3463/files) where he says essentially the same thing. Only AFAICT he's _not_ talking about the message `data`, he's talking about the message `attributes`. The [attributes field is defined in the protobuf file](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L219) as `map<string, string>`, so both the keys and values are indeed  strings and per Google Protobuf will always be encoded as UTF-8. I believe he was taking issue with the following documentation in the python library where `attrs` says it can be provided as `bytes`:

```
        Raises:
            TypeError: If the ``data`` sent is not a bytestring, or if the
                ``attrs`` are not either a ``str`` or ``bytes``.
```

Are we sure he was intending to say that all message `data` should be encoded as UTF-8?

mdietz94:
Okay that SGTM. It wasn't clear to me that ASCII 8BIT left the data alone, I wasn't sure if it changed it or had illegal but sequences itself. If that's the canonical bytes representation then LGTM

PR #1565, opened 2017-06-28
Add schema option to Bigquery::Dataset#load

### LINKED ISSUE #1537: Add public Schema creation, and schema option to Dataset#create_table and Dataset#load
From @AlfonsoUceda's comments in PR #1533:

> is there a change to pass the schema as an option instead a block? for example the following snippet:

```ruby
bigquery.dataset("my_dataset").load "my_table", file, write: "truncate", schema: schema
```

> We create schema objects `#from_gapi(nil)` because then we inject them in the `table#load` using our fork. I know `#from_gapi` method is private but is there a way to create schema objects from a public interface? have you though in that?

More on  @AlfonsoUceda's usage:

> we have our custom reporter component, I mean an ETL process. We have schema separated by files, every schema its class, so we can inject them in our ETL class, every ETL has a file and we define a extractor, transformer and loader (csv). When the CSV file is created then with this file and a generated schema object we inject into our bigquery loaders class to create or truncate the table.

> This is because we can't test our schema in a separate test or because we share schemas by ETLs.

```ruby
bigquery_loader.load(file, schema: schema)

# inside `#load`
table.load file, write: 'truncate', schema: schema # this is because of our fork
```

quartzmo:
This PR adds a `schema` option and supporting `Project#schema` creation method for a way to configure the destination table's schema **in addition to** the existing block support for this configuration. See #1537 for @AlfonsoUceda's use case.

[closes #1537]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -924,12 +924,22 @@ module Google
         #   file that BigQuery will skip when loading the data. The default
         #   value is `0`. This property is useful if you have header rows in the
         #   file that should be skipped.
+        # @param [Google::Cloud::Bigquery::Schema] schema The schema for the
+        #   destination table. Optional. The schema can be omitted if the
+        #   destination table already exists, or if you're loading data from a
+        #   Google Cloud Datastore backup.
+        #
+        #   See {Project#schema} for the creation of the schema for use with
+        #   this option. Also note that for most use cases, the block yielded by
+        #   this method is a more convenient way to configure the schema.
         #
         # @yield [schema] A block for setting the schema for the destination
         #   table. The schema can be omitted if the destination table already
         #   exists, or if you're loading data from a Google Cloud Datastore
         #   backup.
-        # @yieldparam [Schema] schema the object accepting the schema
+        # @yieldparam [Google::Cloud::Bigquery::Schema] schema The schema
+        #   instance provided using the `schema` option, or a new, empty schema
+        #   instance
         #
         # @return [Google::Cloud::Bigquery::LoadJob]
         #
@@ -997,15 +1007,14 @@ module Google
                  projection_fields: nil, jagged_rows: nil, quoted_newlines: nil,
                  encoding: nil, delimiter: nil, ignore_unknown: nil,
                  max_bad_records: nil, quote: nil, skip_leading: nil,
-                 dryrun: nil
+                 dryrun: nil, schema: nil
           ensure_service!
 
-          schema = nil
           if block_given?
-            schema = Schema.from_gapi
+            schema ||= Schema.from_gapi
             yield schema
-            schema = schema.to_gapi
           end
+          schema_gapi = schema.to_gapi if schema
 
           options = { format: format, create: create, write: write,
                       projection_fields: projection_fields,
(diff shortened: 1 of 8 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -752,6 +753,35 @@ def time hour, minute, second
...
+        #     nested_schema.string "place", mode: :required
+        #     nested_schema.integer "number_of_years", mode: :required
+        #   end
+        #
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   gs_url = "gs://my-bucket/file-name.csv"
+        #   load_job = dataset.load "my_new_table", gs_url, schema: schema
+        #
+        def schema
```
blowmage: I think this should yield if a block is given so you can create and configure your schema in a single call.

```ruby
schema = bigquery.schema do |s|
  s.integer   "id",    description: "id description",    mode: :required
  s.string    "breed", description: "breed description", mode: :required
  s.string    "name",  description: "name description",  mode: :required
  s.timestamp "dob",   description: "dob description",   mode: :required
end
```

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1014,7 +1023,7 @@ def load table_id, file, format: nil, create: nil, write: nil,
                       delimiter: delimiter, ignore_unknown: ignore_unknown,
                       max_bad_records: max_bad_records, quote: quote,
                       skip_leading: skip_leading, dryrun: dryrun,
-                      schema: schema }
+                      schema: schema_gapi }
```
blowmage: Will this raise if `schema` is `nil` and `schema_gapi` is undefined?

quartzmo: I believe `schema_gapi` will be `nil` here if `schema` is `nil`.

quartzmo: Acceptance tests are passing.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
quartzmo: I can do this if you want, although the single call has no importance since there is no API interaction.

blowmage: I think we should yield and the doc example should show the block syntax.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
blowmage: Is there a test that doesn't provide a schema and doesn't use the block syntax?

blowmage: I've played around with this and this shouldn't be a problem. It looks like `schema_gapi` is `nil` and not undefined.

quartzmo: I'll check and if not, add one.

blowmage: You learn something new every day...

```ruby
defined? foo #=> nil
foo = "hello" if false
defined? foo #=> "local-variable"

foo #=> nil
bar # NameError: undefined local variable or method `bar'
```

quartzmo: I've added an acceptance test for the case in which an existing table is used and no schema is configured.

I have also added the block to `Project#schema`.

quartzmo:
@AlfonsoUceda Does this look OK to you?

AlfonsoUceda:
@quartzmo it's perfect, thank you for your work, I'll be able to get rid of my fork ;)

quartzmo:
@AlfonsoUceda Great! I'm working on the release now. I'll post on this PR when it's available on rubygems.

quartzmo:
Sorry, I forgot to post here after releasing, but [google-cloud-bigquery 0.27.0 is available on RubyGems](https://rubygems.org/gems/google-cloud-bigquery/versions/0.27.0)

AlfonsoUceda:
Thanks @quartzmo!

PR #1566, opened 2017-06-28
Pub/Sub Message#published_at

blowmage:
This PR adds `Message#published_at`, which was missing. It is also setting this value on messages when they are published. Messages pulled from the API will have this attribute populated.

### DIFF
--- a/google-cloud-pubsub/acceptance/pubsub/async_test.rb
+++ b/google-cloud-pubsub/acceptance/pubsub/async_test.rb
@@ -53,6 +53,7 @@ describe Google::Cloud::Pubsub, :async, :pubsub do
     event = events.first
     event.wont_be :nil?
     event.msg.data.must_equal publish_result.data
+    event.msg.published_at.wont_be :nil?
     # Acknowledge the message
     sub.ack event.ack_id
     # Remove the subscription
--- a/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
+++ b/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
@@ -204,6 +204,7 @@ describe Google::Cloud::Pubsub, :pubsub do
       event = events.first
       event.wont_be :nil?
       event.msg.data.must_equal msg.data
+      event.msg.published_at.wont_be :nil?
       # Acknowledge the message
       subscription.ack event.ack_id
       # Remove the subscription
@@ -231,6 +232,7 @@ describe Google::Cloud::Pubsub, :pubsub do
       event = events.first
       event.wont_be :nil?
       event.msg.data.must_equal msg.data
+      event.msg.published_at.wont_be :nil?
       # Acknowledge the message
       subscription.ack event.ack_id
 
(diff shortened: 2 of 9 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -147,10 +147,12 @@ def publish_batch_async topic_name, batch
             Concurrent::Future.new(executor: @thread_pool) do
               begin
                 grpc = @service.publish topic_name, batch.messages
+                publish_timestamp = Convert.time_to_timestamp Time.now
                 batch.items.zip(Array(grpc.message_ids)) do |item, id|
                   next unless item.callback
 
                   item.msg.message_id = id
+                  item.msg.publish_time = publish_timestamp
```
quartzmo: Might be better to use `published_at` instead of the alias `publish_time`, to make it less likely that some maintainer misses this when searching for usages. Or reverse which is the impl and which is the alias.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/batch_publisher.rb
```diff
@@ -57,9 +57,11 @@ def publish data, attributes = {}
 
           ##
           # @private Create Message objects with message ids.
-          def to_gcloud_messages message_ids
+          def to_gcloud_messages message_ids, published_at = Time.now
+            publish_time = Convert.time_to_timestamp published_at
             msgs = @messages.zip(Array(message_ids)).map do |msg, id|
               msg.message_id = id
+              msg.publish_time = publish_time
```
quartzmo: Might be better to use `published_at` instead of the alias `publish_time`, to make it less likely that some maintainer misses this when searching for usages. Or reverse which is the impl and which is the alias.

blowmage:
> It looks to me that `publish_time` is being set to `Time.now` by the publisher.

It is, but it is set _after_ the message has been published.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/batch_publisher.rb
blowmage: I've updated the variable names to be clearer.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: I've updated the variable names to be clearer.

blowmage:
Currently this PR has a second commit that will set the `published_at` value on messages to the current time _after_ the messages have been published. This is because the `PublishResponse` value returned from the API does not have this value. The only API that returns messages with this value is Pull.

I don't know for sure this is wise. I can see it being helpful, but I can also see it being confusing. Thoughts?

blowmage:
> Also, is it possible to add an assertion demonstrating the presence of a non-nil publish_time in an existing acceptance test?

I've updated the acceptance tests to check for the presence of `published_at`.

quartzmo:
I think it is a bit misleading since the real value is the one set by the server…

blowmage:
Okay, removing the commit from the PR...

PR #1571, opened 2017-06-30
Change Logger#labels to default to empty hash

hxiong388:
Make Logger#labels default to empty hash. So users can set custom labels down stream without need to explicitly initialize logger with empty hash label.

### DIFF
--- a/google-cloud-logging/test/google/cloud/logging/logger_test.rb
+++ b/google-cloud-logging/test/google/cloud/logging/logger_test.rb
@@ -46,6 +46,11 @@ describe Google::Cloud::Logging::Logger, :mock_logging do
     ]
   end
 
+  it "@labels instance variable is default to empty hash if not given" do
+    logger = Google::Cloud::Logging::Logger.new logging, log_name, resource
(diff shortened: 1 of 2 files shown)


quartzmo:
No test coverage for this?

hxiong388:
Right. Added unit test.

google-cloud-logging/test/google/cloud/logging/logger_test.rb
```diff
@@ -46,6 +46,11 @@ def write_req_args severity, extra_labels: {}, log_name_override: nil,
     ]
   end
 
+  it "@labels instance variable is default to empty hash if not given" do
+    logger = Google::Cloud::Logging::Logger.new logging, log_name, resource
+    logger.labels.must_be_kind_of Hash
```
blowmage:
```ruby
logger.labels.must_be_kind_of Hash
logger.labels.must_be :empty?
```

hxiong388: Fixed

PR #1572, opened 2017-06-30
Fix bad retry configuration

geigerj:
This configuration was originally made under the mistaken belief that
UNAVAILABLE indicated that no data had been transmitted to the server.
In fact, this is not necessarily true, so it is unsafe to retry on
that code for non-idempotent RPCs.

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/v1/datastore_client_config.json
+++ b/google-cloud-datastore/lib/google/cloud/datastore/v1/datastore_client_config.json
@@ -6,9 +6,7 @@
           "DEADLINE_EXCEEDED",
           "UNAVAILABLE"
         ],
-        "non_idempotent": [
-          "UNAVAILABLE"
-        ]
+        "non_idempotent": []
       },
       "retry_params": {
         "default": {
--- a/google-cloud-debugger/lib/google/cloud/debugger/v2/controller2_client_config.json
+++ b/google-cloud-debugger/lib/google/cloud/debugger/v2/controller2_client_config.json
@@ -6,9 +6,7 @@
           "DEADLINE_EXCEEDED",
           "UNAVAILABLE"
         ],
-        "non_idempotent": [
-          "UNAVAILABLE"
-        ]
+        "non_idempotent": []
       },
       "retry_params": {
         "default": {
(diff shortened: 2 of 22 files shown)


bmclean:
I find this change a bit odd. Even back in the ndb days we have always retried UNAVAILABLE with an exponential backoff. Will this reference be changing as well? 

https://cloud.google.com/datastore/docs/concepts/errors

Timeouts are different than DEADLINE_EXCEEDED. Should we still be assuming that approximately 1 in 3,000 ordinary datastore operations will result in a timeout? If so, aren't those coming back as UNAVAILABLE?

https://cloud.google.com/appengine/articles/handling_datastore_errors

geigerj:
@bmclean This change is based specifically on @apolcyn's finding that if the gRPC server stub's process terminates at any point before the response message is sent, the client will receive `UNAVAILABLE`. (So it's possible the request message makes it to the backend, but the client still receives `UNAVAILABLE` if something goes wrong at the wrong time.)

This affects only the gRPC clients. You should continue following the per-API documentation for any of the REST-style, HTTP/JSON clients.

geigerj:
@quartzmo Is this change OK with you?

geigerj:
@quartzmo Acknowledged. I think this still makes sense as a precaution against the gRPC issue I noted above. I will check if it is safe to be more liberal with Datastore in particular.

bmclean:
@geigerj Thanks for the info. I am referring to the datastore gRPC client though. I think I confused the issue because we use both clients (ndb datastore on our older stuff, the ruby gRPC datastore on our newer). With the gRPC datastore client we see higher than 1 in 3000 operations return UNAVAILABLE. We simply retry those errors.

quartzmo:
@geigerj Thanks for looking into the particular situation with Datastore.

@bmclean As usual, your feedback is very much appreciated!

geigerj:
Update: from internal discussion, I haven't received confirmation that it is safe to retry non-idempotent Datastore gRPC calls on after receiving UNAVAILABLE. We'll leave this as-is. (The docs should be correct for HTTP/JSON, and I'm asking for an update to the docs for the gRPC case.)

PR #1574, opened 2017-07-04
Accept response content type/disposition as query string parameters for signed URLs

georgeclaghorn:
Google Cloud Storage allows the `Content-Type` and `Content-Disposition` response headers to be overridden via query string parameters on GET requests: `response-content-type` and `response-content-disposition`, respectively. These query string parameters are [explicitly documented](https://cloud.google.com/storage/docs/xml-api/reference-headers) for the XML API, but they work with the JSON API as well.

This PR facilitates passing the `response-content-type` and `response-content-disposition` query string parameters to `Google::Cloud::Storage::File#signed_url` for inclusion in the resulting URL.

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -485,6 +485,27 @@ describe Google::Cloud::Storage::File, :storage do
     end
   end
 
+  it "bucket should create a signed read url with response content type and disposition" do
+    local_file = File.new files[:logo][:path]
+    file = bucket.create_file local_file, "CloudLogoSignedUrlGetBucket.png"
+
+    five_min_from_now = 5 * 60
+    url = bucket.signed_url file.name, method: "GET",
+                                       expires: five_min_from_now,
+                                       query: { "response-content-type" => "application/octet-stream",
+                                                "response-content-disposition" => "attachment; filename=\"google-cloud.png\"" }
+
+    uri = URI url
+    http = Net::HTTP.new uri.host, uri.port
+    http.use_ssl = true
+    http.ca_file ||= ENV["SSL_CERT_FILE"] if ENV["SSL_CERT_FILE"]
+
+    resp = http.get uri.request_uri
+    resp.code.must_equal "200"
+    resp["Content-Type"].must_equal "application/octet-stream"
+    resp["Content-Disposition"].must_equal "attachment; filename=\"google-cloud.png\""
+  end
+
   it "should create a signed delete url" do
     file = bucket.create_file files[:logo][:path], "CloudLogoSignedUrlDelete.png"
 
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -731,6 +731,8 @@ module Google
         #   Private Key.
         # @param [OpenSSL::PKey::RSA, String] private_key Service Account's
         #   Private Key.
+        # @param [Hash] query Query string parameters to include in the signed
+        #   URL, such as response-content-type and response-content-disposition.
         #
         # @example
         #   require "google/cloud/storage"
@@ -778,14 +780,16 @@ module Google
         #
         def signed_url method: nil, expires: nil, content_type: nil,
                        content_md5: nil, headers: nil, issuer: nil,
-                       client_email: nil, signing_key: nil, private_key: nil
+                       client_email: nil, signing_key: nil, private_key: nil,
+                       query: nil
           ensure_service!
           signer = File::Signer.from_file self
           signer.signed_url method: method, expires: expires, headers: headers,
                             content_type: content_type,
                             content_md5: content_md5,
                             issuer: issuer, client_email: client_email,
-                            signing_key: signing_key, private_key: private_key
+                            signing_key: signing_key, private_key: private_key,
+                            query: query
         end
 
         ##
--- a/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
@@ -109,7 +109,7 @@ module Google
             fail SignedUrlUnavailable unless i && s
 
             sig = generate_signature s, signature_str(options)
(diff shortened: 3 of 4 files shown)


google-cloud-storage/lib/google/cloud/storage/file/signer.rb
```diff
@@ -109,7 +109,7 @@ def signed_url options
             fail SignedUrlUnavailable unless i && s
 
             sig = generate_signature s, signature_str(options)
-            generate_signed_url i, sig, options[:expires]
+            generate_signed_url i, sig, options
```
blowmage:
```ruby
generate_signed_url i, sig, options[:expires], options[:query]
```

google-cloud-storage/lib/google/cloud/storage/file/signer.rb
```diff
@@ -120,10 +120,22 @@ def generate_signature signing_key, secret
             Base64.strict_encode64(signature).delete("\n")
           end
 
-          def generate_signed_url issuer, signed_string, expires
-            "#{ext_url}?GoogleAccessId=#{CGI.escape issuer}" \
-              "&Expires=#{expires}" \
+          def generate_signed_url issuer, signed_string, options
```
blowmage:
```ruby
def generate_signed_url issuer, signed_string, expires, query
```

google-cloud-storage/test/google/cloud/storage/signed_url_test.rb
```diff
@@ -102,6 +102,26 @@
...
   end
 
+  it "allows response content type and disposition to be passed in as options" do
+    Time.stub :now, Time.new(2012,1,1,0,0,0, "+00:00") do
+      signing_key_mock = Minitest::Mock.new
+      signing_key_mock.expect :sign, "native-signature", [OpenSSL::Digest::SHA256, "GET\n\n\n1325376300\n/bucket/file.ext"]
+      credentials.issuer = "native_client_email"
+      credentials.signing_key = signing_key_mock
+
+      signed_url = file.signed_url response_content_type: "image/png",
```
blowmage:
```ruby
signed_url = file.signed_url query: { "response-content-type"        => "image/png",
                                      "response-content-disposition" => "attachment; filename=\"test.png\"" }
```

google-cloud-storage/lib/google/cloud/storage/file/signer.rb
```diff
@@ -120,10 +120,22 @@ def generate_signature signing_key, secret
...
+            url = "#{ext_url}?GoogleAccessId=#{CGI.escape issuer}" \
+              "&Expires=#{options[:expires]}" \
               "&Signature=#{CGI.escape signed_string}"
+
+            if options[:response_content_type]
```
blowmage:
```ruby
if query
  query.each do |name, value|
    url << "&#{CGI.escape name}=#{CGI.escape value}"
  end
end
```

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -778,14 +786,19 @@ def public_url protocol: :https
         #
         def signed_url method: nil, expires: nil, content_type: nil,
```
blowmage: Since the additional query parameters don't affect the signed URL's signature, we think its better to add a single optional argument `query` that will apply any additional query parameters:

```ruby
def signed_url method: nil, expires: nil, content_type: nil,
                content_md5: nil, headers: nil, issuer: nil,
                client_email: nil, signing_key: nil, private_key: nil,
                query: nil
```

This is similar to how the `headers` argument is handled.

google-cloud-storage/acceptance/storage/file_test.rb
```diff
@@ -485,6 +485,27 @@
     end
   end
 
+  it "bucket should create a signed read url with response content type and disposition" do
+    local_file = File.new files[:logo][:path]
+    file = bucket.create_file local_file, "CloudLogoSignedUrlGetBucket.png"
+
+    five_min_from_now = 5 * 60
+    url = bucket.signed_url file.name, method: "GET",
```
blowmage:
```ruby
url = bucket.signed_url file.name, method: "GET",
                                   expires: five_min_from_now,
                                   query: { "response-content-type"        => "application/octet-stream",
                                            "response-content-disposition" => "attachment; filename=\"google-cloud.png\"" }
```

georgeclaghorn:
Thanks! Made those changes.

blowmage:
FYI, the acceptance tests show that response-content-disposition works, but response-content-type doesn't.

quartzmo:
@georgeclaghorn Did you see that in #1580, @blowmage apparently found it necessary to remove `response-content-type` from the acceptance test in order to get a passing build?

quartzmo:
Apparently, the service did not change the content type when a different type was passed using the `response-content-type` query param. I just want you to be aware of this; we will release an update containing your addition today. Thanks for your contribution!

georgeclaghorn:
I saw that. Sorry for the trouble!

It looks like GCS ignores `response-content-type` for an object if it was uploaded with a Content-Type; `response-content-type` is only honored if the object has no Content-Type. Likewise, `response-content-disposition` is ignored if the Content-Disposition field is populated in the object’s metadata.

I’m not sure if that’s intended behavior. An aside [in the docs](https://cloud.google.com/storage/docs/access-control/signed-urls) (scroll down to the note just above Signing Strings with the App Engine Identity Service) seems to suggest it is.

quartzmo:
Thanks for the link. The official docs there state:

> Note: Query String Parameters like response-content-disposition and response-content-type are not verified by the signature. To force a Content-Disposition or Content-Type in the response, set those parameters in the object metadata 

I am considering adding these sentences to our documentation for `query`.

blowmage:
Gotcha, thanks for pointing that out. I agree that seems to be the intended behavior.

@quartzmo Should we revisit the docs to make the behavior of these query parameters more clear?

quartzmo:
@blowmage I think we should.

quartzmo:
Released in google-cloud-storage 1.3 and added to activestorage via rails/activestorage#42.

PR #1582, opened 2017-07-07
Update #signed_url query parameter docs

quartzmo:
Follow-up for #1574 to clarify service behavior.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -734,7 +734,13 @@ module Google
         # @param [OpenSSL::PKey::RSA, String] private_key Service Account's
         #   Private Key.
         # @param [Hash] query Query string parameters to include in the signed
-        #   URL, such as response-content-disposition.
+        #   URL. The given parameters are not verified by the signature.
+        #
+        #   Parameters such as `response-content-disposition` and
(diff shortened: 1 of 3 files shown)


google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -734,7 +734,11 @@ def create_file file, path = nil, acl: nil, cache_control: nil,
         # @param [OpenSSL::PKey::RSA, String] private_key Service Account's
         #   Private Key.
         # @param [Hash] query Query string parameters to include in the signed
-        #   URL, such as response-content-disposition.
+        #   URL. The given parameters are not verified by the signature.
+        #   Parameters such as `response-content-disposition` and
+        #   `response-content-type` are only included when the file metadata is
```
blowmage: I think "included" might lead to some confusion. The given parameters will always be applied to the URL. But the behavior of the HTTP response using those parameters will only be applied when the file's metadata is missing those values. Maybe "active" instead?

blowmage: How about:

Query string parameters to include in the signed URL. The given parameters are not verified by the signature.

Parameters such as `response-content-disposition` and `response-content-type` can alter the behavior of the response when using the URL, but only when file resource is missing the corresponding values. (These values can be permanently set using {File#content_disposition=} and {File#content_type=}.)

PR #1594, opened 2017-07-14
Update Spanner cleanup

blowmage:
Update the cleanup task to remove all instances and databases.
We are no longer sharing a project for Spanner developement.


quartzmo:
Great! Glad you remembered about this and removed it!

blowmage:
I found it trying to figure out why `rake acceptance:cleanup` wasn't removing the instances. The way the task was written goes back to when we were using a shared project with the other languages during development.

PR #1595, opened 2017-07-14
Add #error to Spanner Instance::Job and Database::Job

quartzmo:
Also, raise a meaningful error if a job errors in Spanner acceptance setup.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/database/job.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/database/job.rb
@@ -122,6 +122,35 @@ module Google
           end
 
           ##
+          # The status code and message if the operation associated with this
+          # job produced an error, returned as an error object.
+          #
+          # @return [Google::Cloud::Error, nil] An error object containing the
+          #   status code and message, or `nil` if no error occurred.
+          #
+          # @example
+          #   require "google/cloud/spanner"
+          #
+          #   spanner = Google::Cloud::Spanner.new
+          #
+          #   job = spanner.create_instance "my-new-instance",
+          #                                 name: "My New Instance",
+          #                                 config: "regional-us-central1",
+          #                                 nodes: 5,
+          #                                 labels: { production: :env }
+          #
+          #   job.error? # true
+          #
+          #   error = job.error
+          #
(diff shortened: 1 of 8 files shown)


google-cloud-spanner/lib/google/cloud/spanner/database/job.rb
```diff
@@ -122,6 +122,32 @@ def error?
...
+          #   job = spanner.create_database "my-instance",
+          #                                 "my-new-database"
+          #
+          #   job.error? # true
+          #
+          #   error = job.error
+          #
+          def error
+            return nil unless error?
+            fail GRPC::BadStatus.new @grpc.error.code, @grpc.error.message
```
blowmage: In order to set `Error#cause` we need to raise the error. So I suppose this code should look more like this:

```ruby
begin
  begin
    fail GRPC::BadStatus.new @grpc.error.code, @grpc.error.message
  rescue GRPC::BadStatus => grpc_error
    fail Google::Cloud::Error.from_error grpc_error
  end
rescue Google::Cloud::Error => error
  return error
end
```

Or, we could just return the Google::Rpc::Status object, leaking the low level API.

quartzmo: While I'm not in favor of leaking the low-level API, I am beginning to think a status object is more appropriate than returning an un-raised Error, which just seems a little off-pattern.

quartzmo: I'll update this PR to use a status object of some sort.

google-cloud-spanner/lib/google/cloud/spanner/status.rb
```diff
@@ -0,0 +1,76 @@
...
+        def initialize code, message, details
+          @code = code
+          @message = message
+          @details = details
+        end
+
+        ##
+        # @private New Status from a Google::Rpc::Status object.
+        def self.from_grpc grpc
+          code_sym = grpc_code_description_for grpc.code
```
blowmage: I would rather see the code symbol be a separate attribute. The numeric code value will be valuable.

PR #1606, opened 2017-07-21
Update documentation

blowmage:
Update documentation for AsyncPublisher and Subscriber classes. Ensure documented behavior matches requested features, including minor changes for the following:

- [x] Inventory is per Subscriber, not per Stream
- [x] Separate thread pool values
- [x] Add `async` options to `Project#create_topic`

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
@@ -86,11 +86,30 @@ module Google
         # @param [String] topic_name Name of a topic.
         # @param [String] project If the topic belongs to a project other than
         #   the one currently connected to, the alternate project ID can be
-        #   specified here.
+        #   specified here. Optional.
         # @param [Boolean] skip_lookup Optionally create a {Topic} object
         #   without verifying the topic resource exists on the Pub/Sub service.
         #   Calls made on this object will raise errors if the topic resource
-        #   does not exist. Default is `false`.
+        #   does not exist. Default is `false`. Optional.
+        # @param [Hash] async A hash of values to configure the topic's
+        #   {Topic::AsyncPublisher} that is created when {Topic#publish_async}
+        #   is called. Optional.
+        #
+        #   Hash keys and values may include the following:
+        #
+        #   * `:max_bytes` (Integer) The maximum size of messages to be
+        #     collected before the batch is published. Default is 10,000,000
+        #     (10MB).
+        #   * `:max_messages` (Integer) The maximum number of messages to be
+        #     collected before the batch is published. Default is 1,000.
+        #   * `:interval` (Numeric) The number of seconds to collect messages
+        #     before the batch is published. Default is 0.25.
+        #   * `:threads` (Hash) The number of threads to create to handle
+        #     concurrent calls by the publisher:
+        #     * `:publish` (Integer) The number of threads used to publish
+        #       messages. Default is 4.
+        #     * `:callback` (Integer) The number of threads to handle the
+        #       published messages' callbacks. Default is 8.
         #
         # @return [Google::Cloud::Pubsub::Topic, nil] Returns `nil` if topic
         #   does not exist.
@@ -119,6 +138,23 @@ module Google
         #   pubsub = Google::Cloud::Pubsub.new
         #   topic = pubsub.topic "another-topic", skip_lookup: true
         #
+        # @example Configuring AsyncPublisher to increase concurrent callbacks:
+        #   require "google/cloud/pubsub"
+        #
+        #   pubsub = Google::Cloud::Pubsub.new
+        #   topic = pubsub.topic "my-topic",
+        #                        async: { threads: { callback: 16 } }
+        #
+        #   topic.publish_async "task completed" do |result|
+        #     if result.succeeded?
+        #       log_publish_success result.data
+        #     else
+        #       log_publish_failure result.data, result.error
+        #     end
+        #   end
+        #
+        #   topic.async_publisher.stop.wait!
+        #
         def topic topic_name, project: nil, skip_lookup: nil, async: nil
           ensure_service!
           options = { project: project }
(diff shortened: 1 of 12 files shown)


blowmage:
@mdietz94 @lukesneeringer Do you have any input on the documentation changes or the new defaults? Specifically we changed the default number of streams per subscriber to 4, since that was the sweet spot during performance testing. Also, the new max bytes is 8MB, up from 5MB, to allow for larger batches since the API's limit is 10MB. Thoughts?

mdietz94:
Technically the API limits for a single batch is 20 MB, not 10 MB, since it is possible for us to batch two single 10MB user messages together.

mdietz94:
In other languages we set the max bytes to 10 Mb (10^6). Is there a reason you set it lower?

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -111,10 +161,18 @@ def flush
...
+          ##
+          # Checks if the publisher has been started.
+          #
+          # @return [boolean] `true` when started, `false` otherwise.
           def started?
             !stopped?
           end
 
+          ##
+          # Checks if the publisher has been stopped.
```
quartzmo: "Checks if" suggests to me that it makes an RPC. How about "Whether the..." instead for these booleans?

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -50,6 +81,10 @@ def initialize topic_name, service, max_bytes: 5242880,
             super()
           end
 
+          ##
+          # Add a message to the async publisher to be published to the topic.
+          # Messages will be collected in batches and published at together.
```
quartzmo: s/at together/together/

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -23,17 +23,48 @@ module Cloud
     module Pubsub
       class Topic
         ##
-        # # AsyncPublisher
+        # Topic AsyncPublisher object used to publish multiple messages in
```
quartzmo: s/Topic AsyncPublisher object used/Used/

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -308,7 +328,7 @@ def wait_for_messages max: 100
         #
         #   subscriber.start
         #
```
quartzmo: What do you think about adding an example demonstrating the `threads` hash?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -88,11 +143,29 @@ def wait!
...
+        ##
+        # Checks if the subscriber has been started.
+        #
+        # @return [boolean] `true` when started, `false` otherwise.
+        def started?
+          synchronize { @started }
+        end
+
+        ##
+        # Checks if the subscriber has been stopped.
```
quartzmo: "Checks if" suggests to me that it makes an RPC. How about "Whether the..." instead for these booleans?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -66,8 +109,15 @@ def start
           self
         end
 
+        ##
+        # Stops the subscriber.
```
quartzmo: What happens after stop is called? Is there more to the story of calling `stop` on the streams?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
```diff
@@ -55,8 +91,15 @@ def initialize subscription_name, callback, deadline: nil, streams: nil,
           super() # to init MonitorMixin
         end
 
+        ##
+        # Starts the subscriber.
```
quartzmo: Is there anything more to tell about what happens?

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
```diff
@@ -144,10 +182,10 @@ def topic topic_name, project: nil, skip_lookup: nil, async: nil
         #   pubsub = Google::Cloud::Pubsub.new
         #   topic = pubsub.create_topic "my-topic"
         #
```
quartzmo: What do you think about adding an example demonstrating the `async` config?

blowmage:
The original requested maximum was 5MB, so 8MB is an attempt to meet somewhere in the middle.

blowmage:
The 5MB max was requested by @lukesneeringer [here](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1530#issuecomment-307799776). Luke, am I correct in thinking that you would defer to @mdietz94 on this?

google-cloud-pubsub/lib/google/cloud/pubsub/topic/batch_publisher.rb
```diff
@@ -48,8 +48,8 @@ def initialize data = nil, attributes = {}
           end
 
           ##
-          # Add multiple messages to the topic.
-          # All messages added will be published at once.
+          # Add a message to the batch to be published to the topic.
+          # All messages in the batch added will be published at once.
```
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: I intentionally left examples off since I don't expect most users will want to change the default configuration.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/project.rb
blowmage: Same as my Subscription comment, I intentionally left examples off since I don't expect most users will want to change the default configuration. But that's also why you're assigned to review. :)

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: I don't know if the method is the right place for that, or if some guide/overview should explain that. I don't have a great vision for what users want to know when looking at the method, so I kept it as brief as would be expected.

I followed Speech's [Stream#start](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-speech/v0.25.0/google/cloud/speech/stream?method=start-instance) documentation here as well.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb
blowmage: See my comment for `#start`.

mdietz94:
To Luke's point linked above, we enforce a 10MB limit, so using 10^6 is enough to guarantee against rounding errors etc. We use the same serialized proto for checking sizes on both sides, there is no encoding differences, so we've found this to be safe.

blowmage:
FWIW, I agree with you. I've updated these values to 10^6/10MB. But I can change them if others want to make the case for different values.

PR #1608, opened 2017-07-21
Pub/Sub Beta

blowmage:
This PR will merge the pubsub WIP branch into master, and represents another opportunity to review the pending Pub/Sub Beta changes prior to release. It includes the introduction of a Publisher object to collect messages and publish them in batches when a threshold is met (batch count, batch size, batch age).

```ruby
require "google/cloud/pubsub"

topic = pubsub.topic "my-topic"
topic.publish_async "task completed" do |result|
  if result.succeeded?
    log_publish_success result.data
  else
    log_publish_failure result.data, result.error
  end
end

# time passes...

topic.async_publisher.stop.wait!
```

It also includes a Subscriber object, that pulls using GRPC's streaming capability, placing received messages in inventory until they are handled by the user.

```ruby
require "google/cloud/pubsub"

sub = pubsub.subscription "my-topic-sub"

subscriber = sub.listen do |msg|
  # process msg
  msg.ack!
end

subscriber.start

# time passes...

subscriber.stop.wait!
```


blowmage:
@mdietz94 @tbkneeland I don't have permissions to add you to the reviewers, so please add a review if you are able to weigh in on this PR.

mdietz94:
I'll review it now, thanks!

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -0,0 +1,375 @@
...
+
+                    perform_callback_async rec_msg
+                  end
+                end
+                synchronize { pause_streaming! }
+              rescue StopIteration
+                break
+              end
+            end
+          rescue GRPC::DeadlineExceeded
```
mdietz94: This should be the same logic as for Unavailable and Cancelled errors. I would have them all do exponential backoff with a min of immediately. If for some reason you're receiving deadline exceeded because of a server error, which is possible, you would not want to retry as fast as possible.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -286,65 +281,76 @@ def pull immediate: true, max: 100, autoack: false
...
-          pull immediate: false, max: max, autoack: autoack
+        def wait_for_messages max: 100
+          pull immediate: false, max: max
         end
 
         ##
-        # Poll the backend for new messages. This runs a loop to ping the API,
-        # blocking indefinitely, yielding retrieved messages as they are
-        # received.
+        # Create a {Subscriber} object that polls the backend for new messages
```
mdietz94: Subscriber does not 'poll'

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_delayer.rb
```diff
@@ -0,0 +1,173 @@
...
+            return true if ack_ids.empty?
+
+            synchronize do
+              ack_ids.each do |ack_id|
+                if @batch.nil?
+                  @batch = Batch.new max_bytes: @max_bytes
+                  @batch.add deadline, ack_id
+                else
+                  unless @batch.try_add deadline, ack_id
+                    publish_batch!
```
mdietz94: publish_batch seems like a misleading name. Maybe flush_request! or flush_acks!

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_acknowledger.rb
```diff
@@ -0,0 +1,177 @@
...
+                  @cond.wait
+                else
+                  # still waiting for the interval to publish the batch...
+                  @cond.wait(@interval - time_since_first_publish)
+                end
+              end
+            end
+          end
+
+          def publish_batch!
```
mdietz94: This is a misleading name, maybe send_request!, flush_request! etc. but we don't want to confuse this with publishing

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_delayer.rb
```diff
@@ -0,0 +1,173 @@
...
+
+module Google
+  module Cloud
+    module Pubsub
+      class Subscriber
+        ##
+        # @private
+        # # AsyncDelayer
+        #
+        class AsyncDelayer
```
mdietz94: I would prefer we used the term 'modify ack deadline' or 'extend lease' instead of delay. It's inconsistent with our other libraries.

google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
```diff
@@ -51,6 +52,33 @@ def initialize
           @grpc = Google::Pubsub::V1::Topic.new
           @name = nil
           @exists = nil
```
mdietz94: Seems inconsistent that the subscriber is its own file but the publisher is under topic. I would prefer that Publisher is separate from Topic, for consistency within Ruby and with other client libraries.

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
```diff
@@ -0,0 +1,323 @@
...
+            msg = create_pubsub_message data, attributes
+
+            synchronize do
+              fail "Can't publish when stopped." if @stopped
+
+              if @batch.nil?
+                @batch ||= Batch.new self
+                @batch.add msg, block
+              else
+                unless @batch.try_add msg, block
```
mdietz94: We should consider messages that are by themselves larger than the batching limits. i.e. max batched bytes is 10 and we publish a 100 byte message. We would here be adding it to the batch without publish_batch being called immediately. I think it were better that if the message was larger than the batch we published the batch and it by itself immediately.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
```diff
@@ -340,21 +340,6 @@ module Cloud
...
-    #
-    # pubsub = Google::Cloud::Pubsub.new
-    #
-    # sub = pubsub.subscription "my-topic-sub"
-    # sub.listen autoack: true do |msg|
-    #   # process msg
-    # end
-    # ```
-    #
     # ## Configuring timeout
```
blowmage: Thanks. Great catch. This has been updated.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: Good catch, this is updated.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_delayer.rb
blowmage: This class is renamed to `AsyncModifyDeadliner`. (Please note that this class is an internal class, and not part of the public API. Users will not see this in the documentation.)

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_delayer.rb
blowmage: Renamed to `push_batch_request!`. (Also not part of the public API.)

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_acknowledger.rb
blowmage: Renamed to `push_batch_request!`. (Also not part of the public API.)

google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
blowmage: Both `AsyncPublisher` and `BatchPublisher` have been moved from being nested under `Topic` to the top level.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
blowmage: This rescue is specifically to work around a bug in the GAPIC code, where the GAPIC code kills the stream when reaching the timeout. So this rescue here is only a workaround and not part of the standard error handling for API calls.

The UNAVAILABLE and CANCELLED errors are already properly handled with retry and backoff in the GAPIC code when creating the stream by calling the StreamingPull. My understanding is that the code using the stream should not be retrying with backoff _on_ the stream, since the stream is already created.

@mdietz94 Is it be acceptable to handle UNAVAILABLE and CANCELLED errors on the stream object by recreating the stream?

google-cloud-pubsub/lib/google/cloud/pubsub/topic/async_publisher.rb
blowmage: I've updated the code to publish immediately when the batch is ready.

blowmage:
> If there's a good way for me to install this on a VM right now (experimental gem or something to that effect), I would also like to run some load tests. Otherwise I can run them after this PR is submitted.

If you are using bundler the easiest way is to add the following to your Gemfile:

```ruby
source "https://rubygems.org"

gem "google-cloud-pubsub", github: "GoogleCloudPlatform/google-cloud-ruby", branch: "pubsub"
```

Or you can clone the repo and checkout the `pubsub` branch.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
mdietz94: Yes, SGTM

blowmage: Okay, these errors have been added and the stream will restart if they are raised.

blowmage:
@mdietz94 All of your comments have been addressed. Can you please re-review this PR?

mdietz94:
I've been integrating this with the load test framework. I've just run the results for the Publisher. We see reasonable publish to ack latency (50%ile - 30ms, 95%ile - 70ms, 99%ile - 175ms, 99.9%ile 225ms) and the throughput peaks at about 10 MB/s. This is equivalent to Java when only running on a single core, which may be the best we are able to do, though we should discuss if there are strategies users can use to easily run this publisher on multiple processes, since in Java we are able to reach ~500 MB/s on a 16 core VM.

I'll run the subscriber tomorrow and assuming that gives expected results I'll be able to sign off on this PR by the end of the week.

lukesneeringer:
> This is equivalent to Java when only running on a single core, which may be the best we are able to do, though we should discuss if there are strategies users can use to easily run this publisher on multiple processes, since in Java we are able to reach ~500 MB/s on a 16 core VM.

This is a fundamental language difference, and I think that given the timeout issues that we observed, the answer to this is likely to be "use multiple VMs".

mdietz94:
I've been running load tests and it appears that there is an issue clearing the inventory, where messages are leaked even though I am trying to ack every single message. Below is the relevant snippet:

    @subscriber = sub.listen do |msg|
      @semaphore.synchronize do
        @latencies.push (Time.now.to_f * 1000.0 - msg.attributes['sendTime'].to_f).to_i
        @recv_msgs.push MessageIdentifier.new(publisher_client_id: msg.attributes['clientId'].to_i, sequence_number: msg.attributes['sequenceNumber'].to_i)
      end
      msg.ack!
    end
    @subscriber.start


I'm seeing 45 QPS of Ack and 88 QPS of Mod Ack, and after some point the stream stalls. This looks like we're hitting the inventory limits due to some sort of leak. Any ideas? Would there be any logging in an exception was thrown inside this block?

blowmage:
How many streams and what inventory size are you using? I've seen some messages sent to different streams, and acking one one stream won't clear the inventories of the other streams.

mdietz94:
Default inventory size and 4 streams. I'm using that same callback to ack all messages. I think the problem is that if an exception is thrown are we handling it? In other languages we are nacking those messages. Allowing leaks to continuously extend leases is definitely a bug.

blowmage:
Are you sure there is an unhandled exception? Can you add `Thread.abort_on_exception = true` and verify? That will take any unhandled exception on a child thread and throw it on the main thread.

mdietz94:
I'm not sure - I'll try that next - but if there isn't an exception the leak is even more serious, since this means we're somehow not removing from inventory when we ack

mdietz94:
That did not cause the main thread to crash - so I assume that means that it is not throwing an exception. But it's definitely leaking messages somehow. My code is available on the ruby branch of GoogleCloudPlatform/pubsub if you want to take a look.

blowmage:
I'm still looking at this. Nothing to report yet.

mdietz94:
Okay, if there's any logging that would be helpful let me know.

blowmage:
The behavior you are describing seems more likely to be a deadlock, so that is what I'm unraveling now. I observed a deadlock a couple weeks ago, but resolved it. I'm currently investigating but progress has been slow.

blowmage:
@mdietz94 How quickly do you see the streams stop? Is it fairly quick? Like in the first minute?

mdietz94:
Yeah, it's within the first 2 minutes.

mdietz94:
Obviously it depends on QPS, but I'd say at least 25% of the messages leak, so after 4000 messages it gets stuck.

blowmage:
@mdietz94 I think the core issue may not be about deadlocks or memory leaks, but about a bad refactoring made without tests. We have applied a fix to this branch. Can you update to the latest version of this branch and re-run your load tests? More detail below...

---

During a documentation change we noticed that the `inventory` value was being applied to each stream, and not to the subscription. So while we added documentation that the inventory was for all streams on a subscription, we also changed the code to calculate the inventory per-steam. Only we didn't get the [calculation right](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1608/commits/ae9651779d33ba2cce530d4f0674a03be8e86e10#diff-e600415e05effff51b8d7eb4234e84fcR83) and we didn't test it. The end result is that the per-stream inventory was being set to `0`, which caused all the streams to pause their connections soon after starting them. So your described behavior makes total sense, as each of the 4 default streams received 1,000 messages before they paused for being over the inventory limit. The fix (with tests) was applied in [this commit](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1608/commits/8a034ea2cff5f15434101a7a94778ee869b2e120).

Unfortunately, my personal branch with all my benchmarking code was not rebased on the documentation commit, so I was working on code that did not have this bug. However, I have not been able to reproduce a memory leak, an inventory leak, or a deadlock. I'm as confident as someone who made such an egregious mistake can be that the publisher and subscriber should work. Please let me know what you find when you re-run your tests.

mdietz94:
The tests work! I think we should potentially see if there's any way to improve the throughput numbers for GA, but certainly LGTM for Beta.

blowmage:
Excellent! Can you approve the PR?

I think we might be stuck with the current level of performance. The nature of the GIL means that Ruby will have at most one thread active at any point in time. There is no way to get parallelism with a GIL.

blowmage:
Oops, I didn't see your earlier approval. Thanks!

PR #1613, opened 2017-07-26
Storage user_project Changes

blowmage:
I think I have found a couple edge cases where `user_project` wasn't being used when it should have been. When a Bucket object has `user_project` set it will send the value when listing the bucket's files, but the File objects don't also have `user_project` set. The second is `user_project` isn't getting set properly when calling `Bucket#reload!`.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -1132,7 +1132,7 @@ module Google
         # Reloads the bucket with current data from the Storage service.
         def reload!
           ensure_service!
-          @gapi = service.get_bucket name
+          @gapi = service.get_bucket name, user_project: user_project
         end
         alias_method :refresh!, :reload!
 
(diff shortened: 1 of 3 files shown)


quartzmo:
The requirements for this feature are for an option to specify the ID of the project which is willing to be billed for the operation. This option should be opt-in on a per-request basis, not blindly performed for every operation. I think strict adherence to this is difficult, and I think we have already departed from it with `File::List`. So I think I'm in agreement with this PR, but I am wondering if there are other stakeholders who may have a different opinion.

quartzmo:
@jskeet Does the propagation of the `user_project` in this change look OK to you, with regard to the original requirement paraphrased in my comment above?

blowmage:
The requirements also allow for OO libs to set the value on the object that makes the API calls, which is what we have done. So we are in agreement with the requirements. 

I believe a Bucket object with the setting should return File objects with the setting. I think it was an oversight the setting is dropped.

jskeet:
@quartzmo: Not understanding Ruby, I'm not sure what I'm looking at, or what would be idiomatic. I think it would be fine to do whatever feels "right" from a Rubyists perspective - while I'm the point-of-contact for this feature in broad terms, I believe languages have a fair amount of latitude to implement in a language-idiomatic way.

blowmage:
Thanks @jskeet!

PR #1615, opened 2017-07-26
Storage File#delete generation

blowmage:
Allow users to remove older versions of files.

Also update the acceptance cleanup task to remove all generations of files.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -639,6 +639,11 @@ module Google
         # Permanently deletes the file.
         #
         # @return [Boolean] Returns `true` if the file was deleted.
+        # @param [Boolean, Integer] generation Specify a version of the file to
+        #   delete. When `true`, it will delete the version returned by
+        #   {#generation}. The default behavior is to delete the latest version
+        #   of the file (regardless of the version to which the file is set,
+        #   which is the version returned by {#generation}.)
         #
         # @example
         #   require "google/cloud/storage"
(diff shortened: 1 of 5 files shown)


google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -639,6 +639,9 @@ def rotate encryption_key: nil, new_encryption_key: nil
         # Permanently deletes the file.
         #
         # @return [Boolean] Returns `true` if the file was deleted.
+        # @param [Boolean,Integer] generation Select a specific revision of the
+        #   file to delete. When `true` it will use the object's generation. The
```
quartzmo: I'm confused here. What is the value returned by `#generation`, if it is not the "latest version"? In other words, what is the purpose of passing `true` vs. the default. (Also, google-cloud-storage uses "file" rather than "object".)

blowmage: If you call `Bucket#file` with the `generation` argument you can get an object with that specific generation version. It does not have to be the latest generation version. But the current behavior is to always delete the latest version, so this is an attempt to preserve the existing behavior.

quartzmo: Sorry, I don't understand. What is the difference between passing `generation: true` and not passing `generation` at all?

blowmage: Calling `file.delete generation: true` is equivalent to passing `file.delete generation: file.generation`

quartzmo: OK, I understand now. How about this:

> Specify a version of the file to delete. When `true`, it will delete the version returned by {#generation}. The default behavior is to delete the latest version of the file (regardless of the version returned by {#generation}.)

quartzmo: ^ updated

blowmage: Agreed, much better.

blowmage: PR updated.

frankyn: @blowmage, what are the downsides if this parameter isn't used and `File#delete` assumes to delete the current generation used when getting the file?

blowmage: If the File always uses it's local `generation` value then it is a change in behavior. But that complicates things when we can create a File object that has skipped the lookup, because it won't have the generation loaded. Then a looked up object behaves differently than a skipped looked up object. With all the recent Storage changes being made I think this approach is the right way to go.

PR #1616, opened 2017-07-26
Regenerate Speech GAPIC with latest protos

geigerj:
This change includes the word time offset feature.


blowmage:
We have the following style violations on the v1.rb version file:

```
lib/google/cloud/speech/v1.rb:24:81: C: Line is too long. [85/80]

    # In order to use this library, you first need to go through the following steps:

                                                                                ^^^^^

lib/google/cloud/speech/v1.rb:55:81: C: Line is too long. [142/80]

    # - Read the [Google Cloud Speech API Product documentation][Product Documentation] to learn more about the product and see How-to Guides.

                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

lib/google/cloud/speech/v1.rb:56:125: C: Line is too long. [174/80]

    # - View this [repository's main README](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/README.md) to see the full list of Cloud APIs that we cover.
```

Historically this file was has not been provided by the GAPIC generation, and something that we have written. We can either massage this file or we can exclude it from the style check. Thoughts?

geigerj:
@blowmage I just merged a PR to disable rubocop on those comment lines: https://github.com/googleapis/toolkit/pull/1431#pullrequestreview-52478756. I'll regenerate this afternoon.

As a longer-term solution, I'm going to look into running the generated code through a formatter (possibly also rubocop?) before it hits this repo.

google-cloud-speech/test/google/cloud/speech/v1/speech_smoke_test.rb
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# AUTO GENERATED BY GAPIC
+
+require "minitest/autorun"
+require "minitest/spec"
+require "google/cloud/speech/v1/speech_client"
+
+describe "SpeechSmokeTest" do
```
blowmage: This test makes a call to the API and fails when run an environment that does not have gcloud SDK credentials. We haven't seen this test fail on CI because of the style violations, but I believe it will fail because CI is not configured gcloud SDK credentials. 

If a smoke test it wanted, then it should be an acceptance test since it makes an API call. But it should also be written in such a way that it will work without gcloud SDK, since that will not be available on CI.

geigerj: Ack. I will remove this file from the PR and we can make updates in the code generator independently of this Speech change.

quartzmo:
Circle CI seems a bit backed up, but I just ran speech `ci` and `acceptance` locally for this branch and they pass for me.

Issue #1601, opened 2017-07-19
Add code examples demonstrating usage of file generation/versions


frankyn:
Hi! 

Is it possible to list available file generations using the Storage Client? Also is it possible to select a generation when accessing a file from a bucket?

Thank you!

quartzmo:
Yes, both of these are supported:

1. To list available generations for a file, call [`Bucket#files`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/bucket.rb#L493) with the `prefix` option set to the name of the file and the `versions` option set to `true`. The `versions` option:
    > If `true`, lists all versions of an object  as distinct results. The default is `false`. 
1. To select a generation when accessing a file from a bucket, call [`Bucket#file`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/bucket.rb#L534) with the `generation` option set to the desired generation number. This option:
    > When present, selects a specific revision of this object. Default is the latest version.

For general information on this topic, see [Object Versioning](https://cloud.google.com/storage/docs/object-versioning).

frankyn:
Thanks @quartzmo! I missed these two options in the docs. As always your help is greatly appreciated!

quartzmo:
@frankyn Feel free to make a case for adding code examples to demonstrate file versioning if you feel it is important.

frankyn:
Reopened to use as a request: 

I was doing a parity check on functionality of the Storage client. After reviewing the documentation, it would be helpful to have examples of using different generations of a file. The library uses version and generation interchangeably and that confused me when searching for versioning related information about the client. 

Example: `Bucket#file` uses parameter generation and `Bucket#files` uses versions.

quartzmo:
> it would be helpful to have examples of using different generations of a file.

I feel this is a valid request. We can update the title of this issue to match it.

> Bucket#file uses parameter generation and Bucket#files uses versions.

The option names `generation` and `versions` match the names in the service API. For continuity with official documentation, I feel it is important to keep these names unchanged. However, adding a discussion of versioning to our [Storage guide](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/cloud/storage) should help to clarify their meaning.

blowmage:
[...]

PR #1619, opened 2017-07-27
Add File#generations method

blowmage:
The `File#generations` method returns a list of `File` objects for each generation of the path in GCS.
It is the same as calling the following:

```ruby

file = bucket.file "/path/to/my/file.ext"
generations = bucket.files prefix: file.name, versions: true
```

[refs #1601]

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -47,7 +47,7 @@ describe Google::Cloud::Storage::File, :storage do
   end
 
   after do
-    bucket.files.all { |f| f.delete rescue nil }
+    bucket.files(versions: true).all { |f| f.delete generation: true rescue nil }
   end
 
   it "should upload and download a file" do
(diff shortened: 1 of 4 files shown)


google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -123,6 +123,15 @@ def bucket
           @gapi.bucket
         end
```
quartzmo: Missing docs?

google-cloud-storage/acceptance/storage/file_test.rb
```diff
@@ -47,7 +47,7 @@
   end
 
   after do
-    bucket.files.all { |f| f.delete rescue nil }
+    bucket.files(versions: true).all { |f| f.delete generation: true rescue nil }
```
quartzmo: Is this needed to delete the file completely (including previous versions)?

The [DELETE Object](https://cloud.google.com/storage/docs/xml-api/delete-object) docs state:

> If this query parameter is used on a versioned bucket and the object is the current one, no archive of the object is made, but instead the object is deleted completely.

blowmage: Yes, I was unable to delete the bucket because there were versioned files that don't show when calling `bucket.files`, but do show when calling `bucket.files versions: true`. The acceptance helper and cleanup task were already updated to use this approach.

google-cloud-storage/lib/google/cloud/storage/file.rb
blowmage: Yes. I'll fix.

blowmage: Updated.

Issue #1588, opened 2017-07-13
Caller does not have storage.buckets.get access


mtricolici:
```
gcloud auth activate-service-account --key-file "service-account.json"
gsutil ls gs://some-bucket ### this works fine!
```

via ruby it doesn't work using the same service-account.json! why?!

```
storage = Google::Cloud::Storage.new(
      project: 'someproject', keyfile: 'service-account.json')
storage.bucket('some-bucket').files ...
```

...
forbidden: Caller does not have storage.buckets.get access to bucket 'some-bucket'

mtricolici:
I don't need to list buckets! I just want to list all files in a bucket that service account has access to!

quartzmo:
Unfortunately, it is not possible to list files without access to the bucket that contains them. This is due to the current design of the library, which requires that the bucket is loaded before listing its files.

> why?!

I believe one of the founding goals of this project was a "clean, OOP-inspired design", which comes at some cost to flexibility. For more flexibility, the [google-api-client/Google/Apis/StorageV1/StorageService](http://www.rubydoc.info/gems/google-api-client/Google/Apis/StorageV1/StorageService) offers a "flatter" API, although authentication is a bit more involved.

mtricolici:
I've managed to list bucket files/dirs (using the same service account) using the code bellow:
`storage.service.list_files( ...`

quartzmo:
Please note that calling `storage.service.list_files( ...` as suggested above is NOT RECOMMENDED as it uses undocumented access to `storage.service` to use the underlying [google-api-client](http://www.rubydoc.info/gems/google-api-client/Google/Apis/StorageV1/StorageService) implementation. This implementation is subject to change at any time without warning, and has in fact been changed in many of the other packages in google-cloud-ruby. Should it change in the future for google-cloud-storage, the above code will no longer work.

quartzmo:
I can think of two possible solutions for this issue:

1. Expose [`Objects.list`](https://cloud.google.com/storage/docs/json_api/v1/objects/list) as a top-level method (`Project#files`) that accepts the bucket name.
1. Add a mode (option) to `Project#bucket` to return a stub containing the given bucket name without retrieving the bucket metadata. This stub could then be used to call `Bucket#files` without permissions to the bucket.

quartzmo:
[...]

PR #1620, opened 2017-07-27
Add skip_lookup to Storage

blowmage:
This PR adds the `skip_lookup` argument to `Project#bucket` and `Bucket#file`, which will allow for users to make method calls on `Bucket` and `File` without first retrieving the full resource. 

```ruby

bucket = storage.bucket "my-todo-app", skip_lookup: true
file = bucket.file "avatars/heidi/400x400.png", skip_lookup: true
file.download "/var/todo-app/avatars/heidi/400x400.png" # API call
```

This can also be used in conjunction with `user_project`:

```ruby

bucket = storage.bucket "my-todo-app", skip_lookup: true, user_project: true
file = bucket.file "avatars/heidi/400x400.png", skip_lookup: true
file.download "/var/todo-app/avatars/heidi/400x400.png" # API call
```
[refs #1588]


blowmage:
@remi and @frankyn, can you look this over as well?

frankyn:
@blowmage thanks for doing this! I tried using the skip_lookup and hit a bug when I try downloading a file after skipping the lookup. 

Steps to reproduce:
```ruby

bucket  = storage.bucket "bucket_name", user_project: true, skip_lookup: true 
file        = bucket.file "file_name", skip_lookup: true
file.download "local_file_name"
```

Error
```ruby
Google::Cloud::Storage::FileVerificationError:
       The downloaded file failed MD5 verification.
```

frankyn:
I updated my reproduction steps with more detail on permissions used.

remi:
@blowmage Interface LGTM!

I haven't tested to see why this may be breaking for @frankyn. Any thoughts @blowmage?

blowmage:
I think the verifier code is assuming the MD5 hash is always present, and the value is missing for a lazy File object. I'm pretty sure we expect that the MD5 hash value can be missing like it is for encrypted files, so I'm not entirely sure why its raising. I was hoping to look deeper into this today, but it will have to wait until tomorrow. Sorry.

blowmage:
I've added a fix for the verification on download, PTAL.

frankyn:
Thanks @blowmage, when will the storage gem be updated with these changes?

blowmage:
We will likely release Storage at the same time as Pub/Sub. I don't yet know if that will be this week or next.

frankyn:
Ah okay, this is currently blocking Ruby requester pays samples. Is it possible to push it out this sooner sometime this week?

blowmage:
Yeah, I'll plan on releasing Storage tomorrow morning, assuming I can get a green build between now and then. :)

frankyn:
Ah sorry about that. I appreciate your support! Thank you.

blowmage:
This is [released in 1.4.0](https://github.com/GoogleCloudPlatform/google-cloud-ruby/releases/tag/google-cloud-storage%2Fv1.4.0).

PR #1622, opened 2017-07-28
Speech Word Information

blowmage:
This PR adds support for returning word time offsets in the recognition results. The `words` argument is added to all appropriate recognize methods, and `Result#words` is added to return an array of `Result::Word` objects that represent the word with the start and end times.

### DIFF
--- a/google-cloud-speech/acceptance/speech/async_test.rb
+++ b/google-cloud-speech/acceptance/speech/async_test.rb
@@ -166,4 +166,21 @@ describe "Asynchonous Recognition", :speech do
     results.first.confidence.must_be_close_to 0.9, 0.1
     results.first.alternatives.must_be :empty?
   end
+
+  it "recognizes audio with words" do
+    op = speech.process filepath, encoding: :linear16, language: "en-US", sample_rate: 16000, words: true
+
+    op.must_be_kind_of Google::Cloud::Speech::Operation
+    op.wont_be :done?
+    op.wait_until_done!
+    op.must_be :done?
+
+    results = op.results
+    results.count.must_equal 1
+    results.first.transcript.must_equal "how old is the Brooklyn Bridge"
+    results.first.confidence.must_be_close_to 0.9, 0.1
+    results.first.words.wont_be :empty?
+    results.first.words.map(&:word).must_equal %w{how old is the Brooklyn Bridge}
+    results.first.alternatives.must_be :empty?
+  end
 end
--- a/google-cloud-speech/lib/google/cloud/speech/audio.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/audio.rb
@@ -191,6 +191,10 @@ module Google
         #   phrases "hints" so that the speech recognition is more likely to
         #   recognize them. See [usage
         #   limits](https://cloud.google.com/speech/limits#content). Optional.
+        # @param [Boolean] words When `true`, return a list of words with
+        #   additional information about each word. Currently, the only
+        #   additional information provided is the the start and end time
+        #   offsets. See {Result#words}. Default is `false`.
         #
         # @return [Array<Result>] The transcribed text of audio recognized.
         #
@@ -209,14 +213,15 @@ module Google
         #   result.transcript #=> "how old is the Brooklyn Bridge"
         #   result.confidence #=> 0.9826789498329163
         #
-        def recognize max_alternatives: nil, profanity_filter: nil, phrases: nil
+        def recognize max_alternatives: nil, profanity_filter: nil,
+                      phrases: nil, words: nil
           ensure_speech!
 
           speech.recognize self, encoding: encoding, sample_rate: sample_rate,
                                  language: language,
                                  max_alternatives: max_alternatives,
                                  profanity_filter: profanity_filter,
-                                 phrases: phrases
+                                 phrases: phrases, words: words
         end
 
         ##
@@ -239,6 +244,10 @@ module Google
         #   phrases "hints" so that the speech recognition is more likely to
         #   recognize them. See [usage
         #   limits](https://cloud.google.com/speech/limits#content). Optional.
+        # @param [Boolean] words When `true`, return a list of words with
+        #   additional information about each word. Currently, the only
+        #   additional information provided is the the start and end time
+        #   offsets. See {Result#words}. Default is `false`.
         #
         # @return [Operation] A resource represents the long-running,
         #   asynchronous processing of a speech-recognition operation.
@@ -260,7 +269,7 @@ module Google
         #   results = op.results
         #
         def process max_alternatives: nil, profanity_filter: nil,
-                    phrases: nil
+                    phrases: nil, words: nil
           ensure_speech!
 
           speech.process self, encoding: encoding,
@@ -268,7 +277,7 @@ module Google
                                language: language,
                                max_alternatives: max_alternatives,
                                profanity_filter: profanity_filter,
-                               phrases: phrases
+                               phrases: phrases, words: words
         end
         alias_method :long_running_recognize, :process
         alias_method :recognize_job, :process
--- a/google-cloud-speech/lib/google/cloud/speech/convert.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/convert.rb
@@ -1,0 +1,46 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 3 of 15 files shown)


blowmage:
Is there anyone else we should add to review this?

google-cloud-speech/lib/google/cloud/speech/convert.rb
```diff
@@ -0,0 +1,46 @@
...
+        module ClassMethods
+          def number_to_duration number
+            return nil if number.nil?
+
+            Google::Protobuf::Duration.new \
+              seconds: number.to_i,
+              nanos: (number.remainder(1) * 1000000000).round
+          end
+
+          def duration_to_number duration
```
lukesneeringer: Should this return a `rational` in order to match the `-` method on `Date`?

lukesneeringer:
@blowmage Nope, we are enough. No review from the Speech team is required.

google-cloud-speech/lib/google/cloud/speech/convert.rb
blowmage: When we have mapped a Duration protobuf value to a ruby object on other languages we have always used Numeric. This is duration value is consistent with common rails duration methods like `3.hours` or `2.days`. We don't lose any precision since Ruby uses a 64 bit float. Also, Rational inherits Numeric, so the behavior of `Date#-` will be consistent.

google-cloud-speech/lib/google/cloud/speech/audio.rb
```diff
@@ -260,15 +265,15 @@ def recognize max_alternatives: nil, profanity_filter: nil, phrases: nil
         #   results = op.results
         #
         def process max_alternatives: nil, profanity_filter: nil,
-                    phrases: nil
+                    phrases: nil, word_times: nil
```
quartzmo: I strongly feel that this parameter should be named `words` (rather than `word_times`), because it directs the service to return "a list of words". (I know this is a departure from my usual philosophy of maintaining continuity with the underlying API(s), but in this case I believe `words` is a much clearer name.)

google-cloud-speech/lib/google/cloud/speech/result.rb
```diff
@@ -114,10 +123,40 @@ class Alternative
...
           def initialize transcript, confidence
-            @transcript  = transcript
+            @transcript = transcript
             @confidence = confidence
           end
         end
+
+        ##
+        # Word-specific information for recognized words. Word information is
+        # only included in the response when certain request parameters are
```
quartzmo: "certain request parameters..." Is there any way to provide definite information here?

blowmage: That wording comes from the protobuf. It doesn't provide any other context and I'm not aware what other parameters it could mean. I removed that phrase in other documentation entries, but it slipped through here. I'll remove it.

google-cloud-speech/lib/google/cloud/speech/audio.rb
blowmage: K, I'll change this unless someone else feels strongly that `word_times` is better.

google-cloud-speech/lib/google/cloud/speech/result.rb
blowmage: Removed.

google-cloud-speech/lib/google/cloud/speech/audio.rb
blowmage: Updated.

pbogle:
Nit: the description of the CL should be changed to say `words` instead of `word_times` to be consistent with the actual code.

google-cloud-speech/acceptance/speech/async_test.rb
```diff
@@ -166,4 +166,21 @@
     results.first.confidence.must_be_close_to 0.9, 0.1
     results.first.alternatives.must_be :empty?
   end
+
+  it "recognizes audio with words" do
+    op = speech.process filepath, encoding: :linear16, language: "en-US", sample_rate: 16000, words: true
```
pbogle: This should be called "enable_word_time_offsets" for consistency with other APIs and the upcoming auto-generated client libraries, and for future-proofness.

word_time_offsets are enabled individually, and we will add other word level features in the future which will have their own flags, so calling this "words" is too generally.

google-cloud-speech/lib/google/cloud/speech/audio.rb
```diff
@@ -191,6 +191,8 @@ def url?
         #   phrases "hints" so that the speech recognition is more likely to
         #   recognize them. See [usage
         #   limits](https://cloud.google.com/speech/limits#content). Optional.
+        # @param [Boolean] words When `true`, return a list of words and
```
pbogle: As mentioned the naming (enabled_word_time_offsets) and documentation for this field should be made consistent with the proto.  e.g. note the use of "(timestamps)" in the description.

  // *Optional* If `true`, the top result includes a list of words and
  // the start and end time offsets (timestamps) for those words. If
  // `false`, no word-level time offset information is returned. 
  bool enable_word_time_offsets = 8;

google-cloud-speech/lib/google/cloud/speech/project.rb
```diff
@@ -266,6 +266,8 @@ def audio source, encoding: nil, language: nil, sample_rate: nil
         #   phrases "hints" so that the speech recognition is more likely to
         #   recognize them. See [usage
         #   limits](https://cloud.google.com/speech/limits#content). Optional.
+        # @param [Boolean] words When `true`, return a list of words and
```
pbogle: enable_word_time_offsets here as well.

google-cloud-speech/lib/google/cloud/speech/result.rb
```diff
@@ -114,10 +123,39 @@ class Alternative
...
+        #   beginning of the audio, and corresponding to the start of the
+        #   spoken word. This field is only set if `words` was specified
+        #   and only in the top hypothesis. This is an experimental feature
+        #   and the accuracy of the time offset can vary.
+        # @attr_reader [Numeric] end_time Time offset relative to the
+        #   beginning of the audio, and corresponding to the end of the spoken
+        #   word. This field is only set if `words` was specified and
+        #   only in the top hypothesis. This is an experimental feature and
+        #   the accuracy of the time offset can vary.
+        class Word
```
pbogle: Please call this WordInfo for consistency with the proto.

google-cloud-speech/acceptance/speech/async_test.rb
quartzmo: What about grouping the configuration for words in a hash?

`words: { enable_word_time_offsets: true }`

blowmage: `words: { times: true }`

pbogle: We are migrating all speech client libraries to auto-generated client libraries. To minimize the impact on developers when we make that transition, we as service authors would strongly prefer that the manually generated libraries should use names consistent with what the auto-generated libraries will use for this feature; i.e. enable_word_time_offsets.

@lukesneeringer FYI.

quartzmo: `words: { time_offsets: true }` has very little information loss.

lukesneeringer: @quartzmo I think Phil's point here is, pretty soon, developers will have to write `enable_word_time_offsets` soon, so they should write it now to ease migration.

blowmage: The high-level ruby API is already so structurally different than the low-level Protobuf API that I don't think making these names consistent is going to ease any migration between them.

geigerj:
@pbogle When we introduce the partial-GAPIC library, we'll need to migrate users from the existing Ruby library anyway, since the two libraries will have fundamentally different surfaces. I am not sure how much benefit there is in enforcing consistency with the protos for this one particular feature at the cost of creating inconsistency in naming patterns with the rest of the existing library/Ruby convention.

google-cloud-speech/acceptance/speech/async_test.rb
pbogle: If consistency with the OnePlatform reviewed names isn't important,  I like "timestamps" better than "time_offsets", because it's more familiar to our customers. @lukesneeringer What would JJG say about these renamings?

blowmage: We usually use "timestamp" to mean a full date/time value, not a Duration. 

Our current direction of favoring Ruby norms over the API names/structures was initiated by JJ. I assume he is still in favor of it.

pbogle: OK, this will be moot fairly shortly so no point debating if you feel strongly.

pbogle:
LGTM

google-cloud-speech/acceptance/speech/async_test.rb
blowmage: _nods_ Looking forward to seeing what you guys have in store.

blowmage:
Updated the documentation to indicate that there might be more word-specific information in the future.

google-cloud-speech/lib/google/cloud/speech/convert.rb
lukesneeringer: Great. Thanks!

PR #1645, opened 2017-08-03
Dynamic log label API

bestie:
This PR introduces dynamic log labelling and a Rails-like API for configuration them for Rack applications.

Label values can take the form of either static strings or a
user-defined function (any object that responds to `#call`) which is
passed the current Rack request environment at the moment of generating
the log entry.

The label key remains static and the return value of the function
becomes the label value in the log entry.

The tests show an example of writing labelling log entries with a value
from an HTTP header.

We use this in production to tie together all traces from a user session.

### DIFF
--- a/google-cloud-logging/.rubocop.yml
+++ b/google-cloud-logging/.rubocop.yml
@@ -9,7 +9,6 @@ AllCops:
     - "Rakefile"
     - "support/**/*"
     - "test/**/*"
-    - "lib/google/cloud/logging/rails.rb"
 
 Documentation:
   Enabled: false
--- a/google-cloud-logging/.rubocop.yml
+++ b/google-cloud-logging/.rubocop.yml
@@ -9,6 +9,7 @@ AllCops:
     - "Rakefile"
     - "support/**/*"
     - "test/**/*"
+    - "lib/google/cloud/logging/rails.rb"
 
 Documentation:
   Enabled: false
--- a/google-cloud-logging/lib/google/cloud/logging/rails.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/rails.rb
@@ -63,6 +63,7 @@ module Google
           resource_type = Logging.configure.monitored_resource.type
           resource_labels = Logging.configure.monitored_resource.labels
           log_name = Logging.configure.log_name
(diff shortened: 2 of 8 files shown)


google-cloud-logging/.rubocop.yml
```diff
@@ -9,6 +9,7 @@ AllCops:
     - "Rakefile"
     - "support/**/*"
     - "test/**/*"
+    - "lib/google/cloud/logging/rails.rb"
```
hxiong388: Seems like the addition to `Google::Cloud::Logging::Railtie.merge_rails_config` method is exceeding the AbcSize limit. I think it's better to just exempt that one method from the AbcSize check, instead of removing the entire file from being linted. Something like this would do:
```ruby
def self.merge_rails_config rails_config # rubocop:disable AbcSize
  ...
end
```

google-cloud-logging/lib/google/cloud/logging/rails.rb
```diff
@@ -63,14 +63,15 @@ def self.init_middleware app
           resource_type = Logging.configure.monitored_resource.type
           resource_labels = Logging.configure.monitored_resource.labels
           log_name = Logging.configure.log_name
+          labels = Logging.configure.labels
```
hxiong388: Please add an entry in the [configuration instruction doc](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/stackdriver/docs/configuration.md) for this new configuration option.

hxiong388:
@bestie Thanks for submitting the PR. The changes look great, just left couple comments on the change.

google-cloud-logging/.rubocop.yml
bestie: Good point. I did attempt to reduce the size of the method but nothing I could think of was materially better.

google-cloud-logging/lib/google/cloud/logging/rails.rb
bestie: Will do!

bestie:
My pleasure, thanks for the fast feedback.

All comments seem like good ideas to me, will make changes ASAP.

PR #1653, opened 2017-08-07
Pub/Sub test fix

blowmage:
This PR is attempting to diagnose why this test is occasionally failing on CircleCI.

### DIFF
--- a/google-cloud-pubsub/test/google/cloud/pubsub/async_publisher_test.rb
+++ b/google-cloud-pubsub/test/google/cloud/pubsub/async_publisher_test.rb
@@ -315,7 +315,7 @@ describe Google::Cloud::Pubsub::AsyncPublisher, :mock_pubsub do
     mock.expect :publish, big_publish_res, [topic_path(topic_name), [big_message], options: default_options]
     pubsub.service.mocked_publisher = mock
 
-    # 190 is bigger than 10 messages, but less than 11.
+    # the big message's 120 bytes is bigger than the max bytes limit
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/test/google/cloud/pubsub/async_publisher_test.rb
```diff
@@ -344,6 +344,8 @@
 
     callbacks.must_equal 2
 
+    sleep 0.1
```
hxiong388: `sleep` statement in async tests is still flaky, because it doesn't guarantee the child thread runs during the sleep. It's better to use a blocking check to verify some sort of state, such as [this](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/ca518c824e0db4070148604fe1f66a35bd204a51/stackdriver-core/test/stackdriver/core/async_actor_test.rb#L241-L243).

blowmage: Right. The thing is, we have the blocking check in the method `wait!`. I suspect that the problem here is actually with the Mock object. But I don't know for sure. I cannot reproduce locally, so I'm going to use this PR to try to reproduce on CI.

hxiong388: Kk

blowmage:
We skipped the test for now. Unsure why it sometimes fails on CI.

Issue #1623, opened 2017-07-28
Use getQueryResults to wait for queries


quartzmo:
`QueryJob#wait_until_done!` and the high-level methods `Project#query` and `Dataset#query` should use  `jobs.getQueryResults` to wait for completion.

- [x] The implementation should generate a job ID on the client (#1624).
- [x] Call `jobs.insert` (not `jobs.query`)
  - [x] With generated ID to create the job
  - [x] With retry and backoff on any transient error, *including internal error (HTTP 5xx)*.
- [x] Extract the `configuration.query.destinationTable` field from the response of `jobs.insert` and remember it.
- [x] Use `getQueryResults` with `maxResults=0` to wait for the query job to finish. 
  - [x] Retry on the usual errors
  - [x] Also retry if the response's `jobComplete` field is false. (Suggested retry parameters: `initial=1s, max=60s, multiplier=2`.)
- [x] Get the schema from the response of `getQueryResults`.
- [x] Call `Tabledata.list` with the destination table from the earlier step. Use the schema to decode the rows.

quartzmo:
In this requirement:

> Also retry if the response's jobComplete field is false. (Suggested retry parameters: initial=1s, max=60s, multiplier=2.)

Each call to `getQueryResults` with `maxResults=0` blocks for a response. We do not understand the point of adding additional wait time to the loop, so marking this as done. Please open a new issue to discuss if we are mistaken.

quartzmo:
Closed by #1657.

PR #1657, opened 2017-08-07
Bigquery query job

quartzmo:
This PR introduces new implementations of `Project#query` and `Dataset#query` that create a job and wait for it to complete by calling `getQueryResults`, then read data from the temporary table created by the job. To support this, the PR updates `QueryJob#wait_until_done!` to use `getQueryResults`.

Breaking changes:

* Remove timeout and dryrun parameters from `Project#query`
* Change return type of `Project#query` from QueryData to Data
* Remove timeout and dryrun parameters from `Dataset#query`
* Change return type of `Dataset#query` from QueryData to Data   
* Remove timeout parameter from `QueryJob#query_results`
* Change return type of `QueryJob#query_results` from QueryData to Data 
* Remove timeout and dryrun parameters from `View#data`
* Change return type of `View#data` from QueryData to Data
* Remove `QueryData`    

[refs #1623]


quartzmo:
@blowmage Regarding the idea we had about also updating `View#data` to read using `TableDataList`, I don't think it's possible:

> You cannot use the TableDataList JSON API method to retrieve data from a view. 

from [Using Views](https://cloud.google.com/bigquery/docs/views).

quartzmo:
I'm going to see if the same approach of a Query Job writing to a temporary table works for view queries. It seems like it should.

blowmage:
We added `View#data` to be API compatible with `Table#data`, but we could always deprecate/remove it.

quartzmo:
I'm adding a commit to the PR updating `View#data`, but I won't mind removing it (and the method) if you think that's a better long-run choice after looking at the commit.

PR #1674, opened 2017-08-16
Support ETag in BigQuery Dataset and Table updates

### LINKED ISSUE #1629: Support ETags in Dataset CRUD
CRUD operations on datasets should use the ETag header for optimistic locking. See [Using patch in a read-modify-write cycle](https://cloud.google.com/bigquery/docs/api-performance#read-patch-write).

Questions:

* Opt-in or opt-out?
* Raise error on outdated ETag?
* Provide some kind of force overwrite?

### LINKED ISSUE #1630: Support ETags in Table CRUD
CRUD operations on tables should use the ETag header for optimistic locking. See [Using patch in a read-modify-write cycle](https://cloud.google.com/bigquery/docs/api-performance#read-patch-write).

Questions:

* Opt-in or opt-out?
* Raise error on outdated ETag?
* Provide some kind of force overwrite?

quartzmo:
This PR sends the `If-Match` header with the current ETag value on the client for every update to Dataset and Table. The use of the ETag is transparent, but not optional. If the local ETag value does not match the ETag value on the service, a `FailedPreconditionError` will be raised.

The final 3 commits of the PR also add additional RPC `get` calls to load valid ETag values, due to unusual behavior in the service. See discussion on #1629 for details.

[closes #1629, closes #1630]

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/advanced_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/advanced_test.rb
@@ -80,6 +80,7 @@ describe Google::Cloud::Bigquery, :advanced, :bigquery do
     empty_table.schema.field("spells").headers.wont_include :score
     empty_table.schema.field("spells").headers.must_include :properties
     empty_table.schema.field(:spells).field(:properties).headers.wont_include :grade
+    empty_table.reload! # TODO: remove after fixing etag staleness after create_table
     empty_table.schema do |schema|
       # adds to a nested field directly inline
       schema.field(:spells).integer :score, mode: :nullable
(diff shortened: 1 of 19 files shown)


blowmage:
@jskeet wrote: 

> One thing to be aware of is that the etag in the JSON returned when you create a table/dataset is currently different from the etag you'll get if you immediately fetch the object again.

I think we should add a call to `get` after `create` so that the returned resource always has a valid etag. When the server-side implementation changes we can remove the `get` call.

quartzmo:
@blowmage OK, I'll do that, along with acceptance testing that will fail on server-side implementation change and prompt us to remove the `get` call.

quartzmo:
@blowmage I believe this PR is complete and ready for review.

Issue #1634, opened 2017-07-28
Expose Job.statistics.query.queryPlan


quartzmo:
Expose `Job.statistics.query.queryPlan` as a custom class. See [Jobs](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#resource).

quartzmo:
Question: We currently do not expose `statistics` as a custom class, but rather simply as a hash. Should (including the `statistics.query` fields other than `queryPlan`) also be exposed as a custom class?

quartzmo:
> Should (including the statistics.query fields other than queryPlan) also be exposed as a custom class?

The answer to this is no. We expose a number of useful fields from the `statistics` data in custom methods, such as [`QueryJob#bytes_processed`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/query_job.rb#L101-L105).

The array of custom `QueryJob::QueryPlan` objects should be exposed in the same way.

quartzmo:
Closed by #1684.

PR #1684, opened 2017-08-23
Add QueryJob#query_plan

quartzmo:
Expose `Job.statistics.query.queryPlan`.

Add `QueryJob::Stage` and `QueryJob::Step` classes.

[refs #1634]

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/table_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/table_test.rb
@@ -253,6 +253,36 @@ describe Google::Cloud::Bigquery::Table, :bigquery do
     query_job.data.class.must_equal Google::Cloud::Bigquery::Data
     query_job.data.total.wont_be :nil?
 
+    # Query Job - Statistics Query Plan
+    query_job.query_plan.wont_be_nil
+    query_job.query_plan.must_be_kind_of Array
+    query_job.query_plan.wont_be :empty?
+    stage = query_job.query_plan.first
+    stage.must_be_kind_of Google::Cloud::Bigquery::QueryJob::Stage
+    stage.compute_ratio_avg.must_be_kind_of Float
+    stage.compute_ratio_max.must_be_kind_of Float
+    stage.id.must_be_kind_of Integer
+    stage.name.must_be_kind_of String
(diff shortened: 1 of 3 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/query_job.rb
```diff
@@ -106,6 +106,18 @@ def bytes_processed
         end
 
         ##
+        # Describes the execution plan for the query.
+        #
+        # @return [Array<Google::Cloud::Bigquery::QueryJob::Stage>]
+        #
+        def query_plan
+          return nil unless @gapi.statistics.query.query_plan
```
blowmage: What if `@gapi.statistics` is nil? Or `@gapi.statistics.query` is nil?

quartzmo: The existing readers already assume these will always be there. And I think that's correct wrt the API. However, `query_plan` is only available for 7 days per the docs, and therefore may be `nil`.

```ruby
##
        # Checks if the query results are from the query cache.
        def cache_hit?
          @gapi.statistics.query.cache_hit
        end
```

blowmage: That's right. I think we observed that a query job always has `statistics.query`, which is why it isn't guarded anywhere else.

google-cloud-bigquery/acceptance/bigquery/table_test.rb
```diff
@@ -253,6 +253,36 @@
     query_job.data.class.must_equal Google::Cloud::Bigquery::Data
     query_job.data.total.wont_be :nil?
 
+    # Query Job - Statistics Query Plan
+    query_job.query_plan.wont_be_nil
+    query_job.query_plan.must_be_kind_of Array
+    query_job.query_plan.wont_be :empty?
+    stage = query_job.query_plan.first
```
quartzmo: @blowmage Are you OK with `#query_plan` returning an array, even though it sounds like a singular noun? Pretty misleading if you ask me, but I'm just sticking to the underlying API.

blowmage: Yeah, I'd stick with the name.

PR #1698, opened 2017-08-30
Fix GAPIC YARD docs

blowmage:
This change is needed to correct the GAPIC documentation by moving the types to the same line as the method argument. Otherwise the types will be defined as `null` in the JSON documentation files, and the angular app will puke on itself.


blowmage:
@geigerj I already fixed the JSON output manually, and this PR will fix the ruby documentation so the JSON output will be generated correctly. But it would be good to fix this in the GAPIC generator so the the next time these files are generated we don't see this problem again.

geigerj:
Thanks, @blowmage. We'll fix this in the generator. Tracking as https://github.com/googleapis/toolkit/issues/1530

PR #1702, opened 2017-08-31
Add Video Intelligence v1beta2 GAPIC

geigerj:


geigerj:
@blowmage After this is merged, could you create a point release?

blowmage:
@geigerj Do you want `V1beta2` added to the documentation along side `V1beta1`?

geigerj:
Ah, right... I'll update the Rakefile.

blowmage:
We will need to update the dependency on `google-gax` to `0.8.5` in the gemspec since this new code depends on  `Google::Gax::Credentials`.

blowmage:
> After this is merged, could you create a point release?

I'm not sure what you mean by "point release". The current release is `0.21.0`. I would tend towards releasing these changes as a new minor version, like `0.22.0`. Can you confirm that this is _not_ what you want, and would rather see this as a `0.21.1` patch release instead?

blowmage:
I can follow this PR up with another PR making the additional changes needed.

geigerj:
Updated Rakefile and gemspec.

Re: versions, please follow whatever convention you've been following. For GAX, we've used a minor release increment to indicate a breaking change when in major version 0, but it's fine to use it here if that's what you've been doing for large changes like a new API version.

google-cloud-video_intelligence/google-cloud-video_intelligence.gemspec
```diff
@@ -20,7 +20,8 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 0.8.0"
+  gem.add_dependency "google-gax", "~> 0.8.6"
+  gem.add_dependency "googleapis-common-protos", "~> 1.3.1"
```
blowmage: The `google-gax` gem already [declares](https://rubygems.org/gems/google-gax/versions/0.8.6) a dependency on `googleapis-common-protos` with `~> 1.3.5`, so I don't think this line is necessary.

geigerj: This is autogenerated -- it's configured at as an API-level dependency because the API's protos depend on common protos, so this GAPIC depends directly on types defined in the common protos gem. It should be harmless, right?

blowmage: Yes, this is unnecessary and harmless.

geigerj: Acknowledged. I would rather leave this in, then. It probably does not make sense for the code generator to be aware of GAX's dependencies.

blowmage:
I am the author of these commits.

blowmage:
@quartzmo I had to massage the YARD formatting to get these changes to pass the CI build, as well as to add the new classes to the documentation output. Can you look them over and give some feedback? Is this the right approach?

geigerj:
@blowmage Thank you for the fixes. Once we're set on the right YARD formatting, I'll get fixes in to the generator.

quartzmo:
@blowmage Your YARD fixes look good to me!

PR #1706, opened 2017-09-06
Print captured exceptions from child threads

hxiong388:
The Error Reporting, Trace, and Debugger APIs are not enabled by default on GCP projects, so the gems often appear to be broken with no errors when use on new projects. This PR changes the gems to print the captured exceptions in children threads, except in Logging, which may get too spammy if something goes wrong with Logging API.


hxiong388:
cc @frankyn

jmdobry:
Debugger, Logging, Monitoring, and Trace are now enabled by default on new projects.

frankyn:
Thanks @jmdobry

Issue #1635, opened 2017-07-28
Add insert method that creates table if not found


quartzmo:
Add a new streaming insert method that calls `TableData.insertAll` and creates a destination table if the named table is not found.

Behavior: 

* Add `Dataset#insert`
* If insert does not find the table, try creating using exponential backoff with first delay >= 1m
* Retry insert, create table if failed, wait another minute, repeat

quartzmo:
@blowmage 

> Is `Dataset#insert` the correct method?

Should this be an option on `Dataset#insert` or a new method? (And for either choice, what should the new option or method be named?)

blowmage:
I think adding this functionality in `Dataset#insert` makes a lot more sense.

blowmage:
When we've done this in other libraries we have used the `autocreate` method argument.

quartzmo:
Well, this is embarrassing. It turns out that `Dataset#insert` doesn't exist. Should I add it, with this signature?

` def insert table, rows, skip_invalid: nil, ignore_unknown: nil, create: nil`

blowmage:
Yes, please.

quartzmo:
@tswast The suggestions for retry behavior (above) are a bit contradictory/fuzzy (emphasis mine):

They state: "try creating using **exponential** backoff with first delay >= 1m" and "if failed, wait **another minute**, repeat". Do you think the delay should start at 1 minute, double to 2 minutes, and then exponentially increase (4 min, 16 min...) with no limit? Any other ideas or guidance? Afaik, no other platform has implemented this yet.

quartzmo:
@tswast While working on the PR for this issue, I think I discovered why the requirements ask for a retry backoff that begins with 1 min delay. My [initial approach during development](https://github.com/quartzmo/google-cloud-ruby/blob/8688a02516ff7672306d4da87f5ab626e9f2637b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb#L1161-L1169) was to retry the insert just 10 seconds after creating the missing table. The system test for this implementation consistently fails with `NotFoundError`. Even adding a single delay of 1 min consistently fails with this error, implying that only with a subsequent retry of several minutes is the method likely to finally succeed.

This failure to allow inserts for tables created "mid stream" is a known behavior of the service, as discussed in this [SO answer](https://stackoverflow.com/a/30406747/901096).

[...]

PR #1707, opened 2017-09-06
Bigquery insert

quartzmo:
This is an implementation of #1635 that does not precisely follow the suggested steps of first attempting to insert data, and then if the table is not found, creating the table and then reattempting the insert with backoff of >= 1 minute. Instead, if `autocreate` is true, it first performs a `Tables.get` to quickly check for the table, and if the table is not found, it creates the table and then does the insert. Our experience is that if the initial insert data call encounters a not found error, the service will continue to respond with the error for over one minute. See #1635 for discussion.

/cc @jskeet

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery_helper.rb
+++ b/google-cloud-bigquery/acceptance/bigquery_helper.rb
@@ -65,16 +65,16 @@ module Acceptance
       addl.include? :bigquery
     end
 
-    # def self.run_one_method klass, method_name, reporter
-    #   result = nil
-    #   reporter.prerecord klass, method_name
-    #   (1..3).each do |try|
-    #     result = Minitest.run_one_method(klass, method_name)
-    #     break if (result.passed? || result.skipped?)
-    #     puts "Retrying #{klass}##{method_name} (#{try})"
-    #   end
-    #   reporter.record result
-    # end
+    def self.run_one_method klass, method_name, reporter
+      result = nil
+      reporter.prerecord klass, method_name
+      (1..3).each do |try|
+        result = Minitest.run_one_method(klass, method_name)
+        break if (result.passed? || result.skipped?)
+        puts "Retrying #{klass}##{method_name} (#{try})"
+      end
+      reporter.record result
+    end
   end
 end
 
--- a/google-cloud-bigquery/acceptance/bigquery_helper.rb
+++ b/google-cloud-bigquery/acceptance/bigquery_helper.rb
@@ -65,16 +65,16 @@ module Acceptance
       addl.include? :bigquery
     end
 
-    def self.run_one_method klass, method_name, reporter
-      result = nil
-      reporter.prerecord klass, method_name
-      (1..3).each do |try|
-        result = Minitest.run_one_method(klass, method_name)
-        break if (result.passed? || result.skipped?)
-        puts "Retrying #{klass}##{method_name} (#{try})"
-      end
-      reporter.record result
-    end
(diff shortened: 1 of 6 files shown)


google-cloud-bigquery/acceptance/bigquery_helper.rb
```diff
@@ -65,16 +65,16 @@ def setup
...
-      result = nil
-      reporter.prerecord klass, method_name
-      (1..3).each do |try|
-        result = Minitest.run_one_method(klass, method_name)
-        break if (result.passed? || result.skipped?)
-        puts "Retrying #{klass}##{method_name} (#{try})"
-      end
-      reporter.record result
-    end
+    # def self.run_one_method klass, method_name, reporter
```
blowmage: Was this commented out for testing?

quartzmo: Yes, thanks for catching this, I'll restore it.

tswast:
The backend team would prefer that `table.get` get called much less frequently than `tabledata.insertAll`. My understanding is that they propose the following logic:

1. Try to do `tabledata.insertAll` (get 404)
2. Wait a random delay (they didn't define the time, maybe between 1 and 60 seconds?).
3. Attempt to create the table.
4. Wait 1 minute.
5. Retry `tabledata.insertAll`

quartzmo:
@tswast And if 5. fails, repeat 4. and 5. indefinitely until 5. succeeds?

quartzmo:
Or:  if 5. fails, repeat 2., 3., 4. and 5. indefinitely until 5. succeeds?

Sorry, it's just tough to guess at the aggregate, real-world impact of these calls, so the more precisely you can spec it, the better. Anyway, I'll update this PR soon with a pass at it.

tswast:
My understanding is that we should repeat 2, 3, 4, and 5 on 404 errors, continuing indefinitely. Since this could potentially repeat forever, I'd love to see an optional timeout parameter.

quartzmo:
@tswast Thank you, SGTM. I'll update the PR tomorrow.

quartzmo:
I have updated the PR per the guidance above. The [acceptance test](https://github.com/quartzmo/google-cloud-ruby/blob/f2fc5d40a64b9939b4c8f46c71928f0b506efb4a/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb#L274-L304) now takes several minutes to complete. (@blowmage, how do you feel about this additional time?) To show the sequence followed by the test, I added temporary debug logging statements that are now commented out, but need to be removed completely before accepting this PR. The output of my test runs is consistently as follows:

```sh
# Running:

NotFoundError rescue
sleeping before create_table
sleeping before insert
NotFoundError rescue
sleeping before create_table
AlreadyExistsError rescue
sleeping before insert
.

Finished in 205.142495s, 0.0049 runs/s, 0.1999 assertions/s.
```

@tswast I did not yet add the optional timeout parameter. Given the recursive behavior in the method, we believe it would require a separate thread, which will add some complexity and development time. Should we do it?

tswast:
Hmmm... An extra thread does sound more complex. Probably not worth it.

quartzmo:
@blowmage How do you feel about this  [acceptance test](https://github.com/quartzmo/google-cloud-ruby/blob/f2fc5d40a64b9939b4c8f46c71928f0b506efb4a/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb#L274-L304) adding *two to three minutes* to the build?

Issue #1181, opened 2017-01-05
Display the content for GAPIC/GRPC API version such as 'V1' menu item


swcloud:
Right now, if a user clicks 'V1' menu item on the left navigation panel, the right side displays an empty page. We should add content for this menu item or find a way to not change the content on the right side when the menu item is clicked, i.e., when a user clicks 'V1', what it only does is to expand the sub menu items.

quartzmo:
Currently, when any left nav item is clicked, related content is shown in the main page. Clicking any of the top-level "service" nav items ('BigQuery', 'Datastore', etc) results in the display of an overview discussion of the service. Example:

https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.23.0/google/cloud/datastore

> Google Cloud Datastore is a fully managed, schemaless database for storing non-relational data. You should feel at home if you are familiar with relational databases, but there are some key differences to be aware of to make the most of using Datastore.

I think for consistency we should stick to this pattern. Clicking the 'V1' should display an overview discussion informing the user about purpose and expected usage of the GAPIC API.

danoscarmike:
@omaray to flesh out design.  At minimum we need to provide content / redirect.

geigerj:
@landrito Let's put the README content in the version landing page in the meantime.

danoscarmike:
@landrito @geigerj can you please revisit?  Is priority P0 appropriate here - critical issue / imminent release blocking?  Thanks.

landrito:
I think this can be resolved. @eoogbe made this https://github.com/googleapis/toolkit/pull/1358 which should address this. @geigerj @eoogbe can you confirm?

geigerj:
@danoscarmike This is blocking for all new beta-level releases. I'm not clear on the distinction between P0 and P1 that we're using. As @landrito said, @eoogbe is actively working on this.

danoscarmike:
@swcloud can you please verify and close this issue if resolved?  Thanks!

geigerj:
@danoscarmike this issue is not resolved; there is still WIP. It's not blocking _any_ imminent releases, but it is blocking _beta_ releases. I think P0 is probably still accurate.

danoscarmike:
Understood.  Thanks @geigerj

geigerj:
[...]

PR #1714, opened 2017-09-13
Regenerate Video Intelligence v1beta2

geigerj:


geigerj:
@quartzmo We've moved the location of the table in the code generator to the `doc/` directory. Could you verify if this is the correct fix for the `jsondoc` issues we were having earlier?

quartzmo:
@geigerj I pulled the branch and built the jsondoc, and the content looks correct:

```sh
$ cat jsondoc/google/cloud/videointelligence/v1beta2.json 
{"id":"google/cloud/videointelligence/v1beta2","name":"V1beta2","title":["Google","Cloud","Videointelligence","V1beta2"],"description":"<h1 id=\"google-cloud-video-intelligence-api-contents\">Google Cloud Video Intelligence API Contents</h1>\n\n<table class=\"table\">\n  <thead>\n    <tr>\n      <th>Class</th>\n      <th>Description</th>\n    </tr>\n  </thead>\n...
```

google-cloud-video_intelligence/lib/google/cloud/video_intelligence/v1beta2.rb
```diff
@@ -48,51 +48,55 @@ module Cloud
...
+        #   The OAuth scopes for this service. This parameter is ignored if
+        #   an updater_proc is supplied.
+        # @param client_config [Hash]
+        #   A Hash for call options for each method. See
+        #   Google::Gax#construct_settings for the structure of
+        #   this data. Falls back to the default config if not specified
+        #   or the specified config is missing data points.
+        # @param timeout [Numeric]
+        #   The default timeout, in seconds, for calls made through this client.
+        def self.new \
```
quartzmo: So not all of these options will be documented?

geigerj:
Great, thanks for checking about the jsondoc!

google-cloud-video_intelligence/lib/google/cloud/video_intelligence/v1beta2.rb
geigerj: Right. The undocumented options are deprecated and result in a warning if you use them.

google-cloud-video_intelligence/lib/google/cloud/video_intelligence/v1beta2.rb
```diff
@@ -48,51 +48,55 @@ module Cloud
...
+            updater_proc: nil,
+            credentials: nil,
+            scopes: nil,
+            client_config: nil,
+            timeout: nil,
+            lib_name: nil,
+            lib_version: nil
+          kwargs = Hash[
+            *method(__method__).parameters
+              .select { |_, param| binding.local_variable_get(param) != nil }
```
blowmage: Unfortunately Binding#local_variable_get is not available in Ruby 2.0. This is why the build is failing for this PR.

Compare [Binding in 2.1](https://ruby-doc.org/core-2.1.0/Binding.html) to [Binding in 2.0](https://ruby-doc.org/core-2.0.0/Binding.html).

geigerj: Done.

geigerj:
@blowmage Could you take another look? CI is passing now.

blowmage:
Oh, sorry, was waiting for a green build and I missed the notification.

blowmage:
@geigerj There is a discussion on GoogleCloudPlatform/google-cloud-ruby/#1181 discussing some issues with the links added in this PR. Should we manually fix these links so they work and we can release? Or should we wait for a new GAPIC generation? 

Also, do you want to regen the code for V1beta1 so it matches V1beta2?

geigerj:
@blowmage You can wait for a new GAPIC generation before the release. I'm cherry-picking Eva's fix for the links in the code generator (currently under review).

blowmage:
K, FWIW I think those links will be problematic because they always go to the latest release. If you are looking at a previous release (or master) and click those links they will jump you to latest. That might be frustrating for users who aren't expecting it to change the version they are looking at.

geigerj:
@blowmage Noted. (I copied your note over to #1181.) I am OK with that for now because it seems preferable to showing users a blank page. But we should make a fix in the generator to use a relative link that won't jump to the latest version.

Issue #1720, opened 2017-09-14
Add high-level (synchronous) BigQuery operations


quartzmo:
Add new high-level (synchronous) convenience methods that encapsulate waiting for a job to complete. These should be similar to `#query`, except that polling should use `jobs.get` instead of `jobs.getQueryResults`. The naming of the new methods should be consistent with the pairing of `#query` (high-level) and `#query_job` (low-level). The following methods should be added:

- [x] `#copy`
- [x] `#extract`
- [x] `#load`

Existing low-level methods with these names should be renamed with the `_job` suffix, for consistency as noted above.

quartzmo:
I was originally thinking it would be ok for these methods to return `Job` subtypes, but in response to [your comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1722#pullrequestreview-62923233) on the PR, should they all simply return `true` if successful?

blowmage:
I do. I think the "high level" methods should return `true`, and the "low level" methods return the job object.

quartzmo:
Closed by #1722, #1726, #1730.

PR #1722, opened 2017-09-15
Add high-level, synchronous version of copy

quartzmo:
* Rename `Table#copy` to `Table#copy_job`.
* Add high-level, synchronous version as `Table#copy`.

[refs #1720]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
@@ -20,7 +20,7 @@ module Google
       # # CopyJob
       #
       # A {Job} subclass representing a copy operation that may be performed on
-      # a {Table}. A CopyJob instance is created when you call {Table#copy}.
+      # a {Table}. A CopyJob instance is created when you call {Table#copy_job}.
       #
       # @see https://cloud.google.com/bigquery/docs/tables#copyingtable Copying
       #   an Existing Table
@@ -30,7 +30,7 @@ module Google
       class CopyJob < Job
         ##
         # The table from which data is copied. This is the table on
-        # which {Table#copy} was called. Returns a {Table} instance.
+        # which {Table#copy_job} was called. Returns a {Table} instance.
         def source
           table = @gapi.configuration.copy.source_table
           return nil unless table
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
@@ -30,7 +30,7 @@ module Google
       # {CopyJob}, {ExtractJob}, {LoadJob}, and {QueryJob}.
       #
       # A job instance is created when you call {Project#query_job},
-      # {Dataset#query_job}, {Table#copy}, {Table#extract}, {Table#load}, or
+      # {Dataset#query_job}, {Table#copy_job}, {Table#extract}, {Table#load}, or
       # {View#data}.
       #
       # @see https://cloud.google.com/bigquery/docs/managing_jobs_datasets_projects
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
@@ -599,14 +599,18 @@ module Google
         end
 
         ##
-        # Copies the data from the table to another table. The destination table
-        # argument can also be a string identifier as specified by the [Query
-        # Reference](https://cloud.google.com/bigquery/query-reference#from):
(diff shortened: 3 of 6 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -682,6 +686,82 @@ def copy destination_table, create: nil, write: nil, dryrun: nil,
...
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #   table = dataset.table "my_table"
+        #
+        #   copy_job = table.copy "other-project:other_dataset.other_table"
+        #
+        # @!group Data
+        #
+        def copy destination_table, create: nil, write: nil
+          ensure_service!
```
blowmage: Shouldn't this just call `#copy_job` directly? So all this logic isn't duplicated?

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -682,6 +686,82 @@ def copy destination_table, create: nil, write: nil, dryrun: nil,
...
+        #   the destination table. The default value is `empty`.
+        #
+        #   The following values are supported:
+        #
+        #   * `truncate` - BigQuery overwrites the table data.
+        #   * `append` - BigQuery appends the data to the table.
+        #   * `empty` - An error will be returned if the destination table
+        #     already contains data.
+        #
+        # @return [Google::Cloud::Bigquery::CopyJob]
```
blowmage: I would expect this to return the copied Table. Why is it necessary to return the Job object instead?

quartzmo: By the copied table, do you mean the source table or the destination table?

blowmage: The table the data is copied to.

blowmage: My point here is that I don't think these high level methods should return the job object. Only the lower level `*_job` methods should return the job object.

quartzmo: Let's discuss this for all three methods on #1720.

google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
```diff
@@ -30,8 +30,8 @@ module Bigquery
       # {CopyJob}, {ExtractJob}, {LoadJob}, and {QueryJob}.
       #
       # A job instance is created when you call {Project#query_job},
-      # {Dataset#query_job}, {Table#copy}, {Table#extract}, {Table#load}, or
-      # {View#data}.
+      # {Dataset#query_job}, {Table#copy_job}, {Table#copy}, {Table#extract},
```
blowmage: Remove `, {Table#copy}`

google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
```diff
@@ -30,7 +31,8 @@ module Bigquery
       class CopyJob < Job
         ##
         # The table from which data is copied. This is the table on
-        # which {Table#copy} was called. Returns a {Table} instance.
+        # which {Table#copy_job} or {Table#copy} was called. Returns a {Table}
```
blowmage: Remove `or {Table#copy}`

google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
```diff
@@ -20,7 +20,8 @@ module Bigquery
       # # CopyJob
       #
       # A {Job} subclass representing a copy operation that may be performed on
-      # a {Table}. A CopyJob instance is created when you call {Table#copy}.
+      # a {Table}. A CopyJob instance is created when you call {Table#copy_job}
+      # or {Table#copy}.
```
blowmage: Remove `or {Table#copy}`

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -682,6 +686,77 @@ def copy destination_table, create: nil, write: nil, dryrun: nil,
         end
 
         ##
+        # Copies the data from the table to another table using a synchronous
+        # method that blocks for a response. In this method, a {CopyJob} is
```
blowmage: I wouldn't mention CopyJob here. I would just say the method blocks until the copy operation is complete, or an error is raised.

quartzmo: Ok, I'll update for these.

PR #1732, opened 2017-09-19
BigQuery Error Retry and Backoff

### LINKED ISSUE #1715: Update BigQuery retry behavior
The default retry behavior in [`HttpCommand#execute`](https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/http_command.rb#L81-L120) in google-api-ruby-client does not satisfy the requirements for retrying idempotent BigQuery RPCs. The behavior of google-api-ruby-client should be configured or overridden as necessary, but google-api-ruby-client should be kept as a dependency.

The retry behavior should be updated to comply with the following:

- [x] Retry all idempotent RPCs.
- [x] Begin with a 1 second delay, and increase exponentially to a max delay of 32s
- [x] Retry a fixed number of times or indefinitely (depending just on which is more appropriate for Ruby.)
- [x] Use the following logic to determine if an error should be retried:
       `reason == "rateLimitExceeded" || reason == "backendError"`
       Do not look at HTTP response codes.

**Exceptions**: The create-on-insert feature (#1635) will use different logic.

See also [BigQuery SLA Back-off Requirements](https://cloud.google.com/bigquery/sla).

blowmage:
Disable the Google API Client error retry and handle error retry ourselves.
The rules for the retry are from the Google BigQuery SLA:

    After the first error, there is a minimum back-off interval of 1 second
    and for each consecutive error, the back-off interval increases
    exponentially up to 32 seconds.

https://cloud.google.com/bigquery/sla

[closes #1715]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
@@ -61,7 +61,7 @@ module Google
             service.client_options.open_timeout_sec = timeout
             service.client_options.read_timeout_sec = timeout
             service.client_options.send_timeout_sec = timeout
-            service.request_options.retries = @retries || 3
+            service.request_options.retries = 0 # handle retries in #execute
             service.request_options.header ||= {}
             service.request_options.header["x-goog-api-client"] = \
               "gl-ruby/#{RUBY_VERSION} gccl/#{Google::Cloud::Bigquery::VERSION}"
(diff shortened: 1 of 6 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
```diff
@@ -61,7 +61,7 @@ def service
             service.client_options.open_timeout_sec = timeout
             service.client_options.read_timeout_sec = timeout
             service.client_options.send_timeout_sec = timeout
-            service.request_options.retries = @retries || 3
+            service.request_options.retries = 0 # handle retries manually
```
quartzmo: A pointer might be helpful:
`# handle retries via backoff in #execute`

blowmage: That's too long for the line, so I changed it to `# handle retries in #execute`.

Issue #1736, opened 2017-09-20
Batch-and-flush option for Dataset and Table #insert


quartzmo:
* Buffers rows on client for ~10s before sending.
* Option (or separate method.)
* Needs parameters to control batching.
* Needs mechanism to flush accumulated data when done streaming.
* See google-cloud-python for possible implementation.

blowmage:
Does the create on insert behavior apply here as well? It would be unfortunate do have multiple batches attempt to create the table at the same time...

blowmage:
@tswast ^

quartzmo:
@tswast Do you know if any other platform has implemented this feature? I looked at google-cloud-python, but I didn't find it.

tswast:
I know that Java doesn't have it and neither does Python (it was wrongly marked as done in the tracking spreadsheet).

Not sure about create-on-insert. I'll ping Jeremy in the requirements doc. My guess is the existing insert method would be used by the batching method and inherit the same behavior.

blowmage:
I'll continue on the insert-only path I'm on for this then. Thanks!

blowmage:
@tswast We have done a similar asynchonous batch publisher thing for Pub/Sub. They had three thresholds for publishing a batch:

1) How old the batch is (0.25 seconds)
2) How many messages the batch contains (1,000)
3) The total size of the batch (10MB)

Do you want users to be able to specify similar threshold amounts for the asynchronous batch BigQuery inserter? We had to set special GRPC headers to be able to make 10MB RPC calls to Pub/Sub, so I assume BigQuery has a different max size limit. Do you know what that is?

tswast:
The limit is documented to be 10 MB for streaming insert. https://cloud.google.com/bigquery/quotas#streaminginserts

blowmage:
Oh, that's excellent. There is a max of 10,000 rows per batch, but recommended 500. Perfect. Thanks!

PR #1744, opened 2017-09-22
Asynchronous Insert

blowmage:
AsyncInserter is a new class that will create objects that will collect rows to insert in batches.

```ruby

table = dataset.table "my_table"

inserter = table.insert_async do |response|
  # this is called every time a batch is inserted
  log_insert "inserted #{response.insert_count} rows " \
    "with #{response.error_count} errors"
end

# insert multiple rows one at a time
inserter.insert row1
inserter.insert row2
inserter.insert row3
inserter.insert row4

inserter.flush # publish what was given so far

# insert multiple rows at once
inserter.insert [row5, row6, row7, row8]

# stop the inserter, wait for all threads to end
inserter.stop.wait!
```

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table/async_inserter.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table/async_inserter.rb
@@ -1,0 +1,207 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+require "google/cloud/bigquery/convert"
+require "monitor"
+require "concurrent"
+
+module Google
+  module Cloud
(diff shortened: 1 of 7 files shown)


blowmage:
@tswast Can you look this over and let me know if this satisfies the batch-and-flush enhancement described in #1736?

blowmage:
@quartzmo Remaining TODOs have been addressed. PTAL.

google-cloud-bigquery/lib/google/cloud/bigquery/table/async_inserter.rb
```diff
@@ -0,0 +1,207 @@
...
+                  @cond.wait
+                end
+              end
+            end
+          end
+
+          def push_batch_request!
+            return unless @batch
+
+            batch_rows = @batch.rows.dup
```
quartzmo: Is this `dup` necessary? Can the user handle the additional memory usage?

blowmage:
@jjolma Is this something you want to weigh in on?

blowmage:
> Are retries handled the same as the plain insertAll call?

Yes, the inserter object calls the same `insert` method.

> Is it possible to set the insert ID (basically a primary key for deduplication) for the rows?

We already set an insert_id for each row, which is an MD5 hash of the row contents.

tswast:
> We already set an insert_id for each row, which is an MD5 hash of the row contents.

That concerns me slightly. I can definitely imagine cases where the user inserts duplicate rows expects to see duplicates in the results. It's probably not great schema design on the part of the user, but we might want to allow a way to do it.

tswast:
Not sure if it's something we should change, but it's different from what I expected and what I've seen in the other languages. We should start a discussion with some other folks on the BigQuery team to see if we should add similar behavior in the other language implementations.

PR #1750, opened 2017-09-26
Video v1b2

geigerj:
Regenerates GAPIC with the latest code generator to fix bug where overview page is blank

### DIFF
--- a/google-cloud-video_intelligence/lib/google/cloud/video_intelligence.rb
+++ b/google-cloud-video_intelligence/lib/google/cloud/video_intelligence.rb
@@ -43,13 +43,9 @@ module Google
     # input_uri = "gs://cloud-ml-sandbox/video/chicago.mp4"
     # features_element = :LABEL_DETECTION
     # features = [features_element]
-    # input_content = ''
-    # video_context = {}
-    # output_uri = ''
-    # location_id = ''
     #
     # # Register a callback during the method call.
-    # operation = video_intelligence_service_client.annotate_video(input_uri, features, input_content, video_context, output_uri, location_id) do |op|
+    # operation = video_intelligence_service_client.annotate_video(input_uri: input_uri, features: features) do |op|
     #   raise op.results.message if op.error?
     #   op_results = op.results
     #   # Process the results.
--- a/google-cloud-video_intelligence/lib/google/cloud/video_intelligence.rb
+++ b/google-cloud-video_intelligence/lib/google/cloud/video_intelligence.rb
@@ -34,9 +34,46 @@ module Google
     # 2. [Enable the Google Cloud Video Intelligence API.](https://console.cloud.google.com/apis/api/video-intelligence)
     # 3. [Setup Authentication.](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/master/guides/authentication)
     #
-    # ### Installation
-    # ```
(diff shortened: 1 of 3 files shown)


quartzmo:
Unfortunately, this branch is not producing the desired output for me:

```sh
$ cat jsondoc/google/cloud/videointelligence.json 
{"id":"google/cloud/videointelligence","name":"Videointelligence","title":["Google","Cloud","Videointelligence"],"description":"","source":"google-cloud-video_intelligence/lib/google/cloud/video_intelligence/v1beta1/doc/google/cloud/videointelligence/v1beta1/video_intelligence.rb#L17","resources":[],"examples":[],"methods":[]}
```

It appears that the docs are being shadowed by one or more instances of the module that are undocumented. I will attempt to debug this today and report back.

quartzmo:
The current culprit is the undocumented declaration of `module Videointelligence` in `lib/google/cloud/video_intelligence/v1beta1/doc/google/cloud/videointelligence/v1beta1/video_intelligence.rb`. Because this instance is processed last by YARD and the gcloud-jsondoc tool, it overwrites all previous output for the module. As you have probably noticed by now, the output of YARD and the gcloud-jsondoc is NOT mapped to the filepaths of the inputs. It reduces all inputs to the same output paths, which are simply the namespace paths. 

I can think of several solutions:

1.  Duplicate the docs across all declarations of the module or class in YARD's path(s).
1. Modify the gcloud-jsondoc tool to not overwrite output with a non-empty `description` with output with an empty `description`.
1. Completely change the gcloud-jsondoc tool to generate output paths based on input files, not namespaces.

The last option would offer much more control and hopefully solve any issues related to documenting  multiple versions in the same package. However, I am somewhat doubtful it would be compatible with the Angular app, and in any case it would be a lot of work.

geigerj:
> I can think of several solutions:
> 1. Duplicate the docs across all declarations of the module or class in YARD's path(s).

Unfortunately, that's not possible because of fundamental restrictions in how the GAPIC code generator handles versioning. We can copy `overview.rb` on a per-version basis, but we can't insert `v1beta2` docs into the `v1beta1/video_intelligence.rb` file, and we need the method sample to be for `v1beta2`. :(
(That's why I introduced `overview.rb` instead of reusing `doc/.../video_intelligence.rb`.)

> 2. Modify the gcloud-jsondoc tool to not overwrite output with a non-empty description using output with an empty description.

That would be great.

> 3. Completely change the gcloud-jsondoc tool to generate output paths based on input file paths, not namespaces.

That would also be great, but if the cost is likely to be high, #2 seems like the best option.

quartzmo:
I'll work on solution 2 today.

quartzmo:
We pushed an update to the gcloud-jsondoc tool that simply excludes the blank output from the module `Google::Cloud::Videointelligence`, preventing it from overwriting the desired output from `Google::Cloud::VideoIntelligence` (capital "I") in the file `jsondoc/google/cloud/videointelligence.json`. In the longer term, we should pursue changing  `Videointelligence` to `VideoIntelligence`.

geigerj:
Thanks, @quartzmo. Tracking changes to the proto namespace here: https://github.com/googleapis/artman/issues/288.

geigerj:
@quartzmo Once the docs build successfully, could you create a new release and modify the README to mark this client as beta?

quartzmo:
Yes, will do.

PR #1755, opened 2017-09-28
Add user_project option to Project#buckets and #create_bucket

quartzmo:
Also:

* Upgrade to Google API Client dependency to `0.14.2`. 
* Fix some errors in Notification docs.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/project.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/project.rb
@@ -93,21 +93,19 @@ module Google
         # @param [String] token A previously-returned page token representing
         #   part of the larger set of results to view.
         # @param [Integer] max Maximum number of buckets to return.
-        # @param [Boolean, String] user_project If the `requester_pays` flag is
-        #   enabled for any of the returned buckets, and if this parameter is
-        #   set to `true`, transit costs for operations on the enabled buckets
-        #   or their files will be billed to the current project for this
-        #   client. (See {#project} for the ID of the current project.) If this
-        #   parameter is set to a project ID other than the current project, and
-        #   that project is authorized for the currently authenticated service
-        #   account, transit costs will be billed to the given project. The
-        #   default is `nil`.
+        # @param [Boolean, String] user_project If this parameter is set to
+        #   `true`, transit costs for operations on the enabled buckets or their
+        #   files will be billed to the current project for this client. (See
+        #   {#project} for the ID of the current project.) If this parameter is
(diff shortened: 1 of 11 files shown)


google-cloud-storage/lib/google/cloud/storage/project.rb
```diff
@@ -93,6 +93,19 @@ def self.default_project
...
+        #   `true`, transit costs for operations on the enabled buckets or their
+        #   files will be billed to the current project for this client. (See
+        #   {#project} for the ID of the current project.) If this parameter is
+        #   set to a project ID other than the current project, and that project
+        #   is authorized for the currently authenticated service account,
+        #   transit costs will be billed to the given project. The default is
+        #   `nil`.
+        #
+        #   The value provided will be applied to all operations on the returned
+        #   bucket instance and its files.
```
quartzmo: I'll  pluralize this to "instances and their files"

google-cloud-storage/lib/google/cloud/storage/project.rb
```diff
@@ -286,6 +300,16 @@ def bucket bucket_name, skip_lookup: false, user_project: nil
         #   does not exist. For more information, see [How to Host a Static
         #   Website
         #   ](https://cloud.google.com/storage/docs/website-configuration#step4).
+        # @param [String] user_project If this parameter is set to a project ID
```
frankyn: Should this also contains:
```
@param [Boolean, String] user_project If this parameter is set to
```

frankyn:
Looks really great, thanks @quartzmo!

google-cloud-storage/lib/google/cloud/storage/project.rb
quartzmo: I removed the documentation for the boolean from `create_bucket` because I couldn't think of a scenario in which it made sense (although it's still supported.) If you create the bucket, passing `user_project: true` to bill all subsequent operations on it to the "current" project (the same one that created it) should have no effect, correct?

quartzmo: Regarding the docs in `File`, thank you, good catch! Updating...

quartzmo: @frankyn Updated.

frankyn: Good point! Thanks Chris.

quartzmo:
Thanks a lot for all the help, @frankyn!

PR #1767, opened 2017-10-06
Update BigQuery insert_id

blowmage:
Per the BigQuery team:

> The insert ID in tabledata.insertAll should not be a hash of the row's contents. We should support customers intentionally including multiple rows in the same value.

This PR changes the value used for insert_id to `Object#object_id` from an MD5 hash. [`Object#object_id` is guaranteed to be unique on active objects.](https://ruby-doc.org/core-2.4.2/Object.html#method-i-object_id)

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
@@ -37,7 +37,7 @@ describe Google::Cloud::Bigquery::Dataset, :bigquery do
     t = dataset.table table_with_schema_id
     if t.nil?
       t = dataset.create_table table_with_schema_id do |schema|
-        schema.integer  "id",     description: "id description",    mode: :required
+        schema.integer   "id",    description: "id description",    mode: :required
         schema.string    "breed", description: "breed description", mode: :required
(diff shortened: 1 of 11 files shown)


blowmage:
@jjolma Can you review this change?

jjolma:
I think this approach can result in incorrect de-duplication since 1) object_ids can be recycled as objects are garbage collected, and 2) two clients streaming into the same table happen to have object_id collisions.  In either case, the first row would get inserted correctly, but the second one would not.

This is very unlikely, admittedly, but would it be a simple change to use a random ID, e.g. UUID4, instead?

blowmage:
@jjolma I don't think this can happen. If you look at the link in the PR description, it says: 

> The same number will be returned on all calls to object_id for a given object, and no two active objects will share an id.

It is possible for the same `object_id` to be recycled, but not within the scope of the `tabledata.insertAll` RPC call, since all data objects are active during the RPC invocation.

jjolma:
The de-duplication window spans all RPCs within the de-duplication window.

blowmage:
Okay, I have updated the PR to use a UUID for `insert_id`. PTAL.

jjolma:
LGTM

PR #1768, opened 2017-10-06
Consider cancel_job idempotent

blowmage:
Since it is idempotent, it can be retried.


blowmage:
@jjolma Can you review this change?

jjolma:
LGTM

PR #1775, opened 2017-10-13
bigquery: add simple benchmark

pongad:
The benchmark runs queries and reports
1. how long we wait until first row
2. how long we take to iterate all rows

### DIFF
--- a/google-cloud-bigquery/benchmark/benchmark.rb
+++ b/google-cloud-bigquery/benchmark/benchmark.rb
@@ -15,31 +15,31 @@
 require "google/cloud/bigquery"
 require "json"
 
-if ARGV.length < 1 || !ENV.has_key?('BIGQUERY_PROJECT')
+if ARGV.length < 1 || !ENV.key?("BIGQUERY_PROJECT")
   puts "usage: BIGQUERY_PROJECT=<project-id> ruby bench.rb <queries.json>"
   exit 1
 end
 
-bigquery = Google::Cloud::Bigquery.new(project: ENV['BIGQUERY_PROJECT']);
+bigquery = Google::Cloud::Bigquery.new(project: ENV["BIGQUERY_PROJECT"])
 queries = JSON.parse(File.open(ARGV[0]).read)
 
-for query in queries
+queries.each do |query|
   start = Time.now
-  numRows = 0
-  numCols = 0
-  timeToFirstByte = nil
+  num_rows = 0
+  num_cols = 0
+  time_to_first_byte = nil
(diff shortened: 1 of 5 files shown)


blowmage:
A few thoughts:

1) Can we add a rake task as the main entry point to to run this script? We have to run most things through bundler and rake tasks make this easier.
2) Can we pass in both project_id and credentials using the existing environment variables instead of passing values in ARGV? (`BIGQUERY_PROJECT=<your project id> ruby benchmark/benchmark.rb benchmark/queries.json` vs. `ruby benchmark/benchmark.rb -- <your project id> benchmark/queries.json`)
3) Would it be beneficial to format the output? Perhaps using the output of [Benchmark](https://ruby-doc.org/stdlib-2.4.2/libdoc/benchmark/rdoc/Benchmark.html) as an influence?

pongad:
@blowmage Thank you for your thoughts.

1. This sounds good but can you teach me how to do it? :smile: This is probably the first Ruby program I write, and I'm completely unfamiliar with all the tools around it.

2. Good idea. Done.

3. I think we should hold off on formatting for a little bit. I plan to eventually set things up so that we run this benchmark for all languages on schedule and auto-publish results to dashboard. At that point, the "canonical" representation of the data will probably be whatever the dashboard wants it in. To help folks run benchmark locally and save work, we could provide a way to, for example, dump information in JSON and a script to turn JSON data into tables (the script can be written only once and reused everywhere). Do you feel strongly about this?

pongad:
@blowmage ping

blowmage:
@pongad I've pushed a couple commits to your branch that fix some of the usage information and add a rake task. Please look these changes over and let me know what you think.

google-cloud-bigquery/benchmark/benchmark.rb
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "google/cloud/bigquery"
+require "json"
+
+if ARGV.length < 1
```
pongad: We're still reading `ENV["BIGQUERY_PROJECT"]` on line 23 right? Should that change too?

blowmage: Oops, yes, that should be changed as well.

PR #1776, opened 2017-10-15
Add Bucket#compose

### LINKED ISSUE #94: Should we support storage object method Compose?
Does gcloud support composing a new storage object from existing objects? (https://cloud.google.com/storage/docs/json_api/v1/objects/compose)

This API is not called in the node or python projects.

quartzmo:
This PR add `Bucket#compose`, with support for setting the properties of the destination file and for encryption keys.

[closes #94]

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -735,9 +735,6 @@ module Google
         #   of source file names or objects that will be concatenated into a
         #   single file.
         # @param [String] destination The name of the new file.
-        # @param [String] content_type The
-        #   [Content-Type](https://tools.ietf.org/html/rfc2616#section-14.17)
-        #   response header to be returned when the file is downloaded.
         # @param [String] acl A predefined set of access controls to apply to
         #   this file.
         #
@@ -804,7 +801,7 @@ module Google
         #
         #   new_file = bucket.compose [file_1, file_2], "path/to/new-file.ext"
         #
-        def compose sources, destination, content_type: nil, acl: nil
+        def compose sources, destination, acl: nil
           ensure_service!
           sources = Array sources
           if sources.size < 2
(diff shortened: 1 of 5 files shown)


google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -714,6 +714,99 @@ def create_file file, path = nil, acl: nil, cache_control: nil,
...
+        # use customer-supplied encryption keys, you must securely manage your
+        # keys and ensure that they are not lost. Also, please note that file
+        # metadata is not encrypted, with the exception of the CRC32C checksum
+        # and MD5 hash. The names of files and buckets are also not encrypted,
+        # and you can read or update the metadata of an encrypted file without
+        # providing the encryption key.
+        #
+        # @param [Array<String, Google::Cloud::Storage::File>] sources
+        # @param [String, Google::Cloud::Storage::File] destination
+        # @param [String] content_type The
```
blowmage: If `compose` always creates a new file, then `destination` should always be a string, right?

quartzmo: Yes, correct, thanks for catching this error.

blowmage: Ugh, i left that comment on the wrong line. Apologies.

quartzmo:
While writing acceptance tests for the encryption keys support, I observed that the service uses the `x-goog-encryption-key` and related headers to decrypt the source files. I had expected it to use `x-goog-copy-source-encryption-key` for the source files and `x-goog-encryption-key` for the destination file, but that does not seem to be the case. Therefore, I will only be able to accept a key for the sources.

@frankyn Does this sound correct to you?

blowmage:
> Therefore, I will only be able to accept a key for the sources. The destination file will have to be encrypted (if desired) after creation via `objects.rewrite`.

LGTM :+1:

quartzmo:
@frankyn Actually, I was mistaken, it appears that if the sources are encrypted with a customer-supplied key, the destination will be encrypted with the same key.

blowmage:
Even better, IMO.

quartzmo:
@blowmage @frankyn This PR is ready for review.

PR #1781, opened 2017-10-18
Update subscription docs

### LINKED ISSUE #1778: Showcase Subscription#listen in docs
In the guide and the [Subscription](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-pubsub/v0.27.1/google/cloud/pubsub/subscription) docs, we talk about `pull` first, then `listen`. However, we should be encouraging users to use `listen`.

* In any case where we demonstrate both ways to pull messages, we should always highlight/prefer/elevate `listen`.
* In any spot where we only show one mechanism to get messages right now, that mechanism shall be `listen`. (Exception: The method documentation for `pull` itself.)

quartzmo:
Show Subscription#listen above #pull as preferred method.

[closes #1778]

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,8 +327,17 @@ msg = topic.publish "new-message"
 # Retrieve a subscription
 sub = pubsub.subscription "my-topic-sub"
 
-# Pull available messages
-msgs = sub.pull
+# Create a subscriber to listen for available messages
+subscriber = sub.listen do |msg|
+  # process msg
+  msg.ack!
+end
+
+# Start background threads that will call the block passed to listen.
+subscriber.start
+
+# Shut down the subscriber when ready to stop receiving messages.
+subscriber.stop.wait!
 ```
 
 ### Cloud Resource Manager (Alpha)
--- a/README.md
+++ b/README.md
@@ -328,9 +328,9 @@ msg = topic.publish "new-message"
 sub = pubsub.subscription "my-topic-sub"
 
 # Create a subscriber to listen for available messages
-subscriber = sub.listen do |msg|
-  # process msg
-  msg.acknowledge!
+subscriber = sub.listen do |received_message|
+  # process message
+  received_message.acknowledge!
 end
 
 # Start background threads that will call the block passed to listen.
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ sub = pubsub.subscription "my-topic-sub"
 # Create a subscriber to listen for available messages
 subscriber = sub.listen do |msg|
   # process msg
-  msg.ack!
+  msg.acknowledge!
 end
 
 # Start background threads that will call the block passed to listen.
--- a/google-cloud-pubsub/lib/google/cloud/pubsub.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub.rb
@@ -185,45 +185,10 @@ module Google
     # end
     # ```
     #
-    # ## Pulling Messages
+    # ## Receiving messages
     #
-    # Messages are pulled from a Subscription. (See
-    # {Google::Cloud::Pubsub::Subscription#pull})
-    #
-    # ```ruby
-    # require "google/cloud/pubsub"
-    #
-    # pubsub = Google::Cloud::Pubsub.new
-    #
-    # sub = pubsub.subscription "my-topic-sub"
-    # msgs = sub.pull
-    # ```
-    #
-    # A maximum number of messages returned can also be specified:
-    #
-    # ```ruby
-    # require "google/cloud/pubsub"
-    #
-    # pubsub = Google::Cloud::Pubsub.new
-    #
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -327,8 +327,16 @@ msg = topic.publish "new-message"
...
 
-# Pull available messages
-msgs = sub.pull
+# Create a subscriber to listen for available messages
+subscriber = sub.listen do |msg|
+  # process msg
+  msg.ack!
+end
+
+subscriber.start
```
frankyn: Could you add a comment that this method will start a background thread that will call the block passed to the listen method?

quartzmo: Yes, will do.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
```diff
@@ -306,12 +308,17 @@ module Cloud
     # pubsub = Google::Cloud::Pubsub.new
     #
     # sub = pubsub.subscription "my-topic-sub"
-    # received_message = sub.pull.first
-    # if received_message
+    # subscriber = sub.listen do |received_message|
     #   puts received_message.message.data
+    #
     #   # Mark for redelivery by setting the deadline to now
     #   received_message.delay! 0
```
blowmage: should this use `received_message.nack!` instead?

blowmage: I think `.nack!` should be preferred over `.delay! 0`.

quartzmo: Ok, changing it now.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
```diff
@@ -290,12 +289,18 @@ module Cloud
     # pubsub = Google::Cloud::Pubsub.new
     #
     # sub = pubsub.subscription "my-topic-sub"
-    # received_message = sub.pull.first
-    # if received_message
+    # subscriber = sub.listen do |received_message|
```
frankyn: Why is this param different from other block params for listen? `received_message` may be more readable versus `msg`, but it's understandable either way. I'm asking more for consistency.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
```diff
@@ -371,9 +382,10 @@ module Cloud
     #
     # subscriber = sub.listen do |msg|
     #   # process msg
-    #   msg.ack!
+    #   msg.acknowledge!
```
frankyn: Optional Nit: I'm a lazy, and appreciated seeing `msg.ack!` is available as well.

quartzmo: For consistency, we decided to change `ack!` (which is an alias) to `acknowledge!` everywhere.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
quartzmo: I also thought about changing all existing examples to `received_message` for consistency, and would be happy to do so if @blowmage is OK with the change.

blowmage: I'm okay with either `msg` or `received_message` _or_ `rec_msg`...

google-cloud-pubsub/lib/google/cloud/pubsub.rb
blowmage: FWIW, I agree I would rather see `ack!` and `nack!` in these examples. However, we just found out the doc site doesn't list it, so it might be more prudent to not show the aliases in the code examples.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
frankyn: Let's do received_message. I appreciate the verbosity.

quartzmo: OK, I will update all `listen` and `pull` examples to `received_message`.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
quartzmo: It seems we need to create a gcloud-common issue for listing aliases.

frankyn: Ah okay, could you file an issue to track this? 

Understood, I'd rather use what's being documented as to not cause friction when searching the docs for `ack!` versus `acknowledge!`. Keep `acknowledge!`.

google-cloud-pubsub/lib/google/cloud/pubsub.rb
quartzmo: @frankyn @blowmage Updated. (There is one example in which I had to use `rec_message` due to line length.)

PR #1784, opened 2017-10-19
Clarify Spanner staleness documentation

### LINKED ISSUE #1782: [Spanner] Ambiguous documentation with using staleness reads
While using the staleness parameter with `Google::Cloud::Spanner::Client#snapshot`, I found the [documentation](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.44.0/google/cloud/spanner/client?method=snapshot-instance) confusing on what a Numerical value could be for this staleness parameter. 

After reviewing the code, I found that the value provided to the staleness parameter is [converted](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/service.rb#L345) into a Protobuf Duration object using [`Google::Cloud::Spanner::Convert#number_to_duration`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/convert.rb#L212). 

Based on implementation I'm expecting that a Float value will be converted to seconds and nanoseconds. This should be documented with values or explain how values are interpreted, because cloud.google.com [documentation](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#readonly) defines this parameter to be of Protobuf Duration object which can't be used to define this parameter at this time.

blowmage:
Update staleness documentation to provide a use example.

[closes #1782]

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -788,8 +788,8 @@ module Google
         #   [TransactionOptions](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#transactionoptions).)
         # @param [Time, DateTime] read_timestamp Same as `timestamp`.
         # @param [Numeric] staleness Executes all reads at a timestamp that is
-        #   +staleness+ old. The timestamp is chosen soon after the read
-        #   is started.
+        #   `staleness` seconds old. For example, the number 10.1 is translated
+        #   to 10 seconds and 100 milliseconds.
         #
         #   Guarantees that all writes that have committed more than the
         #   specified number of seconds ago are visible. Because Cloud Spanner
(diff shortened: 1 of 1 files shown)


frankyn:
Thanks @blowmage!

vkedia:
Why do we take arbitrary Numeric type? Can we instead take a specific type to avoid confusion of this sort?

vkedia:
Actually I think you should take in [Time](https://ruby-doc.org/core-2.2.0/Time.html) type. This would be consistent with how the spanner's TIMESTAMP type is represented in the library.

frankyn:
@vkedia I thought the type for staleness was a Duration Protobuf object and not a Timestamp?

blowmage:
The argument will mostly be an Integer or a Float, and Numeric is the base class for both. However, I don’t think the type is the what was confusing, I think confusing part was what the value was representing. It wasn’t clear the value was the number of seconds. That’s how I am understanding this request.

blowmage:
This argument is not a TIMESTAMP value.

vkedia:
Yeah sorry I got confused between the staleness and timestamp fields. My bad!

frankyn:
In this PR it would be nice to provide a clarification to the documentation.

@blowmage  I do agree with @vkedia in the sense that it should be allowed to accept a specific type which reflects the cloud documentation for this API.  In short also support the Duration protobuf.

blowmage:
What clarification are you looking for? I don’t follow. 

We don’t expose protobuf types in the Ruby API.  The goal is to be able to change the backend implementations (REST, GRPC) without changing the Ruby API. Users can use the low level API if they would rather deal with the protobuf structures.

frankyn:
Ignore my clarification comment and thanks for the clarity for now exposing the protobuf types.

frankyn:
Is this ready to be merged?

PR #1808, opened 2017-11-07
Bigquery skip lookup

### LINKED ISSUE #1785: Support `skip_lookup` for Dataset and Table
Support the `skip_lookup` argument for Dataset and Table object creation.

quartzmo:
* Add a `skip_lookup` option for `Dataset` and `Table` object creation to avoid unneeded `get` RPCs.
* Support a "reference" mode in `Dataset` and `Table` objects, and state transitions to full-representation objects.
* To avoid ambiguous state in "reference" mode tables, remove the `View` class, moving its methods (except `View#data`) to `Table`.

[closes #1785]


quartzmo:
A couple of outstanding questions on this PR:

* Should `Table#table?`, `Table#view?` and `Table#external?` all return `nil` when a `Table` is in "reference" mode? I'm a little concerned since this result could easily be confused with `false`.
* Is it worth the effort to document the possibility of the `nil` result  in "reference" mode for *each* of the getters in `Dataset` and `Table`?

blowmage:
> Should `Table#table?`, `Table#view?` and `Table#external?` all return `nil` when a Table is in "reference" mode? I'm a little concerned since this result could easily be confused with false.

Unfortunately there is not enough information to be able to return `true` in this instance. Therefore a falsely response seems appropriate.

blowmage:
> Is it worth the effort to document the possibility of the nil result in "reference" mode for each of the getters in `Dataset` and `Table`?

I think so. Possibly listen as a return value. But I don’t know if it needs separate code examples. What do you think?

quartzmo:
@blowmage Updated with additional docs for `nil` return values, and with `skip_lookup` example for `Table#insert`.

Issue #1797, opened 2017-10-31
docs on Google::Cloud::Debugger constructor have "project_id", should be "project"


matti:
All README's + https://cloud.google.com/debugger/docs/setup/ruby

vs 
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-debugger/lib/google/cloud/debugger.rb#L365

blowmage:
@frankyn Is this something you can help with? The code example in the section [Using agent parameters](https://cloud.google.com/debugger/docs/setup/ruby#using_agent_parameters) on the cloud site are incorrectly using `project_id` instead of `project`. The previous code examples are using the Stackdriver configuration approach where the `project_id` argument is correct.

frankyn:
@blowmage, I'll help resolve and update cloud site.

blowmage:
Thanks!

frankyn:
Republished. Closing issue.

matti:
Still multiple project_id’s out there? https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/stackdriver/debugger.ru (linked from the web docs that still also contain project_id) @frankym

frankyn:
Thanks for the follow-up. I'll review, I misunderstood.

quartzmo:
@frankyn Is there anything left to do for this issue? The docs now look correct to me.

frankyn:
@matti I reviewed the client libraries and `project_id` is expected except when using Debugger with `Debugger.new`.

matti:
what about https://cloud.google.com/debugger/docs/setup/ruby

blowmage:
All those examples look correct to me.

PR #1810, opened 2017-11-10
Update Debugger README to use project

frankyn:
This is the only item left for issue #1797. I reviewed the code for `project_id` and Debugger.new uses `project` instead of `project_id`.


frankyn:
Thank you!

Issue #1791, opened 2017-10-24
Expose error details when converting Google::Gax::GaxError to Google::Cloud::Error


qingling128:
When `Google::Gax::GaxError` gets converted to `Google::Cloud::Error`, it does not expose the error details to extract `partial_success` info.

When calling `Google::Cloud::Logging::V2::LoggingServiceV2Client.write_log_entries` directly, we get a `Google::Gax::GaxError` with `partial_success` info.

```ruby
cat write_log_entries.rb
require "google/api/monitored_resource_pb"
require "google/cloud/logging/v2"
require "google/logging/v2/log_entry_pb"
require "google/gax"

LoggingClient = Google::Cloud::Logging::V2::LoggingServiceV2Client
logging_client = LoggingClient.new
entries = [
  Google::Logging::V2::LogEntry.new(
    {
      log_name: 'projects/gapic-test/logs/my-log',
      resource: Google::Api::MonitoredResource.new({type: 'gce_instance'}),
      text_payload: 'pass'
    }),
  Google::Logging::V2::LogEntry.new(
    {
      log_name: 'projects/gapic-test/logs/bad:log',
      resource: Google::Api::MonitoredResource.new({type: 'gce_instance'}),
      text_payload: 'fail'
    }),
]

begin
  logging_client.write_log_entries(
    entries,
    log_name: 'projects/gapic-test/logs/my-log',
    resource: Google::Api::MonitoredResource.new(
      type: 'gce_instance',
      labels: {}
    ),
    labels: {},
    partial_success: true
  )
rescue => error
  puts "error:\n#{error.inspect}\n"
  puts "error details:\n#{error.details[0].log_entry_errors.inspect}\n"
end
```

```
$ ruby write_log_entries.rb
error:
#<Google::Gax::RetryError: GaxError Exception occurred in retry method that was not classified as transient, caused by 7:User not authorized.>
error details:
{
    0=><Google::Rpc::Status: code: 7, message: "User not authorized.", details: []>,
    1=><Google::Rpc::Status: code: 3, message: "Log name contains illegal character :", details: []>
}
```

However, if we follow the official instruction at https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-logging/README.md to use the write_entries function.

```
$ cat cleaner_way.rb
require "google/cloud/logging"

logging = Google::Cloud::Logging.new
entries = [
  logging.entry(
    payload: "Job started.",
[...]

PR #1825, opened 2017-11-13
Expose Error Details

blowmage:
Add the ability for Google Cloud Error objects to relay data from the inner error. This will make it easier to retrieve information such as the gRPC trailing metadata.

[refs #1791]

### DIFF
--- a/google-cloud-core/lib/google/cloud/errors.rb
+++ b/google-cloud-core/lib/google/cloud/errors.rb
@@ -59,6 +59,21 @@ module Google
           504 => DeadlineExceededError
         }[http_status_code] || self
       end
+
+      # @private
+      def respond_to? symbol, include_all = false
+        orig = super
+        return true if orig
+        return cause.respond_to?(symbol, include_all) if super(:cause)
+        false
+      end
+
+      private
+
+      def method_missing m, *args, &block
+        return cause.send(m, *args, &block) if respond_to?(:cause)
+        super
+      end
     end
 
     ##
@@ -204,6 +219,27 @@ module Google
     ##
     # DataLoss indicates unrecoverable data loss or corruption.
     class DataLossError < Error
+    end
+  end
+
+  # @private Add method_missing to Gax::GaxError
+  module Gax
+    # @private
+    class GaxError < StandardError
+      # @private
+      def respond_to? symbol, include_all = false
+        orig = super
+        return true if orig
+        return cause.respond_to?(symbol, include_all) if cause
+        false
+      end
+
+      private
+
+      # @private
+      def method_missing m, *args, &block
+        cause.send(m, *args, &block)
+      end
     end
   end
 end
--- a/google-cloud-core/test/google/cloud/errors/grpc_errors_test.rb
+++ b/google-cloud-core/test/google/cloud/errors/grpc_errors_test.rb
@@ -78,7 +78,7 @@ describe Google::Cloud::Error, :grpc do
   end
 
   it "identifies InternalError" do
-    mapped_error = Google::Cloud::Error.from_error GRPC::BadStatus.new(13, "InternalError")
+    mapped_error = Google::Cloud::Error.from_error GRPC::BadStatus.new(13, "internal")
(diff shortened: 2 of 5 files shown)


blowmage:
The current change includes adding `method_missing` to `Google::Gax::GaxError`, so that the inner `GRPC::BadStatus` data can also be exposed. What do you think about contributing this up to google-gax?

I do find it unfortunate that `Google::Gax::GaxError#details` stomps on `GRPC::BadStatus#details`. If it were up to me I'd rename it to `Google::Gax::GaxError#status_details` so all these values could be shown together.

geigerj:
@blowmage What do you mean by "`Google::Gax::GaxError#details` stomps on `GRPC::BadStatus#details`"? GaxError contains the gRPC error but doesn't inherit from it, so it should not hide any attributes of the gRPC error.

blowmage:
Currently, to get at the error values, you need to have knowledge of all the layers that make up the implementation. A user that has a `Google::Cloud::Error` object and wants to get the metadata, they need to call `Google::Cloud::Error#cause` to get the `Gax::GaxError` object, and call `Gax::GaxError#cause` again to get the `GRPC::BadStatus` object, and then call `GRPC::BadStatus#metadata`. This PR attempts to simplify all that by allowing Google::Cloud::Error to pass calls to missing methods to the inner errors.

```
Google::Cloud::Error#code -> GPRC::BadStatus#code #=> 5
Google::Cloud::Error#details -> GPRC::BadStatus#details #=> "not found"
Google::Cloud::Error#metadata -> GPRC::BadStatus#metadata #=> { "foo" => "bar" }
```

But when you the inner error is a GaxError, it needs to pass those calls to the BadStatus, and looks like this:

```
Google::Cloud::Error#code -> Gax::GaxError -> GPRC::BadStatus#code #=> 5
Google::Cloud::Error#details -> Gax::GaxError#details #=> "not found"
Google::Cloud::Error#metadata -> Gax::GaxError -> GPRC::BadStatus#metadata #=> { "foo" => "bar" }
```

So what I meant by "stomping" was, it would be nice if `Gax::GaxError#details` used a different name so that the Google Cloud Error could expose `GPRC::BadStatus#details` and `Gax::GaxError#details` at the same time.

blowmage:
@geigerj Do you have any thoughts or questions on this?

geigerj:
@blowmage I don't mind renaming `GaxError#details` to `GaxError#status_details` or something similar, although it means going to `Gax 0.10.0`. However, I would prefer that `Google::Cloud::Error` was able to map conflicting attributes to different names, since it goes against the design of Gax for its errors module to know anything, or be designed around, gRPC exceptions. WDYT?

quartzmo:
@blowmage As we discussed via screenshare while reviewing this PR earlier, I would be more in favor of exposing `cause` details explicitly via ordinary methods that appear in documentation, rather than dynamically via `method_missing`, if possible. I understand the difficulties posed by handling errors from multiple underlying libraries, but would appreciate giving the solution more time and thought before proceeding with this dynamic approach.

@qingling128 Can you give us an idea of how urgently we need to resolve #1791?

blowmage:
I think would be great if we can get this in before this next batch of releases. Would you like me to submit a PR?

geigerj:
@blowmage I agree with @quartzmo -- I have some reservations about this dynamic approach, which seems like it could run into similar naming issues in the future. But if you submit a PR to `gax-ruby`, I'm open to that.

I am strongly against tying this to the next batch of releases, which have imminent hard deadlines; this can go out immediately following.

blowmage:
Agree this should not hold up any releases. But IMO if there is to be a breaking change better to get it done quickly.

qingling128:
@quartzmo - To work around this issue, we are using `Google::Cloud::Logging::V2::LoggingServiceV2Client.write_log_entries` directly. I haven't run into any issues using that workaround. So yeah, please don't let this issue block anything else we want to release.

blowmage:
Closing in favor of #1829.

PR #1833, opened 2017-11-15
Add Datastore TransactionOptions

### LINKED ISSUE #1807: Add support for TransactionOptions
## Overview

Datastore has introduced two transaction options that can be passed when creating a transaction (BeginTransaction RPC):

* A transaction can be declared read-only. A read-only transaction cannot modify data.
* A read-write transaction can provide the ID of a previous transaction.

### Transaction Mode

Transactions can optionally be created in read-only or read-write mode. Specifying the mode of the transaction will allow for Datastore to perform future performance optimizations in upcoming changes.

A read-write transaction can read and modify entities. Read-write transactions that operate on the same data can conflict and suffer from reduced throughput due to contention.

A read-only transaction cannot modify entities; in return they do not contend with other read-write or read-only transactions. Using a read-only transaction for transactions that only read data will potentially improve throughput.

### Transaction Retry

Read-write transactions may fail due to contention. A read-write transaction can be retried by specifying previous_transaction when creating the new transaction.

Specifying previous_transaction provides more information that can be used to improve throughput. In particular, if transactional operations A and B conflict, specifying the previous_transaction can help to prevent livelock.

## Requirements

1. Users should be able to specify that a transaction is read-only when they create it.
1. When retrying a read-write transaction, the previous transaction’s ID should be included in the request.
    1. If your client has a RunInTransaction method that takes a function and automatically creates and retries the transaction, then this method should handle setting the previous transaction ID. There is no need to expose the option to the user.
    1. If your client requires that users retry themselves, then expose the option.

### Unit Tests

* When the read-only option is passed, the BeginTransactionRequest is
    ```
    project_id:"project"
    transaction_options:<read_only:<> >
    ```
* When the previous transaction ID “tid” is passed, the BeginTransactionRequest is
    ```
    project_id:"project"
    transaction_options:<read_write:<previous_transaction:"tid" > >
    ```
* If you have a `RunInTransaction` function, simulate a retryable failure to verify that `previous_transaction` is passed on retries.

### Integration Tests

* A read-only transaction can perform reads.
* A read-only transaction fails if it attempts to write. This test currently fails against the production endpoint.

## Resources

[Public protos](https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/datastore.proto#L352)
[Go implementation](https://code-review.googlesource.com/c/gocloud/+/18350)
[C# implementation](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1447)
[Python PR](https://github.com/GoogleCloudPlatform/google-cloud-python/pull/4335)

quartzmo:
* Add support for read-only transactions: Add `read_only` option to Dataset#transaction, and `Transaction#read_only?` and `#read_write?`.
* Add support for `previous_transaction` (pending)

[closes #1807]

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
@@ -544,12 +544,19 @@ module Google
         end
 
         ##
-        # Creates a read-only Datastore transaction that only allows reads.
+        # Creates a read-only transaction that provides a consistent snapshot of
+        # Cloud Datastore. This can be useful when multiple reads are needed to
+        # render a page or export data that must be consistent.
         #
         # A read-only transaction cannot modify entities; in return they do not
         # contend with other read-write or read-only transactions. Using a
         # read-only transaction for transactions that only read data will
         # potentially improve throughput.
+        #
+        # Read-only single-group transactions never fail due to concurrent
+        # modifications, so you don't have to implement retries upon failure.
+        # However, multi-entity-group transactions can fail due to concurrent
+        # modifications, so these should have retries.
         #
         # @see https://cloud.google.com/datastore/docs/concepts/transactions
         #   Transactions
@@ -592,6 +599,7 @@ module Google
             raise TransactionError, "Transaction failed to commit."
           end
         end
+        alias_method :snapshot, :read_only_transaction
 
         ##
         # Create a new Query instance. This is a convenience method to make the
--- a/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
@@ -25,6 +25,7 @@ require "google/cloud/datastore/gql_query"
 require "google/cloud/datastore/cursor"
 require "google/cloud/datastore/dataset/lookup_results"
 require "google/cloud/datastore/dataset/query_results"
+require "google/cloud/datastore/transaction"
 
 module Google
   module Cloud
@@ -542,9 +543,9 @@ module Google
         #   Transactions
         #
         # @yield [tx] a block yielding a new transaction
-        # @yieldparam [Transaction] tx the transaction object
+        # @yieldparam [ReadOnlyTransaction] tx the transaction object
         #
-        # @example Use a read-only transaction when only performing reads:
+        # @example
         #   require "google/cloud/datastore"
         #
         #   datastore = Google::Cloud::Datastore.new
@@ -563,7 +564,7 @@ module Google
         #   end
         #
         def read_only_transaction
-          tx = Transaction.new service, read_only: true
+          tx = ReadOnlyTransaction.new service
           return tx unless block_given?
 
           begin
--- a/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
@@ -466,6 +466,15 @@ module Google
         ##
         # Creates a Datastore Transaction.
         #
+        # @see https://cloud.google.com/datastore/docs/concepts/transactions
+        #   Transactions
+        #
+        # @param [Boolean] read_only Whether the transaction should only allow
+        #   reads. A read-only transaction cannot modify entities; in return
+        #   they do not contend with other read-write or read-only transactions.
+        #   Using a read-only transaction for transactions that only read data
+        #   will potentially improve throughput. Optional. The default is `nil`.
+        #
         # @yield [tx] a block yielding a new transaction
         # @yieldparam [Transaction] tx the transaction object
         #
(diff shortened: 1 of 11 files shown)


google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
```diff
@@ -509,8 +518,26 @@ def run query, namespace: nil, consistency: nil
...
+        #   tasks = nil
+        #
+        #   datastore.transaction read_only: true do |tx|
+        #     task_list = tx.find task_list_key
+        #     if task_list
+        #       tasks = tx.run query
+        #     end
+        #   end
+        #
+        def transaction read_only: nil
```
blowmage: I would strongly prefer that the `read_only` option uses a separate named `snapshot` method, similar to what is done in Spanner's `Client#snapshot`.

quartzmo: I'm afraid that users will not be expecting to create a read-only `Transaction` using a method named `snapshot`, and will struggle to understand the relationship. The official docs do not mention "Snapshot" as a first-class concept.* And we have no `Snapshot` class.

Is consistency with spanner worth it?

* The official docs state: "you can use a transaction to read a consistent snapshot of Cloud Datastore"

blowmage: I think there are many advantages to supporting a read-write transaction in the Transaction class, and supporting a read-only transaction in a Snapshot class. One is that you can omit the `delete`, `insert`, `save`, and `update` methods completely.

Spanner's Snapshot class is a read-only transaction, and is a concept specific to the client library. Meaning, there is no Snapshot resource in the Spanner API. I don't think doing the same in Datastore would cause undue confusion.

In general, I think Snapshot is a useful abstraction and one that I would like to be consistent across Datastore, Spanner, and Firestore.

frankyn: Chiming into the discussion:
As a user having a restricted object that doesn't include write operations would be helpful so I don't incorrectly use a read-only Transaction as a read-write transaction. 

Using a Snapshot class seems like it would be useful considering the work on the Spanner read-only transactions are created using a snapshot method. Datastore [documentation](https://cloud.google.com/datastore/docs/concepts/transactions) does use the word snapshot even though it's not necessarily part of the [API endpoints or parameters](https://cloud.google.com/datastore/docs/reference/rest/v1/projects/beginTransaction#ReadOnly) similar to the Spanner [API reference documentation](https://cloud.google.com/spanner/docs/reference/rest/v1/TransactionOptions#ReadOnly.FIELDS.strong), but Spanner reference documentation documents ReadOnly transactions as "snapshots".

I'm on the fence of using "snapshot" because it's not very self descriptive such as `datastore.transaction read_only: true` or maybe even `datastore.read_only_transaction`. In [Php PR#727](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/727/files#diff-2124f3325a11419ecc0faf5cd6e8c598R524) may use a similar approach as the `datastore.read_only_transaction` method.

I agree there should exist an object that restricts the methods available to a read-only transaction versus a read-write transaction. But I'm not so sure about using Snapshot as the surfaced name.

quartzmo:
> Spanner's Snapshot class is a read-only transaction, and is a concept specific to the client library. Meaning, there is no Snapshot resource in the Spanner API. I don't think doing the same in Datastore would cause undue confusion.

@blowmage I disagree: I believe that departing from the underlying service API, especially adding new vocabulary/naming that is not present in the official API docs, is very likely to cause undue confusion and should be avoided. In this case, I think users familiar with the underlying API will be looking for a method with `transaction` in its name.

However, I respect the work that you put into the design of google-cloud-spanner, and if you don't mind reconfirming your decision in reply to @frankyn's [comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1833#discussion_r151830109), I will update this PR to use a `snapshot` method instead. I just want to mention for the record that I feel the "vertical" integration of staying close to the underlying service API has more value for users than the "horizontal" integration of consistency across the client libraries.

blowmage:
I have two main points. First, I think it is wise to have different objects for read-write transactions and read-only transactions. To set concerns about naming aside, I think it is better to have different methods yielding different objects, with the ability to have different commit logic for these two different use cases. That is why I would prefer to not add a `read_only` argument to `Dataset#transaction`.

Second, I think "snapshot" is a better term to use than "read only transaction". We already have prior art in Spanner using this term. (But, that is just, like, my opinion, man.) I think there are two obvious naming options: Adding a `Snapshot` class and a `Dataset#snapshot` method, or a `ReadOnlyTransaction` class and a `Dataset#read_only_transaction` method.

I have stronger opinions on the first point than I hold on the second. I would support this PR if it delivered the first point but used a different name on the second point.

quartzmo:
@blowmage Docs may need some revising, and I want to add a `snapshot` alias, but PTAL.

google-cloud-datastore/lib/google/cloud/datastore/transaction.rb
```diff
@@ -53,9 +55,7 @@ module Datastore
       #     tasks_in_list = tx.run query
       #   end
       #
-      class Transaction < Dataset
-        attr_reader :id
-
+      class Transaction < ReadOnlyTransaction
```
blowmage: Oops, I didn't realize that Transaction was inheriting from Dataset. Using inheritance here seems like a mistake, and something to move away from instead perpetuating. I know this is a larger issue than the scope of this PR, but I believe the use of inheritance here is likely going to cause more problems than it solves.

For instance, the `Dataset#transaction` and `Dataset#read_only_transaction` methods are going to be inherited. I believe inheritance can also make documentation more difficult than it would otherwise need to be. It would be cleaner to have separate `#run` methods on `Dataset`, `Transaction`, and `ReadOnlyTransaction` with code examples that were specific to the method's class, than having multiple code examples for all the ways the inherited method might be invoked.

I'd like to suggest we remove the use of inheritance, but I can see an argument for not doing that in this PR. I'd support a follow up PR for that. @quartzmo what do you think?

quartzmo: Sounds good. I'll do it in this PR.

PR #1842, opened 2017-11-17
Support public data file list, get, and download.

### LINKED ISSUE #1834: Provide anonymous (unauthenticated) access to public-access buckets
Developers should be able to download public data (files) without specifying any authentication information, in a manner consistent with the rest of the API.

Therefore, provide a simple mechanism to construct a `Storage::Project` instance with no authentication details. This object should be usable in a normal way for whatever actions don't require authentication from a server perspective. It is not required that the client library prohibits other operations without authentication (e.g. upload) - it can rely on the server performing that validation.

#### Acceptance tests

When the client is unauthenticated, it:

1. Should download files from `gs://gcp-public-data-landsat`.
1. Should list files in `gs://gcp-public-data-landsat`.
1. Should **not** download data from a regular private bucket.
1. Should **not** list files in a regular private bucket.
1. Should **not** upload data to a regular private bucket.

quartzmo:
Adds `Google::Cloud::Storage.anonymous`.

[closes #1834]

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage.rb
+++ b/google-cloud-storage/lib/google/cloud/storage.rb
@@ -317,6 +317,23 @@ module Google
     # downloaded.read #=> "Hello world!"
     # ```
     #
+    # Download a public file with an anonymous, unauthenticated client. Use
+    # `skip_lookup` to avoid errors retrieving non-public bucket and file
+    # metadata.
+    #
+    # ```ruby
+    # require "google/cloud/storage"
+    #
+    # storage = Google::Cloud::Storage.anonymous
+    #
+    # bucket = storage.bucket "public-bucket", skip_lookup: true
+    # file = bucket.file "path/to/public-file.ext", skip_lookup: true
+    #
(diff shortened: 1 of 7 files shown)


google-cloud-storage/lib/google/cloud/storage.rb
```diff
@@ -317,6 +317,23 @@ module Cloud
     # downloaded.read #=> "Hello world!"
     # ```
     #
+    # Download a public file with an anonymous, unauthenticated client. Use
+    # `skip_lookup` to avoid errors retrieving non-public bucket and file
+    # metadata.
```
quartzmo: > Use `skip_lookup` to avoid errors retrieving non-public bucket and file metadata.
  
@jskeet This advice is appropriate, isn't it?

jskeet: I don't know what skip_lookup does... I don't think we have any similar concept in the C# code. (When we fetch the data in C#, we don't fetch metadata, and vice versa.)

PR #1848, opened 2017-11-21
Insert migration guide link into top-level docs

geigerj:


google-cloud-language/lib/google/cloud/language.rb
```diff
@@ -28,6 +28,10 @@ module Cloud
     # recognition, and text annotations.
     # - [Product Documentation][]
     #
+    # *This release, 0.28.0, introduces breaking changes relative to the previous
```
quartzmo: Do you want to future-proof the phrasing with something like:

> The 0.28.0 release introduced breaking changes relative to the previous

geigerj: Yeah, that's a good idea. Done.

PR #1849, opened 2017-11-21
Add VideoIntelligence GAPIC client for v1 API

geigerj:


google-cloud-video_intelligence/google-cloud-video_intelligence.gemspec
```diff
@@ -3,7 +3,7 @@
 
 Gem::Specification.new do |gem|
   gem.name          = "google-cloud-video_intelligence"
-  gem.version       = "0.23.0"
+  gem.version       = "0.24.0"
```
blowmage: The [current release](https://github.com/GoogleCloudPlatform/google-cloud-ruby/releases/tag/google-cloud-video_intelligence%2Fv0.24.1) is already `0.24.1`. This should be bumped up to `0.25.0`.

quartzmo:
Failed with:

```sh
Bundler could not find compatible versions for gem
"google-cloud-video_intelligence":
  In Gemfile:
    google-cloud-video_intelligence

    google-cloud was resolved to 0.44.0, which depends on
      google-cloud-video_intelligence (~> 0.23.0)
```

geigerj:
Updated google-cloud dependency, and google-cloud-video_intelligence gemspec/toc.json/Rakefile.

google-cloud-video_intelligence/docs/toc.json
```diff
@@ -6,6 +6,21 @@
       "type": "google/cloud/videointelligence",
       "patterns": ["\/videointelligence", "google\/protobuf", "google\/rpc"],
       "nav": [
+        {
```
blowmage: @quartzmo Won't this also have to be added to the `google-cloud/docs/toc.json`?

quartzmo: @blowmage Good catch. We need to always remember to look outside the scope of the package directory with these GAPIC updates. 

@geigerj  [This section](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud/docs/toc.json#L587-L618) of `google-cloud/docs/toc.json` should also be updated.

geigerj: Updated.

quartzmo: @blowmage This is fixed. Anything else, or is this ready to merge?

PR #1855, opened 2017-11-22
Refactor Google Cloud Speech Module

IlyaUmanets:
Changes:

* Put empty lines to increase readability
* Add `BILLION` constant
* Use `zero?` method instead of `== 0`
* Don't return `nil` directly. `return` returns `nil` by default.
* Remove unused variable declaration
* Replace `return unless object` with `return if object.nil?`

### DIFF
--- a/google-cloud-speech/lib/google/cloud/speech/project.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/project.rb
@@ -190,9 +190,9 @@ module Google
           else
             audio = Audio.from_source source, self
           end
-          audio.encoding = encoding unless encoding.nil?
-          audio.language = language unless language.nil?
-          audio.sample_rate = sample_rate unless sample_rate.nil?
+          audio.encoding = encoding if encoding
+          audio.language = language if language
+          audio.sample_rate = sample_rate if sample_rate
           audio
         end
 
@@ -626,7 +626,6 @@ module Google
         def audio_config encoding: nil, language: nil, sample_rate: nil,
                          max_alternatives: nil, profanity_filter: nil,
                          phrases: nil, words: nil
-          contexts = nil
           contexts = [V1::SpeechContext.new(phrases: phrases)] if phrases
           language = String(language) unless language.nil?
           V1::RecognitionConfig.new({
--- a/google-cloud-speech/lib/google/cloud/speech/audio.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/audio.rb
@@ -293,11 +293,14 @@ module Google
         def self.from_source source, speech
           audio = new
           audio.instance_variable_set :@speech, speech
+
           if source.respond_to?(:read) && source.respond_to?(:rewind)
             source.rewind
             audio.grpc.content = source.read
+
             return audio
           end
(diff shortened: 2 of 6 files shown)


IlyaUmanets:
@googlebot, I signed it!

google-cloud-speech/lib/google/cloud/speech/version.rb
```diff
@@ -16,7 +16,7 @@
 module Google
   module Cloud
     module Speech
-      VERSION = "0.27.0"
+      VERSION = "0.27.1"
```
blowmage: Can you revert the change to the `VERSION`? We change this during the release process, as [documented here](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/RELEASING.md).

IlyaUmanets:
@blowmage, the PR is updated

google-cloud-speech/lib/google/cloud/speech/service.rb
```diff
@@ -29,15 +29,16 @@ class Service
 
         ##
         # Creates a new Service instance.
-        def initialize project, credentials, timeout: nil, client_config: nil
+        def initialize project, credentials, timeout: nil, client_config: {}
```
blowmage: The problem with this change is what happens if `nil` is passed in. The GRPC code expects a hash, and this change could cause errors. @quartzmo what do you think?

quartzmo: @blowmage I agree.

@IlyaUmanets What is your motivation for the changes in this PR? They appear to be stylistic refactorings, but I am concerned that despite the project's automated testing, there is a very slight risk of introducing functional changes that I do not see justified by any benefit, unless you can propose one.

IlyaUmanets:
@blowmage, @quartzmo, you are absolutely right, it was a bad change.

google-cloud-speech/lib/google/cloud/speech/project.rb
```diff
@@ -626,7 +626,6 @@ def operation id
         def audio_config encoding: nil, language: nil, sample_rate: nil,
                          max_alternatives: nil, profanity_filter: nil,
                          phrases: nil, words: nil
-          contexts = nil
```
blowmage: While this doesn’t change functionality, it will issue a warning when `contexts` is used without first being initialized.

quartzmo:
@IlyaUmanets What is your motivation for the changes in this PR? The changes appear to be just stylistic refactorings. I am concerned that despite the project's automated testing, there is a very slight risk of introducing functional changes or bugs. I do not see enough benefit from style changes for this slight risk, personally.

quartzmo:
@IlyaUmanets Thank you for submitting this pull request. For the reason I gave above, we have decided not to accept it at this time. However, we really appreciate the effort: If you have any suggestion for how to functionally improve the google-cloud-ruby packages, please open an issue to discuss, and we will assist you in developing a PR if you want to do so.

IlyaUmanets:
@quartzmo I absolutely agree with you. Thanks a lot !

PR #1865, opened 2017-12-04
Firestore

blowmage:
This PR adds the Firestore Ruby client library.

```ruby
require "google/cloud/firestore"

firestore = Google::Cloud::Firestore.new

# Get the cities collection
cities_col = firestore.col "cities"

# Get the document for NYC
nyc_ref = cities_col.doc "NYC"

# Set the name for NYC
firestore.set(nyc_ref, { "name": "New York City" })
```

The library also implements writing multiple changes in a batch:

```ruby
firestore.batch do |b|
  # Set the data for NYC
  b.col("cities").doc("NYC").set({ name: "New York City" })

  # Update the population for SF
  b.col("cities").doc("SF").update({ population: 1000000 })

  # Delete LA
  b.col("cities").doc("LA").delete
end
```

It also implements support for transactions

```ruby
cities_col = firestore.col("cities").doc("SF")
cities_col.set({ name: "San Francisco",
           state: "CA",
           country: "USA",
           capital: false,
           population: 860000 })

firestore.transaction do |tx|
  new_population = tx.get(cities_col).data[:population] + 1
  tx.update(city, { population: new_population })
end
```

And implements read-only transactions:

```ruby
cities_col = firestore.col "cities"
nyc_ref = cities_col.doc "NYC"
sf_ref  = cities_col.doc "SF"
la_ref  = cities_col.doc "LA"

firestore.read_only_transaction do |rtx|
  # Get each city's population
  nyc_population = rtx.get(nyc_ref).data[:population]
  sf_population  = rtx.get(sf_ref).data[:population]
  ls_population  = rtx.get(la_ref).data[:population]
end
```


blowmage:
@sebastian-schmidt @frankyn Can you please take a look at this now?

google-cloud-firestore/google-cloud-firestore.gemspec
```diff
@@ -0,0 +1,36 @@
+# -*- ruby -*-
+# encoding: utf-8
+
+Gem::Specification.new do |gem|
+  gem.name          = "google-cloud-firestore"
+  gem.version       = "0.6.8"
```
quartzmo: Shouldn't the version be set to `Google::Cloud::Firestore::VERSION`? (currently `0.20.0`)

google-cloud-firestore/docs/toc.json
```diff
@@ -0,0 +1,59 @@
...
+      "patterns": ["\/firestore"],
+      "nav": [
+        {
+          "title": "Project",
+          "type": "google/cloud/firestore/project"
+        },
+        {
+          "title": "V1beta1",
+          "type": "google/cloud/firestore/v1",
+          "patterns": ["\/firestore\/v1$", "\/firestore\/v1\/"],
```
quartzmo: `v1` -> `v1beta1`

frankyn:
Will Review today.

frankyn:
@blowmage I'm seeing a `Gem::LoadError` when I try to require the Firestore gem. 

```
Gem::LoadError: google/rpc/status_pb found in multiple gems: googleapis-common-protos, googleapis-common-protos-types
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:102:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/grpc-1.7.3-x86_64-linux/src/ruby/lib/grpc/google_rpc_status_utils.rb:16:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-gax-0.10.2/lib/google/gax/grpc.rb:31:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-gax-0.10.2/lib/google/gax/errors.rb:32:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-gax-0.10.2/lib/google/gax/api_callable.rb:32:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-gax-0.10.2/lib/google/gax.rb:30:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:120:in `require'
... 8 levels...
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-cloud-firestore-0.6.8/lib/google/cloud/firestore/collection.rb:17:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-cloud-firestore-0.6.8/lib/google/cloud/firestore/database.rb:16:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-cloud-firestore-0.6.8/lib/google/cloud/firestore/project.rb:17:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/google-cloud-firestore-0.6.8/lib/google/cloud/firestore.rb:17:in `<top (required)>'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:133:in `require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:40:in `require'
        from (irb):1
        from /usr/local/google/home/franknatividad/.rbenv/versions/2.3.1/bin/irb:11:in `<main>'
```

frankyn:
@blowmage I'm still reviewing, and should have comments tomorrow. I'm running through samples written for other langues presented on https://cloud.google.com/firestore/docs/ to review the client library. I do have an open question, where are Firestore samples/documentation for the client library similar to other clients?

google-cloud-firestore/google-cloud-firestore.gemspec
blowmage: Oops. That was from the original gemspec file. It should pull the version from `Google::Cloud::Firestore::VERSION`. Fixed now.

google-cloud-firestore/docs/toc.json
blowmage: Fixed. Great catch.

blowmage:
> I'm seeing a Gem::LoadError when I try to require the Firestore gem. 

I am not able to reproduce this. I am able to install a local gem and run it in IRB without any issues. (I use `omg` to create an empty gem-set to ensure I have no other gems installed.) Here are my steps:

```
$ bundle exec rake build
google-cloud-firestore 0.20.0 built to pkg/google-cloud-firestore-0.20.0.gem.
$ cd pkg/
$ omg temp
Switching to temp
$ gem install google-cloud-firestore-0.20.0.gem 
Fetching: concurrent-ruby-1.0.5.gem (100%)
Successfully installed concurrent-ruby-1.0.5
Fetching: little-plugger-1.1.4.gem (100%)
Successfully installed little-plugger-1.1.4
Fetching: multi_json-1.12.2.gem (100%)
Successfully installed multi_json-1.12.2
Fetching: logging-2.2.2.gem (100%)
Successfully installed logging-2.2.2
Fetching: jwt-2.1.0.gem (100%)
Successfully installed jwt-2.1.0
Fetching: memoist-0.16.0.gem (100%)
Successfully installed memoist-0.16.0
Fetching: os-0.9.6.gem (100%)
Successfully installed os-0.9.6
Fetching: public_suffix-3.0.1.gem (100%)
Successfully installed public_suffix-3.0.1
Fetching: addressable-2.5.2.gem (100%)
Successfully installed addressable-2.5.2
Fetching: signet-0.8.1.gem (100%)
Successfully installed signet-0.8.1
Fetching: googleauth-0.6.2.gem (100%)
Successfully installed googleauth-0.6.2
Fetching: google-protobuf-3.5.0-universal-darwin.gem (100%)
Successfully installed google-protobuf-3.5.0-universal-darwin
Fetching: googleapis-common-protos-types-1.0.1.gem (100%)
Successfully installed googleapis-common-protos-types-1.0.1
Fetching: grpc-1.7.3-universal-darwin.gem (100%)
Successfully installed grpc-1.7.3-universal-darwin
Fetching: googleapis-common-protos-1.3.7.gem (100%)
Successfully installed googleapis-common-protos-1.3.7
Fetching: rly-0.2.3.gem (100%)
Successfully installed rly-0.2.3
Fetching: google-gax-0.10.2.gem (100%)
Successfully installed google-gax-0.10.2
Fetching: google-cloud-env-1.0.1.gem (100%)
Successfully installed google-cloud-env-1.0.1
Fetching: google-cloud-core-1.1.0.gem (100%)
Successfully installed google-cloud-core-1.1.0
Successfully installed google-cloud-firestore-0.20.0
20 gems installed
$ irb
irb(main):001:0> require "google/cloud/firestore"
=> true
irb(main):002:0> firestore = Google::Cloud::Firestore.new
=> #<Google::Cloud::Firestore::Database:0x00007fd2e350a0b0 @service=Google::Cloud::Firestore::Service(my-project-id)>
irb(main):003:0> firestore.cols { |col| puts col.collection_id }
gcloud-2017-12-11t23-41-52z-c51d0e80
gcloud-2017-12-12t02-26-42z-0b0001a4
gcloud-2017-12-12t02-31-29z-5ff9de21
gcloud-2017-12-12t03-00-49z-ecef88aa
gcloud-2017-12-12t03-04-24z-49f5ec76
=> nil
irb(main):004:0> exit
$ gem list

*** LOCAL GEMS ***

addressable (2.5.2)
bigdecimal (default: 1.3.0)
bundler (1.16.0, 1.15.4)
concurrent-ruby (1.0.5)
did_you_mean (1.1.0)
faraday (0.13.1)
ffi (1.9.18)
filetree (1.0.0)
google-cloud-core (1.1.0)
google-cloud-env (1.0.1)
google-cloud-firestore (0.20.0)
google-gax (0.10.2)
google-protobuf (3.5.0 universal-darwin)
googleapis-common-protos (1.3.7)
googleapis-common-protos-types (1.0.1)
googleauth (0.6.2)
grpc (1.7.3 universal-darwin)
io-console (default: 0.4.6)
jruby-jars (1.7.27)
jruby-parser (0.5.4)
json (default: 2.0.4)
jwt (2.1.0)
little-plugger (1.1.4)
logging (2.2.2)
memoist (0.16.0)
minitest (5.10.1)
multi_json (1.12.2)
multipart-post (2.0.0)
mustermann (1.0.1)
net-telnet (0.1.1)
openssl (default: 2.0.5)
os (0.9.6)
power_assert (0.4.1)
psych (default: 2.2.2)
public_suffix (3.0.1)
rack (2.0.3)
rack-protection (2.0.0)
rake (12.0.0)
rdoc (default: 5.0.0)
rly (0.2.3)
rsense (0.5.18)
rsense-core (0.6.6)
rsense-server (0.5.18)
signet (0.8.1)
sinatra (2.0.0)
spoon (0.0.6)
test-unit (3.2.3)
thor (0.19.4)
tilt (2.0.8)
xmlrpc (0.2.1)
```

The CI tests are running on Ruby 2.3 as well.

blowmage:
> I do have an open question, where are Firestore samples/documentation for the client library similar to other clients?

Do you mean the shared examples used on the cloud.google.com site? We don't manage those, so I don't know where those are maintained.

If you mean the in-line code documentation and examples, then those were updated and added to the PR today. They had to be updated to reflect recent changes made, and fixing the in-line code example tests took longer to correct than I had hoped.

frankyn:
@blowmage I meant the latter, thanks! 👍

google-cloud-firestore/lib/google/cloud/firestore/document/snapshot.rb
```diff
@@ -0,0 +1,341 @@
...
+          #
+          #   # Get the document data
+          #   nyc_snap[:population] #=> 1000000
+          #
+          # @example Nested data can be accessing with field path:
+          #   require "google/cloud/firestore"
+          #
+          #   firestore = Google::Cloud::Firestore.new
+          #
+          #   frank_snap = firestore.get "users/frank"
```
frankyn: 😸

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1012 @@
...
+        # @return [Query] A query with `where` called on it.
+        #
+        # @example
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   firestore.batch do |b|
+        #     # Create a query
+        #     query = b.where(:population, :>=, 1000000).
```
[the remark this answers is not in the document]
blowmage: There is not a specific order the methods need to be chained, they can be in any order. The order shown in these code examples is to demonstrate the methods being called on the Batch object, and not the Query object that is returned.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,626 @@
...
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a collection reference
+        #   cities_ref = firestore.col "cities"
+        #
+        #   # Get and print all city documents
+        #   cities_ref.docs do |city|
+        #     puts city.document_id
```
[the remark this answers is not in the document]
blowmage: Oops, I think this documentation and example code are incorrect. It says that the returned objects are Document::Reference objects, but in reality they are Document::Snapshot objects, and will have data. I'll update the docs and the examples. Thanks for finding this!

google-cloud-firestore/lib/google/cloud/firestore/read_only_transaction.rb
```diff
@@ -0,0 +1,825 @@
...
+      #   firestore = Google::Cloud::Firestore.new
+      #
+      #   cities_col = firestore.col "cities"
+      #   nyc_ref = cities_col.doc "NYC"
+      #   sf_ref  = cities_col.doc "SF"
+      #   la_ref  = cities_col.doc "LA"
+      #
+      #   firestore.read_only_transaction do |rtx|
+      #     # Get each city's population
+      #     nyc_population = rtx.get(nyc_ref).data[:population]
```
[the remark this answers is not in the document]
blowmage: What do you get when you try this code?

```ruby
firestore.read_only_transaction do |t|
  doc = t.get("hello/startup")
  puts doc.exists?
end
```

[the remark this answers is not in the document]
blowmage: K, I'm investigating.

frankyn:
@blowmage follow-up on the gem install question. I tried requiring the gem again and then I was able to use the gem as expected. I'm trying to understand the `omg` workflow. What is the `omg` command? Is it a Ruby specific tool?

blowmage:
@frankyn The `omg` command is [ohmygems](https://github.com/seattlerb/ohmygems). It is functionally equivalent to rvm gemset and rbenv gemset. It creates an empty environment where you have no gems installed. I try to optimize my ruby response time, so I use a combination of chruby and ohmygems because it is faster than rvm, rbenv, or chruby/chgem.

the-more-you-know

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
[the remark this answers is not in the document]
blowmage: The documentation has been updated to show the correct return type, and the code examples have been updated to show accessing the document data.

google-cloud-firestore/lib/google/cloud/firestore/document/snapshot.rb
blowmage: I wish we could take credit for that code example, but that comes from [here](https://cloud.google.com/firestore/docs/manage-data/add-data#update-data).

google-cloud-firestore/lib/google/cloud/firestore/read_only_transaction.rb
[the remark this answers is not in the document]
blowmage: Welp, you found a legit bug! The problem was that sometimes a response can have an empty result, specifically when it returns a transaction id. I've updated the implementation and tests to be more resilient when receiving an empty result. Thanks for finding this!

google-cloud-firestore/lib/google/cloud/firestore/document/snapshot.rb
frankyn: I'm sure there's a 12 year old out there who feels better about this information being in a sample 😄 

~~ end ~~

google-cloud-firestore/lib/google/cloud/firestore/read_only_transaction.rb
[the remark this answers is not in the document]
frankyn: Tested, and I'm now seeing the expected result! Thanks @blowmage!

quartzmo:
@frankyn fwiw, I just use rbenv gemset.

quartzmo:
@blowmage At which level (alpha, beta) is this library being released? (I noticed the generated `overview.rb` lists "Alpha". And do you want to add Firestore to the top-level README in this PR?

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,626 @@
...
+          #
+          def doc document_path = nil
+            document_path ||= random_document_id
+
+            ensure_context!
+            context.doc "#{collection_path}/#{document_path}"
+          end
+          alias_method :document, :doc
+
+          def add data = nil
```
quartzmo: Missing docs for `#add`?

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,990 @@
...
+        attr_accessor :service
+
+        ##
+        # @private Creates a new Firestore Database instance.
+        def initialize service
+          @service = service
+        end
+
+        ##
+        # The project resource the Cloud Firestore batch belongs to.
```
quartzmo: Does "batch" belong here? (1 of 3 usages in this file.)

blowmage:
I honestly don’t know what level it will be released. I was expecting another PR doing a documentation pass and polish before it is released.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1013 @@
...
+    module Firestore
+      ##
+      # # Batch
+      #
+      # A batch in Cloud Firestore is a set of writes that execute
+      # atomically at a single logical point in time in a database.
+      #
+      # All changes are accumulated in memory until the block passed to
+      # {Database#batch} completes. Unlike transactions, batches are not
+      # automatically retried. See {Database#batch}.
```
quartzmo: Why aren't they retried? 

This might help:

> Batched writes have fewer failure cases than transactions and use simpler code. They are not affected by contention issues, because they don't depend on consistently reading any documents.

Also adding a link to https://firebase.google.com/docs/firestore/manage-data/transactions is probably helpful.

Ditto for `Database#batch`.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1013 @@
...
+        # Write to document with the provided object values. If the document
+        # does not exist, it will be created. By default, the provided data
+        # overwrites existing data, but the provided data can be merged into the
+        # existing document using the `merge` argument.
+        #
+        # @param [String, Document::Reference] doc A string representing the
+        #   path of the document, or a document reference object.
+        # @param [Hash] data The document's fields and values.
+        # @param [true, String|Symbol, Array<String|Symbol>] merge When provided
+        #   and `true` all data is merged with the existing docuemnt data.  When
```
quartzmo: s/docuemnt/document/ (in 12 locations total)

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1013 @@
...
+          end
+
+          nil
+        end
+
+        ##
+        # Write to document with the provided object values. If the document
+        # does not exist, it will be created. By default, the provided data
+        # overwrites existing data, but the provided data can be merged into the
+        # existing document using the `merge` argument.
```
quartzmo: This is a little hard to follow. Maybe adding this line from [
Add Data to Cloud Firestore](https://firebase.google.com/docs/firestore/manage-data/add-data) will help to clarify why one would use `merge`:

> If you're not sure whether the document exists, pass the option to merge the new data with any existing document to avoid overwriting entire documents.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,990 @@
...
+        def batch
+          batch = Batch.from_database self
+          yield batch
+          batch.commit
+        end
+
+        ##
+        # Create a document with the provided object values.
+        #
+        # The batch will fail if the document already exists.
```
quartzmo: It might help to explain why a batch is being used, or simply replace "batch" with "operation" if it can be considered an implementation detail.

quartzmo:
It seems potentially misleading for `Batch` to offer read methods such as `get`, `select`, etc:

```ruby
        #   firestore.batch do |b|
        #     # Get a document reference
        #     nyc_ref = b.doc "cities/NYC"
        #
        #     # Create a document
        #     b.create(nyc_ref, { name: "New York City" })
        #   end
```

The semantics here are identical to that of `Transaction`, but in fact the reads are not part of the batch operation at all: 

> If you do not need to read any documents in your operation set, you can execute multiple write operations as a single batch that contains any combination of set(), update(), or delete() operations.

A batch can be understood as the complement of a read-only transaction. `ReadOnlyTransaction` does not contain write methods, however.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
blowmage: Added.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
blowmage: Added.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
blowmage: Oops, yes it was missed. Added.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
blowmage: Changed to "operation".

blowmage:
> And do you want to add Firestore to the top-level README in this PR?

I do not. I think it would be confusing to add to the top-level README until there is a gem that users can install.

blowmage:
> It seems potentially misleading for `Batch` to offer read methods such as `get`, `select`, etc

I understand why you say that, but the ruby library is doing something extra that I think is more in-line with Rubyist expectations. But yes, the requirements document does indicate that the write batch operations should be specific to write operations:

```ruby
# Get a document reference
nyc_ref = firestore.doc "cities/NYC"

# Create a document in a batch
firestore.batch do |b|
  b.create(nyc_ref, { name: "New York City" })
end
```

This PR delivers that usage. However, I believe a more idiomatic approach is to make all writes that occur within a closure use the batch. Similar to the approach taken by ActiveRecord's transactions. To deliver something like this, objects created or retrieved within the batch will make writes on the batch as long as they are made within the closure:

```ruby
# Functionally equivalent to the previous example
firestore.batch do |b|
  b.doc("cities/NYC").create({ name: "New York City" })
end
```

quartzmo:
>  Similar to the approach taken by ActiveRecord's transactions. 

It is the similarity to transactions that concerns me: Batches are not read-write transactions, but if the semantics are identical, with reads performed on the object yielded to the block, people unfamiliar with the difference may be misled. That's why I recommend removing the read methods from `Batch` and showing examples similar to your first one, above:

```ruby
# Read existing documents
nyc_ref = firestore.doc "cities/NYC"
sf_ref = firestore.doc "cities/SF"

# Update multiple document in a batch rather than in a 
# transaction to avoid contention
firestore.batch do |b|
  b.update(nyc_ref, { name: "Big Apple" })
  b.update(sf_ref, { name: "San Fran" })
end
```

blowmage:
Your code comments are misleading. That code does not read the documents, it creates document reference objects. No reads are happening there.

quartzmo:
That's correct.

google-cloud-firestore/Rakefile
```diff
@@ -0,0 +1,230 @@
+# Copyright 2017, Google Inc. All rights reserved.
```
quartzmo: GAX boilerplate now reads: "Google LLC All rights reserved."

blowmage: I opened GoogleCloudPlatform/google-cloud-ruby#1882 for this.

google-cloud-firestore/lib/google/cloud/firestore/document/reference.rb
```diff
@@ -0,0 +1,418 @@
...
+            else
+              context.database.set self, data, merge: merge
+            end
+          end
+
+          ##
+          # Write to document with the provided object values. If the document
+          # does not exist, it will be created. By default, the provided data
+          # overwrites existing data, but the provided data can be merged into
+          # the existing document using the `merge` argument.
```
frankyn: There doesn't seem to be a merge argument or example. I thought the service did this automatically for updates?

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
frankyn: ~~ Accidentally removed @blowmage's comment on that he added it. ~~

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,663 @@
...
+
+            ensure_context!
+            context.doc "#{collection_path}/#{document_path}"
+          end
+          alias_method :document, :doc
+
+          ##
+          # Create a document with random document identifier.
+          #
+          # The batch will fail if the document already exists.
```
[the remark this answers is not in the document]
quartzmo: This was changed to "the operation" in other usages.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,663 @@
...
+          #   # Get a collection reference
+          #   cities_col = firestore.col "cities"
+          #
+          #   # Get and print all city documents
+          #   cities_col.docs do |city|
+          #     puts "#{city.document_id} has #{city[:population]} residents."
+          #   end
+          #
+          def docs &block
+            query.run(&block)
```
schmidt-sebastian: Can we make CollectionReference inherit from Query, or is that not a thing in Ruby? This would greatly reduce the code and comments in this class.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,663 @@
...
+          #   # Get a collection reference
+          #   cities_col = firestore.col "cities"
+          #
+          #   # Get and print city documents
+          #   cities = ["cities/NYC", "cities/SF", "cities/LA"]
+          #   cities_col.get_all(cities).each do |city|
+          #     puts "#{city.document_id} has #{city[:population]} residents."
+          #   end
+          #
+          def get_all *docs, mask: nil, &block
```
schmidt-sebastian: This should sit on Firestore and accept documents in any collection.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,663 @@
...
+          #
+          #   # Get/run a query
+          #   query.get do |city|
+          #     puts "#{city.document_id} has #{city[:population]} residents."
+          #   end
+          #
+          def query
+            ensure_context!
+
+            Query.start(parent_path, context).from(collection_id)
```
schmidt-sebastian: Note that in other APIs this relationship is reversed. A CollectionReference is a subclass of a Query.

blowmage:
> I had one open question about custom objects with the Ruby client library.

I didn't see that in the original Firestore requirements. Is there any guidance on how this is supposed to work? Does the library automatically map a Document::Snapshot's data to the class? Is this something that is needed to release? Or can it be added afterwards?

Here are some quick thoughts on how to possibly implement this: Probably the best is to include a module in your class. Another option is to see of the object can be represented as a Hash to use it as a Hash. The simplest way is to check if the object responds to the `to_h` as method, but so many objects can respond to that method that it might easily be a false positive. It would be better if it responded to `to_hash`, but common libraries such as `Struct` and `OpenStruct` don't do this by default.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
```diff
@@ -0,0 +1,663 @@
...
+          # {Query#start_at}.)
+          #
+          # @param [Object] values The field value to start the query at.
+          #
+          # @return [Query] A query with `start_at` called on it.
+          #
+          # @example
+          #   require "google/cloud/firestore"
+          #
+          #   firestore = Google::Cloud::Firestore.new
```
schmidt-sebastian: This documentation should point out how this relates to orderBy.

google-cloud-firestore/lib/google/cloud/firestore/convert.rb
```diff
@@ -0,0 +1,545 @@
...
+              content = obj.read.force_encoding "ASCII-8BIT"
+              encoded_content = Base64.strict_encode64 content
+              Google::Firestore::V1beta1::Value.new bytes_value: encoded_content
+            else
+              fail ArgumentError,
+                   "A value of type #{obj.class} is not supported."
+            end
+          end
+
+          def create_writes doc_path, data
```
schmidt-sebastian: I got confused by this name. This is more of a write_for_create than a createWrite(). The last one suggest that it adds a new write, but doesn't necessarily say that it is about the 'create' operation.

schmidt-sebastian:
> I had one open question about custom objects with the Ruby client library.
> 
> I didn't see that in the original Firestore requirements. Is there any guidance on how this is supposed to work? Does the library automatically map a Document::Snapshot's data to the class? Is this something that is needed to release? Or can it be added afterwards?
> 
> Here are some quick thoughts on how to possibly implement this: Probably the best is to include a module in your class. Another option is to see of the object can be represented as a Hash to use it as a Hash. The simplest way is to check if the object responds to the to_h as method, but so many objects can respond to that method that it might easily be a false positive. It would be better if it responded to to_hash, but common libraries such as Struct and OpenStruct don't do this by default.

Whether this is a requirement really depends on how we expect this client to be used. Is it easy to create Firestore objects without it? Or do users commonly have their own types that they want to serialize? We added in Java, since POJOs are very common, but left it out of most other clients.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1020 @@
...
+        # @return [Project] project resource
+        def project
+          @database.project
+        end
+
+        ##
+        # The database resource the Cloud Firestore batch belongs to.
+        #
+        # @return [Database] database resource.
+        def database
```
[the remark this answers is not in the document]
blowmage: This is a bit of future proofing, as my understand is that users will be able to create multiple databases in the future, and we would benefit by being more explicit. But we can alias this method as `firestore` if you would prefer that.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1020 @@
...
+        #
+        #   firestore.batch do |b|
+        #     # Get the root collections
+        #     b.cols.each do |col|
+        #       # Print the collection
+        #       puts col.collection_id
+        #     end
+        #   end
+        #
+        def cols
```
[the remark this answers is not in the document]
blowmage: Yes, I think this is something that will be popular with Rubyists. This is needed to support an additional style coding for using batches. For example, the simplest way to update a couple of documents is as follows:

```ruby
# Update multiple documents, makes multiple API calls
firestore.col("cities").doc("NYC").update({ name: "Big Apple" })
firestore.col("cities").doc("SF").update({ name: "San Fran" })
```

That isn't very efficient, as it makes multiple API calls. Instead, users can change how they perform the updates to make a single API call by using a batch:

```ruby
# create document references
nyc_ref = firestore.doc "cities/NYC"
sf_ref = firestore.doc "cities/SF"

# Update multiple document in the same batch
firestore.batch do |b|
  b.update(nyc_ref, { name: "Big Apple" })
  b.update(sf_ref, { name: "San Fran" })
end
```

While this code is more performant than the first example, it isn't as direct. Rubyists have a strong aesthetic for "beautiful code". So while this PR definitely supports this style, we also want to support the following:

```ruby
# Update multiple document in batch, makes a single API call
firestore.batch do |b|
  b.col("cities").doc("NYC").update({ name: "Big Apple" })
  b.col("cities").doc("SF").update({ name: "San Fran" })
end
```

By allowing a Batch to act like the other context we can support this style of coding. Collections and documents that were created in a Batch will add their updates to the Batch as long as the Batch is open. I think this will be a big benefit for Rubyists by allowing them the ability to structure their code this way.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1020 @@
...
+        end
+
+        ##
+        # @private commit the batch
+        def commit
+          ensure_not_closed!
+          @closed = true
+          return nil if @writes.empty?
+          resp = service.commit @writes
+          Convert.timestamp_to_time resp.commit_time
```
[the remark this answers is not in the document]
blowmage: The reason I didn't do this originally is that I felt it was a confusing result. For example, if the user calls `#update` they don't know if that will create one Write or two (a write and a transform). If the user calls `#create`, `#set`, and `#update` on a batch, and gets 4 write results, how do they know which results are for which operation? WriteResult does not identify what change it is for.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
blowmage: We can, but Rubyists generally prefer composition over inheritance. That is why both Collection::Reference and Query support the same interface. The implementation mechanism isn't as important. This is known as "Duck Typing".

I believe there are advantages to both the implementation and documentation for keeping them separate and not using inheritance here.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
blowmage: Yes, and it does. `Database#get_all` accepts documents of any collection. `Collection::Reference#get_all` also exists as a convenience.

Also, users can call `Collection::Reference#get_all` and pass document reference objects that belong to other collections and the operation will return the appropriate document snapshot.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
blowmage: Agreed. I will update the documentation.

google-cloud-firestore/lib/google/cloud/firestore/convert.rb
blowmage: These methods are all internal private helper methods, so the names aren't as important as they aren't part of the public API. It can return more than one Write value, which is why its plural. We can rename these private helper methods if that is wanted.

blowmage:
> Whether this is a requirement really depends on how we expect this client to be used.

So I looked at the Python example again, and they don't actually accept the custom object, they call `to_dict` on it and pass the `Dict` object to the Firestore code. We can easily do the same with Ruby's `Hash` as well. Here are what our two code examples would be in that case:

```ruby
City = Struct.new(:name, :state, :country, :capital, :population)
```

```ruby
city = City.new("Los Angeles", "CA", "USA", false, 3900000)
db.col("cities").doc("LA").set(city.to_h)
```

Would this approach be acceptable?

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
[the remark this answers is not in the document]
blowmage: Yes, absolutely. But, this code example is demonstrating the `Batch#where` method. So we want to show calling `where` on the `Batch` object.  When written in the order you provided you are calling `where` on a `Query` object.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,1020 @@
...
+
+        ##
+        # Creates a query object with method `order` called on it. (See
+        # {Query#order}.)
+        #
+        # @param [String, Symbol] field A field mask to order results with.
+        # @param [String, Symbol] direction The direction to order the results
+        #   by. Optional. Default is ascending.
+        #
+        # @return [Query] A query with `order` called on it.
```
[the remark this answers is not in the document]
blowmage: I'll add documentation for acceptance values.

google-cloud-firestore/lib/google/cloud/firestore/document/reference.rb
blowmage: The `#update` method always merges, and there is no argument to control the behavior. 

The `#set` method replaces by default, but can merge when the `merge` argument is provided. It does have additional code examples demonstrating how to merge.

blowmage: Oops, I see what you are pointing out now. You are correct, this part of the `#update` method's documentation was accidentally duplicated from the `#set` method's documentation. This has been fixed.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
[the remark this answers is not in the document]
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
schmidt-sebastian: The last behavior you described seems confusing. Is there a reason we need to have this method exposed in two places? We would ideally want all clients to be the same, as this makes porting code from one platform to the others much easier. With this addition, porting from Ruby to other languages is no longer straightforward.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
schmidt-sebastian: Having the same documentation here generates a lot of code duplication. I would rather we concentrate this documentation in one place.

FWIW, every other client uses inheritance to solve this problem.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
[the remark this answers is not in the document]
schmidt-sebastian: You need to strip the extra result that the server returns to make this less confusing for the user.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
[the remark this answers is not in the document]
schmidt-sebastian: Do we have a precedent for this? I think it might make more sense to have to distinct Firestore clients when we have multiple databases. This is what we do for the Firebase RTDB SDK for example.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #   cities_col = firestore.col "cities"
+        #
+        #   # Get the document for NYC
+        #   nyc_ref = cities_col.doc "NYC"
+        #
+        #   # Set the name for NYC
+        #   firestore.set(nyc_ref, { name: "New York City" })
+        #
+        def col collection_path
+          if collection_path.to_s.split("/").count.even?
```
schmidt-sebastian: Does this reject empty strings?

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get and print all city documents
+        #   firestore.docs("cities").each do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def docs collection_path, &block
```
schmidt-sebastian: This is not part of the Firestore API spec.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        end
+        alias_method :document, :doc
+
+        ##
+        # Retrieves a list of documents with data.
+        #
+        # @param [String, Document::Reference] docs One or more strings
+        #   representing the path of the document, or document reference
+        #   objects.
+        # @param [Array<String|Symbol>, String|Symbol] mask A list of field
```
schmidt-sebastian: The mask is not part of the API spec.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a document reference
+        #   nyc_ref = firestore.doc "cities/NYC"
+        #
+        #   # Create a document
+        #   firestore.create(nyc_ref, { name: "New York City" })
+        #
+        def create doc, data
```
schmidt-sebastian: This should only exist on a DocumentReference.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        # @example Set a document and merge only name:
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Set a document
+        #   firestore.set("cities/NYC", { name: "New York City" },
+        #                 merge: [:name])
+        #
+        def set doc, data, merge: nil
```
schmidt-sebastian: The same is true for all these methods - they should be a method on DocumentReference.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        # executed atomically at a single logical point in time in a database.
+        #
+        # All changes are accumulated in memory until the block completes.
+        # Transactions will be automatically retried when possible. See
+        # {Transaction}.
+        #
+        # @see https://firebase.google.com/docs/firestore/manage-data/transactions
+        #   Transactions and Batched Writes
+        #
+        # @param [String] previous_transaction The transaction identifier of a
```
schmidt-sebastian: The transaction retry is an implementation detail that should not be part of the public API.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+            # Sleep with incremental backoff
+            sleep(backoff *= 1.3)
+            # Create new transaction and retry
+            transaction = Transaction.from_database \
+              self, previous_transaction: transaction.transaction_id
+            retries += 1
+            retry
+          rescue Google::Cloud::InvalidArgumentError => err
+            # Return if a previous call has succeeded
+            return nil if retries > 0
```
schmidt-sebastian: Does this block automatically cause retries? We need to make sure we retry on commit failures without backoff when a transaction fails due to a document changes on the backend.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #     # Create a query
+        #     query = rtx.query.select(:population).from(:cities)
+        #
+        #     # Get/run a query
+        #     rtx.get(query).each do |city|
+        #       puts "#{city.document_id} has #{city[:population]} residents."
+        #     end
+        #   end
+        #
+        def read_only_transaction read_time: nil
```
schmidt-sebastian: This is not part of the public API.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #
+        #   # Create a query
+        #   query = firestore.query.select(:population).from(:cities)
+        #
+        #   # Get/run a query
+        #   firestore.get(query).each do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def query
```
schmidt-sebastian: This should sit on CollectionReference.

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #
+        #   # Create a query
+        #   query = firestore.select(:population).from(:cities)
+        #
+        #   # Get/run a query
+        #   firestore.get(query).each do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def select *fields
```
schmidt-sebastian: Sorry, but I really don't think we should be duplicating these comments multiple times. They are super helpful and well-written, but they will just be as useful if they only exist in a single parent class :)

google-cloud-firestore/lib/google/cloud/firestore/database.rb
```diff
@@ -0,0 +1,1013 @@
...
+        #
+        #   # Create a query
+        #   query = firestore.select(:population).from(:cities)
+        #
+        #   # Get/run a query
+        #   firestore.get(query).each do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def get obj
```
schmidt-sebastian: This should be an API of DocumentReference.

google-cloud-firestore/lib/google/cloud/firestore/collection.rb
blowmage: I don't understand. The requested API is still present, we are simply adding additional conveniences that we think Ruby developers would appreciate. Is this not okay to do?

schmidt-sebastian: In my mind (and I think I speak for all of us on the Firestore client side), any API addition on any platform is only justified if it is required to make the client idiomatic for its environment. So if Ruby users expect these methods to exist on every class (sorry, I am exaggerating here, but I also have never really used Ruby before), then this is OK. I do suspect that this is not the case though. 

We have a very clearly defined API (which went to months of API review process and refinement) and we are using the same API in all of our Mobile Clients (Android, Web, iOS, C++) and the six released Server SDKs. For anyone that is familiar with any of the the supported languages, it is currently extremely straightforward to port code from one platform to another. While one language might use Promises and another one may use Tasks, the existing API differences can be explained by the underlying platform. If the Ruby client diverges from this API, Firestore Ruby will no longer adhere to the same API standards and our shared customers will likely start to expect certain API features from the existing implementations that we have explicitly decided not to implemented.

We are still open to suggestions when it comes to our API surface, but we value consistency a lot and would like to ensure that every user of Firestore gets the exact same experience, no matter what language they choose to use. If Ruby supports API endpoints that are not available in other SDKs, then that is no longer the case.

As one data point, here is users asking for Query.select() support on our mobile SDKs because it exists in Node: https://github.com/firebase/firebase-js-sdk/issues/212

blowmage: Well, I think it is a mistake not to take the accepted norms and idioms of each language and environment into account, but I don't think I will be able to convince you that these affordances and enhancements are in addition to the clearly defined api, and not a divergence from it. I'll close this PR and open a separate PR that has all these conveniences removed.

PR #1879, opened 2017-12-14
Datastore Transaction Automatic Retry

blowmage:
Update Datastore to auto retry errors setting previous_transaction.
This brings Datastore to parity with Spanner and Firestore.

### DIFF
--- a/google-cloud-datastore/.rubocop.yml
+++ b/google-cloud-datastore/.rubocop.yml
@@ -32,7 +32,7 @@ Metrics/PerceivedComplexity:
 Metrics/AbcSize:
   Max: 25
 Metrics/MethodLength:
-  Max: 20
+  Max: 25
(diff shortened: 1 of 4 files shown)


quartzmo:
Shouldn't we be concerned about adding this behavior so late in the game? It seems almost certain that existing users will have implemented their own retry logic outside of the library.

blowmage:
I don’t think so. It’s a feature enhancement, like any other enhancement.

bmclean:
@blowmage Would you consider making the deadline configurable? The fixed deadline of 60 seconds could be a problem when running on Heroku (30 second request timeout).

bmclean:
Looks awesome @blowmage. Thank you!

blowmage:
Yeah, that's a good suggestion. I added `deadline` as an argument, same as done in Spanner.

PR #1880, opened 2017-12-15
Pub/Sub Pull Receipt

blowmage:
Add implementation of streaming pull receipt, we well as several threading fixes.

PR #1881, opened 2017-12-15
Add support for GAX 1.0

geigerj:
GAX 1.0 introduces a breaking change in the LRO client instantiation in which the service path argument has been removed to match the general GAPIC template. The GAPICs that use LRO have been updated to support this change.

This PR is in support of a subsequent 1.0 release of Video Intelligence and beta release of Monitoring.


google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb
```diff
@@ -136,6 +136,10 @@ def self.location_path project, location
               )
             end
 
+            class OperationsClient < Google::Longrunning::OperationsClient
+              SERVICE_ADDRESS = SERVICE_ADDRESS
```
blowmage: Is this correct?

google-cloud-bigtable/google-cloud-bigtable.gemspec
```diff
@@ -20,7 +20,8 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 0.10.1"
+  gem.add_dependency "google-gax", "~> 1.0.0"
```
blowmage: Should the dependencies be changed to `~> 1.0`? Is your intent to disallow use of google-gax version 1.1.0?

geigerj: Whoops, updated.

google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb
geigerj: Yes, it's setting `OperationsClient::SERVICE_ADDRESS` to `BigtableInstanceAdminClient::SERVICE_ADDRESS`. (We could probably improve the code generator to be explicit here, since this is confusing...)

quartzmo: Wow, how does this work?

google-cloud-speech/lib/google/cloud/speech/v1/speech_client.rb
```diff
@@ -138,9 +125,11 @@ def initialize \
               updater_proc = credentials.updater_proc
             end
 
+            package_version = Gem.loaded_specs['google-cloud-speech'].version.version
```
blowmage: Why not use `Google::Cloud::Speech::VERSION` here?

google-cloud-bigtable/google-cloud-bigtable.gemspec
quartzmo: What about google-cloud-core, should it remain `~> 1.0.0`?

google-cloud-video_intelligence/Rakefile
```diff
@@ -154,6 +179,16 @@ task :ci do
   header "google-cloud-video_intelligence test", "*"
   sh "bundle exec rake test"
 end
+
+namespace :ci do
```
blowmage: Why is the task `ci:smoke_test` being added? Is this supposed to be used somewhere?

google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb
blowmage: It works. But I agree it looks confusing.

```ruby
class Foo
  VERSION = "1.0"
end

class Bar < Foo
  VERSION = VERSION
end

class Baz < Foo
  # Don't explicitly set the constant
end

Foo::VERSION #=> "1.0"
Bar::VERSION #=> "1.0"
Baz::VERSION #=> "1.0"

# Change Foo's constant
Foo::VERSION.gsub! "1.0", "1.1"

Foo::VERSION #=> "1.1"
Bar::VERSION #=> "1.0"
Baz::VERSION #=> "1.1"
```

google-cloud-bigtable/google-cloud-bigtable.gemspec
blowmage: Yes, this should be changed for all the google-gax dependencies. Once a library that follows SEMVER is moved to 1.0 we should always allow new minor versions as it should be backwards compatible on the major version.

blowmage:
What is the timeline for releasing support for google-gax 1.0?

google-cloud-speech/lib/google/cloud/speech/v1/speech_client.rb
geigerj: GAPIC doesn't generate this constant and isn't aware of manual code.

Agree that it's a cleaner solution for GAPIC to generate the constant (although we would have to be careful not to overwrite it accidentally on refreshes). Filed https://github.com/googleapis/toolkit/issues/1749.

google-cloud-video_intelligence/Rakefile
geigerj: This is spurious. Filed https://github.com/googleapis/toolkit/issues/1750.

google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb
geigerj: Filed https://github.com/googleapis/toolkit/issues/1751.

google-cloud-speech/lib/google/cloud/speech/v1/speech_client.rb
blowmage: This value is the same as `lib_version` being passed into the constructor, right? Why not just use that value here as well?

Or should the `gapic` version be signifying something else?

geigerj: `lib_name/lib_version` exist for historical reasons that don't apply to `google-cloud-ruby`. They actually will not be set for GAPIC-only APIs.

google-cloud-bigtable/google-cloud-bigtable.gemspec
geigerj: Done.

google-cloud-video_intelligence/google-cloud-video_intelligence.gemspec
```diff
@@ -20,7 +20,9 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 0.10.1"
+  gem.add_dependency "google-gax", "~> 1.0"
+  gem.add_dependency "googleapis-common-protos", "~> 1.3.1"
+  gem.add_dependency "googleauth", "~> 0.6.1"
```
blowmage: I don't like setting these dependency if it can be avoided, which I believe it can. What happens if google-gax 1.1.0 is released and changes the dependencies to `googleapis-common-protos ~> 1.4.0` and `googleauth ~> 0.7.0`? That would cause dependency conflicts that could otherwise be avoided.

geigerj: I agree with this. I am checking now to see why these were added to the autogen templates.

geigerj:
> What is the timeline for releasing support for google-gax 1.0?

I can take this on after this is merged by following https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/RELEASING.md, unless you have concerns.

google-cloud-video_intelligence/google-cloud-video_intelligence.gemspec
geigerj: Ah, it looks like maybe these were always there and the reason we're seeing them again in the diff is that you manually removed them in https://github.com/geigerj/google-cloud-ruby/commit/8682c7d8f0849e347ece239f377a51a80765b7f5#diff-0ee68bc1b35563909430d4b42adbcce1.

geigerj: Filed: https://github.com/googleapis/toolkit/issues/1752. For future reference, please tag me when editing autogen files so I know to get the appropriate modifications propagated back into the generator.

google-cloud-language/google-cloud-language.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 0.10.1"
+  gem.add_dependency "google-gax", "~> 1.0"
   gem.add_dependency "googleapis-common-protos", "~> 1.3.1"
```
blowmage: Can you also remove the googleapis-common-protos and googleauth direct dependencies from this gem as well?

blowmage:
@geigerj I'd like to get #1871 and #1880 merged, and #1882 resolved, before the next round of releases. Are you expecting to get this PR merged and do releases today? Or are you looking at next week?

geigerj:
@blowmage I'm fine holding off a few days. Do we have rough ETAs? I am OK holding off until next week to get those PRs merged, but wouldn't want this to wait much more than a week.

blowmage:
Just waiting on #1880 now. I'll submit a PR for #1882 as soon as this is merged. I bet we can get this done early next week.

google-cloud-language/google-cloud-language.gemspec
blowmage: @geigerj ^

geigerj: Updated.

quartzmo: LGTM

PR #1888, opened 2017-12-19
Update RELEASING.md

quartzmo:
Move verification steps into loop for each tag build.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -8,28 +8,6 @@ The Google Cloud Ruby Client project uses [semantic versioning](http://semver.or
 
 After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pulls) for a release have been merged and all [Circle CI builds](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-ruby) are green, you may create a release as follows:
 
-1. Build the gem locally.
-
-    ```sh
-    $ rake build
-    ```
-
-1. Install the gem locally.
-
-    ```sh
-    $ rake install
-    ```
-
-1. Using IRB (not `rake console`!), manually test the gem that you installed in the previous step.
-
-1. Review the changes since the last release.
-
-    ```sh
-    $ bundle exec rake changes[<gem>]
-    ```
-
-1. Review the commits in the changes output, making notes of significant changes. (For examples of what a significant change is, browse the changes in the gem's `CHANGELOG.md`
-
 1. If you haven't already, switch to the master branch, ensure that you have no changes, and pull from origin.
 
     ```sh
@@ -38,15 +16,39 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
     $ git pull <remote> master --rebase
     ```
 
+1. Build the gem locally. (Depending on your environment, you may need to `bundle exec` to rake commands; this will be shown.)
+
+    ```sh
+    $ cd <gem>
+    $ bundle exec rake build
+    ```
+
+1. Install the gem locally.
+
+    ```sh
+    $ bundle exec rake install
+    ```
+
+1. Using IRB (not `rake console`!), manually test the gem that you installed in the previous step.
+
+1. Return to the root directory of the project, and review the changes since the last release.
+
+    ```sh
+    $ cd ..
+    $ bundle exec rake changes[<gem>]
+    ```
+
+1. Review the commits in the changes output, making notes of significant changes. (For examples of what a significant change is, browse the changes in the gem's `CHANGELOG.md`.
+
 1. Edit the gem's `CHANGELOG.md`. Using your notes from the previous step, write bullet-point lists of the major and minor changes. You can also add examples, fixes, thank yous, and anything else helpful or relevant. See google-cloud-node [v0.18.0](https://github.com/GoogleCloudPlatform/google-cloud-node/releases/tag/v0.18.0) for an example with all the bells and whistles.
 
 1. Edit the gem's `version.rb` file, changing the value of `VERSION` to your new version number.
 
 1. Edit the gem's entry in `docs/manifest.json`, adding your new version number to the head of the list, and moving `"master"` to be just below it.
 
-1. If your version change is greater than the [semver](http://semver.org/) patch version, edit the requirement for the gem in `google-cloud/google-cloud.gemspec` and `stackdriver/stackdriver.gemspec`, replacing the old version number for the gem with your new version number.
+1. If the package is `< 1.0.0` and your version change is greater than the [semver](http://semver.org/) patch version, or if the package is `>= 1.0.0` and your version change increments the [semver](http://semver.org/) major version, edit the requirement for the gem in `google-cloud/google-cloud.gemspec` and `stackdriver/stackdriver.gemspec` (if the package is a dependency of stackdriver) , replacing the old version number for the gem with your new version number.
 
-1. Test all the version dependencies are correct.
+1. In the root directory of the project, test that all the version dependencies are correct.
 
     ```sh
     $ bundle update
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -8,45 +8,47 @@ The Google Cloud Ruby Client project uses [semantic versioning](http://semver.or
...
-1. Review the changes since the last release.
+1. Install the gem locally.
 
     ```sh
-    $ bundle exec rake changes[<gem>]
+    $ bundle exec rake install
     ```
 
-1. Review the commits in the changes output, making notes of significant changes. (For examples of what a significant change is, browse the changes in the gem's `CHANGELOG.md`
+1. Using IRB (not `rake console`!), manually test the gem that you installed in the previous step.
```
geigerj: For some reason, with a few of the gems I had to `bundle exec irb` or else I would get a `Gem::MissingSpecError` when I tried to `require` the gem. IIRC, this only happened for `stackdriver` and `debugger`.

RELEASING.md
```diff
@@ -8,45 +8,47 @@ The Google Cloud Ruby Client project uses [semantic versioning](http://semver.or
...
 1. Edit the gem's entry in `docs/manifest.json`, adding your new version number to the head of the list, and moving `"master"` to be just below it.
 
-1. If your version change is greater than the [semver](http://semver.org/) patch version, edit the requirement for the gem in `google-cloud/google-cloud.gemspec` and `stackdriver/stackdriver.gemspec`, replacing the old version number for the gem with your new version number.
+1. If the package is `< 1.0.0` and your version change is greater than the [semver](http://semver.org/) patch version, or if the package is `>= 1.0.0` and your version change increments the [semver](http://semver.org/) major version, edit the requirement for the gem in `google-cloud/google-cloud.gemspec` and `stackdriver/stackdriver.gemspec` (if the package is a dependency of stackdriver) , replacing the old version number for the gem with your new version number.
 
-1. Test all the version dependencies are correct.
+1. In the root directory of the project, test that all the version dependencies are correct.
 
     ```sh
     $ bundle update
```
geigerj: Below (line 64): I was curious about this last `git pull`. If commits have been added between when you last `pull`ed above and here, ISTM you shouldn't include them in the tag, since none of the verification steps you just did apply to the new commits.

RELEASING.md
```diff
@@ -77,28 +79,28 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
...
+1. If your version change is greater than the [semver](http://semver.org/) patch version, then when you are done releasing all individual packages, you should follow these same instructions to release the `google-cloud` umbrella package. Furthermore, if your major releases included at least one dependency of the `stackdriver` umbrella package (currently those dependencies are `google-cloud-debugger`, `google-cloud-error_reporting`, `google-cloud-logging`, and `google-cloud-trace`), then you should also follow these same instructions to release the `stackdriver` umbrella package. It is important that the dependencies of the `google-cloud` and `stackdriver` gems remain compatible so the two can co-exist in the same bundle.
```
geigerj: I would change this condition to "If you updated `google-cloud/google-cloud.gemspec` or `stackdriver/stackdriver.gemspec` in the previous steps" (or something similar) for clarity. It's unclear what "version change" means here because you may have just released multiple gems.

geigerj: Also, why not just make the `google-cloud`/`stackdriver` release part of the individual gem releases, instead of a separate final step?

RELEASING.md
```diff
@@ -77,28 +79,28 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
...
 
 1. On the [google-cloud-ruby releases page](https://github.com/GoogleCloudPlatform/google-cloud-ruby/releases), click [Draft a new release](https://github.com/GoogleCloudPlatform/google-cloud-ruby/releases/new). Complete the form. Include the bullet-point lists of the major and minor changes from the gem's `CHANGELOG.md`. You can also add examples, fixes, thank yous, and anything else helpful or relevant. See google-cloud-node [v0.18.0](https://github.com/GoogleCloudPlatform/google-cloud-node/releases/tag/v0.18.0) for an example with all the bells and whistles.
 
 1. Click `Publish release`.
 
-1. Check again that the [Circle CI build](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-ruby) has passed for the tag. Also, inspect the build logs to confirm that the `release` task completed successfully, and that the docs build succeeded. This can fail on a green build because it is an "after" action in the build.
-
-1. Confirm that the gem for the new version is available on [RubyGems.org](https://rubygems.org/gems/google-cloud).
+1. Repeat steps 1 through 20 if you are releasing multiple gems.
```
geigerj: Thanks -- I see now this is what you were trying to tell me on chat yesterday...

RELEASING.md
quartzmo: > why not just make the google-cloud/stackdriver release part of the individual gem releases, instead of a separate final step?

Because of the order, I think they need to be released after their dependencies?

RELEASING.md
quartzmo: @blowmage This pre-dates me. What's your take on this step?

RELEASING.md
quartzmo: Interesting! I'll take a look at those gems to see what might be different.

quartzmo:
@geigerj Thanks for the input, I will update this PR.

RELEASING.md
blowmage: Something like `bundle exec rake test` will do this. If it runs then you can be fairly confident you don't have any incorrect dependencies.

RELEASING.md
blowmage: Debugger is not included in google-cloud because it requires a more recent version of Ruby than the other gems.

RELEASING.md
blowmage: Right. If you do something like update a dependency that is not compatible with the previous release, then you want to wait until _all_ the individual gems are released before releasing a new umbrella gem like google-cloud or stackdriver. We have to be very careful to avoid bad dependencies.

Also, it means less releases, which is important since they are so time consuming.

geigerj: Ah, right, I didn't mean "release google-cloud in each iteration of the above loop", I meant to consider treating the google-cloud release as "just another gem being released" to simplify the instructions, instead of calling it out in its own special step. But I buy that it's worth having its own step to ensure that it doesn't accidentally get released too early.

RELEASING.md
geigerj: @blowmage Right, you've already effectively done that in a previous step (`bundle exec rake ci[yes]`). My point is that if you run that, and then you pull new commits afterward, you lose the verification you just did.

RELEASING.md
blowmage: Actually, this might be something different. Investigating.

geigerj: Maybe we can recommend `gem install` in a clean gemset for this step, then?

blowmage: In general, I would like to avoid requiring gemset functionality to do releases because the mechanics are different depending on what Ruby version manager folks are using. But, it might be worth noting that Bundler's `install` task may not always work as expected, and to fall back onto using `rake build`, `gem install`, and an empty gemset to verify this if you run into issues with `rake install`.

Chris, what do you think?

quartzmo: I'll try it too, but success apparently won't be proof of anything. I'm open to adding a version of:

> Bundler's install task may not always work as expected, and to fall back onto using rake build, gem install, and an empty gemset to verify this if you run into issues with rake install.

RELEASING.md
quartzmo: @geigerj I think this step needs to be moved up to line 96 (just below the loop). I'll simplify the instructions in this step (using your text), but I won't remove it.

quartzmo:
Updated for @geigerj's comments, please review the new version.

RELEASING.md
blowmage: @quartzmo If you add that then I'll approve.

quartzmo: @blowmage It has been added. See latest.

quartzmo:
```
-1. Install the gem locally.
+1. Build the gem locally. (Depending on your environment, you may need to `bundle exec` to rake commands; this will be shown.)
```

quartzmo:
```
-1. Install the gem locally.
 +1. Install the gem locally. (The `rake install` task shown below may not always work as expected. Fall back to running `gem install` in an empty gemset if needed.)
```

quartzmo:
@blowmage What do I need to add? This?

```
-1. Install the gem locally.
 +1. Install the gem locally. (The `rake install` task shown below may not always work as expected. Fall back to running `gem install` in an empty gemset if needed.)
```

RELEASING.md
blowmage: I see it now. Thanks!

PR #1889, opened 2017-12-19
Remove reference to sync and async queries

tswast:
BigQuery makes no distinction between synchronous and asynchronous queries. This was a misunderstanding that unfortunately propagated to the docs. Such language has now been removed from https://cloud.google.com/bigquery/docs/running-queries

In all cases, BigQuery creates a query "job" to run the query, which can then be polled until it is complete.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery.rb
@@ -92,10 +92,6 @@ module Google
     # as discussed in the guide [Migrating from legacy
     # SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql).
     #
-    # In addition, BigQuery offers both synchronous and asynchronous methods, as
-    # explained in [Querying
-    # Data](https://cloud.google.com/bigquery/querying-data).
-    #
     # ### Standard SQL
     #
     # Standard SQL is the preferred SQL dialect for querying data stored in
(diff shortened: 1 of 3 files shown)


quartzmo:
@tswast Thanks for this PR. We have a queue of release builds that we need to get through, but we'll get this green and merge it soon.

quartzmo:
I noticed that a little lower in the same section of the guide, there are subheadings for `### Synchronous queries` and `### Asynchronous queries` that refer back to the sentence you removed.

```ruby
    # ### Synchronous queries
    #
    # Let's start with the simpler synchronous approach...
```

If orphaned, I think these sections will still be ok, but I just wanted to bring it up.

quartzmo:
The official docs discussion of asynchronous method is also referenced in [project.rb#L98-L99](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb#L98-L99) and [dataset.rb#L671-L672](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb#L671-L672).

Do you want to change: 

```ruby
        # Queries data using the [asynchronous
        # method](https://cloud.google.com/bigquery/querying-data).
```
to
```ruby
         # Queries data using an asynchronous method.    
```

tswast:
Ah yeah, good point.

In a lot of ways, asynchronous / synchronous doesn't really get to the heart of what we want to show. It's more about if we have a reference to a job object or not. I'll follow-up with some more clean-up.

tswast:
Okay, additional references removed. Thanks.

quartzmo:
@tswast Thanks again for this PR!

PR #1894, opened 2017-12-21
Add Ruby Firestore Client

blowmage:
This PR adds the Firestore Ruby client library. It incorporates all the feedback received from the previous PR, and is updated to use the latest google-gax dependency.

```ruby
firestore.batch do |b|
  # Set the data for NYC
  b.set("cities/NYC", { name: "New York City" })

  # Update the population for SF
  b.update("cities/SF", { population: 1000000 })

  # Delete LA
  b.delete("cities/LA")
end
```

It also implements support for transactions

```ruby
cities_col = firestore.col("cities").doc("SF")
cities_col.set({ name: "San Francisco",
           state: "CA",
           country: "USA",
           capital: false,
           population: 860000 })

firestore.transaction do |tx|
  new_population = tx.get(cities_col).data[:population] + 1
  tx.update(city, { population: new_population })
end
```


google-cloud-firestore/lib/google/cloud/firestore/commit_response.rb
```diff
@@ -0,0 +1,139 @@
...
+          # the previous update_time.
+          #
+          # @return [Time] The last update time.
+          attr_accessor :update_time
+
+          ##
+          # The results of applying each change, in the order given.
+          #
+          # @return [Array<Object>] The change results.
+          attr_accessor :transform_results
```
schmidt-sebastian: The way we have structured the API, our users are not actually exposed to the difference between a normal write and a document transform. It feels slightly weird to break this here and expose transformation results.

google-cloud-firestore/lib/google/cloud/firestore/commit_response.rb
```diff
@@ -0,0 +1,139 @@
...
+        #
+        # @return [Array<CommitResponse::WriteResult>] The write results.
+        attr_accessor :write_results
+
+        ##
+        # @private
+        def self.from_grpc grpc, client
+          commit_time = Convert.timestamp_to_time grpc.commit_time
+          write_results = Array(grpc.write_results).map do |write_result|
+            WriteResult.from_grpc write_result, client
```
schmidt-sebastian: Can you file a GitHub issue (or add a TODO) to strip WriteResults for DocumentTransforms?

google-cloud-firestore/lib/google/cloud/firestore/convert.rb
```diff
@@ -0,0 +1,546 @@
...
+          ESCAPED_FIELD_PATH = /\A\`(.*)\`\z/
+
+          def extract_field_paths hash
+            field_paths = hash.keys.map do |path|
+              unless path.is_a? Array
+                path = path.to_s.split(".")
+                path.each do |p|
+                  # Only check for invalid characters when a string
+                  if INVALID_FIELD_PATH_CHARS.match(p) || p["."]
+                    fail ArgumentError, "invalid character"
```
schmidt-sebastian: Can you tell the users here to use the FieldPath class?

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
```diff
@@ -0,0 +1,387 @@
...
+        #
+        # @return [String] document resource path.
+        def path
+          @path
+        end
+
+        # @!group Access
+
+        ##
+        # Retrieves a list of collections nested under the document with data.
```
schmidt-sebastian: I am not sure what you mean by 'with data'. Maybe you can just remove this?

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
```diff
@@ -0,0 +1,387 @@
...
+
+          return enum_for(:cols) unless block_given?
+
+          collection_ids = service.list_collections path
+          collection_ids.each { |collection_id| yield col(collection_id) }
+        end
+        alias_method :collections, :cols
+
+        ##
+        # Retrieves a collection nested under the document with data.
```
schmidt-sebastian: Same as above.

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
```diff
@@ -0,0 +1,387 @@
...
+          end
+
+          CollectionReference.from_path "#{path}/#{collection_path}", client
+        end
+        alias_method :collection, :col
+
+        ##
+        # Retrieve the document data.
+        #
+        # @return [DocumentSnapshot] document with data.
```
schmidt-sebastian: It might not have data if it doesn't exist.

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
```diff
@@ -0,0 +1,387 @@
...
+          ensure_client!
+
+          client.batch { |b| b.set self, data, merge: merge }
+        end
+
+        ##
+        # Update the document with the provided data (fields and values). The
+        # provided data is merged into the existing document data.
+        #
+        # The operation will fail if the document does not exist.
```
schmidt-sebastian: As stated above, please state that update accepts field paths.

google-cloud-firestore/lib/google/cloud/firestore/document_snapshot.rb
```diff
@@ -0,0 +1,367 @@
...
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   nyc_snap = firestore.doc("cities/NYC").get
+        #
+        #   nyc_snap.cols.each do |col|
+        #     puts col.collection_id
+        #   end
+        #
+        def cols &block
```
schmidt-sebastian: DocumentSnapshots should only provide access to the data in a document as well as its DocumentReference (and maybe path).

google-cloud-firestore/lib/google/cloud/firestore/document_snapshot.rb
```diff
@@ -0,0 +1,367 @@
...
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   atlantis_snap = firestore.doc("cities/Atlantis").get
+        #
+        #   # Does Atlantis exist?
+        #   atlantis_snap.missing? #=> true
+        #
+        def missing?
```
schmidt-sebastian: Is this different from `not exists`?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        ##
+        # @private The Google::Firestore::V1beta1::Query object.
+        attr_accessor :query
+
+        ##
+        # @private The firestore client object.
+        attr_accessor :client
+
+        ##
+        # Restricts documents matching the query to return only data for the
```
schmidt-sebastian: Please mention field_paths somewhere.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        #   cities_col = firestore.col "cities"
+        #
+        #   # Create a query
+        #   query = cities_col.all_descendants
+        #
+        #   query.get do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def all_descendants
```
schmidt-sebastian: Sorry, this is an instance of where I am going to ask you to remove this functionality purely because it is not yet available in any of the other SDKs. We would like to figure out how we can expose it in the mobile SDKs first.

Please keep this somewhere (you could even just mark in private if Ruby enforces it), so that we can resurrect this soon.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        #   cities_col = firestore.col "cities"
+        #
+        #   # Create a query
+        #   query = cities_col.direct_descendants
+        #
+        #   query.get do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def direct_descendants
```
schmidt-sebastian: This should be the default.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+          Query.start new_query, parent_path, client
+        end
+
+        ##
+        # Specifies an "order by" clause on a field.
+        #
+        # @param [String, Symbol] name A field mask to order results with.
+        # @param [String, Symbol] direction The direction to order the results
+        #   by. Values that start with "a" are considered `ascending`. Values
+        #   that start with "d" are considered `descending`. Default is
```
schmidt-sebastian: Haha, nice :)

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        end
+        alias_method :order_by, :order
+
+        ##
+        # Skips to an offset in a query. If the current query already has
+        # specified an offset, this will overwrite it.
+        #
+        # @param [Integer] num The number of results to skip.
+        #
+        # @return [Query] New query with `offset` called on it.
```
schmidt-sebastian: BTW, thank you for those docs! They are super helpful for my review (and for the users as well).

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+          new_query = @query.dup
+          new_query ||= StructuredQuery.new
+
+          new_query.limit = Google::Protobuf::Int32Value.new(value: num)
+
+          Query.start new_query, parent_path, client
+        end
+
+        ##
+        # Starts query results at a particular document value. The result set
```
schmidt-sebastian: I would say: At a set of field values (rather than document value).

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        # @example
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a collection reference
+        #   cities_col = firestore.col "cities"
+        #
+        #   # Create a query
+        #   query = cities_col.start_at("NYC")
```
schmidt-sebastian: This won't actually work since you need to specify an order by.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        #
+        #   query.get do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def start_at *values
+          new_query = @query.dup
+          new_query ||= StructuredQuery.new
+
+          values = values.flatten.map { |value| Convert.raw_to_value value }
```
schmidt-sebastian: You need to special handle __name___ (or rather FieldPath.documentId()) to make sure that you pass a resource name.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,613 @@
...
+        end
+
+        ##
+        # Starts query results at a particular document value. The result set
+        # will include the document specified by `values`.
+        #
+        # If the current query already has specified `start_at` or
+        # `start_after`, this will overwrite it.
+        #
+        # When the query is sent to the server, the `values` will be used in
```
schmidt-sebastian: Can you reword this to more clearly state the relationship? I am not sure that I would understand it from this comment.

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
```diff
@@ -0,0 +1,640 @@
...
+    module Firestore
+      ##
+      # # Transaction
+      #
+      # A transaction in Cloud Firestore is a set of reads and writes that
+      # execute atomically at a single logical point in time.
+      #
+      # All changes are accumulated in memory until the block passed to
+      # {Database#transaction} completes. Transactions will be automatically
+      # retried when possible. See {Database#transaction}.
```
schmidt-sebastian: Can you say something like `when documents change before committing the transaction`.

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
```diff
@@ -0,0 +1,640 @@
...
+
+        # @!endgroup
+
+        ##
+        # @private commit the transaction
+        def commit
+          ensure_not_closed!
+          return rollback if @writes.empty?
+          @closed = true
+          return if @writes.empty?
```
schmidt-sebastian: You need to commit even when you have no writes in order to release the read locks.

blowmage:
> As a convenience to other reviewers, might I suggest that you submit the generated code in a separate PR?

By "generated code" do you mean the GAPIC generated files? If so, I don't want to delay the approval effort anymore than we have to. But I have broken those files out into a separate commit in this PR.

google-cloud-firestore/Rakefile
```diff
@@ -0,0 +1,230 @@
...
+
+  ARGV.clear
+  IRB.start
+end
+
+require "yard"
+require "yard/rake/yardoc_task"
+YARD::Rake::YardocTask.new
+
+desc "Generates JSON output from google-cloud-firestore .yardoc"
```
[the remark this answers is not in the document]
blowmage: I believe this is correct. The directory where the cached documentation resides is named `.yardoc`.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,373 @@
...
+    module Firestore
+      ##
+      # # Batch
+      #
+      # A batch in Cloud Firestore is a set of writes that execute
+      # atomically at a single logical point in time in a database.
+      #
+      # All changes are accumulated in memory until the block passed to
+      # {Database#batch} completes. Unlike transactions, batches are not
+      # automatically retried. See {Database#batch}.
```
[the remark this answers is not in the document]
blowmage: Updated.

google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -0,0 +1,356 @@
...
+        #   end
+        #
+        def batch
+          batch = Batch.from_client self
+          yield batch
+          batch.commit
+        end
+
+        ##
+        # Create a transaction to perform multiple reads and changes that are
```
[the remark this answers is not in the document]
blowmage: Fixed. This was already changed in `transaction.rb`.

google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -0,0 +1,356 @@
...
+            # Sleep with incremental backoff
+            sleep(backoff *= 1.3)
+            # Create new transaction and retry
+            transaction = Transaction.from_client \
+              self, previous_transaction: transaction.transaction_id
+            retries += 1
+            retry
+          rescue Google::Cloud::InvalidArgumentError => err
+            # Return if a previous call has succeeded
+            return nil if retries > 0
```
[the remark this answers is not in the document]
blowmage: I don't believe so. The requirements we were given for Retry Behavior state that only `Unavailable` errors are retried. It also has the following pseudocode implementation which does not show retrying on an `InvalidArgument` error:

```
function commit (): error {
  for try := 0 ; true ; try ++ { 
    err := commit ()
    switch code ( err ):
      case Unavailable :
        Sleep with exponential backoff
        continue
      case InvalidArgument :
        if try == 0 :
          return err // on first call, must be a bad request
        else :
          return OK // on later calls, previous call must have succeeded
      default : // do not retry on any other error
        return err
    }
  }
}
```

If you can give us new requirements we can implement them, but AFAICT we have implemented the requirements we have correctly.

google-cloud-firestore/lib/google/cloud/firestore/commit_response.rb
blowmage: Yes, I'll plan on addressing this in a separate PR. But I have many questions about what these requirements are.

google-cloud-firestore/lib/google/cloud/firestore/commit_response.rb
blowmage: So the only think WriteResult has is `update_time`? Would it be possible to just return the time then? And it looks like the [node implementation](https://github.com/googleapis/nodejs-firestore/blob/master/src/write-batch.js#L549) sets `update_time` to `commit_time` when `update_time` is missing. Is that right?

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
blowmage: Changed to "document snapshot".

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
blowmage: Changed.

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
blowmage: Changed to "document snapshot".

google-cloud-firestore/lib/google/cloud/firestore/document_snapshot.rb
blowmage: Removed.

google-cloud-firestore/lib/google/cloud/firestore/document_snapshot.rb
blowmage: No.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: I would remove these, as we have removed the other items you have requested, except that we depend on this functionality to clean up after our acceptance tests. So I'm afraid I will have to keep these methods and mark these as private so they won't appear in the documentation.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: It is the default, but if you previously called `all_descendants` this will change it back. Changed the method to private so it won't appear in the documentation.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Changed.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: As I understand it, when you set a cursor using field values, the values are matched to the field names the query is set it to order by. Is that correct?

@quartzmo, can you help us in working this better?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: The code example has been updated to also specify `order`.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: The field value here is still treated as a string though, right? The field name in `order` is what needs to support `__name__`, isn't it?

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
blowmage: Updated.

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
blowmage: Ick, I thought I had changed this. Fixed now.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,617 @@
...
+        #   cities_col = firestore.col "cities"
+        #
+        #   # Create a query
+        #   query = cities_col.select(:population)
+        #
+        #   query.get do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def select *fields
```
blowmage: @schmidt-sebastian I have a question about this. Can you select a nested field path? If so, are the rules for formatting nested field paths the same rules for formatting that `update` used? e.g.

```
foo.`bar.baz`.biz
```

schmidt-sebastian: Yes, the query API accepts this field path format for where, orderBy and select.

google-cloud-firestore/lib/google/cloud/firestore/batch.rb
```diff
@@ -0,0 +1,373 @@
...
+
+        ##
+        # Update the document with the provided data (fields and values). The
+        # provided data is merged into the existing document data.
+        #
+        # The batch will fail if the document does not exist.
+        #
+        # @param [String, DocumentReference] doc A string representing the
+        #   path of the document, or a document reference object.
+        # @param [Hash] data The document's fields and values.
```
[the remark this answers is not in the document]
blowmage: The documentation has been updated with more information on about using field paths in the data.

google-cloud-firestore/lib/google/cloud/firestore/convert.rb
blowmage: Updated.

google-cloud-firestore/lib/google/cloud/firestore/document_reference.rb
blowmage: The parameter documentation has been updated.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: The documentation has been updated.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: This has been updated. It now says:

> The values provided here are for the field paths provides to `order`. Values provided to `start_after` without an associated field path provided to `order` will result in an error.

What do you think?

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,381 @@ module Cloud
...
+    # high performance, and ease of application development. While the Cloud
+    # Firestore interface has many of the same features as traditional
+    # databases, as a NoSQL database it differs from them in the way it
+    # describes relationships between data objects.
+    #
+    # For more information about Cloud Firestore, read the [Cloud
+    # Firestore Documentation](https://cloud.google.com/firestore/docs/).
+    #
+    # The goal of google-cloud is to provide an API that is comfortable to
+    # Rubyists. Authentication is handled by {Google::Cloud#spanner}. You can
```
blowmage: Use `Google::Cloud#firestore` instead of `Google::Cloud#spanner`.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,381 @@ module Cloud
...
+    # provide the project and credential information to connect to the Cloud
+    # Spanner service, or if you are running on Google Compute Engine this
```
blowmage: Use `Firestore` instead of `Spanner`.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,381 @@ module Cloud
...
+    #
+    # Cloud Firestore stores data in Documents, which are stored in Collections.
+    # Cloud Firestore creates collections and documents implicitly the first
+    # time you add data to the document. You do not need to explicitly create
+    # collections or documents. (For more information, see [Adding Data to Cloud
+    # Firestore](https://cloud.google.com/firestore/docs/manage-data/add-data).
+    #
+    # To create or overwrite a single document, use {Firestore::Client#doc} to
+    # obtain a document reference. (This does not create a document in Cloud
+    # Firestore.) Then, call {Firestore::DocumentReference#set} to create the
```
blowmage: Are you intentionally not showing `create` or `update` here?

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,381 @@ module Cloud
...
+    # firestore = Google::Cloud::Firestore.new
+    #
+    # # Get a document reference
+    # nyc_ref = firestore.doc "cities/NYC"
+    #
+    # nyc_ref.set({ name: "New York City" }) # Document created
+    # ```
+    #
+    # When you use this combination of `doc` and set` to create a new document,
+    # you must specify an ID for the document. Sometimes there isn't a
```
blowmage: Calling `CollectionReference#doc` without an argument will also generate a random ID.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,381 @@ module Cloud
...
+    #
+    # ```ruby
+    # require "google/cloud/firestore"
+    #
+    # firestore = Google::Cloud::Firestore.new
+    #
+    # # Get a document reference
+    # nyc_ref = firestore.doc "cities/NYC"
+    #
+    # field_delete = Google::Cloud::Firestore::FieldValue.delete
```
blowmage: I would use the following for this code example:

```ruby
nyc_ref.update({ name: "New York City",
                 trash: firestore.field_delete })
```

Users can call `Google::Cloud::Firestore::FieldValue.delete` to get the value, but I think we should give higher preference to showing `Google::Cloud::Firestore::Client#field_delete` instead.

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: yikes! good catch

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: I followed the examples provided in [Adding Data to Cloud Firestore](https://cloud.google.com/firestore/docs/manage-data/add-data), which prefer `set` and do not show `create`. However, I need to add an example of `update`, thank you.  I'll do that now.

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: again...

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: OK, i'll change it, thanks.

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: Hmmmmm. Which should we show?

blowmage: How about something like this?

> When you use this combination of `doc` and set` to create a new document, you can specify an ID for the document. If you do not have a meaningful ID for the document, you can omit the ID and Firestore auto-generate an ID for you. You can also perform both of these operations by calling {Firestore::CollectionReference#add}, which will fail if the document already exists.

google-cloud-firestore/lib/google/cloud/firestore.rb
blowmage: Makes sense. It is simpler for new users to follow that path.

blowmage:
@schmidt-sebastian Have you had a chance to look at the requested changes made in this PR?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: This has been implemented and the documentation made more explicit.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,652 @@
...
+            fail "missing collection_id to specify descendants."
+          end
+
+          new_query.from.last.all_descendants = true
+
+          Query.start new_query, parent_path, client
+        end
+
+        ##
+        # @private This is marked private can't be removed.
```
schmidt-sebastian: Nit: Missing word - This is marked private AND/BUT can't be removed.
Here and above.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -0,0 +1,652 @@
...
+
+          new_query.from.last.all_descendants = false
+
+          Query.start new_query, parent_path, client
+        end
+
+        ##
+        # Filters the query on a field.
+        #
+        # @param [String, Symbol] field A field mask to filter results with.
```
schmidt-sebastian: This comment seems out-of-place here.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,442 @@ module Cloud
...
+    # configuration is taken care of for you. You can read more about the
+    # options for connecting in the [Authentication
+    # Guide](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/guides/authentication).
+    #
+    # ## Adding data
+    #
+    # Cloud Firestore stores data in Documents, which are stored in Collections.
+    # Cloud Firestore creates collections and documents implicitly the first
+    # time you add data to the document. You do not need to explicitly create
+    # collections or documents. (For more information, see [Adding Data to Cloud
```
schmidt-sebastian: Nit: I would remove "You do not need to explicitly create collections or documents." since it's stated right before.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,442 @@ module Cloud
...
+    # random_ref.set({ name: "New York City" })
+    #
+    # # The document ID is randomly generated
+    # random_ref.document_id #=> "RANDOMID123XYZ"
+    # ```
+    #
+    # You can perform both of the operations shown above, auto-generating
+    # an ID and creating the document, in a single call to
+    # {Firestore::CollectionReference#add}. This operation will fail if the
+    # document already exists.
```
schmidt-sebastian: "This operation will fail if the document already exists."

This should basically never happen for `add` and hence I would remove this sentence.

google-cloud-firestore/lib/google/cloud/firestore.rb
```diff
@@ -21,6 +21,442 @@ module Cloud
...
+    #
+    # # Get a document reference
+    # nyc_ref = firestore.doc "cities/NYC"
+    #
+    # nyc_ref.set({ name: "New York City" }, merge: true)
+    # ```
+    #
+    # Using {Firestore::DocumentReference#update} instead of `set` ensures that
+    # the provided data is merged into the existing document data. Unlike `set`,
+    # the `update` operation will fail if the document does not exist.
```
schmidt-sebastian: The main differentiator is that update allows for direct deep-updates using field paths.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
```diff
@@ -0,0 +1,211 @@
...
+      #   nested_field_path = Google::Cloud::Firestore::FieldPath.new(
+      #     :favorites, :food
+      #   )
+      #   user_snap.get(nested_field_path) #=> "Pizza"
+      #
+      class FieldPath
+        include Comparable
+
+        ##
+        # Creates a field path object representing the nested fields for
```
schmidt-sebastian: "a nested field"

We only support one field per field path.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
```diff
@@ -0,0 +1,211 @@
...
+        #
+        def initialize *fields
+          @fields = fields.flatten.map(&:to_s)
+          @fields.each do |field|
+            fail ArgumentError, "empty paths not allowed" if field.empty?
+          end
+        end
+
+        ##
+        # The individual field representing the nested fields for document data.
```
schmidt-sebastian: This should be private.

If you intend to keep it public, please update the documentation to state that this is the path to a single nested field.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
```diff
@@ -0,0 +1,211 @@
...
+        #   nested_field_path = Google::Cloud::Firestore::FieldPath.new(
+        #     :favorites, :food
+        #   )
+        #   nested_field_path.fields #=> ["favorites", "food"]
+        #
+        def fields
+          @fields
+        end
+
+        ##
```
schmidt-sebastian: This is private and we should be free to change the escaping mechanism without impacting users.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
```diff
@@ -0,0 +1,211 @@
...
+        #   query.get do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
+        #   end
+        #
+        def self.document_id
+          new :__name__
+        end
+
+        ##
+        # Creates a field path object representing the nested fields for
```
schmidt-sebastian: This is private as well.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
```diff
@@ -0,0 +1,211 @@
...
+
+        protected
+
+        START_FIELD_PATH_CHARS = /\A[a-zA-Z_]/
+        INVALID_FIELD_PATH_CHARS = %r{[\~\*\/\[\]]}
+
+        def escape_field_for_path field
+          field = String field
+
+          return "`#{field}`" if INVALID_FIELD_PATH_CHARS.match field
```
schmidt-sebastian: You need to escape the backtick.

google-cloud-firestore/lib/google/cloud/firestore/field_value.rb
```diff
@@ -0,0 +1,113 @@
...
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a document reference
+        #   nyc_ref = firestore.doc "cities/NYC"
+        #
+        #   field_delete = Google::Cloud::Firestore::FieldValue.delete
+        #
+        #   nyc_ref.update({ name: "New York City",
```
schmidt-sebastian: Nit: I'd suggest to just use `Google::Cloud::Firestore::FieldValue.delete` inline (same below)

blowmage: We can't add it inline and keep the formatting because then the line is too long and the style guide verification will fail. This is why we show `firestore.delete` in the other examples. But here we need to have an example that actually shows calling `Google::Cloud::Firestore::FieldValue.delete`.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: While Ruby can escape the backpack, it doesn't need to be escaped.

```ruby
"`foo`" == "\`foo\`" #=> true
```

schmidt-sebastian: We need to ensure the following escaping:

```
"foo`bar" => "`foo\`bar`
"foo\bar" => `foo\\bar`
"foo\`bar" => `foo\\\`bar`
```

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Fixed here and above.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Yep. It wasn't removed when the new argument text was added. Removed.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: Marked private and updated the description.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: Marked private.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: Marked private.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
blowmage: Precisely which characters in the individual field node need to be escaped?

schmidt-sebastian: Only the ones shown here - backslash and backtick

blowmage: Is this what you are looking for?

```ruby
def escape_field_for_path field
  field = String field

  if INVALID_FIELD_PATH_CHARS.match(field) ||
     field["."] || field["`"] || field["\\"]
    escaped_field = field.gsub /[\`\\]/, "`" => "\\\`", "\\" => "\\\\"
    return "`#{escaped_field}`"
  end

  return field if START_FIELD_PATH_CHARS.match field

  "`#{field}`"
end
```

It is only escaping <code>&#96;</code> and `\`

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: Removed.

google-cloud-firestore/lib/google/cloud/firestore.rb
quartzmo: Removed.

google-cloud-firestore/lib/google/cloud/firestore.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore.rb
blowmage: Fixed.

quartzmo: Text and example have been replaced with demonstration of `update` using `FieldPath`.

blowmage:
@schmidt-sebastian We have made the requested changes and rebased the PR branch on master to pick up a fix for a dependency that was failing the build. PTAL.

google-cloud-firestore/lib/google/cloud/firestore/field_path.rb
schmidt-sebastian: That looks right to me.

schmidt-sebastian:
The client surface LGTM as far as I can tell. I did not review for correctness, as I have never worked with Ruby.

BTW, the changes would be much easier to review if you simply added commits to this PR that contained the more recent changes. Thanks!

blowmage:
@frankyn Do you have time to do a review?

byu:
👏

PR #1895, opened 2017-12-21
Generate Trace V2

landrito:


blowmage:
> I am not sure why this additive change would affect the credentials in the Trace unit tests that are now failing on CI.

The failure is on the doctests. New code examples have been added but the configuration for them has not been added to `support/doctest_helper.rb`, so the doctests are trying to make actual API calls.

The CI environment is not set up with default credentials to connect to any live API services. All connection is configured through the `TEST` environment variables, to catch things like this.

geigerj:
Thanks, @blowmage.

@landrito could you add `doctest.skip "Google::Cloud::Trace::V2::TraceServiceClient"` to [`doctest_helper.rb`](https://github.com/landrito/google-cloud-ruby/blob/d44bb48644d9e2f5cd0d45e4b5cdec5819bd8ab2/google-cloud-trace/support/doctest_helper.rb#L72)?

landrito:
Will do!

landrito:
Done.

PR #1896, opened 2017-12-21
Generate Dataproc V1

landrito:


google-cloud-dataproc/google-cloud-dataproc.gemspec
```diff
@@ -0,0 +1,30 @@
...
+  gem.platform      = Gem::Platform::RUBY
+
+  gem.files         = `git ls-files -- lib/*`.split("\n") +
+                      ["README.md", "LICENSE", ".yardopts"]
+  gem.require_paths = ["lib"]
+
+  gem.required_ruby_version = ">= 2.0.0"
+
+  gem.add_dependency "google-gax", "~> 1.0.0"
+  gem.add_dependency "googleapis-common-protos", "~> 1.3.1"
```
geigerj: This is actually something that needs to be fixed in the generator, but could you hand-edit to remove lines 24-25? The toolkit issue is https://github.com/googleapis/toolkit/issues/1752.

google-cloud-dataproc/acceptance/google/cloud/dataproc/v1/cluster_controller_smoke_test.rb-e
```diff
@@ -0,0 +1,46 @@
...
+# AUTO GENERATED BY GAPIC
+
+require "minitest/autorun"
+require "minitest/spec"
+
+require "google/cloud/dataproc"
+
+describe "ClusterControllerSmokeTest" do
+  it "runs one smoke test with list_clusters" do
+    unless ENV["SMOKE_TEST_PROJECT"]
```
geigerj: `SMOKE_TEST_PROJECT` -> `DATAPROC_TEST_PROJECT`

README.md
```diff
@@ -29,6 +29,7 @@ This client supports the following Google Cloud Platform services at a [Beta](#v
 * [Cloud Pub/Sub](#cloud-pubsub-beta) (Beta)
 * [Stackdriver Monitoring API](#stackdriver-monitoring-api-beta) (Beta)
 * [Stackdriver Trace](#stackdriver-trace-beta) (Beta)
+* [Cloud Dataproc](#cloud-dataproc) (Beta)
```
blowmage: It is generally not a good idea to add these before the gem is released. There may be several days between when this PR is merged and a release happens, and users may be confused why they can't download the gem. I would prefer to see changes to this file removed.

google-cloud/google-cloud.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = ">= 2.0.0"
 
   gem.add_dependency "google-cloud-bigquery", "~> 0.30.0"
+  gem.add_dependency "google-cloud-dataproc", "~> 0.1.0"
```
blowmage: This dependency should not be added until the gem is released.

google-cloud/docs/toc.json
```diff
@@ -122,6 +122,32 @@
         }
       ]
     },
+    {
```
blowmage: This configuration should not be added until the gem is released.

geigerj:
@blowmage Should these steps (and below) become part of RELEASING.md as a special case for first-time releases? Or how do you prefer they be incorporated into the flow?

I would prefer these changes go in, and we merge shortly before release. We're trying to federate as much of the GAPIC process to other googlers as possible, and it would be ideal not to special-case these steps.

quartzmo:
> Should these steps (and below) become part of RELEASING.md as a special case for first-time releases?

Certainly, RELEASING.md could be expanded to cover the addition of new packages, but

blowmage:
> Should these steps (and below) become part of RELEASING.md as a special case for first-time releases? Or how do you prefer they be incorporated into the flow?

I don't have any strong opinions either way. I can see arguments for both of those options. I'm not much of a process wonk, so I would defer to others on how best to document this.

That said, the problem with keeping these in the PR is that we never know what issues we will run into once the PR hits master. Sometimes it takes days to get the acceptance tests working and the build green again. I am just super cautious because we've run into problems like this before.

In fact, I seem to remember having to release an emergency patch for one gem, and we had to remove and re-add dependencies as part of that process. It was a mess and I'd like to avoid that in the future. To me, merging into master is no the same as starting the release process. Those should be two separate steps.

geigerj:
@blowmage OK, I'll request that Googlers not include these changes in their PRs, and will add them when I start the release process.

geigerj:
Re: documentation, let's no-op for now given the unusual state we're in and circle back to process questions at the start of Q1.

google-cloud-dataproc/acceptance/google/cloud/dataproc/v1/cluster_controller_smoke_test.rb-e
landrito: Done.

google-cloud-dataproc/google-cloud-dataproc.gemspec
landrito: Done.

README.md
landrito: Removed.

google-cloud/docs/toc.json
landrito: Removed.

google-cloud/google-cloud.gemspec
landrito: Removed.

blowmage:
@landrito It looks like the file `google-cloud-dataproc/grep` is still in the PR.

landrito:
Woops removed!

blowmage:
RE: the failed build, it seems like we had the same issue of the CI build failing on GAPIC-only gem when running the coverage report just a few months ago. It _looks_ like its a problem with not finding the correct paths, but I don't think that was the actual problem. I'm investigating.

blowmage:
Okay, I figured out why the build is failing. Apparently the gem was removed from the top-level Gemfile. We need the following added back to that file and the build should complete:

```ruby
gem "google-cloud-dataproc", path: "google-cloud-dataproc"
```

The reason it failed was that the top-level rake task didn't know to treat that directory as a gem, so when the tests attempted to run the following code:

```ruby
package_version = Gem.loaded_specs['google-cloud-dataproc'].version.version
```

There gem was not found, and the call to `.version` raised, which is what caused the `LoadError` when trying to load the file during the tests.

So, to be clear, we don't want to add the actual dependency in google-cloud until it is released, but we do need to add the gem to Gemfile so that the rake tasks know to treat it as a gem.

landrito:
Ah makes sense. Thanks for looking into that @blowmage!

quartzmo:
@geigerj @landrito The master build is [currently failing](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-ruby/2979) with: 

```sh
NameError: undefined local variable or method `args' for main:Object

/home/ubuntu/google-cloud-ruby/google-cloud-dataproc/Rakefile:70:in `block in <top (required)>'
```

landrito:
https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1902

This should fix it.

PR #1899, opened 2017-12-21
Release DLP v2beta1

jbolinger:
Created a release for v2beta following playbook instructions.


google-cloud-dlp/google-cloud-dlp.gemspec
```diff
@@ -0,0 +1,28 @@
+# -*- ruby -*-
+# encoding: utf-8
+
+Gem::Specification.new do |gem|
+  gem.name          = "google-cloud-dlp"
+  gem.version       = "0.20.0"
```
geigerj: Just a note: this value seems to come from stale configuration files for DLP. New libraries will now launch at 0.1.0, but we can fix this at release time.

PR #1900, opened 2017-12-22
fixed acceptance tests running from a bad copy paste

jbolinger:
My last change #1899 broke the CI. I think it was just this copy/paste error (missed the first line).

I don't think the docs ask to run the acceptance test locally, should we add that just in case?

Also, double check the version number you mentioned last time.

### DIFF
--- a/google-cloud-dlp/google-cloud-dlp.gemspec
+++ b/google-cloud-dlp/google-cloud-dlp.gemspec
@@ -3,7 +3,7 @@
 
 Gem::Specification.new do |gem|
   gem.name          = "google-cloud-dlp"
-  gem.version       = "0.20.0"
+  gem.version       = "0.2.0"
 
   gem.authors       = ["Google LLC"]
   gem.email         = "googleapis-packages@google.com"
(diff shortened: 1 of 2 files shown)


google-cloud-dlp/google-cloud-dlp.gemspec
```diff
@@ -3,7 +3,7 @@
 
 Gem::Specification.new do |gem|
   gem.name          = "google-cloud-dlp"
-  gem.version       = "0.20.0"
+  gem.version       = "0.2.0"
```
geigerj: `0.1.0`

jbolinger:
Ok, now I'm glad I asked. I'm a little confused - why the difference from
v2beta1? Is it to keep the lib version separate from the backend version?

geigerj:
@jbolinger Correct, lib version is unconnected to API service version. Many of our libraries can connect to a number of service versions.

jbolinger:
makes sense... it might be useful to call that out in the publishing doc since you have to update both.

geigerj:
You actually shouldn't have to set either on the initial generation; DLP's configuration was written several months ago in a previous artman configuration format and was mechanically translated to the current configuration files. I'm not sure at what point the `0.20.0` got introduced, but this shouldn't require intervention for new APIs going forward.

geigerj:
Confirmed `rake acceptance` works locally.

geigerj:
Build still failing because API's not enabled in the test project. Just enabled it and restarted the build.

PR #1916, opened 2018-01-02
BigQuery: Allow a URI as the storage url for a load job

jeremywadsack:
In the [BigQuery `load_job` documentation](http://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-bigquery/v0.30.0/google/cloud/bigquery/table?method=load_job-instance), it states that you can use "a file or the URI of a Google Cloud Storage file", which I understood to be a [`URI`](https://docs.ruby-lang.org/en/2.4.0/URI.html) type object. 

After having this fail, I noticed that the docs do list the types that the object can be and that `URI` isn't listed. But I thought this could be useful to others (and is to us as we're using a `URI`). 

This pull request updates the method to accept a `URI` type parameter. In fact, this alters the method to accept any object that's `#to_s` returns a string representing a GS url.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
@@ -1319,9 +1319,9 @@ module Google
         # file directly. See [Loading Data with a POST Request](
         # https://cloud.google.com/bigquery/loading-data-post-request#multipart).
         #
-        # @param [File, Google::Cloud::Storage::File, String] file A file or the
-        #   URI of a Google Cloud Storage file containing data to load into the
-        #   table.
+        # @param [File, Google::Cloud::Storage::File, String, URI] file A file
+        #   or the URI of a Google Cloud Storage file containing data to load
+        #   into the table.
         # @param [String] format The exported file format. The default value is
         #   `csv`.
         #
@@ -1495,9 +1495,9 @@ module Google
         # file directly. See [Loading Data with a POST Request](
         # https://cloud.google.com/bigquery/loading-data-post-request#multipart).
         #
-        # @param [File, Google::Cloud::Storage::File, String] file A file or the
-        #   URI of a Google Cloud Storage file containing data to load into the
-        #   table.
+        # @param [File, Google::Cloud::Storage::File, String, URI] file A file
+        #   or the URI of a Google Cloud Storage file containing data to load
+        #   into the table.
         # @param [String] format The exported file format. The default value is
         #   `csv`.
         #
@@ -1995,6 +1995,7 @@ module Google
         def load_storage url, options = {}
           # Convert to storage URL
           url = url.to_gs_url if url.respond_to? :to_gs_url
+          url = url.to_s if url.is_a? URI
 
           gapi = service.load_table_gs_url dataset_id, table_id, url, options
           Job.from_gapi gapi, service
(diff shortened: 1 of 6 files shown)


quartzmo:
@jeremywadsack Thank you for submitting this PR. I think this change seems like a good idea. It will also need to be applied to `Table#load`, `Dataset#load_job` and `Dataset#load`, but you might hold off on that until @blowmage has commented.

quartzmo:
I'll look into fixing the broken build. It appears to be unrelated to this PR.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -2010,8 +2014,8 @@ def load_local file, options = {}
 
         def storage_url? file
           file.respond_to?(:to_gs_url) ||
-            (file.respond_to?(:to_str) &&
-            file.to_str.downcase.start_with?("gs://"))
+            (file.respond_to?(:to_s) &&
```
blowmage: This is not a good change. The existence of a `to_s` method is not an indication that the object is a string, and nearly all objects have this method. Non-string objects such as File, Integer, Float, `nil`, etc. all respond to `to_s`.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -1319,9 +1319,9 @@ def extract extract_url, format: nil, compression: nil, delimiter: nil,
         # file directly. See [Loading Data with a POST Request](
         # https://cloud.google.com/bigquery/loading-data-post-request#multipart).
         #
-        # @param [File, Google::Cloud::Storage::File, String] file A file or the
-        #   URI of a Google Cloud Storage file containing data to load into the
-        #   table.
+        # @param [File, Google::Cloud::Storage::File, String, URI] file A file
```
blowmage: To explicitly support an URI object, I would add a conditional to the beginning of the method that checks if the object is a URI and convert it to a String. Or, try to parse any string input with URI and handle the argument as a URI internally.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
jeremywadsack: Understood. This was the easier approach; I didn't know which direction you wanted to go. To me the "ruby" approach would be that anything that "acts like a" gs URL string is acceptable. 

`File`, `Integer`, `Float`, and `nil` would all fail the second half of the check:

```
file.to_s.downcase.start_with?("gs://"))
```

Happy to make the change for more explicit `URI`  support, just suggesting a less constrictive option.

blowmage: That's a fair point. I'll leave it to @quartzmo to decide.

quartzmo: I'm OK with any approach that works. More explicit `URI` support seems fine.

jeremywadsack: The current approach works. Would you like me to change it to be more explicit/restrictive about `URI`? You're call.

quartzmo: Yes, if you don't mind, let's go with the explicit URI support ("a conditional to the beginning of the method that checks if the object is a URI and convert it to a String." as @blowmage suggested). Thanks @jeremywadsack, appreciate your time and work on this!

jeremywadsack:
@quartzmo @blowmage I've fixed this to only allow an explicit `URI` object and updated `Table#load`, `Dataset#load` and `Dataset#load_job` to accept that as well.

If that looks good I can squash and rebase.

quartzmo:
@jeremywadsack Can you rebase on master now for a fix to the CI build that I added yesterday (#1917)?

PR #1920, opened 2018-01-04
Fix acceptance test env var order

blowmage:
The service-specific env vars should be prioritized ahead of the shared env vars.


blowmage:
@geigerj The GAPIC generator may need to be updated to fix this as well

geigerj:
@blowmage Thanks for the heads up. Will do.

quartzmo:
:+1

PR #1924, opened 2018-01-11
Fix YCSB shell script.

haih-g:


google-cloud-spanner/benchmark/bin/ycsb
```diff
@@ -1,5 +1,19 @@
...
+#   --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 \
+#   7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+# curl -sSL https://get.rvm.io | bash -s stable --ruby
+# source $HOME/.rvm/scripts/rvm; gem install bundler
+# sudo apt-get install -y git
+# git clone https://github.com/GoogleCloudPlatform/google-cloud-ruby.git
+# cd google-cloud-ruby; bundle install
 
-bundle exec ruby ${DIR}/../ycsb.rb "${@:1}"
+source $HOME/.rvm/scripts/rvm
```
blowmage: What is users are not using RVM?

google-cloud-spanner/benchmark/bin/ycsb
```diff
@@ -1,5 +1,19 @@
...
+
+cd $HOME/google-cloud-ruby
```
blowmage: What if users are not placing the repo in their home directory? Or their repo directory is named something else?

google-cloud-spanner/benchmark/bin/ycsb
haih-g: This script assumes that the user uses RVM and is following the https://cloud.google.com/ruby/getting-started/hello-world for installing everything.

It would be nice if there is a standard way to install ruby and google-cloud-ruby package. The current installation process has many variables which is hard to map out. On the other hand, this script is for integration with PerfKitBenchmarker and we are going to make sure PerfKitBenchmarker will use RVM and put the repo at the right place.

haih-g:
Can someone take a look at the failure in cl/circleci? I took a look and it seems not related to my change.

blowmage:
@haih-g GitHub is saying that your PR has conflicts and cannot be merged until they are resolved. It looks like your branch is out of date. Can you rebase your branch against the current master branch?

haih-g:
@blowmage I think the conflict is because I renamed the file google-cloud-spanner/benchmark/ycsv.rb to google-cloud-spanner/benchmark/ycsb.rb.

There is a button "Resolve conflicts", but it's grayed out for me. Can you resolve the conflict?

blowmage:
I rebased your branch on current master and that has resolved the conflict.

haih-g:
Thx a lot! I'm trying to understand what caused the conflict. Is that because someone made some change to ycsb.rb? If this happens again before this PR is merged, would we see another conflict?

blowmage:
The build failed because your branch was out of date and missing a fix needed to work with an updated dependency. I assume the conflict was due to your branch being so out of date.

google-cloud-spanner/benchmark/bin/ycsb
quartzmo: I'm glad to see documentation added to this script. Regarding the prerequisites, I think it would be good to add a line after line 3 stating something like:

"It is intended to be run in an <X> environment, with <a>, <b> and <c>.

These dependencies should be installed as follows:"

Basically what you already have, with just a bit more introduction, and most importantly, a list of the dependencies.

haih-g: Thx! Done.

PR #1925, opened 2018-01-12
Regenerate Language GAPIC

### LINKED ISSUE #1866: Generated docs for google-cloud-language don't work
The generated documentation throws a GAX error

```ruby
require "google/cloud/language/v1"

language_service_client = Google::Cloud::Language::V1.new
document = {}
response = language_service_client.analyze_syntax(document)
```
Here's the error
.../google-gax-0.10.2/lib/google/gax/api_callable.rb:351:in `rescue in block in retryable': GaxError Exception occurred in retry method that was not classified as transient, caused by 3:One of content, or gcs_content_uri must be set. (Google::Gax::RetryError)

geigerj:


google-cloud-language/README.md
```diff
@@ -7,17 +7,14 @@ recognition, and text annotations.
 - [Client Library Documentation][]
 - [Product Documentation][]
 
-*The 0.28.0 release introduced breaking changes relative to the previous
```
blowmage: Do we want to keep the 0.28.0 migration notice?

blowmage:
I don't see any updated code examples. Are we sure those docs are correct?

geigerj:
Ah, I see. I had mistaken #1866 to refer to the API sample in the README, which is correct, rather than the method docs, which are incorrect. I'll update the #1866, but unfortunately the method samples are a known issue for the time being.

google-cloud-language/README.md
geigerj: What do you think is a reasonable amount of time to keep it for? It's been available for two months, and it's going to get nuked on every GAPIC refresh.

blowmage: > What do you think is a reasonable amount of time to keep it for?

I have no idea. I just wasn't sure if removing it was a conscious decision or an oversight.

@quartzmo Any input on how long to keep the notice?

quartzmo: I also have no idea. I'd keep it for a few more versions if it's not too much trouble. Maybe @frankyn has an opinion, since he wrote the migration guide?

blowmage:
@geigerj I rebased your branch on current master to pick up the latest Rubocop changes. That will make it much easier to get this PR's build passing.

blowmage:
@geigerj Do you want one of us to push commits to your branch to make CI pass for this PR?

geigerj:
No, let's hold off on this for now. There are configuration updates to
language coming from the API team shortly, and I'll update once they're in.

google-cloud-language/README.md
frankyn: (Thanks @quartzmo for the ping). I'm not sure how long we should have the migration guide in place. I would keep the migration guide link in documentation so users have a direct way of finding it when reading the documentation. Is there a better way to keep this in docs after each refresh?

Given that this is being removed after each refresh. Is the post-install message also being removed?

geigerj: > Is there a better way to keep this in docs after each refresh?

No, it would have to be done manually. In our discussions about the migration, I had originally suggested that this should go away on the first refresh, but since the DPE team is assuming ownership of the libraries' manual components, I'm happy to follow your guidance here.

> Given that this is being removed after each refresh. Is the post-install message also being removed?

Correct.

frankyn: Thanks @geigerj. My preference is to keep it in the docs and the post-install message

Summoning @jonparrott for visibility and input on this case.

jonparrott: Do y'all keep a `CHANGELOG`/`NEWS` or some other release log? If so, it's probably fine to just keep this notice in the `0.28.0` section of whatever that log is. It certainly doesn't need to be in the readme as we shed legacy towards v1.0.0.

geigerj: @jonparrott Yes, the migration guide is linked to from the CHANGELOG.

jonparrott:
@blowmage we can change that in the repo settings so that it's the only merge option if you want.

blowmage:
I don’t think it’s necessary to change the defaults. But the reason we encourage folks to rebase PR branches is to avoid merge these commits.

PR #1957, opened 2018-02-01
Update Spanner Acceptance Test Setup

blowmage:
We have a limited number of instances available for running tests.
The Spanner team has asked that we reuse a instance instead of
creating and deleting an instance for each test run.
Create a shared instance and use a new database per test run.
Remove the acceptance test that creates instances.

### DIFF
--- a/google-cloud-spanner/acceptance/spanner_helper.rb
+++ b/google-cloud-spanner/acceptance/spanner_helper.rb
@@ -246,16 +246,20 @@ $spanner_prefix = "gcruby-#{Date.today.strftime "%y%m%d"}-#{SecureRandom.hex(4)}
 fixture = Object.new
 fixture.extend Acceptance::SpannerTest::Fixtures
 
-job = $spanner.create_instance $spanner_prefix, name: $spanner_prefix, config: "regional-us-central1", nodes: 1
-job.wait_until_done!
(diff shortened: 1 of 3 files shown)


google-cloud-spanner/acceptance/spanner_helper.rb
```diff
@@ -246,16 +246,20 @@ def default_item_rows
 fixture = Object.new
 fixture.extend Acceptance::SpannerTest::Fixtures
 
-job = $spanner.create_instance $spanner_prefix, name: $spanner_prefix, config: "regional-us-central1", nodes: 1
-job.wait_until_done!
-fail GRPC::BadStatus.new(job.error.code, job.error.message) if job.error?
+instance = $spanner.instance "google-cloud-ruby-tests"
+instance ||= begin
+  inst_job = $spanner.create_instance "google-cloud-ruby-tests", name: "google-cloud-ruby-tests", config: "regional-us-central1", nodes: 1
```
quartzmo: What do you think about optionally loading the config from an env variable, with `regional-us-central1` as default if no env var is present?

blowmage: I don't think that is necessary. The CI environments and our development environments don't need that ability. Same as we don't need the ability to change the instance name. I think these are good/sane defaults.

PR #1958, opened 2018-02-02
Fix bug using Time object as key

blowmage:
Reading and deleting take key values. If the value is a `Time` object (for a `TIMESTAMP` value) then the API request was not formatted correctly.

Add test coverage to ensure that Time keys are formatted properly.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/commit.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/commit.rb
@@ -327,14 +327,14 @@ module Google
           keys = [keys] unless keys.is_a? Array
           return Google::Spanner::V1::KeySet.new(all: true) if keys.empty?
           if keys_are_ranges? keys
-            keys = [keys] unless keys.is_a? Array
             key_ranges = keys.map do |r|
               Convert.to_key_range(r)
             end
             return Google::Spanner::V1::KeySet.new(ranges: key_ranges)
           end
-          key_list = Array(keys).map do |i|
-            Convert.raw_to_value(Array(i)).list_value
+          key_list = keys.map do |key|
+            key = [key] unless key.is_a? Array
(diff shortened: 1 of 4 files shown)


blowmage:
The issue here was that `Time` defines a [`#to_a`](http://ruby-doc.org/core-2.5.0/Time.html#method-i-to_a) method. So we can't use `Array()` to ensure that the the value is an `Array`.

google-cloud-spanner/lib/google/cloud/spanner/commit.rb
```diff
@@ -327,14 +327,14 @@ def key_set keys
...
-            keys = [keys] unless keys.is_a? Array
             key_ranges = keys.map do |r|
               Convert.to_key_range(r)
             end
             return Google::Spanner::V1::KeySet.new(ranges: key_ranges)
           end
-          key_list = Array(keys).map do |i|
-            Convert.raw_to_value(Array(i)).list_value
+          key_list = keys.map do |key|
+            key = [key] unless key.is_a? Array
```
quartzmo: I don't get it, why is each key being wrapped in an array?

blowmage: The key, whether a table primary key, or an index key, can be one or many values. So the format for specifying an individual key is a list. This is how Spanner represents this.

quartzmo: Great, thanks.

blowmage:
```
$ irb
irb(main):001:0> now = Time.now
=> 2018-02-01 17:36:50 -0700
irb(main):002:0> [now] # what we want
=> [2018-02-01 17:36:50 -0700]
irb(main):003:0> Array(now) # not what we want
=> [50, 36, 17, 1, 2, 2018, 4, 32, false, "MST"]
```

PR #1962, opened 2018-02-09
BigQuery: add EncryptionConfiguration class

tswast:
This adds partial support for customer-managed encryption keys.

This is the first step towards https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1960

I'm not an expert in Ruby, so I'm just starting with the EncryptionConfiguration class. To get complete support, the `destination_encryption_configuration` needs to be added to query, copy, and load jobs.

Issue #1960, opened 2018-02-06
Add support for BigQuery customer-managed encryption keys


choenden:
BigQuery customer-managed encryption keys allow users to specify a Cloud KMS key to protect their BigQuery table.

- [x] Jobs: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs `destinationEncryptionConfiguration` field, which indicates which Cloud KMS key should be used for the destination.
  - [x] Add `destinationEncryptionConfiguration` field to query job configuration.
  - [x] Add `destinationEncryptionConfiguration` field to copy job configuration.
  - [x] Add `destinationEncryptionConfiguration` field to load job configuration.
- [x] Tables: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables In a table, there is a encryptionConfiguration field, which indicates which Cloud KMS key protects (or should protect in case of CreateTable) a BigQuery table.
  - [x] Add `encryptionConfiguration` to table.
- [ ] (Lower priority) Support for `getServiceAccount` https://cloud.google.com/bigquery/docs/reference/rest/v2/projects/getServiceAccount
Note that unlike the other methods mentioned above, this would generally only be called once and the resulting value (the email address) does not change - so it can easily also be called from UI/API/CLI without much hindrance (hence lower priority).
  - [ ] Add `get_service_account` method to `project` class.

blowmage:
@geigerj @jbolinger The features being asked for here are marked `experimental`. The policy has been not to implement experimental features, but exceptions have been made. What can you tell us about this feature?

geigerj:
This feature request is not currently on our feature roadmap, so I would not make an exception in this case.

@danoscarmike FYI; do you know whether there's any internal tracking of this feature that I'm not aware of?

choenden:
Hey, to clarify, the experimental tag has been removed and this is awaiting rollout. I will send internal bugs and docs your way.

geigerj:
Thanks for the clarification, @choenden!

tswast:
FYI: I won't be able to resume work on this until Wednesday, so if this is still time sensitive someone will need to follow-up my PRs with one that adds the key to query and copy jobs.

quartzmo:
Thanks. Do you want us to take care of it?

tswast:
Yeah, when you get a chance. No rush.

quartzmo:
@tswast OK, will do.

PR #1963, opened 2018-02-10
 BigQuery: add customer managed encryption options to load jobs

tswast:
Another step towards resolving #1960. Still missing copy jobs and query jobs.

Based on #1962

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -432,7 +432,8 @@ module Google
         #   length is 1,024 characters.
         # @param [String] name A descriptive name for the table.
         # @param [String] description A user-friendly description of the table.
-        # @param [EncryptionConfiguration] encryption_configuration A configuration for custom table data encryption.
+        # @param [EncryptionConfiguration] encryption_configuration A
+        #   configuration for custom table data encryption.
         # @yield [table] a block for setting the table
         # @yieldparam [Table] table the table object to be updated
         #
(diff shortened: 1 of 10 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1274,6 +1274,9 @@ def external url, format: nil
         #   dashes. International characters are allowed. Label values are
         #   optional. Label keys must start with a letter and each label in the
         #   list must have a different key.
+        # @param [EncryptionConfiguration] destination_encryption_configuration
```
blowmage: I really don't like the idea of adding even more arguments to these methods, but we can't configure this because are yielding the schema and not the job object. An alternative is to create a `LoadJob::Updater` class and yield it instead, similar to what we have done with `Table::Updater`. Would the effort to add an updater object be worth it to not add yet another argument to an already overloaded method?

quartzmo: @blowmage Is this something that you could do in a follow-up PR?

blowmage: Yes, but it would have to be made before these changes were released.

tswast:
I'll combine this with #1962, as I'll be making changes there and don't particularly want to deal with the git conflicts that'll cause.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
tswast: I started creating a `LoadJob::Updater` class at https://github.com/tswast/google-cloud-ruby/commit/80b23e742e26fdb77cc7a3265435396b392d8823

It has to act a little bit like `Schema` in the same way that `Table::Updater` does since some of the load functions in `Dataset` currently yield a `Schema` but will need to yield a `LoadJob::Updater`.

PR #1966, opened 2018-02-13
Policy Creation Fix

### LINKED ISSUE #1965: [Storage] When a bucket-level IAM policy has no assigned users with roles undefined method error is raised.
Hi all,

When I clear the IAM policy of a bucket using the UI or the Ruby client, I'm unable to view the policy or add new members as I receive the error `NoMethodError: undefined method 'each_with_object' for nil:NilClass`. 

Steps to reproduce:
1. Remove all bucket-level IAM bindings 

```ruby

bucket = storage.bucket "bucket-name"

bucket.policy do |policy|
  copy_roles = policy.roles.clone
  copy_roles.each do |role|
     policy.roles[role[0]] = []
  end
end

# NoMethodError: undefined method `each_with_object' for nil:NilClass
```

2. Trying to `puts` the bucket policy continues to error.

```ruby
puts bucket.policy

# NoMethodError: undefined method `each_with_object' for nil:NilClass

```

Gem Version `google-cloud-storage-1.9.0`.

blowmage:
`Policy` objects returned from the Google API Client may have `bindings` as a `nil` object instead of an empty array. This is because the RESTful resource uses JSON to format the data, and looks like it may omit the bindings data entirely when empty. (I say "may" because I haven't been able to reproduce this from calling the API yet.) We do not see this behavior GRPC endpoints, because of Protobuf's use of default values.

[fixes #1965]

### DIFF
--- a/google-cloud-resource_manager/test/google/cloud/resource_manager/policy_test.rb
+++ b/google-cloud-resource_manager/test/google/cloud/resource_manager/policy_test.rb
@@ -1,4 +1,4 @@
- # Copyright 2016 Google LLC
+# Copyright 2016 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,13 +14,13 @@
 
 require "helper"
 
-describe Google::Cloud::ResourceManager::Policy, :mock_res_man do
+describe Google::Cloud::ResourceManager::Policy do
   let(:etag)       { "etag-1" }
   let(:roles) { { "roles/viewer" => ["allUsers"] } }
   let(:policy)    { Google::Cloud::ResourceManager::Policy.new etag, roles }
 
   it "knows its etag" do
-    policy.roles.must_equal roles
+    policy.etag.must_equal etag
   end
 
   it "knows its roles" do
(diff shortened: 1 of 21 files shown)


quartzmo:
@blowmage Thanks for catching this in Resource Manager as well. Wouldn't similar usages of `Google::Iam::V1::Policy` in Pub/Sub and Spanner also need this fix?

blowmage:
`Google::Iam::V1::Policy` is a protobuf resource, and has correct defaults. So this fix is not needed there.

google-cloud-resource_manager/test/google/cloud/resource_manager/policy_test.rb
```diff
@@ -34,4 +34,35 @@
...
+            role: key,
+            members: val
+          )
+        end
+      )
+
+      policy = Google::Cloud::ResourceManager::Policy.from_gapi gapi
+
+      policy.must_be_kind_of Google::Cloud::ResourceManager::Policy
+      policy.etag.must_equal etag
```
quartzmo: It's not really needed, but adding `policy.roles.wont_be :empty?` here would provide some "extra paranoid" security...

frankyn:
@blowmage thanks for the fast fix!

I have one case which I want to raise. The bucket.policy update returns a policy with a roles Hash containing a role with an empty list and a subsequent call to bucket.policy returns an empty roles Hash. 

Steps to reproduce:
1. Add a member with role to the bucket policy.
2. Delete all members associated to a role. I'm using the following script:
```ruby
response_policy = bucket.policy do |policy|
  copy_roles = policy.roles.clone
  copy_roles.each do |role|
     policy.roles[role[0]] = []
  end
end
```

3. response_policy now contains:
```
#<Google::Cloud::Storage::Policy:0x000055ec042d0a78 @etag="\bF", @roles={"roles/storage.objectViewer"=>[]}
```
4. call `bucket.policy` again:
```
#<Google::Cloud::Storage::Policy:0x000055ec04292f20 @etag="\bG", @roles={}>
```

blowmage:
@frankyn Calling `policy` with a block makes two API calls. The first to get the policy, the second to update and return the policy again. The `response_policy` object is the one returned from the update API call. It looks like there might be another update happening because the `etag` of the `response_policy` object and calling policy again are different. AFAICT the ruby library is doing the right thing.

frankyn:
Thanks for the clarification. Given the implementation I think it would be service side update. Either way the PR worked and lgtm! Thanks @blowmage!

blowmage:
Looking at the policy objects a second time and I think I see the problem. The `policy=` method is returning the policy object returned from the update API call, but ruby is returning the object that was passed to the method. This is because the method ends with `=`. So even though we have test coverage that the API calls are returning the updated policy, we are seeing the original object returned.

The solution is to rename the method to `update_policy`, which works as expected. I will alias it as `policy=` so existing code will continue to update the policy, but ruby will ensure that the original object is returned. This looks to be a language feature and we cannot work around it other than calling the method using the new name. But this will ensure that the object returned from calling `policy` with a block is the object returned from the update API call.

blowmage:
@quartzmo I've updated the PR with the update policy fix. Can you please re-review?

frankyn:
Thanks for looking closer at this @blowmage! I'll review the update as well.

frankyn:
I wanted to make sure I understand this correctly, thanks for the clarification, it makes sense.  So, when the assignment operator `=` is part of a method name it will return the original right-hand value provided and not the return value of the method.

blowmage:
> So, when the assignment operator`=` is part of a method name it will return the original right-hand value provided and not the return value of the method.

Correct.

quartzmo:
I reviewed the update, LGTM! Thanks for identifying and coming up with a fix for the issue of the return value.

PR #1968, opened 2018-02-14
BigQuery: Updates public dataset project name used in docs

alixhami:
Currently, the code samples in the docs use both `publicdata` and `bigquery-public-data` projects. 

This PR:
+ Updates all uses in the docs (not the tests) to `bigquery-public-data` for better consistency across samples and languages
+ Adds backticks around table names where they were missing in order to avoid confusion about when they are required


quartzmo:
Thanks @alixhami!

PR #1973, opened 2018-02-15
BigQuery: Add LoadJob::Updater to allow for additional properties on load jobs

tswast:
As discussed in https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1963, this adds a block to `load_job` so that new properties can be added to LoadJob without needing to add arguments to the `load_job` function.

Notes:

- This does not touch the `load` methods which start and wait for a load job. (I wasn't sure how to pass the block on to the `load_job` method.
- There's quite a bit of copy-pasted code here. I wasn't sure where to put shared private helper methods.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -1269,13 +1269,12 @@ module Google
         #   optional. Label keys must start with a letter and each label in the
         #   list must have a different key.
         #
-        # @yield [schema] A block for setting the schema for the destination
-        #   table. The schema can be omitted if the destination table already
-        #   exists, or if you're loading data from a Google Cloud Datastore
-        #   backup.
-        # @yieldparam [Google::Cloud::Bigquery::Schema] schema The schema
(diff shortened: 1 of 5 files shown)


blowmage:
@tswast Should this PR be changed to merge into the `bigquery` branch instead of `master`?

tswast:
> Should this PR be changed to merge into the `bigquery` branch instead of `master`?

If you're okay with this one going to `master`, I did leave off any `encryption_configuration` changes.

blowmage:
We are very close to releasing new versions, so I'd rather this not merged until after the release. If you don't want to wait until the release before building off this PR for other `encryption_configuration` changes, I'd merge this into the `bigquery` branch instead.

tswast:
Okay, will do.

quartzmo:
@blowmage @tswast This LGTM (and thanks @tswast!!), but if the purpose is be able to support additional configuration (such as `encryption_configuration`) without adding more parameters to the `load_job` method, won't this mean that the configuration options for `load` will fall behind those of `load_job`?

tswast:
@quartzmo You're right. I do need to follow-up with a change to `load`. (I wasn't sure how to do it, so solved the other parts first)

PR #1981, opened 2018-02-22
Add table acceptance test for BigQuery partitioning by field

jeremywadsack:
I realized I missed a test in #1978.

I also removed unused variables in the BigQuery `table_reference_test` (because the variable is re-assigned to the local reference object).

I'm getting a failure in acceptance tests on this:
> Google::Cloud::Bigquery::bigquery#test_0011_should get a list of projects
But I get that on master as well, so I don't think it's related to this change.


quartzmo:
I pulled this branch and ran `rake ci:a` locally, all passing.

PR #1983, opened 2018-02-26
Spanner Batch client

quartzmo:
Add Spanner Batch client.

Batch client is useful when one wants to read or query a large amount of data from Cloud Spanner across multiple processes. It allows to partition specified read/query and then read data over each partition independently yet at the same snapshot.


quartzmo:
All authors are ok with their commits being contributed.

blowmage:
The commits attributed to me are mine and I want them part of this PR.

Issue #1994, opened 2018-03-06
Increasing thread count / RSS


irvingwashington:
Hey there, we’re observing a weird issue with the `google-cloud-pubsub` gem. 
On processes that start a subscriber, sleeping threads count increases without a limit.

The script below is enough for me to reproduce the issue, after 20 minutes the thread count is about 40, if I leave it for few hours it goes into hundreds. The highest count I've seen so far was over 3'000.

(subscriber.rb)
```ruby
require 'google/cloud/pubsub'

keyfile = 'path_to_the_keyfile'
project_id = 'correct_project_id'

topic_name = 'test_topic_name'
subscription_name = 'test-subscription'

@pubsub = pubsub = Google::Cloud::Pubsub.new(project_id: project_id, credentials: keyfile)

@topic = pubsub.topic(topic_name) || pubsub.create_topic(topic_name)
puts "Topic: #{@topic.inspect}"

@subscription = @topic.subscription(subscription_name) || t.create_subscription(subscription_name)
puts "Subscription: #{@subscription.inspect}"

@subscriber = @subscription.listen { |recv_msg| recv_msg.ack! } # Not really called

formatter = ->(sev, dt, _, msg) { "#{Thread.object_id} #{dt.utc.iso8601(3)} #{sev}: #{msg} (Threads num: #{Thread.list.size})\n" }

GRPC::DefaultLogger::LOGGER = Logger.new(STDOUT)
GRPC::DefaultLogger::LOGGER.formatter = formatter

@subscriber.start
```
I run it in irb to play with threads and ObjectSpace (`irb -r ./subscriber.rb`).

The following patch to GRPC seems to put stop to the issue, however I have too little context to know if it's the correct solution (the usage of GRPC might be wrong, also, I don't see anything like this in reported issues).

```
diff --git a/src/ruby/lib/grpc/generic/bidi_call.rb b/src/ruby/lib/grpc/generic/bidi_call.rb
index 3bdcc0062e..75d57744e7 100644
--- a/src/ruby/lib/grpc/generic/bidi_call.rb
+++ b/src/ruby/lib/grpc/generic/bidi_call.rb
@@ -208,6 +208,7 @@ module GRPC
       rescue StandardError => e
         GRPC.logger.warn('bidi: read-loop failed')
         GRPC.logger.warn(e)
+        @enq_th.kill
         raise e
       ensure
         set_input_stream_done.call
```

Some stack traces:
```
[...]

PR #1998, opened 2018-03-07
Pub/Sub Subscriber Thread Fix

blowmage:
This PR addresses a slow load of threads when using a Subscriber. A GRPC streaming response does not stay open indefinitely, and will need to be restarted when it fails. As part of the restart we need to ensure that the streaming request thread in GRPC is also closed. See #1994 for further discussion.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/enumerator_queue.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/enumerator_queue.rb
@@ -30,9 +30,11 @@ module Google
             @queue.push obj
           end
 
-          def dump_queue
+          def quit_and_dump_queue
             objs = []
             objs << @queue.pop until @queue.empty?
+            # Signal that the enumerator is ready to end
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/enumerator_queue.rb
```diff
@@ -30,9 +30,11 @@ def push obj
             @queue.push obj
           end
 
-          def dump_queue
+          def quit_and_dump_queue
```
quartzmo: It seems to me that adding a separate method that just pushes the sentinel might be easier to reason about than this addition of a side-effect to `dump_queue`, which was previously a simple operation.

blowmage: I'd rather have these operations happen simultaneously. This is only to be used as part of the recovery for a failed stream, which is why I felt a longer method name was warranted.

PR #2000, opened 2018-03-08
BigQuery: Allow loading data from a list of URLs

jeremywadsack:
In the BigQuery API documentation, a load job can be given a list or source URIs. This PR updates the client to allow you to send an array of `Storage` objects, or URLs (and `String` or `URI`), when loading data into a table or dataset.

The initial commit adds some table-reference acceptance tests that were missing, but make it complete with the table acceptance tests.

I noticed that there are very few load/load_job acceptance tests for the dataset. Would you like me to also fill that in to cover all the various cases we have for table and table-reference?

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -1148,9 +1148,10 @@ module Google
         # Request](https://cloud.google.com/bigquery/loading-data-post-request#multipart).
         #
         # @param [String] table_id The destination table to load the data into.
-        # @param [File, Google::Cloud::Storage::File, String, URI] file A file
-        #   or the URI of a Google Cloud Storage file containing data to load
-        #   into the table.
+        # @param [File, Google::Cloud::Storage::File, String, URI,
+        #   Array<Google::Cloud::Storage::File, String, URI>] files
+        #   A file or the URI of a Google Cloud Storage file, or an Array of
+        #   those, containing data to load into the table.
         # @param [String] format The exported file format. The default value is
         #   `csv`.
         #
@@ -1312,6 +1313,25 @@ module Google
         #     end
         #   end
         #
+        # @example Pass a list of google-cloud-storage files:
+        #   require "google/cloud/bigquery"
+        #   require "google/cloud/storage"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   file = bucket.file "file-name.csv"
+        #   list = [file, "gs://my-bucket/file-name2.csv"]
+        #   load_job = dataset.load_job "my_new_table", list do |schema|
+        #     schema.string "first_name", mode: :required
+        #     schema.record "cities_lived", mode: :repeated do |nested_schema|
+        #       nested_schema.string "place", mode: :required
+        #       nested_schema.integer "number_of_years", mode: :required
+        #     end
+        #   end
+        #
         # @example Upload a file directly:
         #   require "google/cloud/bigquery"
         #
@@ -1339,7 +1359,7 @@ module Google
         #
         # @!group Data
         #
-        def load_job table_id, file, format: nil, create: nil, write: nil,
+        def load_job table_id, files, format: nil, create: nil, write: nil,
                      projection_fields: nil, jagged_rows: nil,
                      quoted_newlines: nil, encoding: nil, delimiter: nil,
                      ignore_unknown: nil, max_bad_records: nil, quote: nil,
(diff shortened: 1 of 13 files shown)


blowmage:
Can you mix files and URLs?

jeremywadsack:
@blowmage Local files? Not as far as I could tell. To upload local files you send a media-upload request to the `/upload` URI. This can be a multipart upload (which I believe is what the library does), but the [documentation](https://cloud.google.com/bigquery/docs/loading-data-local) says that you cannot upload multiple files:

> Wildcards and comma separated lists are not supported when you load files from a local data source. Files must be loaded individually.

Additionally:

> The body of the request is formatted as a multipart/related content type [RFC2387] and contains exactly two parts. The parts are identified by a boundary string, and the final boundary string is followed by two hyphens.

> Each part of the multipart request needs an additional Content-Type header:

> Metadata part: Must come first, and Content-Type must match one of the accepted metadata formats.
> Media part: Must come second, and Content-Type must match one the method's accepted media MIME types.

That said, I haven't tried to see what the API does if you use a media-upload request _and_ set `configuration.load.sourceUris[]`.

---

On the CI failure, I could use some guidance. I don't know how to address a doc test failure, "This code example is not yet mocked."

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1380,9 +1380,9 @@ def load_job table_id, file, format: nil, create: nil, write: nil,
         # Request](https://cloud.google.com/bigquery/loading-data-post-request#multipart).
         #
         # @param [String] table_id The destination table to load the data into.
-        # @param [File, Google::Cloud::Storage::File, String, URI] file A file
-        #   or the URI of a Google Cloud Storage file containing data to load
-        #   into the table.
+        # @param [File, Google::Cloud::Storage::File, String, URI, Array] files
```
blowmage: @quartzmo Can you offer some suggestions on how to clarify that the array must be fore remote file URLs and not a local file?

blowmage:
@jeremywadsack The doctest failure is because you added a new code example labeled "Pass a list of google-cloud-storage files:", but didn't configure the documentation mocks for it in `support/doctest_helper.rb`. You can probably copy an existing configuration for `Google::Cloud::Bigquery::Table#load_job` in that file. Be sure to run `bundle exec rake ci` in the directory to see what other failures you may have before updating the PR.

jeremywadsack:
@blowmage ba48571 adds doctest mocks (thanks for the pointer). That also reminded me I should document the example for the other three methods.

I ran `rake ci` locally and it passed without these changes, so that commit is a bit of a shot-in-the-dark.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: I think I'd go with something like this:

```ruby
        # @param [File, Google::Cloud::Storage::File, String, URI, 
        #   Array<Google::Cloud::Storage::File>, Array<String|URI>] files
        #   A file or files containing data to load into the table.
        #   This may be a file or Google Cloud Storage file object; a string or 
        #   URI referencing a Google Cloud Storage file; an array of Google 
        #   Cloud Storage file objects; or an array of strings or URIs referencing 
        #   Google Cloud Storage files.
```

It's usually better in these cases not to be too economical with words, but to really spec it out.

blowmage: Could you use `Array<Google::Cloud::Storage::File|String|URI>` instead of `Array<Google::Cloud::Storage::File>, Array<String|URI>`?

jeremywadsack: The latter seems more correct, but I don't know anything about how documentation is produced. Let me know which will work (or how I test that) and I'll make the update.

quartzmo: Run `rake ci`, then open `doc/index.html` in a browser and take a look at your additions. The `Google::Cloud::Storage::File` inside `Array<Google::Cloud::Storage::File>` should be a link to the `File` docs.

quartzmo: @blowmage Hmmm, I think I should have suggested: `Array<Google::Cloud::Storage::File>, Array<String, URI>`, which might be the same as Array<(Google::Cloud::Storage::File|(String, URI))>, although I'm not sure how far you can push this notation. Anyway, the point is: Either an array of files objects, OR an array of string AND/OR  URI objects. (Correct?)

blowmage: I don’t see why you couldn’t mix all 3 types that represent a GCS URL.

jeremywadsack: Programmatically, you can mix all three types. Only a local `File` is different. I suppose it's up to you what you want to expose in the API (and be committed to).

I ran `rake ci` and looked at `docs/index.html` (thanks for the help) but yard doesn't link it either way and also didn't have `Google::Cloud::Storage::File` linked with the syntax that was there before.

[Yard docs](https://yardoc.org/types.html#see_results) show an "array of various objects" as this:

```
Array<String, Symbol, #read>
```

So I went with comma-separated.

jeremywadsack:
@blowmage @quartzmo Let me know if there's anything else you want me to update. If this is good, I can squash/rebase.

quartzmo:
I pulled this branch and ran `rake ci:a` (includes acceptance tests), passes for me locally.

jeremywadsack:
Sqashed and rebased off `master`. I left `58c5960` as a separate commit because it was covering specs that were missing, unrelated to adding this support.

quartzmo:
I'll merge when green (just a formality...)

blowmage:
Thank you for the rebase/squash!

quartzmo:
Thank you @jeremywadsack, really appreciate your contributions!

PR #1977, opened 2018-02-17
Replace manual Speech library with GAPIC and helpers

geigerj:
This PR replaces the existing Vision client library with an
autogenerated library. The autogenerated code is supplemented with a
helper that wraps the autogenerated bidi streaming RPC in the stream
class from the old manual library. (Note that this is a divergence from
the Python library, but one that should substantially improve the
usability of bidi streaming relative to Python.)

As such, it substantially changes the public surface.

(See also #1828, #1959 for analogous changes to Language, Vision.)

cc @jbolinger

PR #2004, opened 2018-03-12
Regenerate v1 to pickup timeout config changes

jbolinger:
We are regenerating this due to some configuration changes.

I left the doc changes from the generated code in but I'm not sure if that makes sense since nothing else has changed.

### DIFF
--- a/google-cloud-speech/lib/google/cloud/speech/v1.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/v1.rb
@@ -1,4 +1,4 @@
-# Copyright 2017 Google LLC
+# Copyright 2018 Google LLC
 #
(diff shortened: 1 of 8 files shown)


blowmage:
Is this meant to replace #1977?

jbolinger:
Oh, good catch @blowmage - it looks like the config values had already been updated when @geigerj created that PR. 

I got pinged by the speech team today and thought that they had made a new change. My bad, I'll close this one.

blowmage:
If you want we can merge this and release a point release of the current library with the updated config.

jbolinger:
I'll ask, but I'm hoping this will motivate them to just push the other one through.

PR #2005, opened 2018-03-13
DLP V2 Release

jbolinger:
Moving from privately staged version.

PR #2009, opened 2018-03-16
Version and documentation updates for v2 release

jbolinger:
Follow up for #2005. I went over it once more and found a few things that I missed the first time around needed for a release.

A few questions for you all:
- Recommendations on the version number for the gem? The API has major changes but it was a beta and is now GA - @geigerj do you have any thoughts on this?
- @blowmage is this the doc update you mentioned last time? Is there a way to test the links or did you mean that I should check after pushing the tag? I believe Jacob mentioned the doc pages are updated by CI when the tag is pushed - is that right?
- Anything else I missed?

### DIFF
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -188,7 +188,6 @@
       "name": "google-cloud-dlp",
       "defaultService": "google/cloud/dlp",
       "versions": [
-        "v0.2.0",
         "v0.1.0",
         "master"
       ]
--- a/docs/manifest.json
+++ b/docs/manifest.json
@@ -188,6 +188,7 @@
       "name": "google-cloud-dlp",
       "defaultService": "google/cloud/dlp",
       "versions": [
+        "v0.2.0",
         "v0.1.0",
         "master"
       ]
--- a/google-cloud/docs/toc.json
+++ b/google-cloud/docs/toc.json
@@ -198,7 +198,7 @@
             },
             {
               "title": "Data Types",
-              "type": "google/dlp/v2/datatypes"
+              "type": "google/privacy/dlp/v2/datatypes"
             }
           ]
         },
@@ -213,7 +213,7 @@
             },
             {
               "title": "Data Types",
-              "type": "google/dlp/v2beta1/datatypes"
+              "type": "google/privacy/dlp/v2beta1/datatypes"
             }
           ]
         }
--- a/google-cloud/docs/toc.json
+++ b/google-cloud/docs/toc.json
@@ -183,6 +183,43 @@
       ]
     },
     {
+      "title": "DLP",
+      "type": "google/cloud/dlp",
+      "patterns": ["\/dlp", "google\/api", "google\/protobuf"],
+      "nav": [
+        {
+          "title": "V2",
+          "type": "google/cloud/dlp/v2",
+          "patterns": ["\/dlp\/v2", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "DLPServiceClient",
+              "type": "google/cloud/dlp/v2/dlpserviceclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/dlp/v2/datatypes"
+            }
+          ]
+        },
+        {
+          "title": "V2beta1",
+          "type": "google/cloud/dlp/v2beta1",
+          "patterns": ["\/dlp\/v2beta1", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "DLPServiceClient",
+              "type": "google/cloud/dlp/v2beta1/dlpserviceclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/dlp/v2beta1/datatypes"
+            }
+          ]
+        }
+      ]
+    },
+    {
       "title": "DNS",
       "type": "google/cloud/dns",
       "nav": [
--- a/Rakefile
+++ b/Rakefile
@@ -293,6 +293,7 @@ namespace :jsondoc do
       "google-cloud-core",
       "google-cloud-datastore",
       "google-cloud-debugger",
+      "google-cloud-dlp",
       "google-cloud-dns",
       "google-cloud-error_reporting",
       "google-cloud-firestore",
--- a/google-cloud-dlp/README.md
+++ b/google-cloud-dlp/README.md
@@ -44,7 +44,7 @@ inspect_config = {
   min_likelihood: :POSSIBLE
 }
 item = { value: "my phone number is 215-512-1212" }
-parent = "projects/#{ENV["DLP_TEST_PROJECT"]}"
+parent = "projects/#{ENV["MY_PROJECT"]}"
 
 response = dlp.inspect_content parent, 
   inspect_config: inspect_config, 
--- a/google-cloud-dlp/README.md
+++ b/google-cloud-dlp/README.md
@@ -37,14 +37,18 @@ See https://www.ruby-lang.org/en/downloads/branches/ for further details.
 ```rb
 require "google/cloud/dlp"
 
-dlp_service_client = Google::Cloud::Dlp.new
-min_likelihood = :POSSIBLE
-inspect_config = { min_likelihood: min_likelihood }
-type = "text/plain"
-value = "my phone number is 215-512-1212"
-items_element = { type: type, value: value }
-items = [items_element]
-response = dlp_service_client.inspect_content(inspect_config, items)
+dlp = Google::Cloud::Dlp.new
+
+inspect_config = { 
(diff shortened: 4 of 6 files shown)


docs/manifest.json
```diff
@@ -188,6 +188,7 @@
       "name": "google-cloud-dlp",
       "defaultService": "google/cloud/dlp",
       "versions": [
+        "v0.2.0",
```
blowmage: We typically don't change version numbers in pull requests. We make this change during the tagged release commit, at the same time as we update the CHANGELOG and add the new version to the main manifest. This is all covered in the RELEASING guide.

geigerj:
@jbolinger re: versioning -- for pre-1.0 releases, we increment the minor version to indicate the equivalent of an ordinary "major version" release, so going `0.1.0` to `0.2.0` is the right thing to do here.

blowmage:
Once changes have been made to master, and the CI build has completed, you can view the documentation on the docs site. Just view the gem and change the version in the left nav dropdown to `master` to see it.

https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-dlp/master/google/cloud/dlp

blowmage:
I thought going GA was the same thing as going 1.0. What is the mechanism for this gem to also be considered GA?

geigerj:
@blowmage I'm speaking only from past experience here, but in the past, we've let gems have a "bake" period in beta, rather than going straight from alpha to GA. We also have not required that a server-side GA declaration necessitate a client library GA.

jbolinger:
@blowmage thanks for the clarification. 

One more thing about the docs: https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/cloud doesn't have DLP listed on the left hand side menu. Is there something I need to do for that?

If not, I'll close this and do everything via the tag.

blowmage:
FYI, it seems that the Data Types links in the side nav of the documentation are not working for both V2 and V2beta1. /cc @quartzmo

blowmage:
It looks like this gem was never added to the document site when it was added as a dependency to the google-cloud gem. I think we want to get this fixed before we cut a release. I’m out for the next couple hours, so maybe @quartzmo can weigh in.

quartzmo:
Sorry, just saw this. Yes, we need to add DLP to the left nav. Need any help with that?

quartzmo:
Steps to add the DLP package to gh-pages site docs:

1. ALREADY DONE: Add an entry for the package in [docs/manifest.json](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/docs/manifest.json#L186-L194).
1. Add the name of the package to [the whitelist](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/Rakefile#L296) in the top-level Rakefile.
1. Copy the JSON fragment from the package's [toc.json](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-dlp/docs/toc.json) to the correct (alphabetical) location in the google-cloud [toc.json](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud/docs/toc.json#L184).

quartzmo:
I figured out the issue with the broken "Data Types" links.

The links are configured in [toc.json](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-dlp/docs/toc.json). It looks like this:

```js
              "title": "Data Types",
              "type": "google/dlp/v2/datatypes"
```

The configuration for building the "Data Types" content is in the [Rakefile](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-dlp/Rakefile#L145) and the part that determines the path of the output looks like this:

```ruby
title: "Google::Privacy::Dlp::V2::DataTypes",
```

The output for this `title` is:

```sh
rake jsondoc
...
jsondoc/google/privacy/dlp/v2/datatypes.json
```

It doesn't match the path (`type` in the world of the Angular app) provided in the TOC.

The situation is the same for `V2beta1`.

jbolinger:
thanks @quartzmo - I've updated the toc and path references - anything else we need to do?

google-cloud/docs/toc.json
```diff
@@ -182,6 +182,43 @@
...
+          "type": "google/cloud/dlp/v2",
+          "patterns": ["\/dlp\/v2", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "DLPServiceClient",
+              "type": "google/cloud/dlp/v2/dlpserviceclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/dlp/v2/datatypes"
```
quartzmo: I think maybe you copied this over before updating the `type` values? Should be `google/privacy/dlp/v2/datatypes`.

google-cloud/docs/toc.json
```diff
@@ -182,6 +182,43 @@
...
+          "type": "google/cloud/dlp/v2beta1",
+          "patterns": ["\/dlp\/v2beta1", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "DLPServiceClient",
+              "type": "google/cloud/dlp/v2beta1/dlpserviceclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/dlp/v2beta1/datatypes"
```
quartzmo: I think maybe you copied this over before updating the `type` values? Should be `google/privacy/dlp/v2beta1/datatypes`.

quartzmo: nm, fixed in subsequent commit.

google-cloud/docs/toc.json
quartzmo: nm, fixed in subsequent commit.

google-cloud/docs/toc.json
```diff
@@ -198,7 +198,7 @@
             },
             {
               "title": "Data Types",
-              "type": "google/dlp/v2/datatypes"
+              "type": "google/privacy/dlp/v2/datatypes"
```
blowmage: Why is `google/privacy/dlp` being used as a path? It doesn't exist AFAICT. The only path I see in this library is `google/cloud/dlp`. Shouldn't we replace `google/privacy/dlp` with `google/cloud/dlp`, and `Google::Privacy::Dlp` with `Google::Cloud::Dlp`?

jbolinger: If I understood @quartzmo's comment correctly this is to reference the following file `jsondoc/google/privacy/dlp/v2/datatypes.json` in the generated docs.

The generator does produce things in both modules `Google::Privacy` and `Google::Cloud` - is that what you mean?

quartzmo: My [comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2009#issuecomment-373859288) was that the links need to match the output. I don't personally know which is correct.

quartzmo: This value in the Rakefile is what controls the output:

```ruby
title: "Google::Privacy::Dlp::V2::DataTypes",
```
But I don't know if `Google::Privacy::Dlp::V2::DataTypes` is really the correct title. I don't know where it came from. Was it generated?

quartzmo: The history of "Data Types" is that someone wanted a "page" with a table listing all the "minor" types such as protobufs. Then that person or someone else requested that minor types from our manual API should appear there as well. The Angular app was never designed to generate a page like this, so I added the generation to our jsondoc tool, controlled by the config in the Rakefile. It's pretty simple, and flexible.

jbolinger: I believe `Google::Privacy::Dlp::V2::DataTypes` is generated. All the generated protobuf classes are under `Google::Privacy` instead of `Google::Cloud` because of how it was setup. So AFAIK this is consistent.

jbolinger:
Is this alright to go out? We're just about out of time if we're going to hit the end of week.

google-cloud/docs/toc.json
quartzmo: OK, SGTM.

quartzmo:
I checked the docs for [`google-cloud-dlp`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-dlp/v0.2.0/google/cloud/dlp) and [`google-cloud` `master`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/master/google/cloud/dlp), and everything looks good to me. However, the current version of [`google-cloud` `v0.51.0`](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.51.0/google/cloud) from March 15 does not have "DLP" in the left nav.

jbolinger:
Thanks for checking @quartzmo - I looked last night and noticed that as well. I can bump `google-cloud` when I get in on Monday. Feeling better now that I've been through the process.

PR #2014, opened 2018-03-21
BigQuery: Add ExtractJob::Updater

tswast:
Also move some shared conversion code to the Convert class.

Another step towards https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1960

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/extract_job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/extract_job.rb
@@ -155,6 +155,122 @@ module Google
         def destinations_counts
           Hash[destinations.zip destinations_file_counts]
         end
+
+        ##
+        # Yielded to a block to accumulate changes for an API request.
+        class Updater < ExtractJob
+          ##
+          # @private Create an Updater object.
+          def initialize gapi
+            @gapi = gapi
+          end
+
+          ##
+          # @private Create an Updater from an options hash.
+          #
+          # @return [Google::Cloud::Bigquery::ExtractJob::Updater] A job
+          #   configuration object for setting query options.
+          def self.from_options table, storage_files, options = {}
+            storage_urls = Array(storage_files).map do |url|
+              url.respond_to?(:to_gs_url) ? url.to_gs_url : url
+            end
+            dest_format = options[:format]
+            if dest_format.nil?
+              dest_format = Convert.derive_source_format storage_urls.first
+            end
+            req = Google::Apis::BigqueryV2::Job.new(
+              configuration: Google::Apis::BigqueryV2::JobConfiguration.new(
+                extract: Google::Apis::BigqueryV2::JobConfigurationExtract.new(
+                  destination_uris: Array(storage_urls),
+                  source_table: table
+                ),
+                dry_run: options[:dryrun]
+              )
+            )
+
+            updater = ExtractJob::Updater.new req
+            updater.compression = options[:compression]
+            updater.delimiter = options[:delimiter]
+            updater.format = dest_format
+            updater.header = options[:header]
+            updater.labels = options[:labels] if options[:labels]
+            updater
+          end
+
+          # Sets the compression type.
+          #
+          # @param [String] value The compression type to use for exported
+          #   files. Possible values include `GZIP` and `NONE`. The default
+          #   value is `NONE`.
+          #
+          # @!group Attributes
+          def compression= value
+            @gapi.configuration.extract.compression = value
+          end
+
+          # Sets the field delimiter.
+          #
+          # @param [String] value Delimiter to use between fields in the
+          #   exported data. Default is <code>,</code>.
+          #
+          # @!group Attributes
+          def delimiter= value
+            @gapi.configuration.extract.field_delimiter = value
+          end
+
+          ##
+          # Sets the destination file format. The default value is `csv`.
+          #
+          # The following values are supported:
+          #
+          # * `csv` - CSV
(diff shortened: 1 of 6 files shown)


tswast:
CI acceptance tests pass:

```
$ rake acceptance[...]
WARNING: You are running Ruby 2.2.8, which is nearing end-of-life.
The Google Cloud API clients work best on supported versions of Ruby. Consider upgrading to Ruby 2.3 or later.
See https://www.ruby-lang.org/en/downloads/branches/ for more info on the Ruby maintenance schedule.
To suppress this message, set the GOOGLE_CLOUD_SUPPRESS_RUBY_WARNINGS environment variable.
Run options: --seed 45328

# Running:

..............................................S..................................................................................

Finished in 638.538364s, 0.2020 runs/s, 2.5652 assertions/s.

129 runs, 1638 assertions, 0 failures, 0 errors, 1 skips

You have skipped tests. Run with --verbose for details.
Cleaning up bigquery datasets after tests.

```

tswast:
This commit is now at https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2016

google-cloud-bigquery/lib/google/cloud/bigquery/extract_job.rb
```diff
@@ -155,6 +155,122 @@ def destinations_file_counts
...
+            updater = ExtractJob::Updater.new req
+            updater.compression = options[:compression]
+            updater.delimiter = options[:delimiter]
+            updater.format = dest_format
+            updater.header = options[:header]
+            updater.labels = options[:labels] if options[:labels]
+            updater
+          end
+
+          # Sets the compression type.
```
quartzmo: A few of these docs comments are missing the first header line (`##`).

tswast: Added.

tswast:
Okay to merge?

I'll probably have a conflict in `service.rb` in my other two PRs, so I can fix those up after this one or https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2015 goes in.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -1342,8 +1344,14 @@ def extract_job extract_url, format: nil, compression: nil,
...
-                      job_id: job_id, prefix: prefix, labels: labels }
-          gapi = service.extract_table table_ref, extract_url, options
+                      labels: labels }
+          updater = ExtractJob::Updater.from_options table_ref, extract_url,
+                                                     options
+
+          yield updater if block_given?
+
+          job_gapi = updater.to_gapi
+          gapi = service.extract_table job_id, prefix, job_gapi
```
blowmage: Instead of passing `job_id` and `prefix` to the service call, shouldn't they already be used and present on the updater object? The updater object seems like a fine place to set a custom `job_id` value.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -1387,13 +1397,32 @@ def extract_job extract_url, format: nil, compression: nil,
...
+        #
+        #   table.extract "gs://my-bucket/file-name.csv" do |extract|
+        #     extract.labels = { "custom-label" => "custom-value" }
+        #   end
+        #
         # @!group Data
         #
         def extract extract_url, format: nil, compression: nil, delimiter: nil,
                     header: nil
-          job = extract_job extract_url, format: format,
```
blowmage: Instead of removing the call to `extract_job`, how about passing the `extract` block to it instead?

```ruby
def extract extract_url, format: nil, compression: nil, delimiter: nil,
                      header: nil, &block
  job = extract_job extract_url, format: format,
                                 compression: compression,
                                 delimiter: delimiter, header: header,
                                 &block

blowmage:
@tswast This question is not directly related to this PR. I'm working on adding Job regionalization support, but it will conflict greatly with these changes. Would you like me to do this work off of your changes in the `bigquery` branch? Which means it would have to wait to be released when these changes are ready. Or would you like me to do this work in `master`, which would mean a pretty large effort to resolve the conflicts?

tswast:
@blowmage I'm hoping these were the last PRs I needed to do, but you're right that it's going to cause a bunch of merge conflicts. I'm fine having these changes go out concurrently with the Job regionalization.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
tswast: Yeah, thinking ahead to the `location` field, this makes sense. I'm missing `job_id` in QueryJob::Updater and LoadJob::Updater, too, though. Maybe let's follow-up with another PR to move this over?

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
tswast: Done in https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2014/commits/45a556d39cbd6dae1e7c50fad9c9b004e443032a

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
blowmage: If I follow up with the location changes I can make these changes then as well.

blowmage:
Okay, I’ll plan on following up this and the other PRs with one that adds location and does some other refactoring. 

I’ll approve as soon as this PR goes green.

quartzmo:
@blowmage It's green.

PR #2020, opened 2018-03-23
Updated generated unit tests for google-gax 1.1.x changes

jbolinger:
Updated the generated unit tests to be compatible with the gRPC mocks for an upcoming gax 1.1.1 release (currently on master).

When doing this, I also noticed that a couple of the gapic modules were pinned to version numbers like 1.0.1 instead of 1.0. I can't think of any reason why we'd want to do that so I updated them as well in this change, but if there's a reason let me know.

### DIFF
--- a/google-cloud-container/google-cloud-container.gemspec
+++ b/google-cloud-container/google-cloud-container.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.1"
+  gem.add_dependency "google-gax", "~> 1.0"
 
   gem.add_development_dependency "minitest", "~> 5.10"
   gem.add_development_dependency "rubocop", "~> 0.50.0"
--- a/google-cloud-dataproc/google-cloud-dataproc.gemspec
+++ b/google-cloud-dataproc/google-cloud-dataproc.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.0"
+  gem.add_dependency "google-gax", "~> 1.0"
 
   gem.add_development_dependency "minitest", "~> 5.10"
   gem.add_development_dependency "rubocop", "~> 0.50.0"
--- a/google-cloud-os_login/google-cloud-os_login.gemspec
+++ b/google-cloud-os_login/google-cloud-os_login.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.0"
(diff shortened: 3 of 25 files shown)


google-cloud-container/google-cloud-container.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.1"
+  gem.add_dependency "google-gax", "~> 1.0"
```
blowmage: Thank you for this change. I don't see why we would want to pin the dependency to >= 1.0 and < 1.1, especially since google-gax is following semver.

google-cloud-dataproc/google-cloud-dataproc.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.0"
+  gem.add_dependency "google-gax", "~> 1.0"
```
blowmage: :+1:

google-cloud-os_login/google-cloud-os_login.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = ">= 2.0.0"
 
-  gem.add_dependency "google-gax", "~> 1.0.0"
+  gem.add_dependency "google-gax", "~> 1.0"
```
blowmage: :+1:

blowmage:
It looks like the CI build for this PR won't pass until google-gax 1.1.0 is released. Is that right?

jbolinger:
That's correct, and I wanted to coordinate with you on the release. I think we are all set to release the updated gax whenever we're ready. We still have a couple of weeks before it has to happen though so let me know when it's convenient.  (not really sure why I re-ran the CI - it was early)

jbolinger:
Hey @blowmage - are you still up for getting this done when you're back? Hoping to get the new gax lib out by the end of the week.

blowmage:
@jbolinger Yep, anytime.

jbolinger:
@blowmage sounds good. Do you mind driving the process? 

I suppose we'll want to push out all the cloud libraries first and then hit gax at the end? I don't typically follow the release plans, but since we've got all week we can time it around whatever else is going on.

blowmage:
We shouldn't have to release anything for a new google-gax version, since it is backwards compatible according to SEMVER. All we need to do it update the mocks in the generated tests to match the new internals. Just let us know when you cut the new release.

jbolinger:
Ok, great. I must have misunderstood your comments on that thread a few weeks ago and though you had wanted to do that for some reason - my bad - I could have done this a long time ago :)

Should land soon and, hopefully, no surprises...

jbolinger:
@blowmage - it's out - I've restarted the tests on this one.

blowmage:
I'm actively working on this, BTW.

PR #2016, opened 2018-03-21
BigQuery: Add encryption properties to jobs classes.

tswast:
Also:
* Add a helper to create an EncryptionConfiguration in project.
* Rename encryption_configuration property to more concise encryption.

Based on https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2014 and https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2015

Almost finishes https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1960, just need to merge to `bigquery` to `master` after this goes in.

PR #2015, opened 2018-03-21
BigQuery: Add CopyJob::Updater class.

tswast:
Another step towards #1960

PR #2003, opened 2018-03-12
BigQuery: Add QueryJob::Updater class

tswast:
As with LoadJob::Updater, the new class QueryJob::Updater will make it easier to add new properties to query jobs without having to get way too many keyword arguments to the `query` and `query_job` functions.

Note: I do not make an attribute for `maximum_billing_tier` as this property is deprecated. According to the BigQuery team, this property is no longer used, as "high-compute" queries are no longer supported.

PR #2022, opened 2018-03-26
BigQuery Job Updaters and EncryptionConfiguration

quartzmo:
This PR merges into master the `bigquery` topic branch that collected the changes from #1973, #2003, #2014, #2015 and #2016.


quartzmo:
@tswast @keylimetoolbox There were a number of conflicts between #1973 and #2000 that I needed to manually resolve while rebasing this branch on master. All tests are passing (including acceptance tests), but PTAL just to be sure the Load Job code still looks good.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -1546,18 +1613,28 @@ def load_job files, format: nil, create: nil, write: nil,
...
+                                     max_bad_records: max_bad_records,
+                                     quote: quote, skip_leading: skip_leading,
+                                     dryrun: dryrun, schema: schema,
+                                     labels: labels, autodetect: autodetect,
+                                     null_marker: null_marker
+
+          yield updater if block_given?
+
+          job_gapi = updater.to_gapi
+          if local_file? files
```
tswast: I assume `local_file?` is okay getting a list now?

blowmage: No, only one local file can be provided. An array of URLs can be provided, either as strings or URI objects or storage file objects.

tswast: Does it return false when given array? That sounds like it should be the correct behavior.

quartzmo: @tswast I struggled to understand this as well while doing the rebase.

I don't think the docs make it clear that only one local file can be provided:

> A file or the URI of a Google Cloud Storage file, or an Array of  those, containing data to load into the table.

However, I'm fairly confident that in this case I preserved the original behavior of #2000 in my manual conflict resolution.

blowmage: Yes, `local_file?` will return `false` if passed an array.

jeremywadsack: Maybe a doc update if this is confusing?

Also , why not use `load_local_or_uri` here?

jeremywadsack:
Looks like this preserves the changes in 2000.

PR #2028, opened 2018-03-29
Update versioning info on documentation

premist:
I compared versioning on [README.md](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/README.md) with docs hosted on [googlecloudplatform.github.io](http://googlecloudplatform.github.io/google-cloud-ruby/#/docs/) and found some inconsistencies. I considered version information on README.md as the latest one and updated documentation on APIs where versioning were outdated.

While I was updating the version information, I also discovered that some GA APIs use "Product Name (GA)" as the title and other GA APIs use "Product Name" without (GA) suffix. I removed (GA) suffix everywhere on docs (except README.md) to be consistent, but feel free to re-add it if you want to.

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/v1/doc/overview.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/v1/doc/overview.rb
@@ -17,7 +17,7 @@ module Google
     # rubocop:disable LineLength
 
     ##
-    # # Ruby Client for Google Cloud Datastore API ([Alpha](https://github.com/GoogleCloudPlatform/google-cloud-ruby#versioning))
+    # # Ruby Client for Google Cloud Datastore API ([GA](https://github.com/GoogleCloudPlatform/google-cloud-ruby#versioning))
     #
     # [Google Cloud Datastore API][Product Documentation]:
     # Accesses the schemaless NoSQL database to provide fully managed, robust,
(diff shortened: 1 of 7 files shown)


blowmage:
@jbolinger Thoughts on this? All these files are generated, and we have been told to not modify them without approval.

jbolinger:
Updates to the alpha/beta/ga labels are not totally automated so I wouldn't be surprised if there is some inconsistency. 

I think it's okay to modify if we can first verify what the correct value is. However, the GA part should not be dropped - @landrito do chime in if that's not always the case, but I'm not aware of any exceptions.

premist:
@jbolinger 

[Google Cloud Storage section](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.51.1/google/cloud/storage) doesn't have a GA label. It doesn't have an API suffix too, though. 

[Google Cloud Translation section](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.51.1/google/cloud/translate) has API suffix and doesn't have a GA label. Seems like it is being generated from README.md, unlike other documentations where documentation is located inside overview.rb.

premist:
Added GA label where it was missing, and rebased.

blowmage:
@jbolinger I've assigned this to you. I will merge when you approve.

premist:
Thanks for the review!

premist:
Thanks for the review! I checked out [one of the docs](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-pubsub/v0.30.1/google/cloud/pubsub) hosted on googlecloudplatform.github.io - which is not updated. Seems like there is a separate deployment process for the docs?

blowmage:
The master docs ave been updated. These changes will have to be released to show up in the release documentation.

Issue #2011, opened 2018-03-19
pubsub: consumer crashes / stops receiving messages after couple minutes


jadekler:
Under load (~30 million messages in the subscription), the app entirely stops receiving new messages.

Here's the app:

```
ruby app.rb

require 'google/cloud/pubsub'

puts 'Ruby starting up'

PROJECT_ID = 'some-project'
TOPIC_ID = 'some-topic'
SUBSCRIPTION_ID = 'some-ruby-subscription'

pubsub = Google::Cloud::Pubsub.new(
    project_id: PROJECT_ID,
    credentials: ENV['GOOGLE_APPLICATION_CREDENTIALS']
)

topic = pubsub.topic TOPIC_ID
subscription = topic.subscription SUBSCRIPTION_ID

puts 'Ruby sub exists/creates'

if subscription == nil || !subscription.exists?
  subscription = topic.subscribe SUBSCRIPTION_ID
end

subscriber = subscription.listen do |received_message|
  puts "Ruby received #{received_message.message_id} #{received_message.data}"
  received_message.acknowledge!
end

puts 'Ruby receiving..'

subscriber.start

sleep

subscriber.stop.wait!
```

Here's a graph show this:

![screen shot 2018-03-17 at 5 40 00 pm](https://user-images.githubusercontent.com/3584893/37561367-3faa517e-2a0a-11e8-8e0e-cdb9b29db474.png)

At this point, the app no longer emits any logs. And, although kubernetes tells me the app is still alive, its CPU is at 0.

![screen shot 2018-03-17 at 5 42 04 pm](https://user-images.githubusercontent.com/3584893/37561376-88368caa-2a0a-11e8-90d0-73d0eef1a392.png)

blowmage:
Does this happen as soon as the Ruby process is started? Or does the Ruby process run for a while, and then the spike happens?

blowmage:
@jadekler Can you answer these questions? Also, about how long does it run, and about how many messages does it acknowledge, before it stops processing?

jadekler:
Hey Mike, sorry for the long delay. I'm out of the country atm (and south africa has very inconsistent wifi availability) so I'll be pretty slow next couple weeks in replying. The failure happened some 10 minutes after the process started. About 300,000 messages were acked at a rate of about 37,000 per minute.

I was able to repro this just now by just restarting the pod. This is all being run on a very minimal / no-special-config docker image that I stuck on GKE.

[...]

Issue #2010, opened 2018-03-17
pubsub: errors warning every 10 minutes


jadekler:
Every 10 minutes, on the dot, I get errors about bidi connections. These do not seem to be preventing the app from pulling messages, or crashing the app, but they are quite concerning. Any idea what's going on here?

![screen shot 2018-03-17 at 4 37 49 pm](https://user-images.githubusercontent.com/3584893/37561047-c203abd8-2a01-11e8-917a-646a43d4a7ec.png)

![screen shot 2018-03-17 at 4 41 46 pm](https://user-images.githubusercontent.com/3584893/37561059-222e63a4-2a02-11e8-8f0e-6a1488e0b602.png)
![screen shot 2018-03-17 at 4 41 51 pm](https://user-images.githubusercontent.com/3584893/37561060-2246ca48-2a02-11e8-830c-ad59f008f609.png)
![screen shot 2018-03-17 at 4 41 54 pm](https://user-images.githubusercontent.com/3584893/37561061-225afcac-2a02-11e8-85fa-c5f1597b8102.png)
![screen shot 2018-03-17 at 4 41 58 pm](https://user-images.githubusercontent.com/3584893/37561062-227013a8-2a02-11e8-9705-05ea20cd8c79.png)

jadekler:
Under load (~30 million messages in the subscription), the app entirely stops receiving new messages.

Here's the app:

```
ruby app.rb

require 'google/cloud/pubsub'

puts 'Ruby starting up'

PROJECT_ID = 'some-project'
TOPIC_ID = 'some-topic'
SUBSCRIPTION_ID = 'some-ruby-subscription'

pubsub = Google::Cloud::Pubsub.new(
    project_id: PROJECT_ID,
    credentials: ENV['GOOGLE_APPLICATION_CREDENTIALS']
)

topic = pubsub.topic TOPIC_ID
subscription = topic.subscription SUBSCRIPTION_ID

puts 'Ruby sub exists/creates'

if subscription == nil || !subscription.exists?
  subscription = topic.subscribe SUBSCRIPTION_ID
end

subscriber = subscription.listen do |received_message|
  puts "Ruby received #{received_message.message_id} #{received_message.data}"
  received_message.acknowledge!
end

puts 'Ruby receiving..'

subscriber.start

sleep

subscriber.stop.wait!
```

Here's a graph show this:

[...]

PR #2030, opened 2018-03-30
Pub/Sub: Recover from GRPC::Core::CallError

blowmage:
As first reported in #2010 and #2011 we are seeing `GRPC::Core::CallError` errors raised regularly on streaming connections under heavy load using the `grpc` gem version 1.7.0 and above. We have discussed this with the GRPC team and opened grpc/grpc#14853 to track this.

The problem adversely affects Google Cloud Pub/Sub Ruby gem because the error is being raised in a thread used by the GRPC library, and is not propagated to the Stream object used to manage the requests and responses to the stream. When the error is raised the stream connection is unexpectedly closed. This PR adds the ability for a Stream object to recover from the stream being closed unexpectedly.

This PR also enhances performance for the batch total byte size calculations.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
@@ -285,16 +285,20 @@ module Google
             @publisher = publisher
             @messages = []
             @callbacks = []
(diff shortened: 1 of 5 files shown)


blowmage:
This change will not stop `GRPC::Core::CallError` from being raised. If users have `Thread.abort_on_exception = true` set their apps will abort when the error is raised.

PR #2032, opened 2018-03-30
Add Bigquery Job geo-regionalization (location)

### LINKED ISSUE #2017: Support BigQuery Job Regionalization
We need to support BigQuery Regionalization by adding `location` to various Job-related methods and API calls:

* `jobs.get`
  * - [ ] Add `location` optional argument to `Project#job`
* `jobs.cancel`
  * - [ ] Send `location` value to API call made in `Job#cancel`
* `jobs.getQueryResults `
  * - [ ] Send `location` value to API call made by `Job#data`
  * - [ ] Send `location` value to API call made by `Job#wait_until_done!`
* `jobs.insert`
  * - [ ] Add `location` optional argument to `Project#query`, `Project#query_job`, `Dataset#query`, and `Dataset#query_job`
  * - [ ] Add `location` optional argument to `Table#copy`, and `Table#copy_job`
  * - [ ] Add `location` optional argument to `Table#extract`, and `Table#extract_job`
  * - [ ] Add `location` optional argument to `Dataset#load `, `Dataset#load_job`, `Table#load`, and `Table#load_job`
  * - [ ] Add `location` to API call made by `Job#rerun!` (possible optional argument as well?)

quartzmo:
Note that this PR does NOT allow the user to "optionally specify a location for job operations"--a client-level property that, once set, is used for dataset creation and the `jobs.insert` operations. Is this required?

[closes #2017]

/cc @jjolma @tmatsuo


google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -904,15 +904,15 @@ def query_job query, params: nil, external: nil,
...
                       maximum_billing_tier: maximum_billing_tier,
                       maximum_bytes_billed: maximum_bytes_billed,
-                      params: params, external: external, labels: labels,
-                      udfs: udfs }
+                      job_id: job_id, prefix: prefix, params: params,
+                      external: external, labels: labels, udfs: udfs }
 
-          updater = QueryJob::Updater.from_options query, options
+          updater = QueryJob::Updater.from_options service, query, options
           updater.dataset = self
```
blowmage: Should the updater's `location` be set to `self.location` before the object is yielded?

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1079,11 +1079,11 @@ def query query, params: nil, external: nil, max: nil, cache: true,
                       legacy_sql: legacy_sql, standard_sql: standard_sql,
                       params: params }
           options[:dataset] ||= self
-          updater = QueryJob::Updater.from_options query, options
+          updater = QueryJob::Updater.from_options service, query, options
```
blowmage: Shouldn't `query` be calling `query_job` instead of creating its own updater?

blowmage:
I think that all jobs that are created on a Dataset or a Table should default the Job's location to the Dataset or Table location. What do you think?

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: Yes, I'll add a commit to do this for all job methods in Dataset and Table.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: I agree that would be better, but may not make this change in this PR.

blowmage:
The requirements document states:

> Allow the user to optionally specify a location for job operations.  Consider having one flag that can be used for dataset creation (the existing location field) and the job operations.

I believe that by adding the optional `location` argument to `Project#job` and allowing the Job Updater objects to set `location` fulfills the first part of requirement. The second part looks to suggest the possibility of having a "client-wide default for location". This looks to have been done in the Go client, which exposes a very different API than the Ruby client. I don't believe a "client-wide default for location" is necessary for the Ruby client, but even if it is something we are asked to implement we would still need all the changes currently in this PR to override an optional default. I think it would be best to address the question of an optional "client-wide default for location" in a separate PR after this one is accepted and merged.

I think we will get more utility from defaulting a job to the `location` to the Dataset or Table that created the job. If a `location` value is not present, e.g. creating a job from a Project or a Table reference, then I can see using a "client-wide default for location". But my preference would be to use the Dataset/Table location before the default location.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: Updated.

google-cloud-bigquery/acceptance/bigquery/location_test.rb
```diff
@@ -58,9 +58,7 @@
     insert_response.error_rows.must_be :empty?
 
     job_id = "test_job_#{SecureRandom.urlsafe_base64(21)}" # client-generated
-    query_job = dataset.query_job query, job_id: job_id do |j|
-      j.location = region
```
blowmage: Why was `location` removed from the tests in this file?

quartzmo: Because now that the location acquired from the dataset or table, this configuration is no longer needed.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1080,6 +1093,7 @@ def query query, params: nil, external: nil, max: nil, cache: true,
                       params: params }
           options[:dataset] ||= self
           updater = QueryJob::Updater.from_options service, query, options
+          updater.location = location if location # may be dataset reference
```
blowmage: I know why you are using a conditional, but users may decide that they want to set the value to `nil`, so we should probably have `QueryJob::Updater#location=` do the right thing when setting the value to `nil`.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -904,15 +910,16 @@ def query_job query, params: nil, external: nil,
...
                       maximum_bytes_billed: maximum_bytes_billed,
-                      params: params, external: external, labels: labels,
-                      udfs: udfs }
+                      job_id: job_id, prefix: prefix, params: params,
+                      external: external, labels: labels, udfs: udfs }
 
-          updater = QueryJob::Updater.from_options query, options
+          updater = QueryJob::Updater.from_options service, query, options
           updater.dataset = self
+          updater.location = location if location # may be dataset reference
```
blowmage: I know why you are using a conditional, but users may decide that they want to set the value to `nil`, so we should probably have `QueryJob::Updater#location=` do the right thing when setting the value to `nil`.

google-cloud-bigquery/acceptance/bigquery/location_test.rb
blowmage: Makes sense.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: What do you mean by "do the right thing"? Prevent a `nil`?

google-cloud-bigquery/acceptance/bigquery/location_test.rb
```diff
@@ -0,0 +1,212 @@
...
+
+require "bigquery_helper"
+
+describe Google::Cloud::Bigquery, :location, :bigquery do
+  let(:region) { "asia-northeast1" }  # Tokyo
+  let(:dataset_id) { "#{prefix}_dataset_location" }
+  let(:dataset) do
+    d = bigquery.dataset dataset_id
+    if d.nil?
+      # Create a temporary dataset (& table) in the EU.
```
jjolma: Stale comment

google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
```diff
@@ -176,6 +178,33 @@ def self.from_options source, target, options = {}
             updater
           end
 
+          ##
+          # Sets the geographic location where the job should run. Required
+          # except for US and EU.
+          #
+          # @param [String] value A geographic location code, such as "US", "EU"
```
jjolma: nit: I would remove "code" here. We typically refer to it as "geographic location".

google-cloud-bigquery/lib/google/cloud/bigquery/job.rb
```diff
@@ -90,6 +90,17 @@ def project_id
           @gapi.job_reference.project_id
         end
 
+        ##
+        # The geographic location where the job resides.
```
jjolma: nit: "where the job runs"

quartzmo: Thanks, fixed.

google-cloud-bigquery/lib/google/cloud/bigquery/copy_job.rb
quartzmo: Thanks, fixed everywhere.

google-cloud-bigquery/acceptance/bigquery/location_test.rb
quartzmo: Thanks, fixed.

quartzmo:
@jjolma Yes, the integration test making real API calls is here: [google-cloud-bigquery/acceptance/bigquery/location_test.rb](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2032/files/5aa65675e34f0a69c14a510d90ae57b07467dab5#diff-8df779055801982efa3445566fbe8d0e)

quartzmo:
@blowmage 

> I have a concern about users setting a location value to nil, and having the API call return the following error: `Google::Cloud::InvalidArgumentError: invalid: Invalid value for: null is not a valid value`

The docs for the `Updater#location=` methods do **not** list `nil` as an accepted value. I think it's fine for our library to expose the validation rules of service API here, isn't it?

blowmage:
The documentation also mentions the behavior of the API when `location` is unset. Setting it to `nil` should be treated as the same thing as being unset, IMO. We know this will raise an error because of the way Google API Client is implemented. I still believe we should work around this, as suggested in the`location=` code I offered.

quartzmo:
@jjolma This PR did not entirely address the following item from the spec document (emphasis mine):

>  Consider having **one flag that can be used for dataset creation** (the existing location field) and the job operations.

In google-cloud-bigquery, I assume this would be an option to set a property in `Project` that would then be used as a default for [`Project#create_dataset`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb#L674)? Regarding "the job operations", all jobs except `Project#query_job` are created via methods on `Dataset` and `Table`, with the location from instances of those classes is already used as a default.

quartzmo:
@blowmage Do you see any issues with adding this optional param to `Project` creation?

quartzmo:
I'm going to release BigQuery now without this addition, since it appears to be optional.

PR #2033, opened 2018-03-31
better error message when transaction failed

Nakilon:
It was hiding the real details of the exception within the transaction block.

Now in my case the
```plain
Transaction failed to commit. (Google::Cloud::Datastore::TransactionError)
```
became
```plain
Transaction failed to commit: undefined method `query' for #<Google::Cloud::Datastore::ReadOnlyTransaction:0x007fd3ee304d90> (Google::Cloud::Datastore::TransactionError)
```

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/dataset.rb
@@ -545,14 +545,14 @@ module Google
             # Create new transaction and retry the block
             tx = Transaction.new service, previous_transaction: tx.id
             retry
-          rescue StandardError
+          rescue StandardError => e
             begin
               tx.rollback
-            rescue StandardError
+            rescue StandardError => e
               raise TransactionError,
-                    "Transaction failed to commit and rollback."
+                    "Transaction failed to commit and rollback: #{e}"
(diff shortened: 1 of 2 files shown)


Nakilon:
Probably there is no need in 2nd and 4th ` => e` because they hide user exception too.

blowmage:
I don't know if this change is needed. Users should use [`Exception#cause`](https://ruby-doc.org/core-2.5.0/Exception.html#method-i-cause) to navigate to the what caused the transaction error.

quartzmo:
@Nakilon As you can see from the unit tests that you modified, the original error is available:

```ruby
     error.cause.wont_be :nil?
     error.cause.message.must_equal "This error should be wrapped by TransactionError."
```

Personally I think we need to add examples showing users how to use Exception#cause to navigate to wrapped errors.

blowmage:
I would be more supportive of a documentation change than a structural change.

Nakilon:
Oh, excuse me, I was stick to Ruby 2.0 for too long. Forgot that there is a new stdlib method..

Nakilon:
Neat but some dependencies ("gax"?) already demand >=2.1

PR #2035, opened 2018-04-02
Refresh DLP v2 library

jbolinger:
Refresh of the DLP v2 library. Primarily includes documentation updates.

@quartzmo or @landrito silly question... I noticed that there are labels for most of the libraries. Should I create one for DLP? Do we use those consistently?

### DIFF
--- a/google-cloud-dlp/.rubocop.yml
+++ b/google-cloud-dlp/.rubocop.yml
@@ -4,8 +4,7 @@ AllCops:
     - "lib/google/**/*"
     - "Rakefile"
     - "test/**/*"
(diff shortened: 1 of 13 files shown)


quartzmo:
@jbolinger I added the label, thanks for pointing that out!

quartzmo:
I added the label to [these older issues and PRs](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues?q=dlp+label%3A%22api%3A+data+loss+prevention%22) as well.

quartzmo:
@landrito I also added a label for BigQuery Data Transfer.

PR #2039, opened 2018-04-03
Add insert_ids option to BigQuery #insert and #insert_async

### LINKED ISSUE #1991: Can't specify insertId for BigQuery to do de-duplication
According to BigQuery's document the user can specify `insertId` for each row. And if the same insertId is inserted twice . BigQuery can de-duplicate the data:
https://cloud.google.com/bigquery/streaming-data-into-bigquery

But actually the user can't. And `insertId` is actually "hardcoded" to some uuid here:
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb#L213

This needs to be fixed.

quartzmo:
[closes #1991]

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/table_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/table_test.rb
@@ -67,6 +67,7 @@ describe Google::Cloud::Bigquery::Table, :bigquery do
         { name: "stephen", breed: "idkanycatbreeds",   id: 6, dob: Time.now.utc }
     ]
   end
+  let(:insert_ids) { Array.new(3) {SecureRandom.uuid} }
   let(:local_file) { "acceptance/data/kitten-test-data.json" }
   let(:target_table_id) { "kittens_copy" }
   let(:target_table_2_id) { "kittens_copy_2" }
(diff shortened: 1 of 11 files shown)


google-cloud-bigquery/acceptance/bigquery/table_test.rb
```diff
@@ -476,6 +498,37 @@
     data.next.must_be :nil?
   end
 
+  it "inserts rows asynchonously with insert_ids option" do
+    insert_result = nil
+
+    inserter = table.insert_async insert_ids: insert_ids do |result|
+      insert_result = result
+    end
+    inserter.insert rows
```
blowmage: Why are `insert_ids` passed to the with the rows? The inserter object could be used to insert thousands of rows, users probably won't know all of the `insert_ids` when the inserter object is created.

```ruby
inserter = table.insert_async do |result|
  insert_result = result
end
inserter.insert rows, insert_ids: insert_ids
```

I was expecting `AsyncInserter#insert` to be the same signature as `Dataset#insert` and `Table#insert` and `Table#async_insert`

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1823,6 +1823,12 @@ def self.new_reference project_id, dataset_id, service
         # @param [Boolean] autocreate Specifies whether the method should create
         #   a new table with the given `table_id`, if no table is found for
         #   `table_id`. The default value is false.
+        # @param [Array<String>] insert_ids A unique ID for each row. BigQuery
```
blowmage: Would you be okay listing the `insert_ids` as the first optional argument, just after the `rows` positional argument?

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -1919,6 +1926,12 @@ def insert rows, skip_invalid: nil, ignore_unknown: nil
         # @param [Boolean] ignore_unknown Accept rows that contain values that
         #   do not match the schema. The unknown values are ignored. Default is
         #   false, which treats unknown values as errors.
+        # @param [Array<String>] insert_ids A unique ID for each row. BigQuery
```
blowmage: I am confused why `insert_ids` is an argument to create the inserter object. I think it should only be present on `AsyncInserter#insert`.

google-cloud-bigquery/acceptance/bigquery/table_test.rb
quartzmo: Oh my, of course, great catch. I'll switch to more realistic testing as well when I fix this.

google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
quartzmo: Yes, see reply above.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
quartzmo: And move it in the method signatures as well?

blowmage: Yes, I think it would be best to have the documentation match the method declaration.

quartzmo:
Updated. Thanks for pairing with me on this @blowmage, that was very efficient!

PR #2002, opened 2018-03-09
Support the use of encryption_key when changing the Storage Class

berlincount:
I was unable to change the Storage Class of an encrypted object using the Ruby gem.

This PR allows setting the `encryption_key` on a file object and passes this to the requisite functions.

Not ideal, but it unblocks me and provides a reasonably intuitive interface other methods might want to get their default key.

PR #2042, opened 2018-04-10
Storage: Add File Rewrite

blowmage:
The existing `File#copy` method supports an encryption key for reads/writes, but it does not allow different encryption keys for both the file being read and the file being written. `File#rotate` allows users to replace encryption keys, but not in a single API call that can give the new file a different name or store in a different bucket. Add `File#rewrite` to support all the existing behavior of `File#copy`, while allowing users to specify both the source and destination encryption keys.

[refs #2002]

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -646,31 +646,155 @@ module Google
         #     f.metadata["copied_from"] = "#{file.bucket}/#{file.name}"
         #   end
         #
-        def copy dest_bucket_or_path, dest_path = nil, acl: nil,
-                 generation: nil, encryption_key: nil
-          ensure_service!
-          options = { acl: acl, generation: generation, key: encryption_key,
-                      user_project: user_project }
-          dest_bucket, dest_path, options = fix_copy_args dest_bucket_or_path,
-                                                          dest_path, options
+        def copy dest_bucket_or_path, dest_path = nil,
+                 acl: nil, generation: nil, encryption_key: nil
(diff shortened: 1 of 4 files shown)


quartzmo:
I think we'll want to add docs and examples to `File#storage_class=` pointing to `File#rewrite` and explaining how `rewrite` with a block can be used to satisfy the use case for #2002. We might add a similar reference from the `#copy` docs to explain how `rewrite` can be used to change the key while copying. Happy to pair on this or to add a commit.

quartzmo:
Add an acceptance test for the new method? (I pulled the branch and all existing acceptance tests pass for me.)

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -1112,30 +1228,36 @@ def gapi_from_attrs *attributes
...
             resp = service.rewrite_file \
-              bucket, name, bucket, name, update_gapi,
-              token: resp.rewrite_token, user_project: user_project
+              bucket, name, new_bucket, new_name, updated_gapi, retry_options
           end
           resp.resource
         end
 
-        def fix_copy_args dest_bucket, dest_path, options = {}
-          if dest_path.respond_to?(:to_hash) && options.empty?
```
quartzmo: What was the story with the `options` param and return value, and the `dest_path` here? I'm assuming this is no longer needed, but what was it for?

blowmage:
> Add an acceptance test for the new method? (I pulled the branch and all existing acceptance tests pass for me.)

`File#copy` and `File#rotate` now call `File#rewrite` for their implementation, so I figured that was sufficient for acceptance tests.

PR #2044, opened 2018-04-13
Dialogflow V2

jbolinger:
This is the first generated release for dialogflow.

I'm trying to see if I can get some real tests before we release, so I've marked not to merge, but I thought I'd go ahead and get it out for review in the meantime.


google-cloud-dialogflow/Gemfile
```diff
@@ -0,0 +1,12 @@
+source "https://rubygems.org"
+
+gemspec
+
+gem "gcloud-jsondoc",
+    git: "https://github.com/GoogleCloudPlatform/google-cloud-ruby.git",
+    branch: "gcloud-jsondoc"
+
```
blowmage: You might consider the jwt workaround as well. CI doesn't always fail on Ruby 2.0 without it, but it is possible.

blowmage:
Besides my comment on the Gemfile, this looks good to me so far. :+1:

jbolinger:
Received confirmation that the samples were run against this and working, so we're okay to proceed.

Thanks for the Gemfile catch @blowmage. This is a new API so @quartzmo I would appreciate it if you could double check the docs.

PR #2048, opened 2018-04-16
Admin table snapshots api

jiren:
- Table snapshot create, get, list and delete, create a table from snapshot API
- GRPC mock test cases

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
@@ -227,6 +227,279 @@ module Google
           )
         end
 
+        # This is a private alpha release of Cloud Bigtable snapshots. This feature
+        # is not currently available to most Cloud Bigtable customers. This feature
+        # might be changed in backward-incompatible ways and is not recommended for
+        # production use. It is not subject to any SLA or deprecation policy.
+        #
+        # Creates a new snapshot in the specified cluster from the specified
+        # source table. The cluster and the table must be in the same instance.
+        #
+        # @param table_id [String]
+        #   The unique name of the table to have the snapshot taken.
+        # @param cluster_id [String]
+        #   The name of the cluster where the snapshot will be created in.
+        # @param snapshot_id [String]
+        #   The ID by which the new snapshot should be referred to within the parent
+        #   cluster, e.g., +mysnapshot+ of the form: +[_a-zA-Z0-9][-_.a-zA-Z0-9]*+
+        # @param description [String]
+        #   Description of the snapshot.
+        # @param ttl [Integer]
+        #   The amount of time in seconds that the new snapshot can stay active
+        #   after it is created. Once 'ttl' expires, the snapshot will get
+        #   deleted. The maximum amount of time a snapshot can stay active is
+        #   7 days. If 'ttl' is not specified, the default value of 24 hours
+        #   will be used.
+        # @param options [Google::Gax::CallOptions]
+        #   Overrides the default settings for this call, e.g, timeout,
+        #   retries, etc.
+        # @return [Google::Longrunning::Operation]
+        # @raise [Google::Gax::GaxError] if the RPC is aborted.
+        # @example
+        #   require "google/cloud/bigtable"
+        #
+        #   client = Google::Cloud::Bigtable.new(
+        #     client_type: :table,
+        #     instance_id: "instance-id"
+        #   )
+        #
+        #   response = client.create_snapshot(
+        #     "table-1",
+        #     "cluster-1",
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
```diff
@@ -227,6 +227,280 @@ def drop_row_range \
...
+        #   end
+        #
+        #   # List all snapshots from all clusters
+        #   client.snapshots.each do |snapshot|
+        #     # Process snapshot
+        #     p snapshot
+        #   end
+
+        def snapshots \
+            cluster_id = "-",
```
quartzmo: I don't see a test for the default `cluster_id` value of `"="`.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
```diff
@@ -227,6 +227,280 @@ def drop_row_range \
...
+        #   cluster, e.g., +mysnapshot+ of the form: +[_a-zA-Z0-9][-_.a-zA-Z0-9]*+
+        # @param description [String]
+        #   Description of the snapshot.
+        # @param ttl [Integer]
+        #   The amount of time in seconds that the new snapshot can stay active
+        #   after it is created. Once 'ttl' expires, the snapshot will get
+        #   deleted. The maximum amount of time a snapshot can stay active is
+        #   7 days. If 'ttl' is not specified, the default value of 24 hours
+        #   will be used.
+        # @param options [Google::Gax::CallOptions]
```
blowmage: We don't usually expose `Google::Gax::CallOptions` in the high level API.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
```diff
@@ -227,6 +227,280 @@ def drop_row_range \
...
+        #   Overrides the default settings for this call, e.g, timeout,
+        #   retries, etc.
+        # @return [Google::Longrunning::Operation]
+        # @raise [Google::Gax::GaxError] if the RPC is aborted.
```
blowmage: We have been wrapping `Google::Gax::GaxError` with a `Google::Cloud::Error`.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
```diff
@@ -227,6 +227,280 @@ def drop_row_range \
...
+        #   )
+
+        def create_snapshot \
+            table_id,
+            cluster_id,
+            snapshot_id,
+            description,
+            ttl: nil,
+            options: nil
+          ttl = Google::Protobuf::Duration.new(seconds: ttl) if ttl
```
blowmage: We use this helper method in some of the other gems:

```ruby
def number_to_duration number
  return nil if number.nil?

  Google::Protobuf::Duration.new \
    seconds: number.to_i,
    nanos: (number.remainder(1) * 1000000000).round
end
```

See [`Convert.number_to_duration`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/convert.rb#L42-L48)

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
jiren: I will add.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
jiren: I will make a change.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
jiren: Got it. I will update in all other APIs also. 

In some of the APIs call options require i.e In `drop_rows` BigTable server is responding with Error - increase timeout limit and try again.
I am planning to add an option for `timeout` for such particular API call.

google-cloud-bigtable/lib/google/cloud/bigtable/table_admin_client.rb
jiren: I will take a reference from it.

sduskis:
Snapshots are an alpha feature on the service.  This R should be considered a low priority.

jiren:
@sduskis I am closing this pull request. All admin apis are covered in https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2128

PR #2049, opened 2018-04-16
Fix issue with labels when nil

tareksamni:
Issue trying to solve:
I've integrated the GCP Stackdriver Trace into my rails project but I was getting an error in some of traces with this message `Transmit to Stackdriver Trace failed:`.

After some traces in the code, I found out that `@service.patch_traces trace` in [middleware.rb](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/middleware.rb#L214) throws an exception that was propagated from  [service.rb](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/service.rb#L81) 

The reason was that some of the traces are passing `labels` in the span with `nil` value which apparently violating the ProtoBuf definition at: [trace_pb.rb](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/devtools/cloudtrace/v1/trace_pb.rb#L25) as it expects a hash while getting a nil.

How to solve?
I think explicitly protecting against nil value in labels in span should be in place to protect against any unexpected passed values to the span's labels which is the case here.

I've written a test case where I passed `labels` parameter with `nil` and I expected `labels` to be `{}` which did not happen. then I fixed the issue inside the span constructor to fix the test.

I hope this was clear enough. Please, feel free to discuss more details. I've monkey patched my project to override the labels if it was nil for now but I'd love to get this update so that I can remove my monkey patching ..

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,6 +2,8 @@
 
 These instructions apply to every gem within the Google Cloud Ruby Client project.
 
+## Releasing individual gems and meta-packages
+
 **Each gem must be released separately.** In order for the docs for the `google-cloud` package to build correctly, the only entry in `docs/manifest.json` that can be updated in the version commit preceding the release tag is the entry for the gem in the tag. The docs build will fail if you attempt to release multiple gems in parallel, since the first tag build will not yet find the listed docs for the other gems in the `gh-pages` branch.
 
 The Google Cloud Ruby Client project uses [semantic versioning](http://semver.org). Replace the `<prev_version>` and `<version>` placeholders shown in the examples below with the appropriate numbers, e.g. `0.1.0` and `0.2.0`. Replace the `<gem>` placeholder with the appropriate full name of the package, e.g. `google-cloud-datastore`.
@@ -42,11 +44,15 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
 
 1. Edit the gem's `CHANGELOG.md`. Using your notes from the previous step, write bullet-point lists of the major and minor changes. You can also add examples, fixes, thank yous, and anything else helpful or relevant. See google-cloud-node [v0.18.0](https://github.com/GoogleCloudPlatform/google-cloud-node/releases/tag/v0.18.0) for an example with all the bells and whistles.
 
-1. Edit the gem's `version.rb` file, changing the value of `VERSION` to your new version number.
+1. Edit the gem's `version.rb` file, if present, or the `version` setting in its `.gemspec` file, changing the value to your new version number.
 
-1. Edit the gem's entry in `docs/manifest.json`, adding your new version number to the head of the list, and moving `"master"` to be just below it.
+1. Edit (or add if new) the gem's entry in `docs/manifest.json`, adding your new version number to the head of the list, and moving `"master"` to be just below it.
 
-1. If the package is `< 1.0.0` and your version change is greater than the [semver](http://semver.org/) patch version, or if the package is `>= 1.0.0` and your version change increments the [semver](http://semver.org/) major version, edit the requirement for the gem in `google-cloud/google-cloud.gemspec` and `stackdriver/stackdriver.gemspec` (if the package is a dependency of stackdriver) , replacing the old version number for the gem with your new version number. Note that the dependency versions of the `google-cloud` and `stackdriver` must gems remain compatible so the two can co-exist in the same bundle.
+1. If the [semver](http://semver.org/) version change for your package requires an increase in the requirement for your package in `google-cloud/google-cloud.gemspec` and/or `stackdriver/stackdriver.gemspec`, replace the old version requirement with your new requirement. Note that because of the use of the [pessimistic operator (`~>`)](https://robots.thoughtbot.com/rubys-pessimistic-operator), only certain version changes will require updating the requirement. Note also that the dependency requirements in the `google-cloud` and `stackdriver` gems must remain compatible so the two can co-exist in the same bundle.
+
+1. If your package is new, ensure that it has been added to the [top-level `Gemfile`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/google-cloud/v0.52.0/Gemfile).
+
+1. If your package is new, ensure that a nav link and a main entry including code example have been added to the [top-level README](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/google-cloud/v0.52.0/README.md).
 
 1. In the root directory of the project, test that all the version dependencies are correct.
 
@@ -85,11 +91,11 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
 
 1. Click `Publish release`.
 
-1. Repeat steps 1 through 19 if you are releasing multiple gems.
+1. Repeat steps 1 through 21 if you are releasing multiple gems.
 
-1. If you updated `google-cloud/google-cloud.gemspec` for a version change to any gem, repeat steps 1 through 19 for the `google-cloud` gem.
+1. If you updated `google-cloud/google-cloud.gemspec` for a version change to any gem, repeat steps 1 through 21 for the `google-cloud` gem. If any gem is new, first follow the steps in [Adding a new gem to meta-packages](#adding-a-new-gem-to-meta-packages), below.
 
-1. If you updated `stackdriver/stackdriver.gemspec` for a version change to any gem, repeat steps 1 through 19 for the `stackdriver` gem.
+1. If you updated `stackdriver/stackdriver.gemspec` for a version change to any gem, repeat steps 1 through 21 for the `stackdriver` gem. If any gem is new, first follow the steps in [Adding a new gem to meta-packages](#adding-a-new-gem-to-meta-packages), below.
 
 1. Wait until the last tag build job has successfully completed on Circle CI. Then push your commits to the master branch. This will trigger another [Circle CI](https://circleci.com/gh/GoogleCloudPlatform/google-cloud-ruby) build on master branch.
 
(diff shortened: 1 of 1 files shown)


quartzmo:
> After some traces in the code, I found out that @service.patch_traces trace in middleware.rb throws an exception that was propagated from service.rb

@tareksamni Do you still have the exception backtrace, and if so, can you post it here?

tareksamni:
Yes, I exactly I went through most paths as well of creating Span and could not figure out how this is happening! but in practice I was getting this exception `Transmit to Stackdriver Trace failed` and after debugging I found that labels value at the time of passing it to protobuf was nil :) 

I will reproduce it once I'm on the same code again and post it .. expect it in the next 24 hours.

tareksamni:
@quartzmo just note that if you passed `labels` as `nil` intentionally then it will override the default value of `labels: {}` in all of the methods that you checked.

tareksamni:
I found that, my monkey patching was the root cause of making `labels = nil`. So this fix is not valid.

but the issue is still there and I now know the root cause for it :) which is the good news .. I think I should close this PR and start a new one. but here you go with the details:

**Reproduce:**

Just integrating the gem for Rails app as advised in the README.md file, I started getting this error on some of the child spans:

```
TypeError: Invalid argument for string field.
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:183:in `initialize'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:183:in `new'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:183:in `to_grpc'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:112:in `block in to_grpc'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:111:in `map'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:111:in `to_grpc'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:81:in `block in patch_traces'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:80:in `each'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:80:in `patch_traces'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/async_reporter.rb:79:in `run_backgrounder'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:291:in `async_run_job'
	/Users/xxxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:315:in `block (2 levels) in ensure_thread'
```

Adding the following code to [`span.rb`](https://github.com/tareksamni/google-cloud-ruby/blob/0a86c2f003fe4cc9274aabb00ddf878edb594674/google-cloud-trace/lib/google/cloud/trace/span.rb#L179):

```
        def to_grpc default_parent_id = 0
          start_proto = Google::Cloud::Trace::Utils.time_to_grpc start_time
          end_proto = Google::Cloud::Trace::Utils.time_to_grpc end_time
          puts '*' * 100
          puts "span_id: #{span_id.to_i}"
          puts "kind: #{kind.to_sym}"
          puts "name: #{name}"
          puts "start_time: #{start_proto}"
          puts "end_time: #{end_proto}"
          puts "parent_span_id: #{parent_span_id || default_parent_id}"
          puts "labels: #{labels}"
          puts '*' * 100

          labels.each do |k, v|
            puts "key: #{k.class}, value: #{v.class}"
          end

          Google::Devtools::Cloudtrace::V1::TraceSpan.new \
            span_id: span_id.to_i,
            kind: kind.to_sym,
            name: name,
            start_time: start_proto,
            end_time: end_proto,
            parent_span_id: parent_span_id || default_parent_id,
            labels: labels
        end
```

I got this output with exception in rails logs:

```
****************************************************************************************************
span_id: 8581987211832177959
kind: SPAN_KIND_UNSPECIFIED
name: render_template.action_view
start_time: #<Google::Protobuf::Timestamp:0x007fdf66b188c8>
end_time: #<Google::Protobuf::Timestamp:0x007fdf66b18850>
parent_span_id: 16663580013895858967
labels: {"/ruby/identifier"=>"/Users/xxx/projects/TestProject/app/views/users/index.html.erb", "/ruby/layout"=>"layouts/application"}
****************************************************************************************************
key: String, value: String
key: String, value: ActionView::Resolver::Path
TypeError: Invalid argument for string field.
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:196:in `initialize'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:196:in `new'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/span.rb:196:in `to_grpc'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:112:in `block in to_grpc'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:111:in `map'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/trace_record.rb:111:in `to_grpc'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:81:in `block in patch_traces'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:80:in `each'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/service.rb:80:in `patch_traces'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/google-cloud-trace-0.31.0/lib/google/cloud/trace/async_reporter.rb:79:in `run_backgrounder'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:291:in `async_run_job'
	/Users/xxx/projects/TestProject/vendor/bundle/ruby/2.4.0/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:315:in `block (2 levels) in ensure_thread'
```

Apparently, a value in `labels` is not a string but an `ActionView::Resolver::Path` which still a child class of `String` but Protobuf does not like it and won't accept it then it throws an exception `TypeError: Invalid argument for string field.`.

These labels are set in [`notifications.rb`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/notifications.rb#L108) as:

```
        def self.payload_to_labels event, maxlen, label_namespace
          labels = {}
          event.payload.each do |k, v|
            if v.is_a? ::String
              v = v[0, maxlen - 3] + "..." if maxlen && v.size > maxlen
              labels["#{label_namespace}#{k}"] = v
            end
          end
          labels
        end
```

`v.is_a? ::String` will return true even for `ActionView::Resolver::Path` because it's a child class of `::String` which will cause the error afterwords as mentioned before.

A nice resolution for this is to always make sure that the value is a String and nothing but string (not even a child class of string). using code like: `labels["#{label_namespace}#{k}"] = v.to_s`.

Does this make sense? should I open another PR for it?

quartzmo:
@tareksamni Thanks, I'll try to take a look at this today.

quartzmo:
@tareksamni Thanks, I just read your comment. The problem makes sense, but regarding the solution, where currently in the library are `labels` entries being set to values that may be `ActionView::Resolver::Path` values?

tareksamni:
These labels are set in [`notifications.rb`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/notifications.rb#L108) as:

```
        def self.payload_to_labels event, maxlen, label_namespace
          labels = {}
          event.payload.each do |k, v|
            if v.is_a? ::String
              v = v[0, maxlen - 3] + "..." if maxlen && v.size > maxlen
              labels["#{label_namespace}#{k}"] = v
            end
          end
          labels
        end
```

when event is `render_template.action_view` then `event.payload` is:

```
{
 "identifier"=>"/Users/xxx/projects/TestProject/app/views/users/index.html.erb",  # string
 "layout"=>"layouts/application" # ActionView::Resolver::Path
}
```

tareksamni:
@quartzmo did you have a chance to look into this?

quartzmo:
@tareksamni Looking again now. Thanks for the ping.

quartzmo:
@tareksamni Sorry this took so long, it all makes sense to me now. Just one question: Does calling `to_s` on `ActionView::Resolver::Path` produce values that are satisfactory? If so, please go ahead with the PR. And thank you for your patience and persistence!

blowmage:
Nearly all objects have `to_s`. I would replace `if v.is_a? ::String` with `if v.responds_to? :to_str` and then add a line to the conditional block that calls it `v = v.to_str`

tareksamni:
@blowmage, as you said nearly all objects have `to_s` but are we really interested in all objects?

for example: `Object.new.to_s` would generate something like: `"#<Object:0x007fe14f0059f0>"`.

Most of the objects that are not meant to have a string data (`human meaningful`) would probably generate the same output.

That's why we only need to stick to `if v.is_a? ::String` because this makes sure that we only pass meaningful human data (strings) there. All we need to do is just to make sure that the object is casted to a string to avoid passing and object that is an instance of a class that is child of ::String which is happening now with `ActionView::Resolver::Path`.

tareksamni:
@quartzmo Yes `layouts/application` is very informative for Rails developer and this is the expected result :)

PR #2065, opened 2018-04-26
Monitoring V3 Documentation Fixes

jbolinger:
Added the missing TOC entries. There still seems to be an issue (blank page) when clicking on the version (V3) on the left hand menu bar, but I'm not sure where that is configured.


quartzmo:
I’ll take a look at the blank page issue, and do a PR to fix if possible.

jbolinger:
thanks @quartzmo - CC me if you figure that out so I don't do it again. I'm sure it'll be painfully obvious in retrospect 😀

PR #2067, opened 2018-04-27
Spanner STRUCT in Query Parameters

blowmage:
This PR adds support for defining and using STRUCT query parameters.

This can be done by sending a Ruby `Hash` of values as a query parameter.

```ruby
require "google/cloud/spanner"

spanner = Google::Cloud::Spanner.new

db = spanner.client "my-instance", "my-database"

results = db.execute "SELECT * FROM users WHERE " \
                     "ID = @user_struct.id " \
                     "AND name = @user_struct.name " \
                     "AND active = @user_struct.active",
                     params: { user_struct: { id: 1, name: "Charlie", active: false } }

results.rows.each do |row|
  puts "User #{row[:id]} is #{row[:name]}"
end
```

Or, the STRUCT can be defined using a `Fields` object and passing values to `Fields#struct`:

```ruby
require "google/cloud/spanner"

spanner = Google::Cloud::Spanner.new

db = spanner.client "my-instance", "my-database"

user_type = db.fields id: :INT64, name: :STRING, active: :BOOL
user_struct = user_type.struct({ id: 1, name: "Charlie", active: false })

results = db.execute "SELECT * FROM users WHERE " \
                     "ID = @user_struct.id " \
                     "AND name = @user_struct.name " \
                     "AND active = @user_struct.active",
                     params: { user_struct: user_struct }

results.rows.each do |row|
  puts "User #{row[:id]} is #{row[:name]}"
end
```

This allows users to create STRUCT values that cannot be defined using a Ruby `Hash`.  e.g. anonymous fields, duplicate field names, etc.

This functionality is not yet widely available, so this PR should not be merged until it is. But it should be reviewed now.

### DIFF
--- a/google-cloud-spanner/acceptance/spanner/client/params/struct_test.rb
+++ b/google-cloud-spanner/acceptance/spanner/client/params/struct_test.rb
@@ -17,19 +17,164 @@ require "spanner_helper"
 describe "Spanner Client", :params, :struct, :spanner do
   let(:db) { spanner_client }
 
-  it "queries and returns a struct parameter" do
-    skip "Sending a STRUCT was working, but now returns an error"
+  describe "Struct Parameters Query Examples" do
+    # Simple field access.
+    # [parameters=STRUCT<threadf INT64, userf STRING>(1,"bob") AS struct_param, 10 as p4]
+    # SELECT @struct_param.userf, @p4;
+    it "Simple field access" do
+      results = db.execute "SELECT @struct_param.userf, @p4",
+                           params: { struct_param: { threadf: 1, userf: "bob" }, p4: 10 }
 
-    results = db.execute "SELECT ARRAY(@value) AS value", params: { value: { message: "hello", repeat: 1 } }
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ userf: :STRING, 1 => :INT64 })
+      results.rows.first.to_h.must_equal({ userf: "bob", 1 => 10 })
+    end
+
+    # # Simple field access on NULL struct value.
+    # [parameters=CAST(NULL AS STRUCT<threadf INT64, userf STRING>) AS struct_param]
+    # SELECT @struct_param.userf;
+    it "Simple field access on NULL struct value" do
+      struct_type = db.fields(threadf: :INT64, userf: :STRING)
+      results = db.execute "SELECT @struct_param.userf",
+                           params: { struct_param: nil },
+                           types:  { struct_param: struct_type }
+
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ userf: :STRING })
+      results.rows.first.to_h.must_equal({ userf: nil })
+    end
+
+    # # Nested struct field access.
+    # [parameters=STRUCT<structf STRUCT<nestedf STRING>> (STRUCT<nestedf STRING>("bob")) AS struct_param]
+    # SELECT @struct_param.structf.nestedf;
+    it "Nested struct field access" do
+      results = db.execute "SELECT @struct_param.structf.nestedf",
+                           params: { struct_param: { structf: { nestedf: "bob" } } }
+
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ nestedf: :STRING })
+      results.rows.first.to_h.must_equal({ nestedf: "bob" })
+    end
+
+    # # Nested struct field access on NULL struct value.
+    # [parameters=CAST(STRUCT(null) AS STRUCT<structf STRUCT<nestedf STRING>>) AS  struct_param]
+    # SELECT @struct_param.structf.nestedf;
+    it "Nested struct field access on NULL struct value" do
+      struct_type = db.fields(structf: db.fields(nestedf: :STRING))
+      results = db.execute "SELECT @struct_param.structf.nestedf",
+                           params: { struct_param: nil },
+                           types:  { struct_param: struct_type }
+
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ nestedf: :STRING })
+      results.rows.first.to_h.must_equal({ nestedf: nil })
+    end
+
+    # # Non-NULL struct with no fields (empty struct).
+    # [parameters=CAST(STRUCT() AS STRUCT<>) AS struct_param]
+    # SELECT @struct_param IS NULL;
+    it "Non-NULL struct with no fields (empty struct)" do
+      struct_type = db.fields({})
+      results = db.execute "SELECT @struct_param IS NULL",
+                           params: { struct_param: {} }
+
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ 0 => :BOOL })
+      results.rows.first.to_h.must_equal({ 0 => false })
+    end
+
+    # # NULL struct with no fields.
+    # [parameters=CAST(NULL AS STRUCT<>) AS struct_param]
+    # SELECT @struct_param IS NULL
+    it "NULL struct with no fields" do
+      struct_type = db.fields({})
+      results = db.execute "SELECT @struct_param IS NULL",
+                           params: { struct_param: nil },
+                           types:  { struct_param: struct_type }
+
+      results.must_be_kind_of Google::Cloud::Spanner::Results
+      results.fields.to_h.must_equal({ 0 => :BOOL })
+      results.rows.first.to_h.must_equal({ 0 => true })
+    end
+
+    # # Struct with single NULL field.
+    # [parameters=STRUCT<f1 INT64>(NULL) AS struct_param]
+    # SELECT @struct_param.f1;
+    it "Struct with single NULL field" do
+      struct_type = db.fields(f1: :INT64)
+      results = db.execute "SELECT @struct_param.f1",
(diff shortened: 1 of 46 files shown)


google-cloud-spanner/lib/google/cloud/spanner/convert.rb
```diff
@@ -164,6 +103,64 @@ def raw_to_value obj
             end
           end
 
+          def field_for_raw obj
+            if NilClass === obj
```
tareksamni: I think using Ruby's `case` is better here. `===` is not widely used in ruby community as it's asymmetric for ex.:
`Symbol === :symbol # => true`
`:symbol === Symbol # => false` 😱

I recommend doing something like:
```
case obj
when NilClass
   raise ArgumentError, "Cannot determine type for nil values."
when Date
  ................................
```

google-cloud-spanner/lib/google/cloud/spanner/convert.rb
```diff
@@ -164,6 +103,64 @@ def raw_to_value obj
...
+          def type_for_field field
+            return field.to_grpc_type if field.respond_to? :to_grpc_type
+
+            if Array === field
+              Google::Spanner::V1::Type.new(
+                code: :ARRAY,
+                array_element_type: type_for_field(field.first)
+              )
+            else
+              Google::Spanner::V1::Type.new(code: field)
```
tareksamni: Guard statement is preferred here than a one statement condition branching ..
something like:

```
return Google::Spanner::V1::Type.new(
                code: :ARRAY,
                array_element_type: type_for_field(field.first)
              ) if Array === field
Google::Spanner::V1::Type.new(code: field)
```

google-cloud-spanner/lib/google/cloud/spanner/fields.rb
```diff
@@ -41,20 +41,75 @@ module Spanner
...
+          types = types.to_a if types.is_a? Hash
+
+          unless types.is_a? Array
             raise ArgumentError, "can only accept Array or Hash"
           end
+
+          # Verify positional arguments are in order
+          types.each_with_index do |type, index|
+            if type.is_a?(Array) && type.first.is_a?(Integer)
+              raise "types must be in order" if type.first != index
```
tareksamni: Should we have custom `Error`s ? instead of raising anonymous/generic errors.

google-cloud-spanner/lib/google/cloud/spanner/convert.rb
blowmage: This is generally true, but for IO-ish objects we need to detect if the object has `read` and `rewind` methods (`elsif obj.respond_to?(:read) && obj.respond_to?(:rewind)`), and we can't do that with a `case` statement.

google-cloud-spanner/lib/google/cloud/spanner/fields.rb
tareksamni: Yes, this what I meant :) not just a raise ..

google-cloud-spanner/lib/google/cloud/spanner/batch_client.rb
```diff
@@ -272,6 +272,77 @@ def load_partition serialized_partition
           Partition.load serialized_partition
         end
 
+        ##
+        # Creates {Fields} object from types.
+        #
+        # This object can be used to create {Data} objects by providing values
+        # that match the field types. See {Fields#struct}.
```
quartzmo: How about:

This method returns a configuration object ({Fields}) that may be provided to queries or may be used to create STRUCT objects. (The STRUCT is represented by the {Data} class.) See {Client#execute} and/or {Fields#struct}.

quartzmo: Updated:

This method returns a configuration object ({Fields}) that may be provided to queries or used to create STRUCT objects. (The STRUCT **will be** represented by the {Data} class.) See {Client#execute} and/or {Fields#struct}.

blowmage: Fixed.

google-cloud-spanner/lib/google/cloud/spanner/convert.rb
blowmage: Due to the amount of code in each conditional I'm fine keeping this method the way it is.

quartzmo:
@blowmage Latest docs update looks great!

google-cloud-spanner/lib/google/cloud/spanner/fields.rb
blowmage: This error was a placeholder until support for unsorted positional types was added. This code has been removed.

quartzmo:
@blowmage Most recent commits LGTM, thanks!

blowmage:
@vkedia Ping!

google-cloud-spanner/lib/google/cloud/spanner/batch_snapshot.rb
```diff
@@ -419,8 +440,72 @@ def close
...
+        #     { id: :INT64, name: :STRING, active: :BOOL }
+        #   )
+        #   user_hash = { id: 1, name: nil, active: false }
+        #
+        #   results = batch_snapshot.execute "SELECT * FROM users WHERE " \
+        #                                    "ID = @user_struct.id " \
+        #                                    "AND name = @user_struct.name " \
+        #                                    "AND active = @user_struct.active",
+        #                                    params: { user_struct: user_hash },
+        #                                    types: { user_struct: user_type }
```
vkedia: what happens if they do not pass the type here? How would the library infer the type of "name"?

google-cloud-spanner/acceptance/spanner/client/params/struct_test.rb
```diff
@@ -17,19 +17,164 @@
...
+      results.fields.to_h.must_equal({ 0 => :BOOL })
+      results.rows.first.to_h.must_equal({ 0 => false })
+    end
+
+    # # Null array of struct field.
+    # [parameters=STRUCT<intf INT64, arraysf ARRAY<STRUCT<threadid INT64>>> (10,CAST(NULL AS ARRAY<STRUCT<threadid INT64>>)) AS struct_param]
+    # SELECT a.threadid FROM UNNEST(@struct_param.arraysf) a;
+    it "Null array of struct field" do
+      struct_value = db.fields(intf: :INT64, arraysf: [db.fields(threadid: :INT64)]).struct([10, nil])
+      results = db.execute "SELECT a.threadid FROM UNNEST(@struct_param.arraysf) a",
```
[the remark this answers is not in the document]
vkedia: Yes you are right. Ignore this comment.

vkedia:
Currently from a row how do we return the struct value?

vkedia:
And how do we return the type of a column which is of type array<struct>?

blowmage:
> And how do we return the type of a column which is of type `array<struct>`?

Each Data object has a Fields object that defines the types. The type of an array of struct objects is represented as an array with a nested Fields object.

danoscarmike:
@vkedia can you sign off on responses to your most recent comments?  Thanks!

vkedia:
So I just want to be sure that the  way we return a struct and an array of struct from a row is same as the way we consume them as query parameters. But it is not clear to me that is true. Can you confirm?

blowmage:
Yes, we return a `STRUCT` and an `ARRAY<STRUCT>` in the same way they are consumed. For additional convenience, we will also map a Ruby Hash (without `nil` values) to a `STRUCT`, since it is a sub-set of all the ways you can model data in a `STRUCT`. But any valid configuration of `STRUCT` (nameless, duplicate names, etc) can be modeled in Ruby using `Client#fields`/`Fields.new` and `Fields#struct` to produce a `Data` object, which is the same Ruby object returned from API calls.

google-cloud-spanner/lib/google/cloud/spanner/convert.rb
```diff
@@ -140,42 +77,110 @@ def raw_to_value obj
...
+            when DateTime
+              :TIMESTAMP
+            when Date
+              :DATE
+            when Array
+              if obj.empty?
+                raise ArgumentError,
+                      "Cannot determine type for empty array values."
+              end
+              non_nil_fields = obj.compact.map { |e| field_for_object e }.compact
```
jabubake: For input:

 ```
client  = spanner.client instance_id, database_id
name_type = client.fields(FirstName: :STRING, LastName: :STRING)
band_members = [name_type.struct(["Elena", "Campbell"]), name_type.struct(["Gabriel", "Wright"])];

client.execute("SELECT SingerId FROM Singers WHERE " \
                  "STRUCT<FirstName STRING, LastName STRING>(FirstName, LastName) IN UNNEST(@names)",
                 params: { names: band_members }).rows.each do |row|
    puts "#{row[:SingerId]}"
                 end
```

This compact does not seem to be working correctly and returns:
```
((FirstName):STRING, (LastName):STRING)
((FirstName):STRING, (LastName):STRING)
```
And hence throws the error : "Cannot determine type for array of different values"

Can you add a test case that uses an array of struct ?

@jsimonweb FYI

blowmage: So after discussing this with @quartzmo we decided that it would be acceptable to use Hash-key comparison with `Fields` since it is an immutable object. This made the code change to fix this very simple. I've added code coverage with this change as well to make sure the `Cannot determine type for array of different values` error was not being raised.

blowmage:
Still waiting to get approval to merge.

blowmage:
We got approval to merge. Removing the label now.

PR #2068, opened 2018-04-29
Facilitate partial file downloads

georgeclaghorn:
[Active Storage](https://github.com/rails/rails/tree/master/activestorage) requires partial download support from each of the clients it uses in order to:
* Peek at objects for content type identification purposes without downloading them in full
* Stream objects from cloud storage to tempfiles on disk in a consistent manner across all services

The Google Cloud Storage client, unlike the other clients Active Storage uses, does not provide this support. We currently work around the lack of partial download support by [manually](https://github.com/rails/rails/blob/3e8c660c11df5f650109fb44d060ea7eb0c7ea26/activestorage/lib/active_storage/service/gcs_service.rb#L51-L56) issuing a GET request with a `Range` header to a signed URL:

```ruby
file = file_for(key)
uri  = URI(file.signed_url(expires: 30.seconds))

Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == "https") do |client|
  client.get(uri, "Range" => "bytes=#{range.begin}-#{range.exclude_end? ? range.end - 1 : range.end}").body
end
```

Because we bypass the GCS client, we miss out on authenticated downloading, connection reuse, error handling, and more. This PR adds partial download support to the client, reducing the code above to the following:

```ruby
file_for(key).download(range: range).string
```

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -440,14 +440,14 @@ module Google
         #   AES-256 encryption key used to encrypt the file, if one was provided
         #   to {Bucket#create_file}.
         #
-        # @param [String] range Optional. The value to use in the Range header
-        #   of the download request. Provide this to perform a partial download.
-        #   When a range is provided, no verification is performed regardless of
-        #   the `verify` parameter's value.
+        # @param [Range] range Optional. The byte range of the file's contents
+        #   to download. Provide this to perform a partial download. When a
+        #   range is provided, no verification is performed regardless of the
+        #   `verify` parameter's value.
         #
         #   Infinite range values are not supported. The provided range must
-        #   have a finite beginning and end. For example, "bytes=3-6" is
-        #   acceptable, while "bytes=3-" is not.
+        #   have a finite beginning and end. For example, `3..6` is acceptable,
+        #   while `3..` and `..6` are not.
         #
         # @param [Boolean] skip_decompress Optional. If `true`, the data for a
         #   Storage object returning a `Content-Encoding: gzip` response header
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -561,9 +561,9 @@ module Google
         #   bucket = storage.bucket "my-bucket"
         #   file = bucket.file "path/to/my-file.ext"
         #
-        #   downloaded = file.download range: 3..6
+        #   downloaded = file.download range: 6..10
         #   downloaded.rewind
-        #   downloaded.read #=> "Hello world!"
+        #   downloaded.read #=> "world"
         #
         def download path = nil, verify: :md5, encryption_key: nil, range: nil,
                      skip_decompress: nil
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -561,9 +561,9 @@ module Google
         #   bucket = storage.bucket "my-bucket"
         #   file = bucket.file "path/to/my-file.ext"
         #
-        #   downloaded = file.download range: "bytes=3-6"
+        #   downloaded = file.download range: 3..6
         #   downloaded.rewind
-        #   downloaded.read #=> "lo w"
+        #   downloaded.read #=> "Hello world!"
         #
         def download path = nil, verify: :md5, encryption_key: nil, range: nil,
                      skip_decompress: nil
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -440,14 +440,11 @@ module Google
         #   AES-256 encryption key used to encrypt the file, if one was provided
         #   to {Bucket#create_file}.
         #
-        # @param [Range] range Optional. The byte range of the file's contents
-        #   to download. Provide this to perform a partial download. When a
-        #   range is provided, no verification is performed regardless of the
-        #   `verify` parameter's value.
-        #
-        #   Infinite range values are not supported. The provided range must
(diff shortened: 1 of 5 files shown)


google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -543,17 +553,32 @@ def update
...
             path = StringIO.new
             path.set_encoding "ASCII-8BIT"
           end
-          file, resp = service.download_file \
-            bucket, name, path, key: encryption_key, user_project: user_project
+          file, resp =
+            service.download_file bucket, name, path,
+                                  key: encryption_key, range: range,
+                                  user_project: user_project
           # FIX: downloading with encryption key will return nil
```
frankyn: @quartzmo could you provide more context on the `Fix` comment?

@georgeclaghorn I'd like to see this PR include a test for a ranged GET when the file is encrypted.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -543,17 +553,32 @@ def update
...
+        # @example Partially download.
+        #
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   file = bucket.file "path/to/my-file.ext"
+        #
+        #   downloaded = file.download range: "bytes=3-6"
+        #   downloaded.rewind
```
frankyn: Looking for context: Why did you perform a `rewind` after download in this example?

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -543,17 +553,32 @@ def update
...
-        def download path = nil, verify: :md5, encryption_key: nil,
+        # @example Partially download.
+        #
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   file = bucket.file "path/to/my-file.ext"
+        #
+        #   downloaded = file.download range: "bytes=3-6"
```
frankyn: I like the simplicity of design, but I'm questioning about exposing `range_start:` and `range_end:` parameters instead of having a developer determine the format correctly `bytes=#-#`. 

Another design could be using a Ruby Range `#...#`, so `range: #..#` and benefit from the lazy enumeration.

google-cloud-storage/lib/google/cloud/storage/file.rb
georgeclaghorn: `download` doesn’t rewind the StringIO it returns after writing. The caller must rewind for a subsequent read to return the downloaded file contents.

google-cloud-storage/lib/google/cloud/storage/file.rb
quartzmo: @frankyn It was @blowmage who added that comment, in 5b6c0355, so he might be able to tell us.

quartzmo: The same comment appears in [file_test.rb#L180](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/5b6c035540da45de4988d85ed34b98cd607963f3/google-cloud-storage/test/google/cloud/storage/file_test.rb#L180).

blowmage: Usually `service.download_file` returns the File/IO object the contents were written to, except that it returns `nil` when using an encryption key. So when the variable is `nil` we are setting it to a File object of the path that the contents were written to.

quartzmo:
The CI build failed with:

```sh
  1) Failure:

Google::Cloud::Storage::File#download#test_0001_Partially download. [/home/ubuntu/google-cloud-ruby/google-cloud-storage/lib/google/cloud/storage/file.rb:568]:

--- expected

+++ actual

@@ -1 +1 @@

-"lo w"

+"#<ArgumentError: unknown keyword: range>"
```

georgeclaghorn:
Fixed the failing test.

georgeclaghorn:
@frankyn, I added an acceptance test for partially downloading an encrypted file in 278f377 and confirmed that the acceptance tests pass locally.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -543,17 +553,32 @@ def update
...
+        #
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   file = bucket.file "path/to/my-file.ext"
+        #
+        #   downloaded = file.download range: 3..6
+        #   downloaded.rewind
+        #   downloaded.read #=> "Hello world!"
```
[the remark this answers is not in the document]
georgeclaghorn: Fixed.

frankyn:
LGTM, thanks @georgeclaghorn!

georgeclaghorn:
@blowmage, I added String header value support. :+1:

google-cloud-storage/lib/google/cloud/storage/service.rb
```diff
@@ -484,9 +484,13 @@ def encryption_key_headers options, key, copy_source: false
         end
 
         def range_header options, range
-          if range
+          case range
+          when Range
             options[:header] ||= {}
             options[:header]["Range"] = "bytes=#{range.min}-#{range.max}"
+          when String
```
blowmage: :+1:

quartzmo:
@georgeclaghorn How urgently do you need this released? We also have #2059 for GCS, but it's temporarily blocked. So we can go ahead with a release for your feature, or wait...

quartzmo:
Thank you for supporting Google Cloud Storage!

quartzmo:
@georgeclaghorn Storage 1.11.0 is now [on rubygems](https://rubygems.org/gems/google-cloud-storage/versions/1.11.0).

PR #2081, opened 2018-05-09
Support BigQuery NUMERIC

blowmage:
This PR adds support for BigQuery NUMERIC. These values will be mapped to Ruby's BigDecimal class.


blowmage:
This feature is not yet widely available, so do not merge until it is.

ElliottBrossard:
This can be merged whenever it is ready now. The NUMERIC data type is in beta: https://cloud.google.com/bigquery/docs/release-notes#may_15_2018. Thanks!

quartzmo:
The official API docs for [NUMERIC type](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric-type).

quartzmo:
@blowmage I guess it might be a good idea to rebase this branch for the google-cloud-bigquery 1.4.0 release.

google-cloud-bigquery/lib/google/cloud/bigquery/schema.rb
```diff
@@ -171,6 +171,24 @@ def float name, description: nil, mode: :nullable
           add_field name, :float, description: description, mode: mode
         end
 
+        ##
+        # Adds a numeric number field to the schema. Numeric is a
+        # fixed-precision numeric type with 38 decimal digits, 9 that follow the
+        # decimal point.
```
quartzmo: Can we include the terms "precision" and "scale" here? For example, the [official docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric-type) state:

> with 38 digits of precision and 9 decimal digits of scale. Precision is the number of digits that the number contains. Scale is how many of these digits appear after the decimal point.

For most users, "precision" and "scale" belong to a familiar vocabulary that is well-defined and therefore reassuring to see.

quartzmo: In the past, we have included official disclaimers on Beta / SLA exclusions. Should we be doing that here? One concern, however, is that we might not be told when to remove the statement.

> This feature is a Beta release and is not subject to any SLA. Please see Launch stages for further information about a Beta release.

quartzmo:
In the past, we have included official disclaimers on Beta / SLA exclusions. Should we be doing that here? One concern, however, is that we might not be told when to remove the statement.

> This feature is a Beta release and is not subject to any SLA. Please see Launch stages for further information about a Beta release. 

Some background on the SLA from [/issuetracker.google.com](https://issuetracker.google.com/issues/35906014#comment44):

> In terms of the SLA, if you run queries in a project that is enrolled in 
the alpha but you don't reference any NUMERIC columns or create NUMERIC 
values, then the usual BigQuery SLA applies. The Google Cloud Alpha terms 
apply for any queries involving NUMERIC, however.

ElliottBrossard:
We aren't doing this in clients for the other languages, but I'm not entirely clear on what the policy is or if one exists for documenting launch stages in the client APIs. My expectation would be that someone will consult the BigQuery documentation related to NUMERIC (or other features for that matter) and see the terms that apply.

quartzmo:
@ElliottBrossard @blowmage In my opinion, we can leave out the SLA notice, since we have had issues in the past with such copied docs not being kept in sync (removed) with official docs.

google-cloud-bigquery/lib/google/cloud/bigquery.rb
```diff
@@ -186,7 +186,8 @@ module Cloud
     # | `BOOL`      | `true`/`false` | |
     # | `INT64`     | `Integer`      | |
     # | `FLOAT64`   | `Float`        | |
-    # | `STRING`    | `STRING`       | |
+    # | `NUMERIC`   | `BigDecimal`   | Will be rounded to 9 decimal places |
```
quartzmo: I think I would use the term "scale" here:

Will be rounded if necessary to fit 9 decimal digits of scale.

It might also be good to state the rounding policy.

blowmage: I actually find "precision" and "scale" more confusing than "decimal places". We have very little space here, which is why I simplified the wording.

quartzmo:
I was looking around for some "Standard SQL" policy on rounding to fit scale, and I found [this](http://www.cs.toronto.edu/~nn/csc309-20085/guide/pointbase/docs/html/htmlfiles/dev_datatypesandconversionsFIN.html):

> NOTE: If you exceed the number of digits expected to the left of the decimal point, an error is thrown. If you exceed the number of expected digits to the right of the decimal point, the extra digits are truncated.

I'm not sure how universal this truncation approach is, but it might be worth investigating.

PR #2083, opened 2018-05-09
Storage: Protect against duplicate members on a Policy role

mdayaram:
The API will reject requests to add a role that already exists.  This has caused us to liter our code with this check every time we're adding a new role.  Figured it would be better to just contribute to the library directly.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/policy.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/policy.rb
@@ -189,7 +189,7 @@ module Google
               next if roles[role_name].empty?
               Google::Apis::StorageV1::Policy::Binding.new(
                 role: role_name,
-                members: roles[role_name]
+                members: roles[role_name].uniq
               )
             end
           )
--- a/google-cloud-storage/test/google/cloud/storage/policy_test.rb
+++ b/google-cloud-storage/test/google/cloud/storage/policy_test.rb
@@ -35,6 +35,22 @@ describe Google::Cloud::Storage::Policy do
     role.frozen?.must_equal false
   end
 
+  describe :to_gapi do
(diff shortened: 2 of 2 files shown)


blowmage:
@mdayaram Your branch is very out of date. Can you please rebase your branch on the latest master? This will help CI pass.

blowmage:
Another way to phrase this request is that `Policy` should guard against duplicate role members. If the user adds a duplicate role member the library should avoid making an API call that it can know is going to be bad.

I haven't reached any conclusions on whether this should be the appropriate behavior for google-cloud-strorage or not, but there is one problem I can see that this PR doesn't address. It only covers `Policy#add`, but there are other ways to add members to roles. For example, the [following code example:](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-storage/v1.11.0/google/cloud/storage/policy?method=role-instance)

```ruby

bucket.policy do |p|
  p.role("roles/storage.admin") << "user:owner@example.com"
end
```

If we are to guard against this in `Policy#add` then we should guard against it in all the ways users can specify duplicate role members.

mdayaram:
@blowmage Good point.  What do you think if I update the `to_gapi` method to dedup the list there? That way at any point where this policy will be used to do an API call, it'll get a deduped list when it calls `to_gapi`.

mdayaram:
I went ahead and rebased and implemented the changes I mentioned in the last comment.  Let me know what you think!

mdayaram:
@blowmage any thoughts on the latest updates?

blowmage:
I think this change makes sense, but I would like @quartzmo to weigh in before merging.

If merged, we will need to make this change to all IAM Policy classes in the other gems as well.

blowmage:
Thanks for your contribution!

mdayaram:
thanks!

PR #2099, opened 2018-05-21
Fix dependency to google-api-client

KrauseFx:
Currently `google-cloud-storage` defines a dependency to a pretty specific feature release of `google-api-client`, breaking the use of it for other libraries that need a newer version of the `google-api-client`.

We from the _fastlane_ team are working on fastlane.ci, which uses [fastlane](https://fastlane.tools) to communicate with the Google Play API, and we use Google Cloud Storage to store build artifacts on Google Cloud.

- Right now, we need at least `google-api-client` of `0.21.2` to support the new Google Play app binaries (via https://github.com/fastlane/fastlane/pull/12568)
- We also need to upload build artifacts to Google Cloud Storage, however we can't add both dependencies due to the strict dependency to `google-api-client` on version `~> 0.19.0` ([source](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/9f1d96c022002a6dad7ee107e39aa818003164d4/google-cloud-storage/google-cloud-storage.gemspec#L22))

### DIFF
--- a/google-cloud-storage/google-cloud-storage.gemspec
+++ b/google-cloud-storage/google-cloud-storage.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gem|
   gem.required_ruby_version = ">= 2.0.0"
 
   gem.add_dependency "google-cloud-core", "~> 1.2"
-  gem.add_dependency "google-api-client", "~> 0.19.0"
+  gem.add_dependency "google-api-client", "~> 0.19"
   gem.add_dependency "googleauth", "~> 0.6.2"
   gem.add_dependency "digest-crc", "~> 0.4"
 


blowmage:
We are cautious to move to an unbound dependency because google-api-client occasionally breaks backwards compatibility. We try to keep up to date with the latest release, but we are behind. The current release is 0.23.0.

KrauseFx:
I totally understand, however it blocks us and probably other development teams from being able to use the latest Google APIs, as well as the `google-cloud-storage` gem. What are the plans for this library, and when do we aim to have a 1.0?

KrauseFx:
I also updated the PR body to add more details on why this change is needed

blowmage:
We are discussing this with the owners of the Google API Client gem about this. The discussion is about updating the narrow dependency to the latest release as we've been doing previously vs. loosening the dependency as you've done in this PR.

blowmage:
It looks like we won't be able to loosen the scope of the dependency, as done in this PR. So we will have to update to the latest release instead, which has been submitted in #2129.

Thanks for your patience. We will try to get these released soon.

KrauseFx:
Perfect, thanks for your help on this @blowmage 👍

Issue #2072, opened 2018-05-03
Blank docs pages for lower-level API links


quartzmo:
Blank pages:

- [ ] [Google::Cloud::Datastore::V1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/datastore/v1)
- [ ] [Google::Cloud::ErrorReporting::V1beta1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/errorreporting/v1beta1)
- [ ] [Google::Cloud::Logging::V2](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/logging/v2) 
- [ ] [Google::Cloud::Monitoring::V3](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/monitoring/v3)
- [ ] [Google::Cloud::Pubsub::V1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/pubsub/v1)
- [ ] [Google::Cloud::Spanner::V1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/spanner/v1)
- [ ] [Google::Cloud::Trace::V1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/trace/v1)

Broken links:

- [ ] [DLP V2beta1](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/dlp/v2beta1)

Incorrect content (lower-level content - see #2075):

- [ ] [Google::Cloud::Datastore](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/datastore)
- [ ] [Google::Cloud::Debugger](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud-debugger/v0.32.0/google/cloud/debugger)
- [ ] [Google::Cloud::Firestore](https://googlecloudplatform.github.io/google-cloud-ruby/#/docs/google-cloud/v0.52.0/google/cloud/firestore)
[...]

PR #2100, opened 2018-05-21
Generate tasks v2beta2

crwilcox:


google-cloud-tasks/Gemfile
```diff
@@ -0,0 +1,12 @@
+source "https://rubygems.org"
+
+gemspec
+
+gem "gcloud-jsondoc",
+    git: "https://github.com/GoogleCloudPlatform/google-cloud-ruby.git",
+    branch: "gcloud-jsondoc"
+
```
blowmage: Probably want to add this as well:

```ruby
if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.1")
  # WORKAROUND: builds are failing on Ruby 2.0.
  # We think this is because of a bug in Bundler 1.6.
  # Specify a viable version to allow the build to succeed.
  gem "jwt", "~> 1.5"
end
```

quartzmo:
The commit b1f88c3 "hand edit tests to return struct, not value, from mock" was needed to fix the following unit test errors:

```sh
Google::Gax::RetryError: GaxError Exception occurred in retry method that was not classified as transient, caused by undefined method `execute' for ...
```

For more information about this issue, see googleapis/gax-ruby#125.

google-cloud-tasks/docs/toc.json
```diff
@@ -0,0 +1,27 @@
...
+          "type": "google/cloud/tasks/v2beta2",
+          "patterns": ["\/tasks\/v2beta2", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "CloudTasksClient",
+              "type": "google/cloud/tasks/v2beta2/cloudtasksclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/tasks/v2beta2/datatypes"
```
quartzmo: I believe this `type` value is incorrect. The correct value is `google/cloud/tasks/v2beta2/datatypes`. (The jsondoc output file is `google-cloud-tasks/jsondoc/google/cloud/tasks/v2beta2/datatypes.json`, derived from the `title` in the `toc_config` in the `Rakefile`: `title: "Google::Cloud::Tasks::V2beta2::DataTypes"`.

google-cloud-tasks/lib/google/cloud/tasks/v2beta2/doc/overview.rb
```diff
@@ -0,0 +1,55 @@
...
+    # ### Next Steps
+    # - Read the [Cloud Tasks API Product documentation][Product Documentation]
+    #   to learn more about the product and see How-to Guides.
+    # - View this [repository's main README](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/README.md)
+    #   to see the full list of Cloud APIs that we cover.
+    #
+    # [Product Documentation]: https://cloud.google.com/cloudtasks
+    #
+    #
+    module Tasks
```
quartzmo: @jbolinger I believe this is another instance of the "blank pages" issue in #2072. Is there some docs content available for the `V2beta2` module? I have confirmed that the jsondoc output for `google/cloud/tasks/v2beta2` will result in a blank page, since in that file: `"description":""`.

jbolinger: Looks like the same issue to me.

@landrito - it looks like you generated Dataproc and it does include the doc comments that are missing here. For example, I think this is what we want (@quartzmo does that look right to you?): 

https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/e52957dd157b0d25d6badb3b5a3a244ff69e7e6f/google-cloud-dataproc/lib/google/cloud/dataproc/v1/doc/google/cloud/dataproc/v1/clusters.rb#L18-L31)

Why are we not seeing the same thing here? Was that manually added or was it changed in the generator?

google-cloud-tasks/docs/toc.json
crwilcox: Is there any way to be sure of this file? This, as best as I can tell, is always handcrafted?

google-cloud-tasks/lib/google/cloud/tasks/v2beta2/doc/overview.rb
quartzmo: @jbolinger I think a table like that is the right content, but tbh I'm not 100% sure.

google-cloud-tasks/docs/toc.json
quartzmo: I'm not sure I understand your question, but the `type` paths should correspond to files in the `jsondoc` directory (after running `rake jsondoc`).

jbolinger: @crwilcox yes, this is something that is always added manually. The best way to be sure is to look through the docs to see if it's being rendered correctly. Once it's merged to master it's easy to simply check the doc site. Prior to that you can preview it locally, but it's a little more work. @quartzmo sent me some good instruction for doing it a few weeks ago.  I can probably dig up for you if you think it would help - I've only done it a few times myself.

Issue #2101, opened 2018-05-22
Bigtable : Protobuf `UpdateAppProfileMetadata` class is missing and Longrunning operation needed Google::Gax::Operation wrapper


jiren:
- Protobuf class missing for `UpdateAppProfileMetadata` in [bigtable_instance_admin_pb.rb](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-bigtable/lib/google/bigtable/admin/v2/bigtable_instance_admin_pb.rb)
Ref: [bigtable_instance_admin.proto](https://github.com/googleapis/googleapis/blob/46890c7e15d18423b26572c0ce90905f5dfefe74/google/bigtable/admin/v2/bigtable_instance_admin.proto#L472)

- Some of the `Google::Longrunning::Operation` jobs are not wrapped into `Google::Gax::Operation`
1. `partial_update_instance`
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/bdfe9bcc71f91bb9c9a6e83c61ec523a7b57644b/google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb#L569

2. `update_app_profile `
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/bdfe9bcc71f91bb9c9a6e83c61ec523a7b57644b/google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client.rb#L1024

3. `snapshot_table`
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/bdfe9bcc71f91bb9c9a6e83c61ec523a7b57644b/google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/bigtable_table_admin_client.rb#L763

I think GAPIC code need to regenerate to fix missing class and longrunning operation wrapper.

jbolinger:
Yes, this is correct. I regenerated locally and it looks like all of those updates were included. 

There were also a few other minor changes that it pulled in. I'll follow up with a PR shortly.

jbolinger:
Fixed by #2102.

PR #2102, opened 2018-05-22
Regenerate Bigtable to address 2101

jbolinger:
This brings the Bigtable generated client up to date with the currently published protos, including items mentioned in #2101

We may need to check on those configuration changes though. I assume they are correct, but I don't really know.

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/doc/google/bigtable/admin/v2/bigtable_instance_admin.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/doc/google/bigtable/admin/v2/bigtable_instance_admin.rb
@@ -311,6 +311,14 @@ module Google
         #   @return [true, false]
         #     If true, ignore safety checks when deleting the app profile.
         class DeleteAppProfileRequest; end
+
+        # This is a private alpha release of Cloud Bigtable replication. This feature
+        # is not currently available to most Cloud Bigtable customers. This feature
+        # might be changed in backward-incompatible ways and is not recommended for
+        # production use. It is not subject to any SLA or deprecation policy.
+        #
+        # The metadata for the Operation returned by UpdateAppProfile.
+        class UpdateAppProfileMetadata; end
       end
     end
   end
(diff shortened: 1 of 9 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/admin/v2/doc/google/bigtable/admin/v2/bigtable_instance_admin.rb
```diff
@@ -311,6 +311,14 @@ class UpdateAppProfileRequest; end
         #   @return [true, false]
         #     If true, ignore safety checks when deleting the app profile.
         class DeleteAppProfileRequest; end
+
+        # This is a private alpha release of Cloud Bigtable replication. This feature
```
blowmage: Does this "private alpha" language mean that these changes should not be released to the general public?

jbolinger: Not sure. I found these comments strange since they are on the public protos, but it doesn't look like we've published this gem yet anyway. There were already comments like this on other classes so I thought it would make sense to get the repos in sync at least.

I don't have any visibility into the release planning, but I suspect we wouldn't want to. Let's check with @theacodes to be sure.

blowmage: The Bigtable team is working on the Ruby client located in this repo. We have given code reviews but don't really have much insight into the design decisions being made.  Perhaps @sduskis can weigh in on whether these updated files are wanted?

sduskis: The comments are out of date.  The "App Profile" feature is now beta.  We need to update those comments in the proto files, and then regenerate GAPIC.

sduskis:
There are new protos in the googleapis/googleapis repo with the updated comments.  Does that mean that we need to regenerate again?

blowmage:
@jbolinger can you regenerate these files?

jbolinger:
@blowmage sure, regenerated!

sduskis:
I took a quick look at the changed files, and things look good from what I've seen.

PR #2108, opened 2018-05-25
Pub/Sub: Subscriber Unary Ack/ModAck

blowmage:
Send Subscriber's Ack/ModAck messages using normal RCP API calls instead of using the GRPC stream. This will relieve some GRPC stream congestion. An empty message is pushed to the GRPC stream every 30 seconds to keep the GRPC stream alive if there are messages being processed.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
@@ -129,14 +129,14 @@ module Google
                   next
                 end
 
-                time_since_first_publish = Time.now - @batch_created_at
-                if time_since_first_publish > @interval
+                time_since_batch_creation = Time.now - @batch_created_at
+                if time_since_batch_creation > @interval
                   # interval met, publish the batch...
                   push_batch_request!
                   @cond.wait
                 else
                   # still waiting for the interval to publish the batch...
-                  @cond.wait(@interval - time_since_first_publish)
+                  @cond.wait(@interval - time_since_batch_creation)
                 end
               end
             end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
@@ -1,4 +1,4 @@
-# Copyright 2017 Google LLC
+# Copyright 2018 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,9 +22,9 @@ module Google
       class Subscriber
         ##
         # @private
-        # # AsyncPusher
+        # # AsyncUnaryPusher
         #
-        class AsyncPusher
+        class AsyncUnaryPusher
           include MonitorMixin
 
           attr_reader :batch
@@ -108,9 +108,7 @@ module Google
               @background_thread.kill if @background_thread
             end
 
-            return nil if @batch.nil?
-
-            @batch.request
+            push_batch_request!
           end
 
           def started?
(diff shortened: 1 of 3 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -53,6 +53,15 @@ def initialize subscriber
             @push_thread_pool = Concurrent::FixedThreadPool.new \
               subscriber.push_threads
 
+            @stream_keepalive_task = Concurrent::TimerTask.new(
+              execution_interval: 30
+            ) do
+              # push empty request every 30 seconds to keep stream alive
+              unless inventory.empty?
```
quartzmo: @blowmage Why this conditional? Why not always send and keep open?

blowmage: I don't know. The directions we got were to keep the stream alive with empty requests as long as there were outstanding messages on the stream.

quartzmo:
@jba PTAL

blowmage:
cc @jba

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
```diff
@@ -0,0 +1,250 @@
...
+
+          def run_background
+            synchronize do
+              until @stopped
+                if @batch.nil?
+                  @cond.wait
+                  next
+                end
+
+                time_since_first_publish = Time.now - @batch_created_at
```
jba: Can you pick a word other than publish? I was confused that this was related to publishing a message to pubsub.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
```diff
@@ -0,0 +1,250 @@
...
+
+                    @batch = Batch.new max_bytes: @max_bytes
+                    @batch.delay deadline, ack_id
+                  end
+                end
+
+                @batch_created_at ||= Time.now
+                @background_thread ||= Thread.new { run_background }
+
+                push_batch_request! if @batch.ready?
```
jba: Could `push_batch_request!` be called concurrently on this line  and line 135 in `run_background`?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
jba: We don't want to keep the stream open artificially if we're not using it.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
blowmage: Changed to `time_since_batch_creation`.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/async_unary_pusher.rb
blowmage: I don't know if that would make much of a difference. All the time consuming work is done asynchronously within the method. We want to make the call to `push_batch_request!` synchronously so that it will use the same mutex lock as the code that called it, to ensure that the @batch object hasn't been modified before the method call completes.

jba: Sorry, I wasn't suggesting that we make concurrent calls. I was concerned that concurrent calls would result in a race condition. I just want to make sure you considered the possibility.

blowmage: Ah, I understand now. The mutex lock prevents concurrent calls to `push_batch_request!` from happening. This is provided by the `synchronous` block that this code is running in. I believe this code and the background thread provide the proper protection.

PR #2110, opened 2018-05-29
Release google-cloud-tasks 0.1.0

crwilcox:


README.md
```diff
@@ -798,6 +798,36 @@ $ gem install google-cloud-monitoring
  end
 ```
```
quartzmo: I think this section should be added just above "Cloud Translation API". (I know some of the services are out of alphabetical order, but starting from the top, I think that's the correct spot...)

Also, a "nav" link to this section needs to be added to the top of the file.

google-cloud/google-cloud.gemspec
```diff
@@ -37,6 +37,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "google-cloud-spanner", "~> 1.3"
   gem.add_dependency "google-cloud-speech", "~> 0.29"
   gem.add_dependency "google-cloud-storage", "~> 1.10"
+  gem.add_dependency "google-cloud", "~> 0.1.0"
```
quartzmo: Replace "google-cloud" with "google-cloud-tasks" ?

quartzmo:
@crwilcox Have you followed all of the steps in [Adding a new gem to meta-packages](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/RELEASING.md#adding-a-new-gem-to-meta-packages)?

crwilcox:
@quartzmo , I am following https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/RELEASING.md as it was specified in the playbook for releasing ruby.

As that is not in the list, and I am currently on step 14, I have not done that. It seems to me perhaps this doc needs tweaking to include that as step 10?

crwilcox:
I proposed some edits to the doc to reflect suggestions made here. https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2111

google-cloud/docs/toc.json
```diff
@@ -712,6 +712,28 @@
...
+          "type": "google/cloud/tasks/v2beta2",
+          "patterns": ["\/tasks\/v2beta2", "google\/api", "google\/protobuf"],
+          "nav": [
+            {
+              "title": "CloudTasksClient",
+              "type": "google/cloud/tasks/v2beta2/cloudtasksclient"
+            },
+            {
+              "title": "Data Types",
+              "type": "google/tasks/v2beta2/datatypes"
```
quartzmo: As I mentioned in [this comment on the original PR](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2100#discussion_r190072465), I don't think this path will match the output of the `rake jsondoc` task.

crwilcox: Why is that? @blowmage seemed to think it would and I have no real understanding of what this is supposed to be. Is there a regression in the generator or jsondoc?

quartzmo: The correct value is `google/cloud/tasks/v2beta2/datatypes`. (The jsondoc output file is `google-cloud-tasks/jsondoc/google/cloud/tasks/v2beta2/datatypes.json`, derived from the title in the `toc_config` in the Rakefile: `title: "Google::Cloud::Tasks::V2beta2::DataTypes"`.

Here is the [DataTypes page title in the first PR](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/f7e83f88b355cbab187b347c6d10d86f265323f2/google-cloud-tasks/Rakefile#L78). The value of `type` needs to match.

The generator output for these values has not matched for some time now, perhaps always. It needs to be fixed.

blowmage:
Will merge when CI is green.

google-cloud/docs/toc.json
jbolinger: @crwilcox @quartzmo AFAIK the generator never did this and it's something I was hoping to address once DPE assumed all the publishing roles. How does this relate to the other documentation changes that you want to make though? Should we fix this `toc.json` issue or can we do more to move that along?

quartzmo: This configuration (and problem) will go away if we change the documentation to standard YARD output.

PR #2115, opened 2018-05-30
Update RELEASING.md adding notes on version.rb

crwilcox:
version.rb files aren't necessarily at the root. Add language to help make that clear.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -44,7 +44,7 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
 
 1. Edit the gem's `CHANGELOG.md`. Using your notes from the previous step, write bullet-point lists of the major and minor changes. You can also add examples, fixes, thank yous, and anything else helpful or relevant. See google-cloud-node [v0.18.0](https://github.com/GoogleCloudPlatform/google-cloud-node/releases/tag/v0.18.0) for an example with all the bells and whistles.
 
-1. Edit the gem's `version.rb` file, if present, or the `version` setting in its `.gemspec` file, changing the value to your new version number. Note: the `.gemspec` file may reference the `version.rb` file. For instance, `google-cloud`'s `.gemspec` references a `version.rb` file at `../lib/google/cloud/version`
+1. Edit the gem's `version.rb` file, if present, or the `version` setting in its `.gemspec` file, changing the value to your new version number. Note: The `.gemspec` file may reference the `version.rb` file. For instance, `google-cloud`'s `.gemspec` references a `version.rb` file at `../lib/google/cloud/version`.
 
 1. Edit (or add if new) the gem's entry in `docs/manifest.json`, adding your new version number to the head of the list, and moving `"master"` to be just below it.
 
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -44,7 +44,7 @@ After all [pull requests](https://github.com/GoogleCloudPlatform/google-cloud-ru
 
 1. Edit the gem's `CHANGELOG.md`. Using your notes from the previous step, write bullet-point lists of the major and minor changes. You can also add examples, fixes, thank yous, and anything else helpful or relevant. See google-cloud-node [v0.18.0](https://github.com/GoogleCloudPlatform/google-cloud-node/releases/tag/v0.18.0) for an example with all the bells and whistles.
 
-1. Edit the gem's `version.rb` file, if present, or the `version` setting in its `.gemspec` file, changing the value to your new version number.
+1. Edit the gem's `version.rb` file, if present, or the `version` setting in its `.gemspec` file, changing the value to your new version number. Note: the `.gemspec` file may reference the `version.rb` file. For instance, `google-cloud`'s `.gemspec` references a `version.rb` file at `../lib/google/cloud/version`
```
quartzmo: Needs capitalization after colon and a period at the end:

Note: The `.gemspec` file may reference the `version.rb` file. For instance, `google-cloud`'s `.gemspec` references a `version.rb` file at `../lib/google/cloud/version`.

quartzmo:
@crwilcox I'm on the fence about whether we should add this. The use and location of a `version.rb` file is something I've taken for granted for a long time in Ruby. I perceive it as a quite familiar convention (if not universal), which I think is supported by the default behavior of Bundler's scaffolding for new gems:

```sh
$ bundle gem foo
Creating gem 'foo'...
...
      create  foo/Gemfile
      create  foo/lib/foo.rb
      create  foo/lib/foo/version.rb
      create  foo/foo.gemspec
      create  foo/Rakefile
      create  foo/README.md
      create  foo/bin/console
      create  foo/bin/setup
      create  foo/.gitignore
      create  foo/LICENSE.txt
      create  foo/CODE_OF_CONDUCT.md
```

@dazuma @blowmage Any opinion on this?

blowmage:
> version.rb files aren't necessarily at the root

The `version.rb` file is never at the root directory for the gem, or even a direct child of the `lib` directory. It is always nested in a file path that matches the gem name. I don't think the guide implies it is at the root.

crwilcox:
If it isn't at the root ever I can reword this. I can tell you the way it is written I would expect it to be in the base directory.

blowmage:
> Any opinion on this?

File paths are not required more than once, so it would be problematic to not have a unique file path for each gem's `version.rb` file. The `version.rb` file path used by each gem in this repo is a common Ruby idiom. I guess the question is whether we need to be explicit about every Ruby idiom in the RELEASING guide.

crwilcox:
I am alright not changing this if you find this detracts from the purpose of the guide. I am not a ruby developer and have however been tasked with managing package releases for google-cloud-ruby, so for me this is helpful. I would imagine it also may be helpful for anyone not familiar with ruby packaging also.

blowmage:
I usually defer to @quartzmo on decisions about documentation. My opinion is not as trustworthy as his judgement. :)

quartzmo:
> Edit the gem's `.gemspec` file, changing the value to your new version number.

At this point I'd personally recommend we abandon the PR, because it's still not accurate: In most of our packages, the `.gemspec` file does *not* need to be edited, just the `version.rb` file. (Which I believe is the whole point of the `version.rb` file, to reduce churn on `.gemspec` file.)

But I'm also open to another update if you're up for it. I can imagine a version that improves upon the original without adding many more words.

PR #2122, opened 2018-06-06
Enable frozen_string_literal magic comment in pubsub

ianks:
When memory profiling an app using this gem, I noticed that a large number
(hundreds of thousands )of "@grpc" and "ASCII-8BIT" strings allocated. This
would happen with each message, whic

This patch addresses the issue and future issues by enforcing a frozen string
literals for all files in this library, and adding a rubocop rule to enforce
this pattern in new files.


ianks:
For a reference, after receiving 3000 messages, here is the string allocation report we get from `memory_profiler`

```
     99059  ""
     45000  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/google-cloud-pubsub-0.30.2/lib/google/cloud/pubsub/message.rb:67
     30000  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/google-cloud-pubsub-0.30.2/lib/google/cloud/pubsub/message.rb:68
     21000  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/grpc-1.12.0-x86_64-linux/src/ruby/lib/grpc/generic/rpc_desc.rb:47
      3000  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/google-cloud-pubsub-0.30.2/lib/google/cloud/pubsub/received_message.rb:57
        42  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/google-cloud-pubsub-0.30.2/lib/google/cloud/pubsub/subscriber/async_pusher.rb:164
        16  /home/ianks/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/google-cloud-pubsub-0.30.2/lib/google/cloud/pubsub/subscriber/stream.rb:314
```

blowmage:
Great catch! This was an oversight and should be fixed.

My understanding is the `frozen_string_literal: true` comment works on Ruby 2.3 and newer. Since we are supporting Ruby 2.0 and newer I have created #2123 to address the strings reported in this PR on all supported Ruby versions. If there are more strings that should not be created please report them there.

That said, I can see an argument for also adding the `frozen_string_literal: true` comment to google-cloud-* gems in addition to not creating more strings than is necessary. At a minimum, we should know where we are mutating strings so we can avoid those mutations in preparation of Ruby 3.0.

In the past I've run the following, only to have it error in Bundler:

`RUBYOPT="--enable-frozen-string-literal" bundle exec rake test`

quartzmo:
@ianks Can you please rebase your branch on master in order to get this [commit](https://github.com/GoogleCloudPlatform/google-cloud-ruby/commit/9819e686f149016b8b6ebad0ab712c5e7c594bc4) which should fix the broken CI build for Ruby 2.0 for this PR? Thanks!

ianks:
@quartzmo Done!

blowmage:
@quartzmo I'm considering putting this PR on hold until #2130 is unblocked and merged. Pairing the frozen string comments with dropping unsupported rubies, and updating rubocop. Thoughts

quartzmo:
@blowmage Agreed. I'll create a new issue for adding frozen string comments to **all** packages.

@ianks Are you interested in expanding this PR to add the comment to all packages, or would you rather close it and leave that to us? In either case, we really appreciate your push to make this happen.

PR #2123, opened 2018-06-06
Pub/Sub: Reduce String GC

blowmage:
Reduce the number of throwaway strings created by the Pub/Sub implementation.

[refs #2122]


blowmage:
> This looks good, so I'm approving, but how does this PR relate to #2122? Does it replace it?

It provides a different solution to the same problem.

ianks:
@quartzmo I think this PR will benefit user's who are not on Ruby 2.3 yet.

PR #2126, opened 2018-06-08
Correct the argument type for the schema in Schema::dump

jeremywadsack:
I had a copy-paste error in the documents for `Schema::dump` that listed the `schema` argument as `IO, String`. This corrects that.


quartzmo:
@jeremywadsack Thanks for the follow-up! Have you been using Schema `dump` and `load`?

jeremywadsack:
We haven't put it into production yet, but we'll do that soon. We're using it our test suites though and it's super helpful. Although given what the methods do I think we're using it much more in test that we will in production.

quartzmo:
CI is a bit backed up but I'll merge this PR soon.

PR #2127, opened 2018-06-08
Add Session labels to Project#client and #batch_client

### LINKED ISSUE #1817: Updated Spanner features
The recent PR #1804 updated the Spanner GAPIC/Protobuf code and a new API call to list sessions. Should these be supported in the high-level Ruby API, even though we don't expose Session currently?

The following attributes were also added to the Session resource:

* `Session#labels`
* `Session#create_time`
* `Session#approximate_last_use_time`

quartzmo:
Also:

* Convert labels argument from `Google::Protobuf::Map`.
* Rescue `GRPC` error in `Results.from_enum`.
* Update expected errors in acceptance tests.
* Add error case unit tests for `Session#reload!` and `#keepalive!`

[closes #1817]

### DIFF
--- a/google-cloud-spanner/acceptance/spanner/client/edge_test.rb
+++ b/google-cloud-spanner/acceptance/spanner/client/edge_test.rb
@@ -26,13 +26,13 @@ describe "Spanner Client", :edge, :spanner do
   end
 
   it "reads with invalid table fails" do
-    assert_raises GRPC::NotFound do
+    assert_raises Google::Cloud::NotFoundError do
       db.read "invalid_table", [:id, :int]
     end
   end
 
   it "reads with invalid column fails" do
-    assert_raises GRPC::NotFound do
+    assert_raises Google::Cloud::NotFoundError do
       db.read table_name, [:id, :invalid]
     end
   end
@@ -59,19 +59,19 @@ describe "Spanner Client", :edge, :spanner do
   end
 
   it "queries to a non-existing table fails" do
-    assert_raises GRPC::InvalidArgument do
+    assert_raises Google::Cloud::InvalidArgumentError do
       db.execute "SELECT id, name FROM invalid_table"
     end
   end
 
   it "queries to a non-existing column fails" do
-    assert_raises GRPC::InvalidArgument do
+    assert_raises Google::Cloud::InvalidArgumentError do
       db.execute "SELECT id, name FROM #{table_name}"
     end
   end
 
   it "queries with bad SQL fails" do
-    assert_raises GRPC::InvalidArgument do
+    assert_raises Google::Cloud::InvalidArgumentError do
       db.execute "SELECT Apples AND Oranges"
     end
   end
--- a/google-cloud-spanner/lib/google/cloud/spanner/project.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/project.rb
@@ -219,6 +219,8 @@ module Google
         def create_instance instance_id, name: nil, config: nil, nodes: nil,
                             labels: nil
           config = config.path if config.respond_to? :path
+          # Convert from possible Google::Protobuf::Map
+          labels = Hash[labels.map { |k, v| [String(k), String(v)] }] if labels
           grpc = service.create_instance \
             instance_id, name: name, config: config, nodes: nodes,
                          labels: labels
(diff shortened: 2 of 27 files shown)


google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -88,6 +88,26 @@ def path
...
+        # * Label keys must be between 1 and 63 characters long and must conform
+        #   to the following regular expression: `[a-z]([-a-z0-9]*[a-z0-9])?`.
+        # * Label values must be between 0 and 63 characters long and must
+        #   conform to the regular expression `([a-z]([-a-z0-9]*[a-z0-9])?)?`.
+        # * No more than 64 labels can be associated with a given resource.
+        #
+        # @return [Hash{String=>String}] The label keys and values in a hash.
+        #
+        def labels
+          @grpc.labels
```
blowmage: This will return a `Google::Protobuf::Map` object. It is probably better to call `@grpc.labels.to_h` instead.

quartzmo: Good catch. Does this need to change in `Instance#labels` too?

blowmage: Yes, especially if those labels are being passed into a Protobuf constructor, because the constructor won't accept a Map, it only takes a Hash.

blowmage: Although if we do convert the Instance Map to a Hash, we may need to provide lifecycle management for the Hash, since the Hash could be updated and `Instance#save` called.

blowmage: Another option is to keep the values using Map, but always cast Map to Hash when passing to the API.

```ruby
if labels
  labels_hash = Hash[labels.map { |key, value| [String(key), String(value)] }]
  session = Google::Spanner::V1::Session.new labels: labels_hash 
end
```

google-cloud-spanner/acceptance/spanner/client/edge_test.rb
```diff
@@ -26,13 +26,13 @@
   end
 
   it "reads with invalid table fails" do
-    assert_raises GRPC::NotFound do
+    assert_raises Google::Cloud::NotFoundError do
```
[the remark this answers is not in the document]
blowmage: Hah. That's awful.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -219,6 +219,8 @@ def instance instance_id
         def create_instance instance_id, name: nil, config: nil, nodes: nil,
                             labels: nil
           config = config.path if config.respond_to? :path
+          # Convert from possible Google::Protobuf::Map
+          labels = Hash[labels.map { |k, v| [String(k), String(v)] }] if labels
```
blowmage: :+1:

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -435,6 +437,21 @@ def create_database instance_id, database_id, statements: []
         #     becomes available. The default is `true`.
         #   * `:threads` (Integer) The number of threads in the thread pool. The
         #     default is twice the number of available CPUs.
+        # @param [Hash] session_labels The labels to be applied to all sessions
```
blowmage: The only bone I have to pick is with the choice of `session_labels` instead of `labels`. Can you help me understand why `session_labels` is a better choice?

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -206,6 +206,8 @@ def self.from_enum enum, service
             results.instance_variable_set :@enum,     enum
             results.instance_variable_set :@service,  service
           end
+        rescue GRPC::BadStatus => e
```
blowmage: Very excited for this fix.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
quartzmo: Yes. The `labels` name is perfect for `Project#create_instance` because the method clearly identifies the type that will receive the labels. However, `Project#client` and `#batch_client` do not in any way identify that `Session` will receive the labels. Although it is mentioned in the docs, the `session_labels` naming is important to clarify this for users.

quartzmo: Fwiw, `sessionLabels` is also used in [SpannerOptions.java](https://github.com/vkedia/google-cloud-java/blob/0bec2b43b749ca630160760ce23562d0a1f3b377/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java#L161-L176).

blowmage: We use `pool` even though it is not specifically prefixed with `session_`. I am just a tad uncomfortable with the inconsistency.

quartzmo: Changed to `labels`.

PR #2131, opened 2018-06-13
Regenerate Speech for v1p1beta1

jbolinger:
Regenerated v1 and v1p1beta1. The manual test needs to be updated, so I didn't try to run the smoke tests yet.

Also note that it looks like the `credentials.rb` change isn't included. I think we have another release coming this week if you want to include it in this update - your call. I thought we had already released that change but it looks like I was wrong.


blowmage:
@frankyn Are you okay waiting for the upcoming change to generated `credentials.rb` file before a release?

google-cloud-speech/README.md
```diff
@@ -24,7 +24,7 @@ $ gem install google-cloud-speech
 ```rb
 require "google/cloud/speech"
 
-speech_client = Google::Cloud::Speech.new
+speech_client = Google::Cloud::Speech::V1p1beta1.new
```
blowmage: Is the intention that the beta API version is the one used in the README? The default version is still showing `:v1` on `Google::Cloud::Speech.new`.

google-cloud-speech/Rakefile
```diff
@@ -121,6 +121,24 @@ task :jsondoc => :yard do
 
   toc_config = {
     documents: [
+      {
```
blowmage: Do you want `V1p1beta1` to be listed above `V1` when the default API still looks to be `V1`?

jbolinger: Hmm, good question. I don't really know what message we're trying to send with the v1p1beta1. Should probably be consistent with your other comment about the `README.md`.

I'll update to prefer V1 unless someone tells us differently.

frankyn:
Let's wait so we don't have to update the migration guide when credentials are released. I'm not sure I'm in the right place to define release timing. However, I have one issue I'm working on resolving so waiting a bit would help.

frankyn:
Sent an email out on release requirements. Pending response.

google-cloud-speech/test/google/cloud/speech/v1/speech_client_test.rb
```diff
@@ -25,7 +25,7 @@
 class CustomTestError < StandardError; end
 
 # Mock for the GRPC::ClientStub class.
-class MockGrpcClientStub
+class MockGrpcClientV1Stub
```
blowmage: @jbolinger Check out these changes I had to make to get the tests passing. The test class created for the V1p1beta test was stepping on the V1 class, leading to test failures. I named all these things differently and that helped get the tests passing.

blowmage: i was seeing failures expecting a V1p1beta1 object but getting a V1 object.

jbolinger: Strange, I can't seem to reproduce it. I changed those back let's see what the CI does and I can revert it if needed.

frankyn:
Where are we with this PR? Is it ready?

blowmage:
@jbolinger Can you give us an update? When can you update these files again to get the `credentials.rb` changes included?

jbolinger:
@blowmage I can manually update it if you want to get this out now. The version change hasn't been (correctly) released in the generator yet. Is it ready to go otherwise?

blowmage:
We would like to get this PR merged as soon as is reasonable. We were waiting on updated generated files to get the `credentials.rb` change.

jbolinger:
Sorry I didn't realize this was the only open item - didn't mean to be the bottleneck.

@blowmage could you take a look? Is this the way you were thinking of aliasing the `credentials.rb`?

While doing this I noticed that our unit tests have some cross talk. I had to rename the `MockSpeechCredentials` in one of the versions to get things working. I wonder if there's a better way to run the tests? Could it be related to the errors that you mentioned that I couldn't reproduce? No reason to block this PR on it, but looks like we've got another issue to solve.

google-cloud-speech/lib/google/cloud/speech/credentials.rb
```diff
@@ -12,19 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "googleauth"
+require "google/cloud/speech/v1/credentials"
```
blowmage: This gem isn't 1.0 yet, and the next release is definitely a breaking change. Do we really need to maintain this file and constant? For GA gems yes, but should we do that here?

@quartzmo what do you think?

blowmage: The constant being `Google::Cloud::Speech::Credentials`.

quartzmo: If `google/cloud/speech/credentials.rb` can be removed from the public API, this would be a good time to do it, wouldn't it?

blowmage: I think it can be, yes.

I just checked all the code examples and I don't see any examples of how to use the environment variables or the Credentials classes. But perhaps that should be solved in a separate change?

quartzmo: Hmmm, yes, that would probably be needed in other gems as well, correct?

blowmage: Yeah, probably a good idea. The docs site has the general authentication guide, but the GAPIC generated gems should probably list what environment variables are used somewhere in the gem documentation.

PR #2133, opened 2018-06-14
Build: Remove yard-doctest autoloading

blowmage:
This was adding the plugin multiple times, and would print the following warning multiple times when a rake task was run:

`[error]: Error loading plugin 'yard-doctest'`

### DIFF
--- a/google-cloud-bigquery/Rakefile
+++ b/google-cloud-bigquery/Rakefile
@@ -134,7 +134,6 @@ end
 
 desc "Run yard-doctest example tests."
 task doctest: :yard do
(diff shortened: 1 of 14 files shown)


blowmage:
I miss the `build` label...

blowmage:
If you still see the warning message after this change delete your `~/.yard.config` file.

blowmage:
@jbolinger Is this something you would want to remove from the GAPIC project files?

jbolinger:
@blowmage Sure, I don't think those were all set manually anyway. That line doesn't appear to be in the generator.

blowmage:
Okay, that's good to hear. Thanks!

PR #2146, opened 2018-06-21
[Speech] Update documented quickstarts.

frankyn:
* I think this may need to move into the autogen versus being updated this way. 
* @jbolinger is this something you can help with?

### DIFF
--- a/google-cloud-speech/lib/google/cloud/speech/v1.rb
+++ b/google-cloud-speech/lib/google/cloud/speech/v1.rb
@@ -40,7 +40,7 @@ module Google
     # ```rb
     # require "google/cloud/speech"
     #
-    # speech_client = Google::Cloud::Speech(version: :v1)
+    # speech_client = Google::Cloud::Speech.new(version: :v1)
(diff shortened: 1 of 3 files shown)


google-cloud-speech/lib/google/cloud/speech/v1.rb
```diff
@@ -38,9 +38,9 @@ module Cloud
     # ### Preview
     # #### SpeechClient
     # ```rb
-    # require "google/cloud/speech/v1"
+    # require "google/cloud/speech"
     #
-    # speech_client = Google::Cloud::Speech::V1.new
+    # speech_client = Google::Cloud::Speech(version: :v1)
```
blowmage: You still need to call `#new`:

```ruby
speech_client = Google::Cloud::Speech.new(version: :v1)
```

jbolinger:
@frankyn no problem, just created https://github.com/googleapis/gapic-generator/issues/2099. Feel free to add any more details. I'll try and take a look at it later today or tomorrow.

PR #2148, opened 2018-06-21
Add schema_update_options and time_partitioning to Jobs

### LINKED ISSUE #2147: Add schemaUpdateOptions to BigQuery LoadJob
Add support for the `schemaUpdateOptions` array to BigQuery `LoadJob` and `LoadJob::Updater`.

quartzmo:
- [x] Add `schema_update_options` to `LoadJob`
- [x] Add `time_partitioning` to `LoadJob`
- [x] Add `time_partitioning` to `QueryJob`

[closes #2147, closes #2152]

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
@@ -170,18 +170,26 @@ describe Google::Cloud::Bigquery::Dataset, :bigquery do
 
   it "imports data from a local file and creates a new table with specified schema in a block with load_job" do
     job_id = "test_job_#{SecureRandom.urlsafe_base64(21)}" # client-generated
-    job = dataset.load_job "local_file_table", local_file, job_id: job_id do |schema|
-      schema.integer   "id",    description: "id description",    mode: :required
-      schema.string    "breed", description: "breed description", mode: :required
-      schema.string    "name",  description: "name description",  mode: :required
-      schema.timestamp "dob",   description: "dob description",   mode: :required
-      schema.schema_update_options = schema_update_options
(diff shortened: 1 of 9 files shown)


google-cloud-bigquery/acceptance/bigquery/dataset_test.rb
```diff
@@ -168,16 +170,26 @@
...
-      schema.string    "breed", description: "breed description", mode: :required
-      schema.string    "name",  description: "name description",  mode: :required
-      schema.timestamp "dob",   description: "dob description",   mode: :required
+    job = dataset.load_job "local_file_table", local_file, job_id: job_id do |job|
+      job.schema.integer   "id",    description: "id description",    mode: :required
+      job.schema.string    "breed", description: "breed description", mode: :required
+      job.schema.string    "name",  description: "name description",  mode: :required
+      job.schema.timestamp "dob",   description: "dob description",   mode: :required
+      job.schema_update_options = schema_update_options
+      job.time_partitioning_type = "DAY"
```
blowmage: What do you think about creating a method that takes these 4 values as arguments and represents them as an object instead of exposing 5 prefixed values?

quartzmo: I'd be happy to do that if you think it's better. Our past design has been to flatten, but I agree that the number of methods makes it seem reasonable to have a value object.

blowmage: I don't think we've been 100% consistent. In some other gems like Storage we would have a method that sets all the values, and an object that represents the collected values. I think it makes sense here, as all these values are dependent on each other.

quartzmo:
@blowmage This PR is complete and ready for review, PTAL.

PR #2160, opened 2018-06-27
[Bigtable] Table consistency and app profile APIs

jiren:
- App Profile create/update/delete/list APIs with unit mock tests and acceptance test
- Table replication consistancy check APIs
- Unit test using mocks and acceptance test

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/app_profile.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/app_profile.rb
@@ -184,7 +184,8 @@ module Google
         # Delete app profile.
         #
         # @param ignore_warnings [Boolean]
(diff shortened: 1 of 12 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/app_profile.rb
```diff
@@ -0,0 +1,371 @@
...
+            name,
+            ignore_warnings: ignore_warnings
+          )
+          true
+        end
+
+        # Update app profile.
+        #
+        # @param ignore_warnings [Boolean]
+        #   If true, ignore safety checks when updating the app profile.
```
blowmage: Probably a good idea to state that the default value is `false`.

PR #2166, opened 2018-06-29
Firestore: remove unneeded base64 encoding applied to BLOB fields

aleh:
It looks like fields of type 'bytes' can be set using binary streams. Before the change however the contents of a stream would be base64-encoded in `google-cloud-firestore/lib/google/cloud/firestore/convert.rb` which would lead to the actual value of the corresponding 'bytes' field to be base64 encoded as well (in addition to the base64 encoding
the Web Console does, for example).

### DIFF
--- a/google-cloud-firestore/test/google/cloud/firestore/convert/value_to_raw_test.rb
+++ b/google-cloud-firestore/test/google/cloud/firestore/convert/value_to_raw_test.rb
@@ -78,7 +78,7 @@ describe Google::Cloud::Firestore::Convert, :value_to_raw, :mock_firestore do
 	contents.write("c\0ntents")
 	contents.close_write()
 	
-    raw.must_equal contents.string
+    raw.string.must_equal contents.string
   end
 
   it "converts a reference value" do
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/test/google/cloud/firestore/convert/value_to_raw_test.rb
```diff
@@ -69,11 +69,11 @@
   end
 
   it "converts a bytes value" do
-    value = Google::Firestore::V1beta1::Value.new(bytes_value: Base64.encode64("contents"))
+    value = Google::Firestore::V1beta1::Value.new(bytes_value: "c\0ntents")
 
     raw = Google::Cloud::Firestore::Convert.value_to_raw value, firestore
-    raw.must_be_kind_of StringIO
-    raw.read.must_equal "contents"
+	raw.must_be_kind_of StringIO
```
blowmage: Please match indenting of the previous line.

aleh:
@blowmage Replaced tabs with spaces.

blowmage:
@schmidt-sebastian Can you weigh on on this. When the client was first implemented both the [RPC `Value`](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1beta1#google.firestore.v1beta1.Value#google.firestore.v1beta1.Value.FIELDS.bytes.google.firestore.v1beta1.Value.bytes_value) and the [REST `Value`](https://cloud.google.com/firestore/docs/reference/rest/v1beta1/Value#FIELDS.bytes_value) documentation said the value should be Base64 encoded. But now the RPC `Value` documentation is not showing that. I think it makes sense the RPC endpoints would not need Base64 encoding. Can you confirm?

schmidt-sebastian:
We use Protobuf-encoding in our GRPC based API. You should not need to Base64-encode any values. All encoding should be taken care of by Protobuf.

blowmage:
Yeah, that's what we've done for other GRPC libraries as well. Thanks!

PR #2170, opened 2018-07-03
Fix Acceptance test for northamerica-northeast1-a zone and added Europe zones

jiren:
- `northamerica-northeast1-a` listed in https://cloud.google.com/bigtable/docs/locations but not supported yet.
- Added Europe zones `europe-west1` and `europe-west2`.


quartzmo:
I've pulled this branch and am running the change locally.

PR #2172, opened 2018-07-03
Tasks: set google-gax dependency to v1.3

TheRoyalTnetennba:
Fix for issue [https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/2171](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/2171)


quartzmo:
Would it be possible to add a test covering the error in #2171?

PR #2175, opened 2018-07-03
Propagate Logging list arguments

### LINKED ISSUE #2163: logging reading next does not work
https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-logging/lib/google/cloud/logging/entry/list.rb#L73-L76

Because none of the instance variables are passed into the from_grpc method, the api responds with an error any time you call next with any parameters:

```
logging = Google::Cloud::Logging.new project_id: project_id
entries = logging.entries(filter:filter_thats_not_null)
entries.next
Google::Cloud::InvalidArgumentError: 3:page_token doesn't match arguments from the request
```

it seems like the from_grpc in the list.next should be called with all the params as https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-logging/lib/google/cloud/logging/project.rb#L152

blowmage:
This change fixes a bug in Logging's list classes. The arguments needed for subsequent list calls would not propagate to the second list object, so the API call for the third list object would raise an error.

[fixes #2163]

### DIFF
--- a/google-cloud-logging/lib/google/cloud/logging/metric/list.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/metric/list.rb
@@ -72,7 +72,7 @@ module Google
             return nil unless next?
             ensure_service!
             grpc = @service.list_metrics token: token, max: @max
-            self.class.from_grpc grpc, @service
+            self.class.from_grpc grpc, @service, @max
           end
 
           ##
--- a/google-cloud-logging/lib/google/cloud/logging/resource_descriptor/list.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/resource_descriptor/list.rb
@@ -75,7 +75,7 @@ module Google
             list_grpc = @service.list_resource_descriptors(
               token: token, max: @max
             )
-            self.class.from_grpc list_grpc, @service
+            self.class.from_grpc list_grpc, @service, @max
           end
 
           ##
--- a/google-cloud-logging/lib/google/cloud/logging/sink/list.rb
+++ b/google-cloud-logging/lib/google/cloud/logging/sink/list.rb
@@ -71,7 +71,7 @@ module Google
             return nil unless next?
             ensure_service!
             list_grpc = @service.list_sinks token: token, max: @max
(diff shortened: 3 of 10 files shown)


google-cloud-logging/lib/google/cloud/logging/metric/list.rb
```diff
@@ -72,7 +72,7 @@ def next
             return nil unless next?
             ensure_service!
             grpc = @service.list_metrics token: token, max: @max
-            self.class.from_grpc grpc, @service
+            self.class.from_grpc grpc, @service, @max
```
quartzmo: Want to change the `max` param to keyword arg to match `Entry::List` and `Log::List`?

google-cloud-logging/lib/google/cloud/logging/resource_descriptor/list.rb
```diff
@@ -75,7 +75,7 @@ def next
             list_grpc = @service.list_resource_descriptors(
               token: token, max: @max
             )
-            self.class.from_grpc list_grpc, @service
+            self.class.from_grpc list_grpc, @service, @max
```
quartzmo: Want to change the `max` param to keyword arg to match `Entry::List` and `Log::List`?

google-cloud-logging/lib/google/cloud/logging/sink/list.rb
```diff
@@ -71,7 +71,7 @@ def next
             return nil unless next?
             ensure_service!
             list_grpc = @service.list_sinks token: token, max: @max
-            self.class.from_grpc list_grpc, @service
+            self.class.from_grpc list_grpc, @service, @max
```
quartzmo: Want to change the `max` param to keyword arg to match `Entry::List` and `Log::List`?

google-cloud-logging/lib/google/cloud/logging/metric/list.rb
blowmage: Possibly, but I don't want to make a change like that in this PR.

PR #2178, opened 2018-07-05
Update google-gax dependency

blowmage:
Update the google-gax dependency for gems with GAPIC files that call `Google::Gax::Grpc.create_stub` with the `interceptors` named argument. This argument was added in google-gax version 1.3, so the dependency needs to be updated to ensure that the right version is used.


blowmage:
This is for changes that have already hit master. Future changes, like those currently pending for trace in #2157, should be accounted for in their own PRs.

blowmage:
Only `tasks` and `vision` have releases that are missing these updated dependencies. The other gems do not yet have releases that include the updated GAPIC files.

PR #2180, opened 2018-07-06
Initial gapic gen for kms

TheRoyalTnetennba:


google-cloud-kms/.ruby-version
```diff
@@ -1 +1 @@
-2.4.2
```
blowmage: Files like this should not be committed to the repo. Developers should ignore files that are specific to the developer environment in the developer's' global git ignore configuration, and not a `.gitignore` file in the repo.

blowmage: Oops, I was looking at a commit and not the PR. This file was removed in a later commit. Thanks!

PR #2188, opened 2018-07-13
Support DocumentSnapshot objects as cursors

### LINKED ISSUE #1956: Firestore Use DocumentSnapshot as Cursors
Allow DocumentSnapshot objects to be passed into `Query#start_at`, `Query#start_after`, `Query#end_before`, `Query#start_at`.

blowmage:
Allow a DocumentSnapshot object as argument to Query cursor methods.

* Ensure consistency between cursor methods and order.
* Add argument validation and errors to cursor methods.
* Add precondition checks to where and order when cursors are present in the query.

[closes #1956]

### DIFF
--- a/google-cloud-firestore/lib/google/cloud/firestore/query.rb
+++ b/google-cloud-firestore/lib/google/cloud/firestore/query.rb
@@ -213,6 +213,11 @@ module Google
         #   end
         #
         def where field, operator, value
+          if query_has_cursors?
+            raise "cannot call where after calling " \
+                  "start_at, start_after, end_before, or end_at"
+          end
+
           new_query = @query.dup
           new_query ||= StructuredQuery.new
 
@@ -274,6 +279,11 @@ module Google
         #   end
         #
         def order field, direction = :asc
+          if query_has_cursors?
+            raise "cannot call order after calling " \
+                  "start_at, start_after, end_before, or end_at"
+          end
+
           new_query = @query.dup
           new_query ||= StructuredQuery.new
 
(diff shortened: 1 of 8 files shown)


blowmage:
@schmidt-sebastian Can you look at this?

quartzmo:
I think the docs need some additional updates. Here's an example for `#start_at` which can be copied to the other 3 methods. I would also add examples showing the `DocumentSnapshot` args (not shown here).

```ruby
        ##
        # Creates and returns a new query that starts at the document that matches
        # the specified set of field values or document snapshot. The result set will 
        # include this document.
        #
        # The starting position is relative to the order of the query. The 
        # provided values (or document snapshot) must contain all of the 
        # fields provided to `order`. 
        #
        # If the current query already has specified `start_at` or
        # `start_after`, this will overwrite it.
        #
        # @param [Object, Array<Object>] values The set of field values or
        #   the document snapshot to start the query at.
```

Is the following sentence correct? It seems to me that the `document_id` is only used to identify the document snapshot to the service, but then the values are extracted from the document snapshot and matched to the fields in `orderBy`.

```ruby
        # ... If a {DocumentSnapshot}
        # object is provided the `document_id` value will be used and should be
        # paired with a call to `#order` using {Client#document_id}.
```

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -361,9 +361,11 @@ def limit num
         # If the current query already has specified `start_at` or
         # `start_after`, this will overwrite it.
         #
-        # The values provided here are for the field paths provides to `order`.
+        # The values provided here are for the field paths provided to `order`.
         # Values provided to `start_at` without an associated field path
-        # provided to `order` will result in an error.
+        # provided to `order` will result in an error. If a {DocumentSnapshot}
+        # object is provided the `document_id` value will be used and should be
```
schmidt-sebastian: We should not need to expose the fact that we use the `document_id`. If a DocumentSnapshot is provided, we will do this automatically without enforcing this on the user.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -389,7 +391,13 @@ def start_at *values
           new_query = @query.dup
           new_query ||= StructuredQuery.new
 
-          values = values.flatten.map { |value| Convert.raw_to_value value }
+          values = values.flatten.map do |value|
+            if value.respond_to? :document_id
```
schmidt-sebastian: I don't know much about Ruby - but this seems like it might break if a user-specified object responds to `document_id`?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -636,13 +668,26 @@ def default_filter
...
+            last_direction = order_by.direction
           end
+
+          new_query.order_by << StructuredQuery::Order.new(
+            field: StructuredQuery::FieldReference.new(
+              field_path: "__name__".freeze
+            ),
+            direction: last_direction
+          )
+
```
schmidt-sebastian: Please take a look how this is done in the Node or Java client. You also need to add an order by constraint if there are any inequalities. 

If you change:
`query.where('foo', '<=', 'bar')`
to
`query.where('foo', '<=', 'bar').orderBy('__name__')`
the query becomes invalid. Instead it needs to be:
`query.where('foo', '<=', 'bar').orderBy('foo').orderBy('__name__')`

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Okay.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Ruby doesn't have user-specific objects, we instead use `Hash`. If we did have user-specific objects, we would perform the conversion to `Hash` before the check for `document_id`.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: Removed.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: This hash been removed in the latest code change. PTAL.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
blowmage: This PR has been updated and should implement the logic as you have explained it.

blowmage:
@schmidt-sebastian This has been updated. Can you take another look at this?

blowmage:
@schmidt-sebastian Do you have time to look at this? As soon as this PR is accepted and merged we can open the PR for Firestore Watch.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -636,13 +890,123 @@ def default_filter
...
+          values_to_cursor values, query
+        end
+
+        def ensure_inequality_field_paths_in_order_by! query
+          inequality_paths = inequality_filter_field_paths query
+          orig_order = order_by_field_paths query
+
+          inequality_paths.reverse.each do |field_path|
+            next if orig_order.include? field_path
```
schmidt-sebastian: Note that there should only ever be one `inequality_paths` and hence the direction doesn't matter here. You could even error if there is more than one.

blowmage: Agreed that it could raise. I don't see this being raised in some of the other client implementations though, so we would rely on the API to give an error for multiple inequalities.

I implemented it this way so it could be more declarative and less imperative.

PR #2201, opened 2018-07-23
Data read apis, read and mutation apis acceptance tests

jiren:
- Read APIs with unit and acceptance test cases
- Check and mutate APIs
- Drop row range API
- Sample row keys API
- Reduce admin API call frequency in acceptance tests
- Added env flag to run cluster create and updated acceptance test case.


jiren:
#2201

google-cloud-bigtable/acceptance/bigtable/column_family_test.rb
```diff
@@ -28,6 +28,10 @@
     end
   end
 
+  after(:all) do
```
blowmage: Minitest doesn't have `after(:all)`, so this will run after each test in this file is run.

jiren: Got it.

PR #2207, opened 2018-07-26
Add Firestore conformance tests

quartzmo:
In addition to adding dynamic conformance testing based on generated tests from `test-definition_pb.rb`, this PR makes the following changes to the library:

- `Query#select` is no longer additive, only the argument to the final invocation will be kept. [BREAKING CHANGE]
- Validate field paths arguments and raise `ArgumentError` for duplicate field paths and if one field is the prefix of another.
- Operators unknown to Firestore (such as `!=`) will raise `ArgumentError`.


blowmage:
These are my commits, and I consent to them being used in this pull request.

quartzmo:
Thanks @jbolinger!

PR #2232, opened 2018-08-03
Specify YARD Markdown provider

blowmage:
All gems use Markdown syntax for documentation, and we have configured YARD to generate documentation accordingly. However, we also use some Markdown options, such as fenced code blocks, that are not standard Markdown and not supported in all Markdown providers. Set the Markdown provider to redcarpet to ensure that documentation generates properly.

* Specify redcarpet in the .yardopts file.
* Add redcarpet as a development dependency. This will ensure that users who generate documentation from the gem will produce properly formatted documentation.
* Add YARD as a development dependency to gems that were missing it.


blowmage:
If accepted, I assume this something that should be changed to the new gem generated files, yes?

blowmage:
We have sort of lucked out that rubydoc.info generates markdown using a provider that supports fenced code blocks, but the default markdown provider that the yard CLI uses does not.

blowmage:
I fixed a typo and the build is green now.

blowmage:
Great! Thanks!

FWIW, after we have moved all documentation back to YARD HTML files we will no longer need any of the `jsondoc` tasks or `toc_config` data in the Rakefile, or the `gcloud-jsondoc` gem in the Gemfile. So more changes are coming.

PR #2241, opened 2018-08-06
[Bigtable] Get table admin api changes 

jiren:
- Set default `skip_lookup` to true. This will reduce admin api calls for data operations
- Updated test cases for skip lookup
- Updated smart retry read. Now it if will fail only if successive 3 failure and updated tests
- Doc update and fixes

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
@@ -448,10 +448,10 @@ module Google
         #   * `:SCHEMA_VIEW` - Only populates `name` and fields related to the table's schema
         #   * `:REPLICATION_VIEW` - Only populates `name` and fields related to the table's replication state.
         #   * `:FULL` - Populates all fields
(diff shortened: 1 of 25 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
```diff
@@ -451,7 +451,7 @@ def tables
         # @param skip_lookup [Boolean] Create table object without verifying
         #   that the table resource exists.
         #   Calls made on this object will raise errors if the table.
-        #   does not exist. Default is `false`. Optional.
+        #   does not exist. Default value is `true`. Optional.
```
blowmage: If you want to flip the default behavior you might consider renaming the argument to something like `perform_lookup` so it can stay `nil` in the default case.

jiren: Yes this a good idea.

jiren:
I ran the acceptance tests and working fine. Please merge it.

PR #2264, opened 2018-08-07
update auth urls

TheRoyalTnetennba:
per https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/2254


TheRoyalTnetennba:
Thanks @blowmage

blowmage:
I just realized that the service account I was using specifies the URL to use, and so my test didn't use the URL changed in this PR. Would you like me to re-run my test, or are we confident enough that this change is okay?

TheRoyalTnetennba:
I'd feel better retesting.

blowmage:
I retested using a service account JSON file with those URLs removed. Everything passed again.

PR #2278, opened 2018-08-11
Refactor Subscriber Stream shutdown

blowmage:
Change the shutdown process to no longer process pending callbacks. When the number of pending callbacks was too high the thread pool would hang. This new behavior also seems to be more expected.

Change the wait behavior to only wait on pending change requests. Limit waiting on pending change requests, so the Subscriber does not block indefinitely.

We can make this change because we are using the Unary pusher and not the Stream pusher, because the Unary pusher does not require that we keep the stream open to send change requests.


blowmage:
I don't understand why Kokoro is failing. It doesn't look related to this PR though. Perhaps because #2276 is not yet merged?

@TheRoyalTnetennba should we be concerned about the Kokoro failures?

TheRoyalTnetennba:
Yeah, it just can't find the build config's because they haven't merged yet.

PR #2279, opened 2018-08-11
Fix bug in AsyncUnaryPusher

blowmage:
The `modify_ack_deadline` requests being created by AsyncUnaryPusher were malformed.


blowmage:
We have lots of changes in master since the last release, but I would like to get this fixed ASAP. I am thinking about cherry-picking this change on the latest release tag and release from there, and then yanking the latest release. Thoughts?

PR #2280, opened 2018-08-13
Fix Storage GZIP test

blowmage:
Update test to set GZIP encoding to BINARY. This looks to be causing a failure on the new Kokoro builds.

[fixes #2276]


blowmage:
I don't understand why the build is failing. I can't get it to fail locally.

PR #2286, opened 2018-08-14
Update CSV mime type

blowmage:
The mime-types 3.2.0 gem has changed the type for .csv files. This is needed to get the builds passing again.


quartzmo:
Is this the correct [changelog entry](https://github.com/mime-types/ruby-mime-types/blob/master/History.md#32--2018-08-12) for mime-types 3.2.0? I don't see the change listed there.

blowmage:
Yes, that is the correct changelog entry. See the issue the changelog references, it describes the change using the CSV mime types.

https://github.com/mime-types/ruby-mime-types/pull/132

> One such example are "text/comma-separated-values" and "text/csv" MIME types. The former is obsolete and unregistered, while the latter is not obsolete and registered, but mime-types would still rank "text/comma-separated-values" higher because it just looks at the alphabetic order.

PR #2300, opened 2018-08-15
Fix middleware documentation warning

blowmage:
Correct "unknown parameter name" warning when generating documentation.

[refs #2299]


blowmage:
It looks like JSONDOC is wanting the param name to be in a format that YARD raises a warning about. So until JSONDOC is retired these changes are at an impasse.

blowmage:
Now that JSONDOC has been removed, this change can be considered. Please take a look.

PR #2338, opened 2018-08-23
Updated Documentation Content

blowmage:
This PR makes several changes to the gem's packaged content in order to provide more value both on the [GCP GitHub site](
https://googlecloudplatform.github.io/google-cloud-ruby) and on [vanilla YARD sites](https://www.rubydoc.info).

It defaults the main content to a new file, `CONTENT.md`, which contains the "guide" content that was previously located at `Google::Cloud::Datastore`. This solves a couple of problems. YARD isn't configurable to link to a Ruby constant for the main page, unlike RDoc. YARD defaults to README.md, which we shouldn't use YARD links (`{Dataset#entity}`) in since it is primarily intended for use on GitHub. So we created a new file that can handle YARD links and provide the guide content.

The JSONDOC site has some content that is missing from the new YARD-based documentation. Content such as contributing, fan, troubleshooting, etc. Most of these files are in the root-directory, but again intended for viewing on GitHub. Copying these files to the gem directory allows support of YARD links, so we can link to/from this content in the documentation.

You can see an example of the documentation for these changes [here](http://blowmage.com/google-cloud-ruby/docs/google-cloud-datastore/master/), and you can compare and contrast this to the documentation without these changes [here](http://blowmage.com/google-cloud-ruby/docs/google-cloud-datastore/latest/).

Thoughts? Should I make this change to the other veneer gems?


quartzmo:
Do we really need CODE_OF_CONDUCT for each gem?

blowmage:
I think "need" is subjective. I added it because CONTRIBUTING links to it. So to make the link work we need the file.

I don't think we are worse off for having this be part of the gem's documentation, but this is exactly the type of thing I think needs a discussion. Does this PR add too many files? How much is too much?

dazuma:
It's probably a good idea to have CODE_OF_CONDUCT present for each gem. Especially because I still think we should be looking towards splitting the repo at some point in the future (after we think most of our cross-library refactorings have stabilized).

I'm also fine with the number of files and changes.

One open question I have is whether the markdown files should live in a subdirectory.

blowmage:
> One open question I have is whether the markdown files should live in a subdirectory.

If we want these new files in a directory I propose `guides`. The directory `doc` is the default directory for YARD writing the HTML files, and I always get confused by that dir and JSONDOC's `docs` dir, so I would avoid `docs` because of that.

TheRoyalTnetennba:
Leaving the markdown files where they are makes any future transition to one gem per repo easier.

royrobertrobert1974:
I need to talk to someone

dazuma:
@royrobertrobert1974 Consider joining the GCP slack at https://gcp-slack.appspot.com/ . There's a "ruby" channel if you have Ruby-specific questions.

google-cloud-datastore/AUTHENTICATION.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Google Cloud Platform environments
+
+While running on Google Cloud Platform environments such as Google Compute
+Engine, Google App Engine and Google Kubernetes Engine, no extra work is needed.
+The **Project ID** and **Credentials** and are discovered automatically. Code
+should be written as if already authenticated. Just be sure when you [set up the
+GCE instance][gce-how-to], you add the correct scopes for the APIs you want to
+access. For example:
+
```
quartzmo: Since we are effectively "inlining" previously shared content here, what about customizing this to the gem and just showing the scopes for Datastore?

google-cloud-datastore/CONTRIBUTING.md
```diff
@@ -0,0 +1,209 @@
...
+
+To run the Datastore acceptance tests:
+
+``` sh
+$ cd google-cloud-datastore/
+$ bundle exec rake acceptance[\\{my-project-id},\\{/path/to/keyfile.json}]
+```
+
+Or, if you prefer you can store the values in the `GCLOUD_TEST_PROJECT` and
+`GCLOUD_TEST_KEYFILE` environment variables:
```
quartzmo: Replace all `GCLOUD_` with `GOOGLE_CLOUD_`?

google-cloud-datastore/OVERVIEW.md
```diff
@@ -0,0 +1,579 @@
...
+
+```ruby
+require "google/cloud/datastore"
+
+datastore = Google::Cloud::Datastore.new
+
+task = datastore.find "Task", "sampleTask"
+```
+
+Optionally, {Google::Cloud::Datastore::Dataset#find} can be given a key object:
```
quartzmo: All of these links to code constructs will definitely not work when this file is viewed in github's markdown rendering, but what about in the YARD-generated HTML?

google-cloud-datastore/TROUBLESHOOTING.md
```diff
@@ -0,0 +1,31 @@
+# Troubleshooting
+
+## Where can I get more help?
+
```
quartzmo: Should we add GCP slack to this doc?

google-cloud-datastore/AUTHENTICATION.md
blowmage: Yes.

google-cloud-datastore/CONTRIBUTING.md
blowmage: The test env vars all begin with `GCLOUD_TEST_`.

google-cloud-datastore/OVERVIEW.md
blowmage: The links will work in the YARD-generated HTML, which is a big part of why I'm proposing this.

google-cloud-datastore/TROUBLESHOOTING.md
blowmage: Yes.

google-cloud-datastore/CONTRIBUTING.md
quartzmo: Oh yeah...

quartzmo:
> One open question I have is whether the markdown files should live in a subdirectory.

I don't think so, but we should probably try to avoid having a lot of small docs files, and consider merging `FAQ.md` and `TROUBLESHOOTING.md` into `OVERVIEW.md`. If we do decide to have a subdirectory for these files, I think `/docs` is the conventional name.

blowmage:
All, this PR has been updated to add files for all the hand-written gems. We didn't change any code in the generated files or gems. PTAL.

PR #2341, opened 2018-08-23
Update partial download test

blowmage:
This test fails occasionally on CI. We aren't sure why.
We have a custom message to help dignose, but we have seen some
download verification errors when it gets proc'd.
Set the download to verify: :none to avoid the verification errors.
Update the custom message to show a bit more information.
Also set file path to .txt so the mime types are handled correctly.

### DIFF
--- a/google-cloud-storage/acceptance/storage/file_test.rb
+++ b/google-cloud-storage/acceptance/storage/file_test.rb
@@ -355,24 +355,25 @@ describe Google::Cloud::Storage::File, :storage do
 
   it "should upload and partially download text" do
     inmemory = StringIO.new "Hello world!"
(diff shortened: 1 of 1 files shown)


google-cloud-storage/acceptance/storage/file_test.rb
```diff
@@ -355,24 +355,25 @@
 
   it "should upload and partially download text" do
     inmemory = StringIO.new "Hello world!"
-    uploaded = bucket.create_file inmemory, "uploaded/with/inmemory.png"
+    uploaded = bucket.create_file inmemory, "uploaded/with/inmemory.txt"
```
blowmage: I wonder if the reason we've seen this test fail occasionally is because the mime-type was being set incorrectly due to the file extension.

PR #2349, opened 2018-08-24
Add Array FieldValue types

blowmage:
Add array_union and array_delete FieldValue configuration.
This allow document arrays to be added to or removed from.

### DIFF
--- a/google-cloud-firestore/lib/google/cloud/firestore/client.rb
+++ b/google-cloud-firestore/lib/google/cloud/firestore/client.rb
@@ -302,6 +302,56 @@ module Google
           FieldValue.server_time
         end
 
+        ##
+        # Creates a sentinel value to indicate the union of the given values
+        # with an array.
+        #
+        # @param [Object] values The values to add from the array. Required.
+        #
+        # @return [FieldValue] The array union field value object.
+        #
+        # @example
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a document reference
+        #   nyc_ref = firestore.doc "cities/NYC"
+        #
+        #   array_union = firestore.field_array_union 1, 2, 3
+        #
+        #   nyc_ref.update({ name: "New York City",
+        #                    lucky_numbers: array_union })
+        #
+        def field_array_union *values
+          FieldValue.array_union(*values)
+        end
+
+        ##
+        # Creates a sentinel value to indicate the removal of the given values
+        # with an array.
+        #
+        # @param [Object] values The values to remove from the array. Required.
+        #
+        # @return [FieldValue] The array delete field value object.
+        #
+        # @example
+        #   require "google/cloud/firestore"
+        #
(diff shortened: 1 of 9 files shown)


blowmage:
@schmidt-sebastian I don't have permissions to add you as a reviewer to this PR. Can you add yourself?

google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -302,6 +302,56 @@ def field_server_time
           FieldValue.server_time
         end
 
+        ##
+        # Creates a sentinel value to indicate the union of the given values
+        # with an array.
+        #
+        # @param [Object] values The values to add from the array. Required.
```
schmidt-sebastian: s/to add from/to add to/

google-cloud-firestore/lib/google/cloud/firestore/field_value.rb
```diff
@@ -62,6 +63,33 @@ def type
...
+        #
+        #   # Get a document reference
+        #   nyc_ref = firestore.doc "cities/NYC"
+        #
+        #   array_union = Google::Cloud::Firestore::FieldValue.array_union(
+        #     1, 2, 3
+        #   )
+        #   array_union.type #=> :array_union
+        #   array_union.values #=> [1, 2, 3]
+        #
```
schmidt-sebastian: The members of this class should not need to be part of the public interface. Exposing this will lock us into an API, while not actually improving the usability of the client.

google-cloud-firestore/lib/google/cloud/firestore/field_value.rb
```diff
@@ -107,6 +135,72 @@ def self.delete
...
+        #   nyc_ref.update({ name: "New York City",
+        #                    lucky_numbers: array_union })
+        #
+        def self.array_union *values
+          # We can flatten the values because arrays don't support sub-arrays
+          values.flatten!
+          raise ArgumentError, "values must be provided" if values.nil?
```
schmidt-sebastian: Please note that you need to validate that there are no invalid field values in `values`. The first version that we shipped in Node silently dropped field value sentinels if there were contained inside field value sentinels. `FieldValue.arrayUnion(FieldValue.arrayUnion('foo'))` created an empty array. Instead, we should throw an error.

google-cloud-firestore/lib/google/cloud/firestore/client.rb
blowmage: Fixed.

google-cloud-firestore/lib/google/cloud/firestore/field_value.rb
blowmage: Marked private.

google-cloud-firestore/lib/google/cloud/firestore/field_value.rb
blowmage: Added this check, and tests to ensure that an error is raised.

PR #2351, opened 2018-08-24
Add array-contains operator

blowmage:
Add `array-contains` as an operator to the `Query#where` method.

### DIFF
--- a/google-cloud-firestore/test/google/cloud/firestore/query/where/array_contains_test.rb
+++ b/google-cloud-firestore/test/google/cloud/firestore/query/where/array_contains_test.rb
@@ -17,21 +17,6 @@ require "helper"
 describe Google::Cloud::Firestore::Query, :where, :array_contains, :mock_firestore do
   let(:query) { Google::Cloud::Firestore::Query.start nil, "#{firestore.path}/documents", firestore }
 
-  it "using array_contains" do
-    expected_query = Google::Firestore::V1beta1::StructuredQuery.new(
-      where: Google::Firestore::V1beta1::StructuredQuery::Filter.new(
-        field_filter: Google::Firestore::V1beta1::StructuredQuery::FieldFilter.new(
-          field: Google::Firestore::V1beta1::StructuredQuery::FieldReference.new(field_path: "foo"),
-          op: :ARRAY_CONTAINS,
-          value: Google::Firestore::V1beta1::Value.new(integer_value: 42)
-        )
-      )
-    )
-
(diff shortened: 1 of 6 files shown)


blowmage:
@schmidt-sebastian I don't have permissions to add you as a reviewer to this PR. Can you add yourself?

google-cloud-firestore/test/google/cloud/firestore/query/where/array_contains_test.rb
```diff
@@ -0,0 +1,153 @@
...
+      where: Google::Firestore::V1beta1::StructuredQuery::Filter.new(
+        field_filter: Google::Firestore::V1beta1::StructuredQuery::FieldFilter.new(
+          field: Google::Firestore::V1beta1::StructuredQuery::FieldReference.new(field_path: "foo"),
+          op: :ARRAY_CONTAINS,
+          value: Google::Firestore::V1beta1::Value.new(integer_value: 42)
+        )
+      )
+    )
+
+    generated_query = query.where(:foo, :contains, 42).query
```
schmidt-sebastian: Where does `:contains` come from? Is that a constant that the SDK exposes? Reason I am asking is that I would like to only expose the name "array-contains". If under the hood, you also accept "array_contains", "contains" or "include", that's fine. I would just prefer if this was not publicly documented, otherwise users will ask us to the same in all SDKs.

blowmage: Add operation mapping is defined [here](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2351/files#diff-58dd15514cde954c379bfef85f858eb3R903). We can leave the other mappings undocumented, but I would like to document `array_contains` because ruby doesn't allow `-` in symbols. So we can say `where("field", "array-contains", 42)` or `where(:field, :array_contains, 42)`. Is that okay?

schmidt-sebastian: Yes, sounds good to me.

blowmage: Great. I removed `contains?` and `contains`. But kept `include?` and `include` because a very common Ruby idiom is [`Enumerable#include?`](https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-include-3F).

I also added the missing documentation. The only operators that are documented are `array-contains` and `array_contains`.

Issue #2406, opened 2018-09-05
Pubsub: Client config is not propagated to the publisher


apoloval:
In Pubsub initialization, it's possible to provide a client config that will override some of the settings at the GAX level. This is how the client config is propagated across the different layers:
* In `Google::Cloud::Pubsub.new` ([code](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub.rb#L563-L565)), the `client_config` param is received. There, it's used to invoke the constructor of `Google::Cloud::Pubsub::Service`. 
* In `Google::Cloud::Pubsub::Service.initialize` ([code](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb#L40)), the `client_config` param is stored in a object attribute. 
* In `Google::Cloud::Pubsub::Service.subscriber`, the service initializes and memoizes the subscriber. The client config is used in the initialization (that's correct):
```ruby
        def subscriber
          return mocked_subscriber if mocked_subscriber
          @subscriber ||= begin
            V1::SubscriberClient.new(
              credentials: channel,
              timeout: timeout,
              client_config: client_config,
              lib_name: "gccl",
              lib_version: Google::Cloud::Pubsub::VERSION
            )
          end
        end
```
* Nevertheless, in `Google::Cloud::Pubsub::Service.publisher`, the `client_config` param is not specified:
```ruby
        def publisher
          return mocked_publisher if mocked_publisher
          @publisher ||= begin
            V1::PublisherClient.new(
              credentials: channel,
              timeout: timeout,
              lib_name: "gccl",
              lib_version: Google::Cloud::Pubsub::VERSION
            )
          end
        end
```

This is causing the **publisher to not receive the client config**. As result, it's not possible to configure client settings for publishing (in particular, backoff/retry settings).

apoloval:
It seems @quartzmo forgot to include the param in commit 580b8feb16e980f50327cba20d88a325712e689e.

blowmage:
Might be a good idea to double-check the other gems to make sure this isn’t happening anywhere else.

quartzmo:
[...]

PR #2407, opened 2018-09-05
Add missing client_config in Pubsub publisher instantiation

apoloval:
Possible fix of #2406.

When GAX was introduced in Pubsub, the possibility of passing a `client_config` value from `Google::Cloud::Pubsub.new` was given. The point is that such config should be propagated to the subscriber and publisher objects. That's right for the subscriber (lines 61 to 72). But it seems like publisher was not also updated in commit 580b8feb16e980f50327cba20d88a325712e689e. 

The fix is as simple as passing the `client_config` value while calling to `PublisherClient.new`.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -78,6 +78,7 @@ module Google
             V1::PublisherClient.new(
               credentials: channel,
               timeout: timeout,
+              client_config: client_config,
               lib_name: "gccl",
               lib_version: Google::Cloud::Pubsub::VERSION
             )


quartzmo:
I noticed that for emulator usage, the `Pubsub::Service` initialization is [also missing the `client_config` argument](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub.rb#L574-L581).

However, in google-cloud-datastore, the emulator initialization of `Datastore::Service` [does receive `client_config`](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-datastore/lib/google/cloud/datastore.rb#L666). Although I'm not certain that it is actually useful, should we also add that in this PR? @blowmage?

apoloval:
> I noticed that for emulator usage, the Pubsub::Service initialization is also missing the client_config argument.

Yes, this is something that also bit me: why the timeout property was not there in `service.timeout` when emulator was used (both `timeout` and `client_config` are not passed in that case). Actually I read the `new` method many times until I noticed that early return and assumed we don't care about emulator. 

In the sake of consistency and the principle of least astonishment, I'd say client config and timeout values should be propagated when using the emulator. Do you want me to fix that in an additional commit?

quartzmo:
Yes, that sounds good to me. Thank you~

quartzmo:
Now we have the question, can the `Datastore::Service` for emulator use use `timeout` as well? @blowmage?

blowmage:
Yes, I think the emulator configuration should be consistent across gems.

google-cloud-datastore/lib/google/cloud/datastore.rb
```diff
@@ -660,26 +660,16 @@ def self.new project_id: nil, credentials: nil, scope: nil, timeout: nil,
...
         emulator_host ||= configure.emulator_host
         if emulator_host
-          return Datastore::Dataset.new(
-            Datastore::Service.new(
-              project_id, :this_channel_is_insecure,
-              host: emulator_host, timeout: timeout,
-              client_config: client_config
-            )
-          )
+          return dataset_for_emulator(project_id, emulator_host, timeout,
```
blowmage: I understand why these methods were extracted, because of the Rubocop failures, but I don't want to add even more methods to the metaclass. I would rather see some creative formatting:

```ruby
return Datastore::Dataset.new(Datastore::Service.new(
  project_id, :this_channel_is_insecure,
  host: emulator_host, timeout: timeout,
  client_config: client_config
))
```

Or add exclusion comments around the method to disable the Rubocop rule:

```ruby
# rubocop:disable Metrics/MethodLength

def self.new project_id: nil, credentials: nil, scope: nil, timeout: nil,
             client_config: nil, emulator_host: nil, project: nil,
             keyfile: nil
  #...
end

# rubocop:enable Metrics/MethodLength
```

apoloval: Totally. I wasn't sure you'd be OK with disabling the rubocop metric. Fixed in cce5a7944f5a1ec626ce445591a7a84bed10b56d.

apoloval:
Thank you for your kindly and fast code review 👏 👏 👏

PR #2441, opened 2018-09-13
Fix circular require warnings

blowmage:
Place a conditional on the google/cloud require.
This will stop the circular require warnings when gems are
loaded by the google-cloud-core auto-load mechanism.

### DIFF
--- a/google-cloud-bigquery/lib/google-cloud-bigquery.rb
+++ b/google-cloud-bigquery/lib/google-cloud-bigquery.rb
@@ -19,7 +19,7 @@
 
 
 gem "google-cloud-core"
-require "google/cloud"
+require "google/cloud" unless defined? Google::Cloud.new
 require "google/cloud/config"
 require "googleauth"
 
--- a/google-cloud-bigtable/lib/google-cloud-bigtable.rb
+++ b/google-cloud-bigtable/lib/google-cloud-bigtable.rb
@@ -21,7 +21,7 @@ gem "google-cloud-core"
 
 require "googleauth"
 require "grpc"
-require "google/cloud"
+require "google/cloud" unless defined? Google::Cloud.new
 require "google/cloud/config"
 
 module Google
--- a/google-cloud-datastore/lib/google-cloud-datastore.rb
+++ b/google-cloud-datastore/lib/google-cloud-datastore.rb
@@ -20,7 +20,7 @@
 
 
 gem "google-cloud-core"
-require "google/cloud"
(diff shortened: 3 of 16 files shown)


blowmage:
Warnings are currently disabled on all the tests, partially because there are many warnings in dependencies that we can't address. But also partially because of these circular require warnings. Resolving the circular require warnings will get us one step closer to being able to enable warnings on our test runs.

blowmage:
@dazuma I would really like your input on this. WDYT?

dazuma:
Does this syntax work? It doesn't seem to work for me:

```
% irb
irb(main):001:0> module Foo
irb(main):002:1> def foo
irb(main):003:2> end
irb(main):004:1> end
=> :foo
irb(main):005:0> defined? Foo.foo
=> nil
```

dazuma:
Oh, never mind, I forgot `def self.foo`...

blowmage:
Yeah, it works. It is safer to check for `.new` than it is to check for the module to exist, because gemspec loads the version and that creates the namespace. Even checking for `Google::Cloud::Core` wasn't as reliable as `Google::Cloud.new`.

dazuma:
Yeah, this seems fine, and it seems to work in my tests. I wonder if it would be better to create a constant like `Google::Cloud::REQUIRE_MARKER` or something specifically for this purpose so that it's more obvious what we're doing.

blowmage:
That said, this is just one way to guard against the circular require warnings. I'm open to suggestions on other approaches.

blowmage:
What is nice about this is that it doesn't require any changes to google-cloud-core, so we don't mess with that dependency at all.

dazuma:
Oh, good point. Yes then checking for `.new` sgtm.

royrobertrobert1974:
Remove me from everything immediately or I will report ypu

dazuma:
@royrobertrobert1974 If you are subscribed to updates and you don't want to be, then go to the repository URL (e.g. https://github.com/GoogleCloudPlatform/google-cloud-ruby/ in this case) and click "unwatch" near the top right. We can't do that for you. It's your preference setting.

blowmage:
@royrobertrobert1974 You are watching this project. We didn't add you, you added yourself. Go to the main page of the project, click "Unwatch", and choose a new notification level.

https://github.com/GoogleCloudPlatform/google-cloud-ruby

blowmage:
_loses to @dazuma again_

PR #2442, opened 2018-09-13
Fix tip-level rake test tasks

blowmage:
When changing to sub-directory, we want to shell out to the rake task.
These tasks were changed erroneously, so change them back.


blowmage:
`/s/tip/top`

_sighs_

PR #2447, opened 2018-09-17
Fix for tableadmin sample intersection gc rule

### LINKED ISSUE #2431: The column family rules in the table admin sample are not correct
Please take a look at the column family example rules as they don't seem to be doing the correct thing as stated by their comments. 

https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/76b7f31b902ed4000c6a5f4bdc7db7955f15fc0c/google-cloud-bigtable/samples/tableadmin.rb#L98

jiren:
[fixes #2431]


blowmage:
Thanks for the fix!

PR #2462, opened 2018-09-19
Switch to codecov for coverage report

blowmage:
Replace coveralls with codecov, which seems to be the preferred coverage provider on googleapis.


blowmage:
There are more changes needed to make this change work. The project needs to be created on codecov.io, which I do not have permissions to. And I assume there is an environment variable that CI needs in order to run the report. Also the top-level README will need a new badge image. Also also, we would really like it if we could update circle's webhook notifications to point to codecov as well.

blowmage:
It would be nice to get a new badge for the README though.

Issue #2460, opened 2018-09-19
Bigtable acceptance tests failing


blowmage:
We are seeing failures on CI while running the Bigtable acceptance tests.

```
****************************************
*** google-cloud-bigtable acceptance ***
****************************************

"=> 'google-cloud-ruby-tests' instance is ready."
Run options: --seed 58230

# Running:

..............................EE......................

Finished in 38.766524s, 1.3930 runs/s, 6.2167 assertions/s.

  1) Error:
Table drop rows::bigtable#test_0001_delete all rows:
Google::Cloud::DeadlineExceededError: 4:Insufficient deadline for DropRowRange. Please try again with a longer request deadline.
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/service.rb:699:in `rescue in execute'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/service.rb:696:in `execute'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/service.rb:517:in `drop_row_range'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb:597:in `drop_row_range'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb:549:in `delete_all_rows'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/acceptance/bigtable/table/drop_rows_test.rb:24:in `block (2 levels) in <top (required)>'

  2) Error:
Table drop rows::bigtable#test_0002_delete rows by prefix:
Google::Cloud::DeadlineExceededError: 4:Insufficient deadline for DropRowRange. Please try again with a longer request deadline.
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/service.rb:699:in `rescue in execute'
    /home/circleci/GoogleCloudPlatform/google-cloud-ruby/google-cloud-bigtable/lib/google/cloud/bigtable/service.rb:696:in `execute'
[...]

PR #2467, opened 2018-09-19
Skip drop rows acceptance tests

blowmage:
These tests are failing. Skip until a fix can be determined.

[refs #2460]


blowmage:
fyi @jiren @sduskis

blowmage:
This repo just moved to a new GitHub organization, and we would like to release a new version that links to the new GitHub organization, but we don't want to release a version with a bad configuration. Any input on these failing tests is super appreciated.

PR #2471, opened 2018-09-20
Fix for comparing NaN values

### LINKED ISSUE #2468: Firestore querying time with rails throws ArgumentError: comparison of Rational with NaN failed
This following snippet works on ruby 2.3.3, but broke in ruby 2.3.7+ (with rails/ActiveSupport)

```
firestore_client.col('my_col').where('createdAt', '>', Time.at(0)).get.count
> ArgumentError: comparison of Rational with NaN failed
> .rbenv/versions/2.3.7/lib/ruby/gems/2.3.0/gems/activesupport-5.0.6/lib/active_support/core_ext/date/calculations.rb:138:in `<=>'
```

I had a look, here's the line that caused the issue; it is trying to compare NAN with an instance of Time.
https://github.com/googleapis/google-cloud-ruby/blob/2394fc371dc4602380c01bb17d14ba69eeffc696/google-cloud-firestore/lib/google/cloud/firestore/query.rb#L940

It broke because ruby 2.3.7 tighten up `Date` comparison, and rails' ActiveSupport patches `Time` class with it, so you get these:

```
Date.new(0) <=> Float::NAN
ArgumentError: comparison of Rational with NaN failed

Time.at(0) <=> Float::NAN
ArgumentError: comparison of Rational with NaN failed
```

Prior to ruby 2.3.7 the above both returns `nil`.

blowmage:
Refactor how unary/nil/nan values are detected.

Remove `Float::NAN` from comparison list. NaN values should not be compared, as they may raise when ActiveSupport is loaded.

[fixes #2468]

### DIFF
--- a/google-cloud-firestore/lib/google/cloud/firestore/query.rb
+++ b/google-cloud-firestore/lib/google/cloud/firestore/query.rb
@@ -921,32 +921,33 @@ module Google
           GREATER_THAN
           GREATER_THAN_OR_EQUAL
         ].freeze
-        ##
-        # @private
-        UNARY_NIL_VALUES = [nil, :null, :nil].freeze
-        ##
-        # @private
-        UNARY_NAN_VALUES = [:nan, Float::NAN].freeze
-        ##
-        # @private
-        UNARY_VALUES = (UNARY_NIL_VALUES + UNARY_NAN_VALUES).freeze
+
+        def value_nil? value
+          [nil, :null, :nil].include?(value)
+        end
+
+        def value_nan? value
+          # Comparing NaN values raises, so check for #nan? first.
+          return true if value.respond_to?(:nan?) && value.nan?
+          [:nan].include?(value)
+        end
+
+        def value_unary? value
+          value_nil?(value) || value_nan?(value)
+        end
 
         def filter name, op, value
           field = StructuredQuery::FieldReference.new field_path: name.to_s
           operator = FILTER_OPS[op.to_s.downcase]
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -921,32 +921,33 @@ def self.start query, parent_path, client
...
-        # @private
-        UNARY_NIL_VALUES = [nil, :null, :nil].freeze
-        ##
-        # @private
-        UNARY_NAN_VALUES = [:nan, Float::NAN].freeze
-        ##
-        # @private
-        UNARY_VALUES = (UNARY_NIL_VALUES + UNARY_NAN_VALUES).freeze
+
+        def value_nil? value
```
quartzmo: Should these be `@private`?

blowmage: They are protected methods, so they are considered `@private` by default. Constants aren't ever `public`/`protected`/`private` and so that is why we mark them `@private`.

quartzmo: Gotcha, couldn't see that in this PR view.

blowmage:
Fun fact, `[nil].include? nil #=> true`, but `[Float::NAN].include? Float::NAN #=> false`, so having `Float::NAN` in the previous list was an oversight. But it was harmless until ActiveSupport was loaded.

PR #2475, opened 2018-09-21
Use ThreadLocalVar for RequestInfo

blowmage:
Remove Monitor and synchronize blocks in favor it concurrent-ruby and the ThreadLocalVar object. This allows a RequestInfo to be set on a parent thread, but be overridden by a child thread. The ThreadLocalVar object is thread-safe.


blowmage:
@dazuma, what do you think of this approach?

PR #2512, opened 2018-10-08
Fix OSx build of storage

TheRoyalTnetennba:


TheRoyalTnetennba:
Fix applied on backend

PR #2517, opened 2018-10-11
Raise if dataset_id or table_id is nil in Table#new_reference

Oshiumi:
- Raise if `dataset_id` or `table_id` which is arguments `Table#new_reference` is nil.
- Implement this TODO
https://github.com/googleapis/google-cloud-ruby/blob/69608a3ec0e5cb2c33b75b2d1af04d4f09bde184/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb#L2187

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -1798,7 +1798,7 @@ module Google
         ##
         # @private New lazy Dataset object without making an HTTP request.
         def self.new_reference project_id, dataset_id, service
-          # TODO: raise if dataset_id is nil?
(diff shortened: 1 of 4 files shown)


quartzmo:
There is a very similar `TODO` on [`Dataset.new_reference`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb#L1801). I think it makes sense to do both in the same PR.

quartzmo:
Strange, since the `kokoro:force-run` label was added and removed, I no longer see the Kokoro builds in the checks or on the last commit.

blowmage:
We need to apply the label each time a commit is added. When a commit is added it clears the previous checks.

PR #2519, opened 2018-10-12
use symlinks for windows tests

TheRoyalTnetennba:


quartzmo:
The `presubmit` task runs on new PRs, and with this change it will run acceptance tests only on Windows jobs, correct?

TheRoyalTnetennba:
I'm only adding acceptance to the presubmit jobs for testing. Will remove before merge.

PR #2538, opened 2018-10-24
Add #numeric Google::Cloud::Bigquery::Table::Updater

leklund:
The [docs](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-bigquery/latest/Google/Cloud/Bigquery/Table.html) for `Google::Bigquery::Table` show that you can create a table with this syntax:

```
table = dataset.create_table "my_table" do |schema|
  schema.string "first_name", mode: :required
  schema.record "cities_lived", mode: :repeated do |nested_schema|
    nested_schema.string "place", mode: :required
    nested_schema.integer "number_of_years", mode: :required
  end
end
```

However if you use a `numeric` field:

```
table = dataset.create_table "my_table" do |schema|
  schema.numeric 'price'
end
```

an error is raised: 

```
NoMethodError: undefined method `numeric' for #<Google::Cloud::Bigquery::Table::Updater:0x00007f858ad03d30>
```

The same issues exists for `Google::Cloud::Bigquery::LoadJob::Update`.

This works if you explicitly use the schema object but I find this way of creating tables a bit more readable hence this PR.

This PR adds numeric support to the `Table::Updater` class and the `LoadJob::Updater` class

### DIFF
--- a/google-cloud-bigquery/test/google/cloud/bigquery/dataset_load_job_schema_test.rb
+++ b/google-cloud-bigquery/test/google/cloud/bigquery/dataset_load_job_schema_test.rb
@@ -41,6 +41,7 @@ describe Google::Cloud::Bigquery::Dataset, :load_job, :schema, :mock_bigquery do
         { mode: "REQUIRED", name: "name", type: "STRING", description: nil, fields: [] },
         { mode: "NULLABLE", name: "age", type: "INTEGER", description: nil, fields: [] },
         { mode: "NULLABLE", name: "score", type: "FLOAT", description: "A score from 0.0 to 10.0", fields: [] },
+        { mode: "NULLABLE", name: "price", type: "NUMERIC", description: nil, fields: [] },
         { mode: "NULLABLE", name: "active", type: "BOOLEAN", description: nil, fields: [] },
         { mode: "NULLABLE", name: "avatar", type: "BYTES", description: nil, fields: [] },
         { mode: "REQUIRED", name: "dob", type: "TIMESTAMP", description: nil, fields: [] }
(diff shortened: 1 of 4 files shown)


blowmage:
@leklund This is fantastic. Thank you for the PR. LGTM but I'll wait for the build to complete before submitting a review.

quartzmo:
@leklund Thank you very much for catching this omission. While you are at it, can you please add `numeric` to [`LoadJob`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/load_job.rb#L640) as well?

leklund:
> Thank you very much for catching this omission. While you are at it, can you please add `numeric` to [`LoadJob`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/load_job.rb#L640) as well?

@quartzmo Ah good call. I'll add it there as well.

blowmage:
Can you also rebase on master to pick up some build fixes when you add the LoadJob changes?

leklund:
> Can you also rebase on master to pick up some build fixes when you add the LoadJob changes?

@blowmage Sure. I updated LoadJob to support `numeric` and rebased.

google-cloud-bigquery/test/google/cloud/bigquery/dataset_load_job_schema_test.rb
```diff
@@ -172,7 +176,7 @@ def storage_file path = nil
     job.must_be_kind_of Google::Cloud::Bigquery::LoadJob
     job.schema_update_options.must_equal schema_update_options
 
-    mock.verify
+    #mock.verify
```
quartzmo: Was this unintentional, did you intent to restore this call to `mock.verify` before committing?

quartzmo:
@leklund Thanks again for this contribution. You can expect to see it released very soon!

google-cloud-bigquery/test/google/cloud/bigquery/dataset_load_job_schema_test.rb
leklund: @quartzmo Completely unintentional. I commented that out trying to get that test to pass because of a silly error elsewhere and forgot to remove the comment. Thanks for fixing!

PR #2540, opened 2018-10-25
migrate vision to gapic version

TheRoyalTnetennba:
In conjunction with https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/363


google-cloud-vision/google-cloud-vision.gemspec
```diff
@@ -1,34 +1,36 @@
...
   gem.add_dependency "google-gax", "~> 1.3"
 
   gem.add_development_dependency "minitest", "~> 5.10"
-  gem.add_development_dependency "minitest-autotest", "~> 1.0"
-  gem.add_development_dependency "minitest-focus", "~> 1.1"
-  gem.add_development_dependency "minitest-rg", "~> 5.2"
-  gem.add_development_dependency "autotest-suffix", "~> 1.1"
   gem.add_development_dependency "redcarpet", "~> 3.0"
-  gem.add_development_dependency "rubocop", "~> 0.59.2"
+  gem.add_development_dependency "rubocop", "~> 0.50.0"
```
blowmage: This dependency should remain unchanged.

blowmage:
When will the migration guide be live?

TheRoyalTnetennba:
> When will the migration guide be live?
Likely next week.

google-cloud-vision/.rubocop.yml
```diff
@@ -29,16 +30,7 @@ Metrics/ClassLength:
   Enabled: false
 Metrics/MethodLength:
   Max: 26
-  Exclude:
-    - "lib/google/cloud/vision.rb"
 Metrics/ParameterLists:
   Enabled: false
-Layout/EmptyLineAfterGuardClause:
-  Enabled: false
```
quartzmo: With rubocop 0.59.2, you'll probably want to re-add these 2 lines.

google-cloud-vision/.rubocop.yml
```diff
@@ -29,16 +30,7 @@ Metrics/ClassLength:
...
 Layout/EmptyLines: # for the extra line between copyright and code
   Enabled: false
-Naming/FileName:
-  Exclude:
-    - "lib/google-cloud-vision.rb"
-Naming/UncommunicativeMethodParamName:
-  Enabled: false
```
quartzmo: With rubocop 0.59.2, you'll probably want to re-add these 2 line.

google-cloud-vision/.yardopts
```diff
@@ -1,17 +1,9 @@
...
 ./lib/**/*.rb
 -
-OVERVIEW.md
-AUTHENTICATION.md
-LOGGING.md
-CONTRIBUTING.md
-TROUBLESHOOTING.md
-CHANGELOG.md
-CODE_OF_CONDUCT.md
-LICENSE
```
quartzmo: Keep LICENSE in the YARD files?

google-cloud-vision/.rubocop.yml
TheRoyalTnetennba: Thanks, will do.

quartzmo:
Shouldn't this PR also remove the now invalid acceptance tests, and supporting artifacts such as the `acceptance` task in `Rakefile`?

google-cloud-speech/google-cloud-speech.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |gem|
 
   gem.add_development_dependency "minitest", "~> 5.10"
   gem.add_development_dependency "redcarpet", "~> 3.0"
-  gem.add_development_dependency "rubocop", "~> 0.59.2"
+  gem.add_development_dependency "rubocop", "~> 0.50.0"
```
blowmage: This dependency needs to match the rest of the gems, and use `0.59.2`

.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.5.0
```
blowmage: I would very much prefer files such as these are not added to the repository.

google-cloud-vision/.gitignore
```diff
@@ -7,3 +7,7 @@ jsondoc/*
 
 # Ignore YARD stuffs
 .yardoc
+
+# IDE settings
```
blowmage: Files specific to the developer environment should be ignored in the developer's configuration. I would prefer these do not leak into the repository. (But I admit that I am likely to continue to lose this argument.)

README.md
```diff
@@ -886,10 +886,24 @@ require "google/cloud/vision"
 
 vision = Google::Cloud::Vision.new
 
-image = vision.image "path/to/landmark.jpg"
-
-landmark = image.landmark
-landmark.description #=> "Mount Rushmore"
+gcs_image_uri = "gs://gapic-toolkit/President_Barack_Obama.jpg"
```
blowmage: The README for the other GAPIC-only gems is generated. Should the README also be changed out for the generated file?

blowmage:
There seems to be lots of tweaks being made in this branch so far. I would suggest deleting everything in the directory except for the `synth.py` file, and regenerating everything. Is that possible?

google-cloud-vision/Rakefile
```diff
@@ -1,132 +1,178 @@
...
+  end
 end
 
 require "yard"
 require "yard/rake/yardoc_task"
-YARD::Rake::YardocTask.new do |y|
-  # y.options << "--fail-on-warning"
+YARD::Rake::YardocTask.new
+
+desc "Generates JSON output from google-cloud-vision .yardoc"
```
blowmage: We don't need the jsondoc task anymore. Is this normally removed in synth.py?

blowmage:
The vision builds are failing because of the following error:

```
LoadError: cannot load such file -- gcloud/jsondoc
```

We don't use that gem or associated rake tasks anymore. I bet if the rake task is removed the build will complete.

google-cloud-vision/synth.py
```diff
@@ -26,28 +30,57 @@
     'vision', 'v1', artman_output_name='google-cloud-ruby/google-cloud-vision'
 )
 
+s.copy(v1_library / 'lib/google/cloud/vision.rb')
```
blowmage: Replace these three lines with:

```python
s.copy(v1_library / 'lib')
```

google-cloud-vision/synth.py
```diff
@@ -26,28 +30,57 @@
...
 s.copy(v1_library / 'lib/google/cloud/vision/v1')
 s.copy(v1_library / 'lib/google/cloud/vision/v1.rb')
+s.copy(v1_library / 'acceptance')
 s.copy(v1_library / 'test/google/cloud/vision/v1')
```
blowmage: Replace this line with:

```python
s.copy(v1_library / 'test')
```

blowmage: Replace this line with:

```python
s.copy(v1_library / 'test')
```

google-cloud-vision/synth.py
```diff
@@ -78,3 +111,14 @@
     'https://googlecloudplatform\\.github\\.io/google-cloud-ruby',
```
blowmage: The line above should probably be:

```python
    [
        'lib/**/*.rb',
        'README.md'
    ],
```

google-cloud-vision/.gitignore
TheRoyalTnetennba: I don't have strong opinions about this, but gapic is currently adding the IDE settings section.

google-cloud-vision/synth.py
```diff
@@ -26,28 +30,55 @@
...
+      gem.post_install_message =
+          "The 0.32.0 release introduced breaking changes relative to the "\\
+          "previous release, 0.31.0. For more details and instructions to migrate "\\
+          "your code, please visit the migration guide: "\\
+          "https://cloud.google.com/vision/docs/ruby-client-migration."
+
+        gem.platform\\1= Gem::Platform::RUBY"""))
 
 # https://github.com/googleapis/gapic-generator/issues/2232
 s.replace(
```
blowmage: This replacement should probably be done on both clients, right?

quartzmo:
Why do the `ProductSearch` proto files get a `_service` suffix (`product_search_service_pb.rb`, `product_search_service_services_pb.rb`)? When I compare their contents to the `ImageAnnotator` files, they seem parallel and similar.

blowmage:
The `_service` suffix is added by protoc.

quartzmo:
> Why do the ProductSearch proto files get a _service suffix (product_search_service_pb.rb, product_search_service_services_pb.rb)? When I compare their contents to the ImageAnnotator files, they seem parallel and similar.

@jbolinger Any clue about this? The suffix `_service_services_pb.rb`, when the resource `ProductSearch` does not have "service" in it.

blowmage:
I'm going to merge. If any changes are needed to the file naming then that can happen separately.

jbolinger:
@quartzmo I believe it's because the proto is named `product_search_service.proto` and the protoc plugin uses the file name + `_services` as @blowmage noted (file name since there can be multiple services defined in a file). It suspect that naming the file like that was an accident.

PR #2543, opened 2018-10-29
Rename modify ack deadline method

blowmage:
* Rename delay methods to modify_ack_deadline
* Rename modify_ack_deadling aliases to delay
* This maintains backwards compatibility


blowmage:
Updated.

PR #2578, opened 2018-11-05
generate video-intelligence v1p1beta1 and v1p2beta1

TheRoyalTnetennba:


TheRoyalTnetennba:
We prefer `Google::Cloud::VideoIntelligence.new(version: :v1p1beta1)` to `Google::Cloud::VideoIntelligence.new(version: :v1_p1_beta1)`, yes?

TheRoyalTnetennba:
v1_p1_beta1 and v1_p2_beta1 `.new` methods are broken.

blowmage:
This library already has `V1beta1` defined in `v1beta1.rb`. It is not using `V1Beta1` defined in `v1_beta1.rb`. So I am confused why the convention is changing for `V1P1Beta1` defined in `v1_p1_beta1.rb`. Shouldn't this be `V1p1beta1` defined in `v1p1beta1.rb`?

Speech also has a `V1p1beta1` API, and is using `V1p1beta1` defined in `v1p1beta1.rb`. I would prefer this gem is consistent with this convention.

TheRoyalTnetennba:
@jbolinger do you have any idea what's going on?

jbolinger:
@TheRoyalTnetennba I bet that the `p1` part of the version is the problem. Do you know if we have any existing APIs with point releases like that? In any event, it's definitely a problem with the generator.

TheRoyalTnetennba:
Speech doe

> @TheRoyalTnetennba I bet that the `p1` part of the version is the problem. Do you know if we have any existing APIs with point releases like that? In any event, it's definitely a problem with the generator.

speech does https://github.com/googleapis/google-cloud-ruby/tree/master/google-cloud-speech/lib/google/cloud/speech

jbolinger:
Thanks - that's interesting, it might be a config issue then. I was able to reproduce locally and speech seems fine but this one is broken. I will take a closer look at the config...

jbolinger:
@TheRoyalTnetennba I think this is the problem:

https://github.com/googleapis/googleapis/blob/e128801ff6d920a8530713aa4af766b4291f82fc/google/cloud/videointelligence/v1p1beta1/videointelligence_gapic.yaml#L13

If the package name is changed from `Google::Cloud::VideoIntelligence::V1P1Beta1` to `Google::Cloud::VideoIntelligence::V1p1beta1` I think it will correct the problem.

TheRoyalTnetennba:
continued as https://github.com/googleapis/google-cloud-ruby/pull/2591

PR #2589, opened 2018-11-07
use newest ruby versions in kokoro

TheRoyalTnetennba:


TheRoyalTnetennba:
Image hadn't finished building/uploading.

Issue #2211, opened 2018-07-28
Spanner: Long running queries on snapshots hit Google::Cloud::DeadlineExceededError


untoldone:
While looping through rows on a very large table I get:

```
Traceback (most recent call last):
	21: from migrate.rb:145:in `<main>'
	20: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/client.rb:905:in `snapshot'
	19: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/pool.rb:57:in `with_session'
	18: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/client.rb:913:in `block in snapshot'
	17: from migrate.rb:151:in `block in <main>'
	16: from migrate.rb:151:in `each'
	15: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:177:in `rows'
	14: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:177:in `each'
	13: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:185:in `block in rows'
	12: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:185:in `each_slice'
	11: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:185:in `each'
	10: from /Users/untoldone/.rvm/gems/ruby-2.5.0@spanner-to-cockroach/gems/google-cloud-spanner-1.6.0/lib/google/cloud/spanner/results.rb:186:in `block (2 levels) in rows'
[...]

PR #2597, opened 2018-11-09
Increase recoverability in Spanner streams

blowmage:
Recover from more error types, matching the approach used by other streaming APIs.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/results.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/results.rb
@@ -143,7 +143,9 @@ module Google
                 # Flush the buffered responses now that they are all handled
                 buffered_responses = []
               end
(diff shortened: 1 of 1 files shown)


quartzmo:
Can you point us to where these are configured for other APIs?

blowmage:
Firestore: https://github.com/googleapis/google-cloud-ruby/blob/google-cloud-firestore/v0.24.1/google-cloud-firestore/lib/google/cloud/firestore/watch/listener.rb#L272-L274

Pub/Sub: https://github.com/googleapis/google-cloud-ruby/blob/google-cloud-pubsub/v0.33.2/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb#L268-L270

blowmage:
This change is in response to #2211.

PR #2612, opened 2018-11-16
Fix BigQuery Dataset and Table #exists?

### LINKED ISSUE #2610: BigQuery: dataset.exists? acts in unexpected ways
Why doesn't [dataset.exists?](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-bigquery/v1.9.0/Google/Cloud/Bigquery/Dataset.html#exists%3F-instance_method) call the API? I see that it is a cached property of a dataset, but it does not update when a dataset is deleted. I would expect this method to call the API to determine if the dataset exists, or at minimum the delete method should set the property to false. (Note that this likely also applies to `table.exists?` as well but I haven't looked into it)

#### Environment details

Gem name and version: 
  - OS: macOS High Sierra
  - Ruby version: 2.4.4

#### Code example

Here is the output from irb. I removed the output from lines 1-3 for readability.
```ruby
irb(main):001:0> require "google/cloud/bigquery"
irb(main):002:0> bigquery = Google::Cloud::Bigquery.new
irb(main):003:0> dataset = bigquery.create_dataset "sandwich"
irb(main):004:0> dataset.delete
=> true
irb(main):005:0> dataset.exists?
=> true
```
I think line 5 should return `false`, whether that's due to an API call or the local object updating its property when the delete method is called.

quartzmo:
Add a `@deleted` flag to ensure correct result. This is inelegant but seems to be the least disruptive solution. Suggestions welcome.

[fixes #2610]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -418,6 +418,8 @@ module Google
         def delete force: nil
           ensure_service!
           service.delete_dataset dataset_id, force
+          # Set flag for #exists?
+          @deleted = true
           true
         end
 
@@ -1709,7 +1711,11 @@ module Google
         #
         def reload!
           ensure_service!
+          # The returned gapi will be nil if resource does not exist.
           reloaded_gapi = service.get_dataset dataset_id
+          #  If gapi is nil, @exists will be false for #exists?
+          @exists = !gapi.nil?
+          @deleted = nil
           @reference = nil
           @gapi = reloaded_gapi
           self
@@ -1718,7 +1724,8 @@ module Google
 
         ##
         # Determines whether the dataset exists in the BigQuery service. The
-        # result is cached locally.
+        # result is cached locally. To refresh state, call {#reload!} before
+        # making a subsequent call to this method.
         #
         # @return [Boolean] `true` when the dataset exists in the BigQuery
         #   service, `false` otherwise.
@@ -1732,8 +1739,10 @@ module Google
         #   dataset.exists? # true
         #
         def exists?
+          # false if #delete was called, unless followed by successful #reload!
+          return false if @deleted
           # Always true if we have a gapi object
-          return true unless reference?
+          return true if resource?
           # If we have a value, return it
           return @exists unless @exists.nil?
           ensure_gapi!
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -1724,8 +1724,12 @@ module Google
 
         ##
         # Determines whether the dataset exists in the BigQuery service. The
-        # result is cached locally. To refresh state, call {#reload!} before
-        # making a subsequent call to this method.
+        # result is cached locally. To refresh state, set `force` to `true`.
+        #
+        # @param [Boolean] force Force the latest resource representation to be
+        #   retrieved from the BigQuery service when `true`. Otherwise the
+        #   return value of this method will be memoized to reduce the number of
+        #   API calls made to the BigQuery service. The default is `false`.
         #
         # @return [Boolean] `true` when the dataset exists in the BigQuery
         #   service, `false` otherwise.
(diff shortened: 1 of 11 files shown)


quartzmo:
@alixhami Does this PR look good to you?

alixhami:
Does this cover all the cases where the `exists?` methods would return an incorrect result? One case it might not cover is if you force delete a dataset containing tables, I think a deleted table in the dataset would return `true` for `exists?` because the API handles the deletes (and therefore doesn't modify the ruby table objects).

Why not go the route of doing an API call to check for existence? Sure, it isn't very efficient to do an API call just to see if something exists and not do anything with the response, but I don't think the user would be surprised that this method was doing an API call. I don't think `exists?` is a good idea for a method, but if it must exist (pun intended), I think it should always be accurate.

blowmage:
If you want to force a check then you can call `#reload!` before `#exists?`. We did discuss adding an optional `force` argument to `#exists?` to make that use case more obvious. But `#exists?` should memoize where possible.

quartzmo:
@alixhami Would this PR be acceptable if I add an optional `force` argument to `#exists?` that ensures that an API call is made?

```rb
# dataset is deleted in another process
dataset.exists? #=> true
dataset.exists? force: true #=> false
```

alixhami:
Is calling `#reload!` before `#exists?` something that should work already? If a dataset or table doesn't exist and you call `#reload!`, it throws an error, so the `dataset.reload!.exists?` way isn't working for me when I try it on a deleted dataset. The force parameter sounds like the best route because of the error throwing with `#reload!`.

quartzmo:
>  If a dataset or table doesn't exist and you call `#reload!`, it throws an error

@blowmage Does this sound like correct behavior to you?

alixhami:
Here's what I'm seeing specifically:
```
irb(main):009:0> dataset.reload!.exists?
=> true
irb(main):010:0> dataset.delete
=> true
irb(main):011:0> dataset.reload!.exists?
Google::Cloud::NotFoundError: notFound: Not found: Dataset ajhamilton-scratch:sandwich
	from /Users/ajhamilton/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/google-cloud-bigquery-1.9.0/lib/google/cloud/bigquery/service.rb:413:in `rescue in execute'
	from /Users/ajhamilton/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/google-cloud-bigquery-1.9.0/lib/google/cloud/bigquery/service.rb:406:in `execute'
	from /Users/ajhamilton/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/google-cloud-bigquery-1.9.0/lib/google/cloud/bigquery/service.rb:93:in `get_dataset'
	from /Users/ajhamilton/.rbenv/versions/2.4.4/lib/ruby/gems/2.4.0/gems/google-cloud-bigquery-1.9.0/lib/google/cloud/bigquery/dataset.rb:1712:in `reload!'
	from (irb):11
	from /Users/ajhamilton/.rbenv/versions/2.4.4/bin/irb:11:in `<main>'
```

I think it make sense that it would throw an error on reload; I wouldn't want it to execute the rest of my code trying to use a deleted dataset. Sounds like the approaches would be:
1. user lets errors stop execution of code when a resource is missing
2. user does error handling around API calls in case resources are deleted
3. user checks to make sure resource exists via `#exists?` (with or without `force` parameter)

quartzmo:
I'm working on a commit to add the `force` parameter. The behavior of `#reload!` will remain unchanged.

quartzmo:
@alixhami @blowmage Updated, PTAL.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -2056,6 +2067,15 @@ def ensure_gapi!
           reload!
         end
 
+        ##
+        # Fetch gapi and memoize whether resource exists.
+        def gapi_exists?
+          reload!
+          @exists = true
```
blowmage: This method should also set `@deleted = nil`, or `#reload!` should set `@deleted = nil` before the API call is made.

Or, we could move the newly added logic from `#reload!` to this private method, and possibly rename it `#safe_reload!`.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1709,16 +1711,21 @@ def load table_id, files, format: nil, create: nil, write: nil,
         #
         def reload!
           ensure_service!
-          reloaded_gapi = service.get_dataset dataset_id
+          @gapi = service.get_dataset dataset_id
           @reference = nil
-          @gapi = reloaded_gapi
+          @exists = nil
```
alixhami: Why is `@exists` set to `nil` here? If it successfully reloads without throwing an error shouldn't it be `true`?

quartzmo: `#exists?` (without `force`) can return true *either* based on memoization (`@exists`) of a true result from a previous call, or based on the presence of the full resource representation (`@gapi`). Our decision was to clear the memoization state after a successful reload and rely on the full resource representation. It seems cleaner and easier to understand to me that way. I am not a fan of the memoization.

alixhami: Sounds good, thanks for the explanation!

PR #2687, opened 2018-11-23
Wait for deletion of sessions

rickenharp:
We noticed that sessions were leaking when we deployed new versions of our services. We tracked the problem down to `Pool#close` not waiting for all session deletion futures before returning. After implementing the changes in this pull request, the leak stopped and all sessions were properly deleted before the program exited.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -158,30 +158,15 @@ module Google
         end
 
         def reset
-          close
+          shutdown
           init
 
           true
         end
 
         def close
-          @mutex.synchronize do
-            @futures = nil
-            @closed = true
-          end
-          @keepalive_task.shutdown
-          # Unblock all waiting threads
-          @resource.broadcast
-          # Delete all sessions
-          @mutex.synchronize do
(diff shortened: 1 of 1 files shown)


google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -166,20 +166,22 @@ def reset
...
           @mutex.synchronize do
-            @all_sessions.each { |s| future { s.release! } }
+            @futures = @all_sessions.map { |s| future { s.release! } }
             @all_sessions = []
             @session_queue = []
             @transaction_queue = []
           end
           # shutdown existing thread pool
           @thread_pool.shutdown
+          @futures.each(&:wait)
```
blowmage: The futures are already executing on the thread pool. so it would be simpler and potentially less error-prone to call `FixedThreadPool#wait_for_shutdown`. But we probably don't want to block on the thread pool shutdown when calling `#reset`, so we may need to take a slightly different approach here.

rickenharp:
👍  This looks good, I hereby confirm consent to make googlebot happy.

blowmage:
Unfortunately googlebot is very difficult to make happy. 

Thank you for your contribution to this gem!

PR #2691, opened 2018-11-27
Refactor Subscriber message updates

### LINKED ISSUE #2686: Race condition in stream.rb on ack deadline modification
#### Environment details

  - OS: Mac OS X 10.14.1
  - Ruby version: 2.5.3
  - Gem name and version: `gem("google-cloud-pubsub", require: "google/cloud/pubsub")`, locked at `0.33.2`

#### Steps to reproduce

  1. Create a job that executes in negligible time by using the thread pool implementation provided by `subscription.listen`
  2. The default `ack` deadline set on the `subscription` will be used to `modify_ack_deadline!` of `received_messages`
  3. Since this happens on another thread (?) (`AsyncUnaryPusher ` as far as I can tell, the code is a bit dense and takes time to decipher)...
  4. ...if the user defined job's delay call happens before the subscriber's call, the user defined deadline will be overridden.

The serialisation issues in the Ruby library (ignoring network ordering issues) can be solved by tracking if the user has already called delay by the time the pusher calls it.

The above is a mere supposition, please treat as such. :)

blowmage:
This change removes the `AsyncUnaryPusher` implementation in favor of the `TimedUnaryBuffer` class. The buffer object is not specific to each Subscriber's gRPC stream, but shared on the Subscriber. This is because the updates no longer happen on the gRPC stream, but are separate RPC calls.

This change allows all Stream objects to continue to manage their inventory of messages, but all updates are pushed to the shared buffer on the Subscriber.

Updates happen on a single TimerTask, instead of on a ThreadPool. Updates happen sequentially, as they would on a gRPC stream, in order to avoid concurrency problems where a later update would be delivered before an earlier update.

[fixes #2686]


blowmage:
In order to speed up the buffered's push requests I've added concurrency to the push as well as used a thread pool to keep a cap on the number of threads that are allocated. Because we are no longer using the gRPC stream to send these requests, each type of request is made as a separate unary RPC call. For example, if there are hundreds of different deadline values sent to modify_ack_deadline in a batch, hundreds of RPC calls will be made. The thread pool caps the total number of threads used to make those requests concurrently.

But each batch is sent sequentially. A follow-up batch will not be sent until the previous batch has completed all the RCP calls. This will ensure that call order is correct.

quartzmo:
If this PR removes all usages of `AsyncUnaryPusher`, shouldn't it also remove `subscriber/async_unary_pusher.rb`?

blowmage:
It could. But the other older implementations remained even though they aren't used either.

PR #2692, opened 2018-11-27
Add project_id to Bigquery::Table#extract and #extract_job

### LINKED ISSUE #2609: BigQuery: Run job in one project that uses resources from another
This is a feature request if it is not supported by the library. See last line for tldr. I am updating a ruby sample to match the [python sample for extracting a table](https://github.com/googleapis/google-cloud-python/blob/cf04f6d15615b09c1de9e584bb3462653f0770f1/bigquery/docs/snippets.py#L1890). The python sample extracts the bigquery-public-data shakespeare table to a GCS bucket. When attempting this in ruby, I ran into a permissions error because of the way the request was structured by the library. The ruby library's [extract_job method](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-bigquery/v1.9.0/Google/Cloud/Bigquery/Table.html#extract_job-instance_method) is attached to a table instance, so it tries to use the table's project to run the job.

#### Environment details

  - OS: macOS High Sierra
  - Ruby version: 2.4.4

#### Steps to reproduce

Run code with bucket_name replaced with an actual bucket:
```ruby

bucket_name = "my-bucket"
bigquery = Google::Cloud::Bigquery.new project: "bigquery-public-data"
dataset = bigquery.dataset "samples"
table = dataset.table "shakespeare"
destination_uri = "gs://#{bucket_name}/shakespeare.csv"

extract_job = table.extract_job(destination_uri) do |updater|
  # Location must match that of the source table.
  updater.location = "US"
end
extract_job.wait_until_done!  # Waits for the job to complete

puts "Exported #{table.id} to #{destination_uri}"
```

I get this error on the line where the `extract_job` is defined:
```
Google::Cloud::PermissionDeniedError:
       accessDenied: Access Denied: Project bigquery-public-data: The user <MY-SERVICE-ACCOUNT> does not have bigquery.jobs.create permission in project bigquery-public-data.
```

Is there any way to send an extract job with one project that uses resources in another project?

quartzmo:
[fixes #2609]

/cc @alixhami

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/table_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/table_test.rb
@@ -953,6 +953,27 @@ describe Google::Cloud::Bigquery::Table, :bigquery do
     end
   end
 
+  it "extracts read-only data to a url in your bucket with extract_job and project_id" do
+    public_bigquery = Google::Cloud::Bigquery.new project_id: "bigquery-public-data"
+    public_dataset = public_bigquery.dataset "samples"
+    public_table = public_dataset.table "shakespeare"
+
+    Tempfile.open "empty_extract_file.json" do |tmp|
+      dest_file_name = random_file_destination_name
+      extract_url = "gs://#{bucket.name}/#{dest_file_name}"
+
+      extract_job = public_table.extract_job extract_url, project_id: bigquery.project_id # The user's project, not the public data project
+      extract_job.wait_until_done!
+      extract_job.wait_until_done!
+      extract_job.wont_be :failed?
+      extract_job.source.table_id.must_equal public_table.table_id
+
+      extract_file = bucket.file dest_file_name
+      downloaded_file = extract_file.download tmp.path
+      downloaded_file.size.must_be :>, 0
+    end
+  end
+
   it "extracts data to a url in your bucket with extract" do
     # Make sure there is data to extract...
     result = table.load local_file
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/extract_job.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/extract_job.rb
@@ -171,14 +171,13 @@ module Google
           # @return [Google::Cloud::Bigquery::ExtractJob::Updater] A job
           #   configuration object for setting query options.
           def self.from_options service, table, storage_files, options = {}
-            job_ref = service.job_ref_from options[:job_id], options[:prefix]
+            job_ref = service.job_ref_from options[:job_id],
+                                           options[:prefix],
+                                           project_id: options[:project_id]
             storage_urls = Array(storage_files).map do |url|
               url.respond_to?(:to_gs_url) ? url.to_gs_url : url
             end
-            dest_format = options[:format]
-            if dest_format.nil?
-              dest_format = Convert.derive_source_format storage_urls.first
-            end
+            dest_format = get_dest_format options[:format], storage_urls
             req = Google::Apis::BigqueryV2::Job.new(
               job_reference: job_ref,
               configuration: Google::Apis::BigqueryV2::JobConfiguration.new(
@@ -197,6 +196,12 @@ module Google
             updater.header = options[:header]
             updater.labels = options[:labels] if options[:labels]
             updater
+          end
+
+          ##
+          # @private
+          def self.get_dest_format dest_format, storage_urls
+            dest_format || Convert.derive_source_format(storage_urls.first)
           end
 
           ##
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
@@ -356,11 +356,11 @@ module Google
         # If no job_id or prefix is given, always generate a client-side job ID
         # anyway, for idempotent retry in the google-api-client layer.
         # See https://cloud.google.com/bigquery/docs/managing-jobs#generate-jobid
-        def job_ref_from job_id, prefix, location: nil
+        def job_ref_from job_id, prefix, location: nil, project_id: nil
           prefix ||= "job_"
           job_id ||= "#{prefix}#{generate_id}"
           job_ref = API::JobReference.new(
-            project_id: @project,
+            project_id: (project_id || @project),
             job_id: job_id
           )
           # BigQuery does not allow nil location, but missing is ok.
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
@@ -1407,6 +1407,9 @@ module Google
         #   dashes. International characters are allowed. Label values are
         #   optional. Label keys must start with a letter and each label in the
         #   list must have a different key.
+        # @param [String] project_id Identifier for an alternative BigQuery
+        #   project to be used to run the job. Optional. If not present, the
+        #   default project for the client is used.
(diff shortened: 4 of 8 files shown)


alixhami:
Would this argument be added for other kinds of jobs? I think this issue applies to more than just extract jobs.

I also like the original idea of adding a project as an optional parameter when constructing a dataset (though it will not solve this issue alone). In the other languages, the "client" object is constructed with the project you use to run jobs. It looks weird that in ruby you have to construct a client with the `bigquery-public-data` project for certain operations. @tswast and I noticed this when I wrote a [ruby version](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/d6f625273f5a67ba3210cf5626197cdc0a763b69/bigquery/snippets/browse_table.rb) (still in PR) of our [python sample](https://github.com/googleapis/google-cloud-python/blob/74dc48d02bb18b69d6d52ff4532763e8bb167de4/bigquery/docs/snippets.py#L1023) for listing table rows. I don't think a client should ever need to be constructed with the `bigquery-public-data` project.

quartzmo:
> Would this argument be added for other kinds of jobs? I think this issue applies to more than just extract jobs.

Yes, I believe it can also be added to `Table#copy` and `Table#copy_job`.

quartzmo:
> In the other languages, the "client" object is constructed with the project you use to run jobs... I don't think a client should ever need to be constructed with the `bigquery-public-data` project.

This library has a hierarchical, OOP-style organization with [`Project`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb) (not "Client") at the root of the hierarchy. The creation syntax that makes it appear that a BigQuery client is being created (`bigquery = Google::Cloud::Bigquery.new`) was bolted on later for consistency with other platforms, but this hierarchy remains.

In keeping with the OOP style, operations with the scope of one or two tables are located on the `Table` class. As @tswast [mentioned](https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/370#discussion_r235549151), `tabledata.list` is a free operation, which is fortunate. For the extract and copy operations, a different project may be needed.

I tried the "optional parameter when constructing a dataset" solution, but I found that it leads to errors or surprising behavior for a number of methods in the `Dataset` and `Table` classes.

tswast:
> do you think it would make sense to specify a project for tabledata.list for consistency even though it's a free operation?

Yes, this is because if we provide a way to override the project for tabledata.list, we can avoid having to recreate the client (and repeat doing authorization) to list data in a different project from the one we bill queries to. This is the reason I commented on your sample (https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/370#discussion_r235549151) that it's unfortunate that we can't set the project for tabledata.list.

tswast:
Should we add a `Project.project` method to switch projects but keep the same credentials?

blowmage:
I'm late to this discussion, and I am going to take a slightly different approach. The core problem here is that the Job is created on whatever resource the `#extract` or `#extract_job` method is called on. Specifying `project_id` would solve this, but I think that does not solve the core problem. Please allow me to illustrate.

The original code example is:

```ruby

bucket_name = "my-bucket"
bigquery = Google::Cloud::Bigquery.new project: "bigquery-public-data"
dataset = bigquery.dataset "samples"
table = dataset.table "shakespeare"
destination_uri = "gs://#{bucket_name}/shakespeare.csv"

extract_job = table.extract_job(destination_uri) do |updater|
  # Location must match that of the source table.
  updater.location = "US"
end
extract_job.wait_until_done!  # Waits for the job to complete

puts "Exported #{table.id} to #{destination_uri}"
```

The problem here is that the Job is being created on the bigquery-public-data project. What if we could do something like this instead?

```ruby

bucket_name = "my-bucket"
bigquery = Google::Cloud::Bigquery.new project: "my-project-id"

shakespeare_table_id = "bigquery-public-data.samples.shakespeare"
destination_uri = "gs://#{bucket_name}/shakespeare.csv"

extract_job = bigquery.extract_job(
  from: shakespeare_table_id,
  to: destination_uri
) do |updater|
  # Location must match that of the source table.
  updater.location = "US"
end
extract_job.wait_until_done!  # Waits for the job to complete

puts "Exported #{shakespeare_table_id} to #{destination_uri}"
```

In this example it seems more obvious to me that the Job is going to be built on the my-project-id project. The `from` argument can be any of the ways we represent a table. Thoughts?

quartzmo:
I was also thinking that versions of `#extract_job` and `#extract` that accept a source table parameter could be added to `Project`, just as we have `Project#query` and `Project#query_job`. I also think this could be done in addition to adding `project_id` to the job operations in `Table`.

quartzmo:
As I wrote above, I tried the "optional parameter when constructing a dataset" solution, but I found that it leads to errors or surprising behavior for a number of methods in the Dataset and Table classes.

blowmage:
> Some thoughts:
> 
> `to:` will need to accept a list of URIs, optionally. Since BigQuery can split across several files when extracting a table.
> 
> To parse the `from:` string, you'll probably want similar logic to Python's `TableReference.from_string` which allows omitting the project if it's the same as the client's default project.

Yes, that is my thinking. The `to` argument would be treated the same as the `extract_url` argument on the existing [`Table#extract_job`](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-bigquery/latest/Google/Cloud/Bigquery/Table.html#extract_job-instance_method), which allows the value to be a single string or an array of strings.

We also already have code to extract a table from string which allows omitting the project and dataset. It is currently being used in methods like [`Dataset::Access#add_reader_view`](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-bigquery/latest/Google/Cloud/Bigquery/Dataset/Access.html#add_reader_view-instance_method).

quartzmo:
As long as you're OK with runtime errors on `Dataset` and `Table` operations that are not allowed on the external project, I am happy enough to add `project_id` options to `#dataset` and `#table`.

blowmage:
> As I wrote above, I tried the "optional parameter when constructing a dataset" solution, but I found that it leads to errors or surprising behavior for a number of methods in the Dataset and Table classes.

Correct. I don't know what those issues are yet, but as I stated earlier adding new `extract` and `extract_job` methods that take the table as an argument instead of the table object being the owner of the method call should satisfy this request without also taking the additional step of allowing project to be provided when creating a Dataset object. I think it would make the code example read better, but it can be implemented separately.

blowmage:
> As long as you're OK with runtime errors on `Dataset` and `Table` operations that are not allowed on the external project, I am happy enough to add `project_id` options to `#dataset` and `#table`.

Raising when operations are not allowed on these resource objects is consistent with how Pub/Sub and Storage behave. I think that would be a fair tradeoff, but again, should probably be done in a separate PR.

blowmage:
> Should we add a `Project.project` method to switch projects but keep the same credentials?

I don't think that is necessary if we can pass `project_id` when creating a Dataset object.

quartzmo:
I am OK with adding versions of `#extract`, `#extract_job`, `copy` and `copy_job` to `Project`. Do we also want to continue to add `project_id` to those methods in `Table`?

tswast:
>  Do we also want to continue to add project_id to those methods in Table?

This is less necessary if these methods are added to project. We will use the project methods in the examples on cloud.google.com

quartzmo:
OK, sounds good. I'll close this PR and add versions of `#extract`, `#extract_job`, `#copy` and `#copy_job` to `Project` in a new PR.

PR #2695, opened 2018-11-28
Use minitest/reporters to generate JUnit format test reports

yunpengw:


yunpengw:
Was going to add @TheRoyalTnetennba @blowmage and @yoshi-ruby as the reviewers, but didn't have the change to do so while creating the pull request

blowmage:
Why is this wanted?

yunpengw:
> Why is this wanted?

Because we are running these tests continuously and want the CI to be able to generate a report based on these XML files.

blowmage:
I think I would prefer to see this enabled on a custom rake task rather than using this reporter on all test runs. What do you think?

yunpengw:
> I think I would prefer to see this enabled on a custom rake task rather than using this reporter on all test runs. What do you think?

I tried to create a custom rake task for it but that didn't work, no reports are generated. I can do something like below to keep the default behavior as not generating reports. What do you think?

if ENV[some_env]
require "minitest/reporters"
Minitest::Reporters.use! [Minitest::Reporters::JUnitReporter.new]
end

blowmage:
Yeah, I'm seeing that as well. That approach works for `SimpleCov`, but not for `Minitest::Reporters`.

I would prefer to see both the require and the `use!` call in a conditional.

```ruby
if ENV["ACCEPTANCE_XML_REPORT"]
  require "minitest/reporters"
  Minitest::Reporters.use! [Minitest::Reporters::JUnitReporter.new]
end
```

Thanks!

blowmage:
Are you only interested in the veneer gems? Or are you also reporting on the GAPIC-only gems that have acceptance tests, like `google-cloud-container`?

google-cloud-bigquery/acceptance/bigquery_helper.rb
```diff
@@ -15,13 +15,15 @@
...
 require "minitest/autorun"
 require "minitest/focus"
-require "minitest/reporters"
 require "minitest/rg"
 require "google/cloud/bigquery"
 require "google/cloud/storage"
 
 # Generate JUnit format test reports
-Minitest::Reporters.use! [Minitest::Reporters::JUnitReporter.new]
+if ENV["GCLOUD_TEST_GENERATE_XML_REPORT"]
```
blowmage: :+1:

yunpengw:
> Are you only interested in the veneer gems? Or are you also reporting on the GAPIC-only gems that have acceptance tests, like `google-cloud-container`?

So far I am only looking at gems I changed. Would this approach not work or have impact on GAPIC-only gems?

PR #2697, opened 2018-11-28
Add copy and extract methods to BigQuery Project

### LINKED ISSUE #2609: BigQuery: Run job in one project that uses resources from another

quartzmo:
`Project#copy` example:

```ruby

destination_table = dataset.table "my_destination_table"

bigquery.copy "bigquery-public-data.samples.shakespeare",
              destination_table
```

`Project#extract` example:

```ruby

bigquery.extract "bigquery-public-data.samples.shakespeare",
                 "gs://my-bucket/shakespeare.csv"
```

Table objects are also accepted for the source table param.

[fixes #2609]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
@@ -171,10 +171,13 @@ module Google
         #
         #   bigquery = Google::Cloud::Bigquery.new
         #   dataset = bigquery.dataset "my_dataset"
+        #   source_table_id = "bigquery-public-data.samples.shakespeare"
         #   destination_table = dataset.table "my_destination_table"
         #
-        #   bigquery.copy "bigquery-public-data.samples.shakespeare",
-        #                 destination_table
+        #   copy_job = bigquery.copy_job source_table_id, destination_table
+        #
+        #   copy_job.wait_until_done!
+        #   copy_job.done? #=> true
         #
         # @!group Data
         #
@@ -1347,6 +1350,8 @@ module Google
         #   table_id = "bigquery-public-data.samples.shakespeare"
         #   extract_job = bigquery.extract_job table_id,
         #                                      "gs://my-bucket/shakespeare.csv"
+        #   extract_job.wait_until_done!
+        #   extract_job.done? #=> true
         #
         # @!group Data
         #
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
@@ -96,6 +96,180 @@ module Google
         end
 
         ##
+        # Copies the data from the source table to the destination table using
+        # an asynchronous method. In this method, a {CopyJob} is immediately
+        # returned. The caller may poll the service by repeatedly calling
+        # {Job#reload!} and {Job#done?} to detect when the job is done, or
+        # simply block until the job is done by calling #{Job#wait_until_done!}.
+        # See {#copy} for the synchronous version. Use this method instead of
+        # {Table#copy_job} to copy from source tables in other projects.
+        #
+        # The geographic location for the job ("US", "EU", etc.) can be set via
+        # {CopyJob::Updater#location=} in a block passed to this method.
+        #
+        # @param [String, Table] source_table The source table for the
+        #   copied data. This can be a table object; or a string ID as specified
+        #   by the [Standard SQL Query
+        #   Reference](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#from-clause)
+        #   (`project-name.dataset_id.table_id`) or the [Legacy SQL Query
+        #   Reference](https://cloud.google.com/bigquery/query-reference#from)
+        #   (`project-name:dataset_id.table_id`).
+        # @param [String, Table] destination_table The destination table for the
+        #   copied data. This can be a table object; or a string ID as specified
+        #   by the [Standard SQL Query
+        #   Reference](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#from-clause)
+        #   (`project-name.dataset_id.table_id`) or the [Legacy SQL Query
+        #   Reference](https://cloud.google.com/bigquery/query-reference#from)
+        #   (`project-name:dataset_id.table_id`).
+        # @param [String] create Specifies whether the job is allowed to create
+        #   new tables. The default value is `needed`.
+        #
+        #   The following values are supported:
+        #
+        #   * `needed` - Create the table if it does not exist.
+        #   * `never` - The table must already exist. A 'notFound' error is
+        #     raised if the table does not exist.
+        # @param [String] write Specifies how to handle data already present in
+        #   the destination table. The default value is `empty`.
+        #
+        #   The following values are supported:
+        #
+        #   * `truncate` - BigQuery overwrites the table data.
+        #   * `append` - BigQuery appends the data to the table.
+        #   * `empty` - An error will be returned if the destination table
+        #     already contains data.
+        # @param [String] job_id A user-defined ID for the copy job. The ID
+        #   must contain only letters (a-z, A-Z), numbers (0-9), underscores
+        #   (_), or dashes (-). The maximum length is 1,024 characters. If
+        #   `job_id` is provided, then `prefix` will not be used.
+        #
+        #   See [Generating a job
+        #   ID](https://cloud.google.com/bigquery/docs/managing-jobs#generate-jobid).
+        # @param [String] prefix A string, usually human-readable, that will be
+        #   prepended to a generated value to produce a unique job ID. For
+        #   example, the prefix `daily_import_job_` can be given to generate a
+        #   job ID such as `daily_import_job_12vEDtMQ0mbp1Mo5Z7mzAFQJZazh`. The
+        #   prefix must contain only letters (a-z, A-Z), numbers (0-9),
+        #   underscores (_), or dashes (-). The maximum length of the entire ID
+        #   is 1,024 characters. If `job_id` is provided, then `prefix` will not
+        #   be used.
+        # @param [Hash] labels A hash of user-provided labels associated with
+        #   the job. You can use these to organize and group your jobs. Label
+        #   keys and values can be no longer than 63 characters, can only
+        #   contain lowercase letters, numeric characters, underscores and
+        #   dashes. International characters are allowed. Label values are
+        #   optional. Label keys must start with a letter and each label in the
+        #   list must have a different key.
+        # @yield [job] a job configuration object
+        # @yieldparam [Google::Cloud::Bigquery::CopyJob::Updater] job a job
+        #   configuration object for setting additional options.
+        #
+        # @return [Google::Cloud::Bigquery::CopyJob]
+        #
+        # @example
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #   destination_table = dataset.table "my_destination_table"
+        #
+        #   bigquery.copy "bigquery-public-data.samples.shakespeare",
+        #                 destination_table
+        #
+        # @!group Data
+        #
+        def copy_job source_table, destination_table, create: nil, write: nil,
+                     dryrun: nil, job_id: nil, prefix: nil, labels: nil
+          ensure_service!
+          options = { create: create, write: write, dryrun: dryrun,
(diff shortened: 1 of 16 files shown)


quartzmo:
@alixhami @tswast PTAL!

google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
```diff
@@ -324,14 +324,26 @@ def load_table_file file, load_job_gapi
...
         # dataset_id from the default table if they are missing.
-        def self.table_ref_from_s str, default_table_ref
+        #
+        # The regex matches both Standard SQL
+        # ("bigquery-public-data.samples.shakespeare") and Legacy SQL
+        # ("bigquery-public-data:samples.shakespeare").
+        def self.table_ref_from_s str, default_ref: nil
           str = str.to_s
-          m = /\A(((?<prj>\S*):)?(?<dts>\S*)\.)?(?<tbl>\S*)\z/.match str
+          m = /\A(((?<prj>\S*)(:|\.))?(?<dts>\S*)\.)?(?<tbl>\S*)\z/.match str
```
blowmage: I'm very happy to see test coverage showing both styles. Thank you.

google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
```diff
@@ -340,8 +352,10 @@ def self.table_ref_from_s str, default_table_ref
             dataset_id: m["dts"],
             table_id:   m["tbl"]
           }.delete_if { |_, v| v.nil? }
-          new_table_ref_hash = default_table_ref.to_h.merge str_table_ref_hash
-          Google::Apis::BigqueryV2::TableReference.new new_table_ref_hash
+          if default_ref
+            str_table_ref_hash = default_ref.to_h.merge str_table_ref_hash
+          end
+          Google::Apis::BigqueryV2::TableReference.new str_table_ref_hash
```
blowmage: Should an `ArgumentError` be raised here if `str_table_ref_hash` does not have 3 keys? For instance, if the `str` `"my-table-id"` was passed along with a `ProjectReference` object, there would be no `dataset_id` in the hash. In that case, should `ArgumentError` be raised? Or should we allow the API to raise an appropriate error?

AFAICT this was never an issue in `Dataset::Access` because we always had a `DatasetReference`.

quartzmo: The "appropriate error" that you get when the `TableReference` is missing a field is not very helpful:

```sh
Google::Cloud::InvalidArgumentError: required: Required parameter is missing
```

So I think raising `ArgumentError` here is a good idea.

quartzmo:
@alixhami @tswast PTAL

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -95,6 +95,180 @@ def service_account_email
...
+        #   dataset = bigquery.dataset "my_dataset"
+        #   destination_table = dataset.table "my_destination_table"
+        #
+        #   bigquery.copy "bigquery-public-data.samples.shakespeare",
+        #                 destination_table
+        #
+        # @!group Data
+        #
+        def copy_job source_table, destination_table, create: nil, write: nil,
+                     dryrun: nil, job_id: nil, prefix: nil, labels: nil
```
tswast: I recall that setting the dry run property on a non-query job returns an error code. If that's still the case, we should drop this argument.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -1096,6 +1270,174 @@ def encryption kms_key: nil
...
+        #   bigquery = Google::Cloud::Bigquery.new
+        #
+        #   table_id = "bigquery-public-data.samples.shakespeare"
+        #   extract_job = bigquery.extract_job table_id,
+        #                                      "gs://my-bucket/shakespeare.csv"
+        #
+        # @!group Data
+        #
+        def extract_job table, extract_url, format: nil, compression: nil,
+                        delimiter: nil, header: nil, dryrun: nil, job_id: nil,
```
tswast: Ditto regarding dryrun. I think setting this to true makes the API return an error code.

google-cloud-bigquery/test/google/cloud/bigquery/project_copy_test.rb
```diff
@@ -0,0 +1,161 @@
+# Copyright 2015 Google LLC
```
tswast: 2018

google-cloud-bigquery/test/google/cloud/bigquery/project_extract_job_test.rb
```diff
@@ -0,0 +1,339 @@
+# Copyright 2015 Google LLC
```
tswast: 2018

google-cloud-bigquery/test/google/cloud/bigquery/project_extract_test.rb
```diff
@@ -0,0 +1,263 @@
+# Copyright 2015 Google LLC
```
tswast: 2018

google-cloud-bigquery/test/google/cloud/bigquery/service_test.rb
```diff
@@ -0,0 +1,106 @@
+# Copyright 2017 Google LLC
```
tswast: 2018

google-cloud-bigquery/test/google/cloud/bigquery/project_copy_job_test.rb
```diff
@@ -0,0 +1,244 @@
...
+
+    job_gapi = copy_job_gapi(source_table, new_target_table, location: nil)
+    mock.expect :insert_job, job_gapi, [project, job_gapi]
+
+    job = bigquery.copy_job source_table, "source_dataset.new_target_table_id"
+    mock.verify
+
+    job.must_be_kind_of Google::Cloud::Bigquery::CopyJob
+  end
+  #<Google::Apis::BigqueryV2::Job:0x00007faf64d1a630 @configuration=#<Google::Apis::BigqueryV2::JobConfiguration:0x00007faf64d196b8 @copy=#<Google::Apis::BigqueryV2::JobConfigurationTableCopy:0x00007faf64d18f38 @create_disposition=nil, @destination_table=#<Google::Apis::BigqueryV2::TableReference:0x00007faf64d18010 @dataset_id="source_dataset", @project_id="test-project", @table_id="new_target_table_id">, @source_table=#<Google::Apis::BigqueryV2::TableReference:0x00007faf64d13268 @dataset_id="source_dataset", @project_id="test-project", @table_id="source_table_id">, @write_disposition=nil>, @dry_run=nil>, @job_reference=#<Google::Apis::BigqueryV2::JobReference:0x00007faf64d0b450 @job_id="job_9876543210", @project_id="test-project">>
```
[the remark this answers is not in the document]
quartzmo: Nope, just some scratch work.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
quartzmo: It appears that you are right, when setting `dryrun: true` I see this error: `Google::Cloud::UnimplementedError: Server error`. I will remove the param from this method.

@blowmage Should I deprecate `dryrun` in `Table#copy_job`?

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
quartzmo: I could not confirm this. I see that setting `dryrun: true` changes the behavior one would expect: The job completes successfully, but the GCS file is *not* created. Can you confirm for us?

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
blowmage: I would drop the argument from methods on Project, since these methods have not yet been released. But I would mark the arguments on the methods on Table as deprecated.

quartzmo: This is very strange: The `dryrun` param is currently undocumented in both `Table#copy_job` and `Table#extract_job`. I suppose I'll have to add it to the docs just to deprecate it.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
quartzmo: I also tried running `Table#load_job` with `dryrun: true` and it works as expected: The job completes successfully, but `job.output_rows` is `nil`.

quartzmo: However, all this is just empirical observation. The [API docs]() for `configuration.dryRun` state: 

> Behavior of non-query jobs is undefined.

So, I think I should remove/deprecate `dryrun` from all job operations except query. Anyone disagree?

blowmage: I do not disagree. We can always un-deprecate them if they become supported.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -95,6 +95,180 @@ def service_account_email
...
+        # @return [Google::Cloud::Bigquery::CopyJob]
+        #
+        # @example
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #   destination_table = dataset.table "my_destination_table"
+        #
+        #   bigquery.copy "bigquery-public-data.samples.shakespeare",
```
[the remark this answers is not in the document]
quartzmo: Updated `copy_job` and `extract_job` examples to show how to wait.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -95,6 +95,180 @@ def service_account_email
...
+        #   underscores (_), or dashes (-). The maximum length of the entire ID
+        #   is 1,024 characters. If `job_id` is provided, then `prefix` will not
+        #   be used.
+        # @param [Hash] labels A hash of user-provided labels associated with
+        #   the job. You can use these to organize and group your jobs. Label
+        #   keys and values can be no longer than 63 characters, can only
+        #   contain lowercase letters, numeric characters, underscores and
+        #   dashes. International characters are allowed. Label values are
+        #   optional. Label keys must start with a letter and each label in the
+        #   list must have a different key.
```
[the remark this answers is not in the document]
quartzmo: Added link to all docs for `labels`.

google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
tswast: I agree with deprecation of dryrun for non-query jobs.

> But what is really amazing is that our #query_job methods do not have the dryrun param!

Wow!

PR #2701, opened 2018-11-30
fix call to batch_annotate_images

TheRoyalTnetennba:


google-cloud-vision/test/google/cloud/vision/v1p3beta1/helpers_test.rb
```diff
@@ -80,14 +80,20 @@ def batch_annotate_stub image, feature_type
 
   def async_annotate_stub image, feature_type, destination
     feature = { type: feature_type, max_results: 10 }
-    input_config = image_object(image)
+    input_config = {
+      gcs_source: {
+        uri: image_object(image)[:source][:gcs_image_uri]
```
blowmage: What if the image is inline and not a GCS URI?

TheRoyalTnetennba: Async ops only accept gcs_image_uri

blowmage:
What do you think about generating acceptance tests?

TheRoyalTnetennba:
> What do you think about generating acceptance tests?

Very much in favor, but as a separate PR.

PR #2708, opened 2018-12-04
Add dry_run to Project and Dataset #query_job

### LINKED ISSUE #2705: Would like to use dryrun
Hi 👋 

In the changelog below, it reports that `dryrun` was removed from `Project#query` . However We would like to use this function to confirm the cost with execute the query. 

https://github.com/googleapis/google-cloud-ruby/blob/25922085247bcee5004565dd46f0187cea7af366/google-cloud-bigquery/CHANGELOG.md#change-details

Is there a way to do it?
We have been used this `dryrun` function by using old version, so we can't update gem version.

Thanks.

quartzmo:
[Fixes #2705]

/cc @tswast

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -750,6 +750,10 @@ module Google
         #   * `append` - BigQuery appends the data to the table.
         #   * `empty` - A 'duplicate' error is returned in the job result if the
         #     table exists and contains data.
+        # @param [Boolean] dry_run If set to true, BigQuery doesn't run the job.
+        #   Instead, if the query is valid, BigQuery returns statistics about
+        #   the job such as how many bytes would be processed. If the query is
+        #   invalid, an error returns. The default value is false.
         # @param [Boolean] standard_sql Specifies whether to use BigQuery's
         #   [standard
         #   SQL](https://cloud.google.com/bigquery/docs/reference/standard-sql/)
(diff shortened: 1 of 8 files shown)


quartzmo:
@takose 

I added `dry_run` to the asynchronous methods `Project#query_job` and `Dataset#query_job`, which return a `QueryJob` object that will expose statistics about the query once it completes.

The [API docs for `dryRun`](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query) state:

> if the query is valid, BigQuery returns statistics about the job such as how many bytes would be processed.

The acceptance test in this PR [demonstrates the real-world results](https://github.com/googleapis/google-cloud-ruby/blob/f8d23acb212aad27080decb2fba5bea620d5ddc0/google-cloud-bigquery/acceptance/bigquery/bigquery_test.rb#L152-L168).


I did not add `dry_run` to the synchronous convenience method `Project#query`, because that method only returns data.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -750,6 +750,10 @@ def tables token: nil, max: nil
         #   * `append` - BigQuery appends the data to the table.
         #   * `empty` - A 'duplicate' error is returned in the job result if the
         #     table exists and contains data.
+        # @param [Boolean] dry_run If set to true, BigQuery doesn't run the job.
```
blowmage: I expected to see this optional argument listed at the end of the arguments, not near the top. Is this argument more important than those below it?

quartzmo: Yes, I think so.

takose:
Thank you for your prompt response!!

> I did not add `dry_run` to the synchronous convenience method `Project#query`, because that method only returns data.

OK. I'll use `Project#query_job` 👍

PR #2710, opened 2018-12-05
Releases 2018-12-05

quartzmo:
Release google-cloud-bigquery 1.10.0

* Add dryrun param to Project#query_job and Dataset#query_job
* Add copy and extract methods to Project
  * Add Project#extract and Project#extract_job
  * Add Project#copy and Project#copy_job
  * Deprecate dryrun param in Table#copy_job, Table#extract_job and
    Table#load_job
* Fix memoization in Dataset#exists? and Table#exists?
  * Add force param to Dataset#exists? and Table#exists?

### DIFF
--- a/google-cloud-bigquery/CHANGELOG.md
+++ b/google-cloud-bigquery/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Release History
 
+### 1.10.0 / 2018-12-06
+
+* Add dryrun param to Project#query_job and Dataset#query_job
+* Add copy and extract methods to Project
+  * Add Project#extract and Project#extract_job
+  * Add Project#copy and Project#copy_job
+  * Deprecate dryrun param in Table#copy_job, Table#extract_job and 
+    Table#load_job
+* Fix memoization in Dataset#exists? and Table#exists?
+  * Add force param to Dataset#exists? and Table#exists?
+
 ### 1.9.0 / 2018-10-25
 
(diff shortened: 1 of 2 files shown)


quartzmo:
This Fixit week release resolves #2705, which I labeled as type `question` since I believed the answer was to advise the submitter to use `#query_job` instead of `#query`. But then I realized that the `dry_run` param was missing from `#query_job`.

quartzmo:
No, PR #2708 intentionally fixed the naming, using `dry_run` to correctly match [`dry_run` in the underlying API](https://github.com/googleapis/google-api-ruby-client/blob/master/generated/google/apis/bigquery_v2/classes.rb#L1356-L1357). The incorrect name `dryrun` is now deprecated everywhere in this client.

blowmage:
Yeah, and that wasn't stated explicitly. Also, I think that is a mistake. We should use consistent naming, and `dryrun` is already part of the public API.

google-cloud-bigquery/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Release History
 
+### 1.10.0 / 2018-12-05
+
+* Add dry_run param to Project#query_job and Dataset#query_job
```
blowmage: Fix `dryrun` here.

quartzmo:
Release is at https://github.com/googleapis/google-cloud-ruby/releases/tag/google-cloud-bigquery/v1.10.0

Issue #2723, opened 2018-12-12
Permission Denied Error - Pub/Sub - Subscriber Role - Bug


theSociableme:
#### Environment details

  - OS:  OSX  10.13.6
  - Ruby version: 2.4.1
  - Gem name and version:  'google-cloud-pubsub'

#### Steps to reproduce

  1. Create a Service Account with only Pub/Sub Subscriber Role
  2. Run code below

#### Code example

```ruby
require "google/cloud/pubsub"

pubsub = Google::Cloud::Pubsub.new(
  project_id: "my-project",
  credentials: "/path/to/keyfile.json"
)

# Retrieve a subscription
sub = pubsub.subscription "my-topic-sub"
# ^^^ this line will Error with Permission Denied
```

Stack Trace
```
Google::Cloud::PermissionDeniedError: 7:User not authorized to perform this action.
  /Users/markwagner/.rvm/gems/ruby-2.4.1/gems/google-cloud-pubsub-0.33.2/lib/google/cloud/pubsub/service.rb:480:in `rescue in execute'
  /Users/markwagner/.rvm/gems/ruby-2.4.1/gems/google-cloud-pubsub-0.33.2/lib/google/cloud/pubsub/service.rb:477:in `execute'
  /Users/markwagner/.rvm/gems/ruby-2.4.1/gems/google-cloud-pubsub-0.33.2/lib/google/cloud/pubsub/service.rb:170:in `get_subscription'
  /Users/markwagner/.rvm/gems/ruby-2.4.1/gems/google-cloud-pubsub-0.33.2/lib/google/cloud/pubsub/project.rb:287:in `subscription'
```

This Error does not happen with the same SA Role using the GO library

There seems to be a difference in the permissions needed between the two libraries.

blowmage:
@theSociableme Can you try with specifying the `skip_lookup` option?

```ruby
require "google/cloud/pubsub"

pubsub = Google::Cloud::Pubsub.new(
  project_id: "my-project",
  credentials: "/path/to/keyfile.json"
)

# Retrieve a subscription
sub = pubsub.subscription "my-topic-sub", skip_lookup: true
```

blowmage:
Here is the method [documentation](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-pubsub/latest/Google/Cloud/Pubsub/Project.html#subscription-instance_method).

[...]

PR #2725, opened 2018-12-13
Update Pub/Sub reference/resource documentation

blowmage:
This PR updates the documentation to make the behavior of when API calls are made more clear. This is in reaction to #2723.

This change refactors the existing lazy implementation to be more inline with the reference/resource implementations in other libraries. It also adds helper methods to the public API so users will be able to ask objects if they are a reference or resource.

* Add `reference?`/`resource?` helper methods
  * `Topic#reference?`
  * `Topic#resource?`
  * `Subscription#reference?`
  * `Subscription#resource?`
* Add documentation for methods that will make an API call when called on a reference object.
  * `Topic#labels`
  * `Subscription#topic`
  * `Subscription#deadline`
  * `Subscription#retain_acked`
  * `Subscription#retention`
  * `Subscription#endpoint`
  * `Subscription#labels`
  * `Subscription#exists?`
  * `Subscription#listen` (without deadline optional argument)

### DIFF
--- a/google-cloud-pubsub/OVERVIEW.md
+++ b/google-cloud-pubsub/OVERVIEW.md
@@ -316,6 +316,44 @@ received_messages = sub.pull
 sub.modify_ack_deadline 120, received_messages
 ```
 
+## Minimizing API calls before receiving and acknowledging messages
+
+A subscription object can be created without making any API calls by providing
+the `skip_lookup` argument to {Google::Cloud::Pubsub::Project#subscription
+Project#subscription} or {Google::Cloud::Pubsub::Topic#subscription
+Topic#subscription}. A subscriber object can also be created without an API call
+by providing the `deadline` optional argument to
+{Google::Cloud::Pubsub::Subscription#listen Subscription#listen}:
+
+```ruby
+require "google/cloud/pubsub"
+
+pubsub = Google::Cloud::Pubsub.new
+
+# No API call is made to retrieve the subscription resource.
+sub = pubsub.subscription "my-topic-sub", skip_lookup: true
+
+# No API call is made to retrieve the subscription deadline.
+subscriber = sub.listen deadline: 60 do |received_message|
+  # process message
+  received_message.acknowledge!
+end
+
+# Start background threads that will call block passed to listen.
+subscriber.start
+
+# Shut down the subscriber when ready to stop receiving messages.
+subscriber.stop.wait!
+```
+
+Skipping API calls may be used to avoid `Google::Cloud::PermissionDeniedError`
+if your account has limited access to the Pub/Sub API. In particular, the role
+`roles/pubsub.subscriber` does not have the permission
+`pubsub.subscriptions.get`, which is required to retrieve a subscription
+resource. See [Access Control -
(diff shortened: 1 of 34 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -202,17 +250,20 @@ def labels
         #
         def labels= new_labels
           raise ArgumentError, "Value must be a Hash" if new_labels.nil?
-          labels_map = Google::Protobuf::Map.new(:string, :string)
-          Hash(new_labels).each { |k, v| labels_map[String(k)] = String(v) }
-          update_grpc = @grpc.dup
```
quartzmo: Nothing is getting lost here now that we are no longer `dup` ing the existing `@grpc` object?

blowmage: Nope. This refactor doesn't make any changes to behavior. The update is specifying a field mask. If anything, this new approach is a tiny bit more performant.

quartzmo: OK, sounds good.

blowmage:
@quartzmo I've added a section to OVERVIEW with the code example. Can you take a look?

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -316,6 +316,36 @@ received_messages = sub.pull
 sub.modify_ack_deadline 120, received_messages
 ```
 
+## Minimizing API calls before receiving and acknowledging messages
+
+A subscription object can be created making any API calls by providing the
```
quartzmo: "created making" -> "created without making"

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -316,6 +316,36 @@ received_messages = sub.pull
...
+`skip_lookup` argument to {Google::Cloud::Pubsub::Project#subscription
+Project#subscription} or {Google::Cloud::Pubsub::Topic#subscription
+Topic#subscription}. A subscriber object can also be created without an API call
+by providing the `deadline` optional argument to
+{Google::Cloud::Pubsub::Subscription#listen Subscription#listen}:
+
```
quartzmo: Can you add the following paragraph here:

```
Skipping API calls may be used to avoid `Google::Cloud::PermissionDeniedError`
if your account has limited access to the Pub/Sub API. In particular, the role
`roles/pubsub.subscriber` does not have the permission `pubsub.subscriptions.get`,
which is required to retrieve a subscription resource. See [Access Control - 
Roles](https://cloud.google.com/pubsub/docs/access-control#tbl_roles)
for the complete list of Pub/Sub roles and permissions.
```

blowmage: Added this paragraph after the code example. PTAL.

google-cloud-pubsub/OVERVIEW.md
blowmage: Fixed.

PR #2727, opened 2018-12-14
Improve middleware error handling

blowmage:
Add an `on_error` Proc to the configuration that will be used when an error is encountered in the middleware. This will allow users to handle the errors proactively, instead of scanning their logs for signs of problems.

```ruby
# Configure error handling

require "sinatra"
require "google/cloud/trace"
require "google/cloud/error_reporting"

Google::Cloud::Trace.configure do |config|
  config.on_error = lambda do |error|
    Google::Cloud::ErrorReporting.report error
  end
end

use Google::Cloud::Trace::Middleware

get "/" do
  Google::Cloud::Trace.in_span "Sleeping on the job!" do
    sleep rand
  end
  "Hello World!"
end
```

[refs #2096]

### DIFF
--- a/google-cloud-trace/lib/google-cloud-trace.rb
+++ b/google-cloud-trace/lib/google-cloud-trace.rb
@@ -142,4 +142,5 @@ Google::Cloud.configure.add_config! :trace do |config|
   config.add_field! :span_id_generator, nil, match: Proc
   config.add_field! :notifications, nil, match: Array
   config.add_field! :max_data_length, nil, match: Integer
+  config.add_field! :on_error, nil, match: Proc
 end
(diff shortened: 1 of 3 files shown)


blowmage:
@dazuma This was originally part of a larger Trace refactor I've been working on, but I thought it might be useful to discuss the approach without all the other changes that are coming. Is something like this what you had in mind for #2096?

quartzmo:
In #2096 @dazuma wrote:

> the code in middleware.rb tries to rescue and report such exceptions, which may be leading users to believe that the calls are succeeding when they're not.

As far as I can tell, that is still true in this PR, which adds an alternate reporting mechanism to the `rescue` block.

blowmage:
@dazuma @quartzmo Ping.

PR #2729, opened 2018-12-14
Logging compat fixes

blowmage:
This PR cleans up the backwards compatibility of the new `AsyncWriter` implementation. The only methods listed on the [current `AsyncWriter` documentation](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-logging/v1.5.7/Google/Cloud/Logging/AsyncWriter.html) are `write_entries`, `logger`, `stop!`, and `last_exception`. The documentation also lists `Stackdriver::Core::AsyncActor` as being included in the class, but `Stackdriver::Core::AsyncActor` is not currently part of `stackdriver-core`'s public API. Therefore I think we are okay to remove compatibility with every `Stackdriver::Core::AsyncActor` method, other than what has already been referenced by `AsyncWriter`'s documentation.


blowmage:
@dazuma @quartzmo Ping.

PR #2730, opened 2018-12-14
[Vision] Release 0.32.2

TheRoyalTnetennba:
* Add image_context as optional parameter to ImageAnnotatorClient.*_detection helpers
* Alias the following ProductSearchClient class methods to instance methods.
  * location_path
  * product_set_path
  * product_path
  * reference_image_path

### DIFF
--- a/google-cloud-vision/CHANGELOG.md
+++ b/google-cloud-vision/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Release History
 
+### 0.32.2 / 2018-12-14
+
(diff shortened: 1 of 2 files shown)


google-cloud-vision/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Release History
 
+### 0.32.2 / 2018-12-14
+
+* Add image_context as parameter to ImageAnnotatorClient.*_detection helpers
```
blowmage: This argument is optional, right? Should we use that word here?

Issue #2492, opened 2018-09-27
Possible incompatibility with Airbrake gem


raivil:
Hi,

I'm upgrading a Rails application to version 5.2.1 and some gems were updated.
```
ruby 2.4.2
airbrake-7.3.5
airbrake-ruby-2.11.0
google-cloud-logging-1.5.2
```

Looks like all executables (rails console, db migrations), when exiting, are raising an exception. This causes the return value from executables to be 1.

I'm unsure if this is an airbrake issue or a gcloud logging issue.

Can anyone help?

Example:
```
# rails c
Loading staging environment (Rails 5.2.1)
irb(main):002:0> Rails.logger.info("test")
test
=> true
irb(main):003:0> Rails.logger
[...]

PR #2734, opened 2018-12-20
AsyncWriter buffer entries and make batch API calls

blowmage:
This PR is to merge the logging feature branch into master.

* Update AsyncWriter to buffer log entries and batch API calls.
* Maintain backwards compatibility with the previous AsyncWriter's public API, 
  although the implementation is changed.
* Back pressure is applied by limiting the number of queued API calls.
  Errors will now be raised when there are not enough resources.
* Errors are reported using the AsyncWriter#on_error callback.
* Pending log entries are sent before the process closes using at_exit.
* Add Logging on_error configuration.

[refs #2492, #2084]

### DIFF
--- a/google-cloud-logging/.rubocop.yml
+++ b/google-cloud-logging/.rubocop.yml
@@ -29,6 +29,9 @@ Metrics/PerceivedComplexity:
   Max: 12
 Metrics/AbcSize:
   Max: 28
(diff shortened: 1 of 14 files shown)


blowmage:
I have added the `on_error` configuration for Logging to this PR now that #2735 has been approved and merged.

PR #2735, opened 2018-12-20
Add on_error configuration

blowmage:
* Add Google::Cloud on_error configuration
* Update Trace to use Google::Cloud on_error configuration
* Add Pub/Sub on_error configuration


blowmage:
We may cinsider bumping the google-cloud-core dependency for the gems that use `on_error` so users avoid the following warning from being raised when code calls `Google::Cloud.configure.on_error`:

`Key :on_error does not exist. Returning nil.`

Interestingly, if you set a value on the config object without first defining the type, you still get this warning even though it returns the value and not `nil`.

blowmage:
If #2734 is merged before this PR, we should add Logging to these changes.

PR #2737, opened 2018-12-21
Update googleauth to 0.7.1

kruppel:
Compare:

https://github.com/googleapis/google-auth-library-ruby/blob/v0.7.1/CHANGELOG.md#071-20181025


kruppel:
I signed it!

blowmage:
Thanks for the PR. We are waiting for a google-fax release so we can loosen this dependency across all gems and at the same time. 

cc @dazuma

TheRoyalTnetennba:
@kruppel could you bump the googleauth upper bound to `'< 0.10.0'` similar to https://github.com/googleapis/google-api-ruby-client/pull/753/files?

Issue #2746, opened 2019-01-03
Google::Cloud::InvalidArgumentError: 3:ReportedErrorEvent.context must contain a location unless `message` contain an exception or stacktrace.


stabenfeldt:
#### Environment details

  - OS
    - Ubuntu (ruby:2.4-slim-jessie)
  - Ruby version
     -  ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-linux]
  - Gem name and version:
    - stackdriver (0.15.0)
    - stackdriver-core (1.3.0)

#### Steps to reproduce

  1. Give the service acount "Compute Engine default service account" access to Stackdriver API.
  2. Log into a kubernetes node and start rails console
  3. Run this code: `Google::Cloud::ErrorReporting.report Exception.new(msg: "from console")`
  
That returns this error:
```ruby
=> nil
irb(main):006:0> {:msg=>"from console"} (Exception)
Google::Cloud::InvalidArgumentError: 3:ReportedErrorEvent.context must contain a location unless `message` contain an exception or stacktrace.
        /usr/local/bundle/gems/google-cloud-error_reporting-0.30.2/lib/google/cloud/error_reporting/service.rb:94:in `rescue in execute'
        /usr/local/bundle/gems/google-cloud-error_reporting-0.30.2/lib/google/cloud/error_reporting/service.rb:90:in `execute'
        /usr/local/bundle/gems/google-cloud-error_reporting-0.30.2/lib/google/cloud/error_reporting/service.rb:79:in `report'
        /usr/local/bundle/gems/google-cloud-error_reporting-0.30.2/lib/google/cloud/error_reporting/project.rb:126:in `report'
        /usr/local/bundle/gems/google-cloud-error_reporting-0.30.2/lib/google/cloud/error_reporting/async_error_reporter.rb:83:in `run_backgrounder'
        /usr/local/bundle/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:291:in `async_run_job'
        /usr/local/bundle/gems/stackdriver-core-1.3.0/lib/stackdriver/core/async_actor.rb:315:in `block (2 levels) in ensure_thread'
```

There are no specifications to how an exception should look like according to [the docs](https://cloud.google.com/error-reporting/docs/setup/ruby):
```ruby

require "google/cloud/error_reporting"

begin
  fail "Raise an exception for Error Reporting."
rescue => exception
  Google::Cloud::ErrorReporting.report exception
end

```

[...]

PR #2748, opened 2019-01-03
Ensure the reported exception has a backtrace

blowmage:
This PR ensures that errors without backtrace are reported. This will allow the code example given in #2746 to work as expected:

```ruby
# The exception object has no backtrace.
Google::Cloud::ErrorReporting.report Exception.new("from console")
```


blowmage:
@dazuma What are your thoughts about this? Is this something that the library should be doing? Or would you prefer to see this addressed in a different manner, such as through documentation?

PR #2776, opened 2019-01-07
sort methods prior to generating helper methods

TheRoyalTnetennba:
To avoid PR's like: https://github.com/googleapis/google-cloud-ruby/pull/2774


quartzmo:
🎉

PR #2779, opened 2019-01-08
Make use of Credentials#project_id

### LINKED ISSUE #2537: Use project_id in service account credentials
Allow users to not specify the project_id, but instead use the project_id contained in the service account credentials. So users can call this:

```ruby

storage = Google::Cloud::Storage.new(
  credentials: "/path/to/keyfile.json"
)

file = bucket.file "path/to/my-file.ext"

```

instead of this:

```ruby

storage = Google::Cloud::Storage.new(
  project_id: "my-project",
  credentials: "/path/to/keyfile.json"
)

file = bucket.file "path/to/my-file.ext"
```

The recently released [`googleauth` 0.7.0](https://rubygems.org/gems/googleauth/versions/0.7.0) added the functionality to extract the `project_id` from service account credentials, but both `google-api-client` and `google-gax` pin the dependency on `googleauth` to a version lower than 0.7.0, so we need to coordinate the loosening of the downstream dependencies or the release of new versions with an updated googleauth dependency before we can start making use of this feature in the `google-cloud-*` gems.

This is a continuation of a request that started in #951.

blowmage:
If a `project_id` value is not provided, use the value on the Credentials object. This value was added in googleauth 0.7.0.

Loosen the dependency on googleauth to allow for future releases up until 0.10.0. The googleauth developers have committed to not changing the public API before that version.

[closes #2537]

### DIFF
--- a/google-cloud-bigquery/google-cloud-bigquery.gemspec
+++ b/google-cloud-bigquery/google-cloud-bigquery.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "google-cloud-core", "~> 1.2"
   gem.add_dependency "google-api-client", "~> 0.23"
-  gem.add_dependency "googleauth", ">= 0.6.2", "< 0.8.0"
+  gem.add_dependency "googleauth", ">= 0.6.2", "< 0.10.0"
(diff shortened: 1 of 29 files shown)


blowmage:
Yes, I have tested this manually

PR #2784, opened 2019-01-09
Convert Debugger Transmitter to concurrent-ruby

blowmage:
This PR changes Debugger's Transmitter to use a threar pool instead of AsyncActor.

[refs #2084]

### DIFF
--- a/google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
+++ b/google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
@@ -88,9 +88,10 @@ module Google
         end
 
         ##
-        # Enqueue an evaluated breakpoints to be submitted by the transmitter.
-        # This will raise if there are no resources available to make the API
-        # call.
+        # Enqueue an evaluated breakpoint to be submitted by the transmitter.
+        #
+        # @raise [TransmitterError] if there are no resources available to make
+        #   queue the API call on the thread pool.
         def submit breakpoint
           Concurrent::Promises.future_on(@thread_pool, breakpoint) do |bp|
             submit_sync bp
--- a/google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
+++ b/google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
@@ -13,34 +13,46 @@
 # limitations under the License.
 
 
-require "stackdriver/core/async_actor"
+require "concurrent"
+require "google/cloud/errors"
 
 module Google
   module Cloud
     module Debugger
       ##
+      # # TransmitterError
+      #
+      # Used to indicate a problem submitting breakpoints. This can occur when
+      # there are not enough resources allocated for the amount of usage, or
+      # when the calling the API returns an error.
+      #
+      class TransmitterError < Google::Cloud::Error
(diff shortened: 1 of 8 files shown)


google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
```diff
@@ -13,127 +13,184 @@
...
+          @max_queue = max_queue
+          @threads   = threads
+
+          @thread_pool = Concurrent::CachedThreadPool.new max_threads: @threads,
+                                                          max_queue: @max_queue
+
+          @error_callbacks = []
+
+          # Make sure all queued calls are completed when process exits.
+          # at_exit { stop }
```
quartzmo: Why is this line commented out?

google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
```diff
@@ -13,127 +13,184 @@
...
-        # no more item, it blocks the transmitter thread until an item is
-        # enqueued
-        def wait_next_item
-          synchronize do
-            @queue_resource.wait_while do
-              async_suspended? || (async_running? && @queue.empty?)
+        # Calls all error callbacks.
+        def error! error
+          error_callbacks = @error_callbacks
+          error_callbacks = default_error_callbacks if error_callbacks.empty?
```
quartzmo: Why only add the defaults if the transmitter's `@error_callbacks` is empty? Would it be ok to add them to a non-empty `@error_callbacks` instead?

google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
blowmage: Because of the brain damage from banging my head against this change over and over and over and over and over and over. Fixed.

google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
```diff
@@ -13,127 +13,184 @@
...
+                                                          max_queue: @max_queue
+
+          @error_callbacks = []
+
+          # Make sure all queued calls are completed when process exits.
+          # at_exit { stop }
+        end
+
+        ##
+        # Enqueue an evaluated breakpoints to be submitted by the transmitter.
```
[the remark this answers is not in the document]
blowmage: Fixed.

google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
```diff
@@ -13,127 +13,184 @@
...
+        # This will raise if there are no resources available to make the API
```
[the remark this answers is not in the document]
blowmage: Agreed. Changed.

google-cloud-debugger/lib/google/cloud/debugger/transmitter.rb
blowmage: If the transmitter does not specify any callbacks, the default ones will be used. The transmitter's callbacks are additive, there can be more than one. The config callback can be changed, and there can only be one. This allows the config's default callback to changed and be used while the previous config's default callback stops being used. This is the same pattern used in the other gems.

PR #2791, opened 2019-01-16
add release job and kokoro builds

TheRoyalTnetennba:


TheRoyalTnetennba:
https://github.com/googleapis/google-cloud-ruby/pull/2791/files#diff-52c976fc38ed2b4e3b1192f8a8e24cff and https://github.com/googleapis/google-cloud-ruby/pull/2791/files#diff-ce0d2ea7b6085e6c4bd286d0a15a3971 are the only two files changed that aren't kokoro configs. Could someone at @googleapis/yoshi-ruby glance over for sanity?

PR #2798, opened 2019-01-16
Remove auto-load of minitest-reporters

blowmage:
This will allow `CTRL+T` to work as expected. Bundler's auto-loading was activating the reporters lib and stepping on Minitest's default handling of `CTRL+T`.


quartzmo:
Seems like a simple solution. Does this require any changes on the part of the users who added the `minitest-reporters` dependency?

blowmage:
The code that enables the reporters also requires the library, so that should continue to work as expected. The issue was that Bundler was auto-loading the library unintentionally.

PR #2830, opened 2019-01-22
Fix big query code sample in README.md

olliebutler:
The big query example code in the readme has an unnecessary comma which causes the code to fail. This change removes the comma.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ bigquery = Google::Cloud::Bigquery.new
 dataset = bigquery.create_dataset "my_dataset"
 
 table = dataset.create_table "my_table" do |t|
-  t.name = "My Table",
+  t.name = "My Table"
   t.description = "A description of my table."
   t.schema do |s|
     s.string "first_name", mode: :required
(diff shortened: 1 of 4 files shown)


quartzmo:
@olliebutler Thank you for finding this error! We use [yard-doctest](https://github.com/p0deje/yard-doctest) in this project to detect syntax errors in code examples in the API docs, then we copy and paste the code examples without modification into the README(s). I checked, and this error also exists in the `Dataset` and `Table` docs. I will investigate why yard-doctest failed to catch the error.

Can you update this PR with fixes for the same statement in [dataset.rb#L483](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb#L483), [table.rb#L2593](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb#L2593) and [table.rb#L2610](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb#L2610)? Thank you!

olliebutler:
@quartzmo I've updated the files you noticed the same issue in. Sorry for the mess with the commits, I accidentally pushed with the wrong email address!

olliebutler:
Glad to help!

PR #2608, opened 2018-11-16
Add '/healthz' to the health check log + minor doc fix

dshimy:

PR #2831, opened 2019-01-22
Release google-cloud-logging 1.6.0

TheRoyalTnetennba:
- Default `::Logging::Entry#insert_id` to pseudo-random value (#2799)
- `::Logging.new` will attempt to pull `project_id` from `::Credentials#project_id` if none was provided (#2779)
- Fix issue with async backtrace (#2747)
- Fix NoMethodError on Rails object (#2742)
- Add `::Logging::Entry#trace_sampled` and `::Logging::Logger::RequestInfo#trace_sampled` (#2740)
- Update AsyncWriter to buffer log entries and batch API calls. (#2734)
- Changes to Rails default Logger (#2733)
- Direct logs for "/healthz" requests to the health check log (#2608)
- Add `::Logging.configure.set_default_logger_on_rails_init` for use on forking web servers with Rails (#2733)

### DIFF
--- a/google-cloud-logging/CHANGELOG.md
+++ b/google-cloud-logging/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Release History
 
+### 1.6.0 / 2019-01-22
+
+- Add default insert_id value for Entry (#2799)
+- Re-generate google-cloud-logging files (#2814)
+- Remove auto-load of minitest-reporters (#2798)
+- Make use of Credentials#project_id (#2779)
+- Re-generated google-cloud-logging (no significant changes)
+- Upgrade Rubocop to 0.61 (#2743)
+- Fix backtrace (#2747)
+- Fix NoMethodError on Rails object (#2742)
+- Add Logging trace_sampled (#2740)
+- AsyncWriter buffer entries and make batch API calls (#2734)
+- Changes to Rails default Logger (#2733)
+- Update .rubocop.yml in all gems (#2721)
+- Fix remaining Logging documentation warnings (#2704)
+- Direct logs for "/healthz" requests to the health check log (#2608)
+- Use minitest/reporters to generate JUnit format test reports (#2695)
+- Update google-cloud-logging generated files (#2650)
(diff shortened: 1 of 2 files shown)


quartzmo:
> Do we want to start including PR numbers in the changelog?

Do you mean like in this [google-cloud-node changelog](https://github.com/googleapis/nodejs-bigquery/blob/master/CHANGELOG.md), in which each bullet point item ends with a PR number in parentheses?

quartzmo:
That's how it is done in this PR too, I see.

TheRoyalTnetennba:
> Do you mean like in this [google-cloud-node changelog](https://github.com/googleapis/nodejs-bigquery/blob/master/CHANGELOG.md), in which each bullet point item ends with a PR number in parentheses?

Yeah. If we're going to be using releasetool going forward, we won't have to do any additional work to add this, and it could improve QOL in some cases.

google-cloud-logging/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Release History
 
+### 1.6.0 / 2019-01-22
+
+- Default `::Logging::Entry#insert_id` to pseudo-random value (#2799)
```
blowmage: I do not like listing pull requests in the changelog.

We have used `*` for bullet lists before this entry, why the change?

blowmage: When you say `::Logging` you refer to the `Logging` in the top-level namespace, not `Google::Cloud::Logging`.

blowmage:
> it could improve QOL in some cases

I don't think referring to GitHub PRs in the changelog is beneficial because it makes it easy to not put enough information in the changelog, and the changelog loses value. It also ties the gem to GitHub, and the links are only generated when viewing the file on GitHub. If the file is viewed locally, or on the hosted docs, or a third party doc host, those links aren't created. Not to mention if the project ever changes where or how it is hosted.

The changelog's purpose is to describe the changes to the library since the last release, not link to how the changes were made.

PR #2834, opened 2019-01-23
Add field_mask argument to get_all method

### LINKED ISSUE #2794: Firestore: Add field_mask argument to get_all method
From the design document:

> The getAll() method takes an array of DocumentReferences and returns a list of results in order. If a document is not found, a DocumentSnapshot with no backing data is returned in its place. An optional FieldMask can be provided, which translated to a DocumentMask in the BatchGetDocuments RPC.

blowmage:
Allows specific portions of the document data to be returned.

[closes #2794]

### DIFF
--- a/google-cloud-firestore/lib/google/cloud/firestore/client.rb
+++ b/google-cloud-firestore/lib/google/cloud/firestore/client.rb
@@ -173,6 +173,17 @@ module Google
         # @param [String, DocumentReference, Array<String|DocumentReference>]
         #   docs One or more strings representing the path of the document, or
         #   document reference objects.
+        # @param [Array<String|FieldPath>] field_mask One or more field path
+        #   values, representing the fields of the document to be returned. If a
+        #   document has a field that is not present in this mask, that field
+        #   will not be returned in the response. All fields are returned when
+        #   the mask is not set.
+        #
+        #   A field path can either be a {FieldPath} object, or a dotted string
+        #   representing the nested fields. In other words the string represents
+        #   individual fields joined by ".". Fields containing `~`, `*`, `/`,
+        #   `[`, `]`, and `.` cannot be in a dotted string, and should provided
+        #   using a {FieldPath} object instead. (See {#field_path}.)
         #
         # @yield [documents] The block for accessing the document snapshots.
         # @yieldparam [DocumentSnapshot] document A document snapshot.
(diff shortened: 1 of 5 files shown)


google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -173,6 +173,17 @@ def doc document_path
         # @param [String, DocumentReference, Array<String|DocumentReference>]
         #   docs One or more strings representing the path of the document, or
         #   document reference objects.
+        # @param [Array<String|FieldPath>] field_mask One or more field path
```
quartzmo: Should `FieldPath` be fully-qualified here?

google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -190,16 +201,37 @@ def doc document_path
...
-        def get_all *docs
+        # @example Get docs using a field mask:
+        #   require "google/cloud/firestore"
+        #
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get and print city documents
+        #   cities = ["cities/NYC", "cities/SF", "cities/LA"]
+        #   firestore.get_all(cities, field_mask: [:population]).each do |city|
+        #     puts "#{city.document_id} has #{city[:population]} residents."
```
quartzmo: Just to be clear, `document_id` not a field, correct? Is it worth a comment to explain this since otherwise the example seems to be a contradiction of the docs? ("If a document has a field that is not present in this mask, that field will not be returned in the response.")

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
```diff
@@ -102,18 +113,41 @@ def firestore
...
             coalesce_doc_path_argument doc_path
           end
+          mask = Array(field_mask).map do |field_path|
+            if field_path.is_a? FieldPath
+              field_path.formatted_string
+            else
+              FieldPath.parse(field_path).formatted_string
+            end
+          end
+          mask = nil if mask.empty?
```
quartzmo: Refactor this (and the similar code in client.rb) to `Convert` to remove duplication?

google-cloud-firestore/lib/google/cloud/firestore/client.rb
blowmage: It does not need to be. YARD finds the constant in the current namespace. This is consistent with how existing YARD docs reference `FieldPath`.

google-cloud-firestore/lib/google/cloud/firestore/client.rb
blowmage: Correct, `document_id` is not a field. I think the code example is pretty clear. If you disagree can you make a suggestion?

google-cloud-firestore/lib/google/cloud/firestore/transaction.rb
blowmage: I don't think this benefits that much from a well named abstraction. I considered it and decided to leave it inline, matching other existing instances of these calls in the code.

```ruby
# Compare:
FieldPath.parse(field_path).formatted_string
# to:
Convert.field_path_to_formatted_string(field_path)
```

PR #2841, opened 2019-01-24
generate talent gem

TheRoyalTnetennba:
* Initial gapic generation for talent
* Generate helper methods and tests
* Add kokoro configs for talent


google-cloud-logging/.autotest
```diff
@@ -1,15 +0,0 @@
-# -*- ruby -*-
```
blowmage: Why was this file deleted?

TheRoyalTnetennba: It isn't included in the other gapic-only gems, and it seems less useful for gapic-only gems. What are your thoughts?

blowmage: I don't understand why this PR adding a new gem is removing files from an existing gem.

TheRoyalTnetennba: Lol didn't notice the path. Adding it back.

Rakefile
```diff
@@ -699,15 +699,6 @@ def generate_kokoro_configs
           config = ERB.new(File.read("./.kokoro/templates/#{os_version}.cfg.erb"))
           f.write(config.result(binding))
         end
-
```
blowmage: Why was this removed?

TheRoyalTnetennba: It should've been removed earlier. For kokoro tasks there is a github side and a google3 side. In my ideal world you could be able to generate configs for both from this repo. I'm not at all sure that that is allowed. The code I deleted is an incomplete, leftover fragment from that exploration.

blowmage: kk

PR #2879, opened 2019-02-04
Google::Cloud::Dialogflow::V2::IntentsClient#batch_update_intents example

fuminori-ido:
Google::Cloud::Dialogflow::V2::IntentsClient#batch_update_intents document was too short for me to understand so that I added an example for future reference.


fuminori-ido:
I signed it again with my github account

PR #2880, opened 2019-02-04
Google::Cloud::Dialogflow::V2::IntentsClient#batch_update_intents example

fuminori-ido:
Google::Cloud::Dialogflow::V2::IntentsClient#batch_update_intents document was too short for me to understand so that I added an example for future reference.


quartzmo:
We can probably accept this change, but because this client is generated, I am guessing that it will not survive the next client update.

fuminori-ido:
> because this client is generated, I am guessing that it will not survive the next client update.

Oh, that is unfortunate... In that case, It would be good for all of `batch_*` API user to see detailed example (like I added here) in source of the generated code.

quartzmo:
@fuminori-ido I think the best thing to do is for you to move this request to an issue. I am closing the PR because it will not be a permanent solution for the generated code.

PR #2891, opened 2019-02-07
Update google-cloud-storage.gemspec

quartzmo:
Upgrade `google-api-client` dependency to `~> 0.26` to provide `Bucket::IamConfiguration`.

I visually inspected and ran automated tests against the `google-api-client` versions `0.25.0` and `0.26.0` to confirm the need for this change.


frankyn:
Thanks @quartzmo!

PR #2894, opened 2019-02-07
Fix test coverage task

blowmage:
* Add talent gem to appropriate Gemfile files.
* Add talent gem to google-cloud gemspec.
* Add irm gem to appropriate Gemfile files.


blowmage:
Agreed.

blowmage:
It might be worth repeating the information in both places. This isn't the first time the step has been skipped.

quartzmo:
I think moving it to the stepset will improve its visibility a lot and fix the problem.

blowmage:
The problem is actually adding a new gem to the repo, not the initial release. The gem needs to be added to the top-level Gemfile when a new gem is added to the repo. It doesn't need to be added to the google-cloud or glcoud Gemfiles until the gem is added as a dependency to google-cloud's gemspec, which one should do during the gem's first release.

The problem we are seeing is new gems are added to the repo without adding the gem to the top-level Gemfile, so the coverage task breaks.

PR #2904, opened 2019-02-08
Update RELEASING.md

quartzmo:
Add top-level Gemfile to new gem stepset.

See discussion in #2894.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -50,10 +50,9 @@ green, you may create a release as follows:
    significant changes. (For examples of what a significant change is, browse
    the changes in the gem's `CHANGELOG.md`.)
 
-1. If your gem is new, ensure that it has been added to the [top-level
-   `Gemfile`](https://github.com/googleapis/google-cloud-ruby/blob/master/Gemfile).
-   Follow the steps in [Adding a new gem to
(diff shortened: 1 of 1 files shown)


blowmage:
I'll repeat my comment from #2894: The problem is new gems are being added to the repo but not added to the top-level Gemfile. My biggest concern here is not releasing a new gem, but ensuring new gems are configured properly when they are first added to the repo. Some gems live in the repo for months before their initial release. The problem is new gems in repo without being added to the top-level Gemfile causes the test coverage task to break.

Perhaps we need a different guide, specific to adding new gems to the repo, but before they are released?

quartzmo:
That would seem to belong in .github/CONTRIBUTING.md.

blowmage:
I don't know. Most contributors are making changes to an existing gem, I think instructions for adding a new gem might be too much information to throw on them. I think we should have a separate guide/checklist for adding a new gem, whether it is done by a Googler or an external contributor. We can link to this new guide from CONTRIBUTING if we want.

TheRoyalTnetennba:
I agree with mentioning/linking this in CONTRIBUTING.

quartzmo:
A new guide with a stepset dedicated to adding a new gem sounds good to me. Any objections to merging this PR now?

PR #2906, opened 2019-02-08
Remove circle and simplify release process

TheRoyalTnetennba:
* Remove .circleci folder
* Remove :circleci namespace from Rakefile
* Rewrite kokoro:release task to rely on :release task
* Kokoro builds now check to make sure that every library has been added to the top-level Gemfile


Rakefile
```diff
@@ -624,29 +596,28 @@ namespace :kokoro do
...
-        sh "bundle exec rake build"
-        gem = Dir.entries("pkg").select { |entry| File.file? "pkg/#{entry}" }.first
-        path = FileUtils.mkdir_p(File.expand_path("~") + "/.gem")
-        File.open("#{path.first}/credentials", "w") do |f|
-          f.puts "---"
-          f.puts ":rubygems_api_key: #{ENV["RUBYGEMS_API_TOKEN"]}"
-        end
-        sh "chmod 0600 #{path.first}/credentials"
-        sh "gem push pkg/#{gem}"
+        version = `bundle exec gem list`
```
blowmage: We know what `ENV["PACKAGE"]` is, presumably because it is set when starting the release build. We know what the version is from the release PR. Why don't we have `ENV["VERSION"]` here that is populated when the release build is started and to make sure that the version matches. Grabbing the version from the gem list negates the benefit of having the check, doesn't it?

TheRoyalTnetennba: I agree that having `ENV["VERSION"]` set by kokoro is a better flow. It isn't currently and I believe would require some adjustments to autorelease/releasetool to make it work. I will try to get things working that way in the long run. For now, this at least makes it unnecessary to publish docs manually for each release.

Issue #2885, opened 2019-02-06
ResourceManager does not support Folders or specifying Parents for Projects


wfarr:
**Is your feature request related to a problem? Please describe.**

Our GCP organization makes use of IAM on a folder basis. Some of these libraries (eg. ResourceManager) do not appear to be parent-aware at the moment, which means we're best off digging into lower-level abstractions.

 **Describe the solution you'd like**

I'd like for ResourceManager to be both support CRUD operations on Folders, as well as allow specifying a parent when creating projects (and I suppose for moving projects between folders too).

 **Describe alternatives you've considered**

While there are definitely alternatives available (such as not using folders), it's not a design choice I have control over in my organization.

 **Additional context**

I'm certainly happy to contribute towards making this better. I've got monkey-patches we're using locally for the moment, but I'd like to understand how I can best work with upstream on getting this support mainline.

If the answer is simply "sign the CLA and send a Pull Request", I'd like to suss out any design concerns the team might have about adding `parent` as an optional parameters to things like `create_project`, etc.

wfarr:
Sure thing!

``` ruby
resource_manager = Google::Cloud::ResourceManager.new

# this works today
resource_manager.create_project("tokyo-rain-123", name: "test project")

# this does not work because parent is not passed down to the underlying `Google::Apis::CloudresourcemanagerV1::CloudResourceManagerService`
resource_manager.create_project("tokyo-rain-123", name: "test project", parent: { "type": "folder", "id": "12345"})

# this does not work today, because Folders are not implemented in the ResourceManager here yet
resource_manager.create_folder("this is a display_name")

# this similarly does not work
resource_manager.create_folder("this is a display_name", parent: { "type": "folder", "id": "12345" })

# this also does not work yet
resource_manager.create_folder("this is a display_name", parent: { "type": "organization", "id": "12345" })
```

Happy to answer any further questions!

blowmage:
Because you say you have monkey-patches in place already, do you have a preference for remaining on the V1 API? Or would you prefer to use the V2 API?

wfarr:
> Because you say you have monkey-patches in place already, do you have a preference for remaining on the V1 API? Or would you prefer to use the V2 API?

A brief glance seems to indicate V2 has the Folder support we'd hope for, but does not contain Projects. V1 contains projects, but lacks Folder support for CRUD operations (as best I can tell with a quick glance through the auto generated code).

Seems like we'll probably end up needing to use both in this case.

[...]

PR #2910, opened 2019-02-11
Resource Manager: Project Parent Resource

blowmage:
The PR adds the parent resource to Project. This was added to the Project API a while ago, but never exposed in the Ruby gem. The parent resource to be set on a project's creation, as well an a project's update.

* Add `Project#parent`
* Add `project` optional named argument to `Manager#create_project`
* Add `Resource` class
* Add `Manager#resource` convenience method

This PR is a partial implementation for #2885, but it only affects the aspects of the API already covered by the gem. The other API features such as Folders and Organizations are still not supported by the Ruby gem.

[refs #2885]


wfarr:
Tested this out in a project where we're making projects underneath folders and it works as advertised. 👍

PR #2926, opened 2019-02-13
Release google-cloud-datastore 1.5.1

blowmage:
* Add `ReadOnlyTransaction` convenience methods:
  * Add `ReadOnlyTransaction#query`
  * Add `ReadOnlyTransaction#gql`
  * Add `ReadOnlyTransaction#key`

### DIFF
--- a/google-cloud-datastore/lib/google/cloud/datastore/version.rb
+++ b/google-cloud-datastore/lib/google/cloud/datastore/version.rb
@@ -16,7 +16,7 @@
 module Google
   module Cloud
     module Datastore
-      VERSION = "1.5.0".freeze
+      VERSION = "1.5.1".freeze
     end
   end
 end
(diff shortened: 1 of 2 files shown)


google-cloud-datastore/lib/google/cloud/datastore/version.rb
```diff
@@ -16,7 +16,7 @@
 module Google
   module Cloud
     module Datastore
-      VERSION = "1.5.0".freeze
+      VERSION = "1.5.1".freeze
```
quartzmo: `1.6.0` since new functionality is introduced to the public API?

blowmage: I considered that. But ultimately felt like this wasn't adding new functionality as much as a new way to access existing functionality. So I felt that a patch release was appropriate.

PR #2936, opened 2019-02-20
Fix label_detection helper smoke test

blowmage:
This change changes the helper tests so they pass and the build will succeed.


quartzmo:
What do you think about relaxing the assertion, maybe changing to `refute labels.empty?` or something like that?

blowmage:
@TheRoyalTnetennba ^ What do you think about Chris' question?

TheRoyalTnetennba:
`refute labels.empty?` SGTM

blowmage:
Okay, updated to use `refute_empty labels`. PTAL.

PR #2962, opened 2019-02-26
Fetch credentials for magic github proxy

busunkim96:
These credentials will be used by the releasetool reporter.


busunkim96:
@TheRoyalTnetennba Would you be able to merge this? I don't have write permission.

PR #2979, opened 2019-03-04
Protobuf 3.7

blowmage:
The google-protobuf 3.7.0 release over the weekend fixes many issues, but it also changes the JSON structure used in our tests. This PR updates the tests to use the new JSON structure for `Timestamp` and `Google::Protobuf::Duration`, and moves Spanner tests off of JSON due to the numerous changes for JSON serialization and deserialization of `Google::Protobuf::Type`.

### DIFF
--- a/google-cloud-spanner/test/google/cloud/spanner/pool/keepalive_or_release_test.rb
+++ b/google-cloud-spanner/test/google/cloud/spanner/pool/keepalive_or_release_test.rb
@@ -30,19 +30,16 @@ describe Google::Cloud::Spanner::Pool, :keepalive_or_release, :mock_spanner do
   let :results_hash do
     {
       metadata: {
-        rowType: {
+        row_type: {
           fields: [
-            { type: { code: "INT64" } }
+            { type: { code: :INT64 } }
           ]
         }
       },
-      values: [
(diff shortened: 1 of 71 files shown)


google-cloud-spanner/test/google/cloud/spanner/pool/keepalive_or_release_test.rb
```diff
@@ -1,126 +1,123 @@
...
+#
+#     mock = Minitest::Mock.new
+#     session.service.mocked_service = mock
+#
+#     pool.keepalive_or_release!
+#
+#     shutdown_pool! pool
+#
+#     mock.verify
+#   end
```
quartzmo: Why is this file commented out?

TheRoyalTnetennba:
OSx is passing 🎉

google-cloud-spanner/test/google/cloud/spanner/pool/keepalive_or_release_test.rb
blowmage: I was getting segfaults on Ruby 2.5, so I commented this test out to run some tests. I should have restored the file before committing. Thanks for pointing this out. Fixed now.

PR #2992, opened 2019-03-07
generate scheduler v1 library

TheRoyalTnetennba:


blowmage:
Are there acceptance tests for this API? Do we know if the V1 endpoints work?

PR #2993, opened 2019-03-08
generate scheduler v1 library

TheRoyalTnetennba:


TheRoyalTnetennba:
> Are there acceptance tests for this API? Do we know if the V1 endpoints work?

@blowmage looking into acceptance tests

PR #2996, opened 2019-03-08
Experimental: generate container v1beta1 with google-style

TheRoyalTnetennba:
* Generate container v1beta1
* Add google-style gem
* Remove excluded, run `rubocop -a`, and re-add excluded

For a baseline generation, see: https://github.com/googleapis/google-cloud-ruby/pull/2995


google-cloud-container/lib/google/cloud/container.rb
```diff
@@ -90,13 +90,13 @@ module Cloud
     module Container
       # rubocop:enable LineLength
 
-      FILE_DIR = File.realdirpath(Pathname.new(__FILE__).join("..").join("container"))
+      FILE_DIR = File.realdirpath Pathname.new(__FILE__).join("..").join("container")
```
blowmage: These changes are just going to be overwritten on the next generation, aren't they?

google-cloud-container/lib/google/cloud/container/v1/cluster_manager_client.rb
```diff
@@ -47,7 +47,7 @@ class ClusterManagerClient
           DEFAULT_SERVICE_PORT = 443
 
           # The default set of gRPC interceptors.
-          GRPC_INTERCEPTORS = []
+          GRPC_INTERCEPTORS = [].freeze
```
blowmage: We don't want to freeze this. This is where users configure interceptors to use.

google-cloud-container/lib/google/cloud/container.rb
TheRoyalTnetennba: No, I added `rubocop -a` to the actual synth process. Your other comment points out a good example of why this can be problematic. I thought it would at least be interesting.

PR #3034, opened 2019-03-19
add instance aliases for class methods

TheRoyalTnetennba:
* Alias `Google::Cloud::Tasks::{Version}::CloudTasksClient.location_path` as instance method.
* Alias `Google::Cloud::Tasks::{Version}::CloudTasksClient.project_path` as instance method.
* Alias `Google::Cloud::Tasks::{Version}::CloudTasksClient.queue_path` as instance method.
* Alias `Google::Cloud::Tasks::{Version}::CloudTasksClient.task_path` as instance method.

### DIFF
--- a/google-cloud-tasks/lib/google/cloud/tasks/v2beta2/helpers.rb
+++ b/google-cloud-tasks/lib/google/cloud/tasks/v2beta2/helpers.rb
@@ -1,0 +1,56 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


google-cloud-tasks/lib/google/cloud/tasks/v2beta2/helpers.rb
```diff
@@ -0,0 +1,56 @@
...
+ module Google
+  module Cloud
+    module Tasks
+      module V2beta2
+        class CloudTasksClient
+          # Alias for Google::Cloud::Tasks::V2beta2::CloudTasksClient.location_path.
+          # @param project [String]
+          # @param location [String]
+          # @return [String]
+          def location_path project, location
```
blowmage: We had discussed switching the helper methods to named arguments instead of positional arguments. Do we want to define these custom methods using named parameters?

TheRoyalTnetennba: I feel like I missed that discussion. I'm fine with that... though it would be inconsistent to do that now - irm, kms, scheduler, talent, and vision are all using positional arguments

blowmage: Ah, makes sense then.

Issue #3064, opened 2019-03-28
Can not obtain a signed url for a file with a space in its name


prapicault:
#### Environment details
  - OS: MacOS
  - Ruby version: 2.4.4
  - Gem name and version: google-cloud-storage 1.17.0

#### Steps to reproduce
1 - Create a bucket `my-todo-app`
2 - Create a folder called `foo`
3 - Add a file called `file with space.txt`
4 - Generate a signed_url following the code at http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-storage/latest/#Using_Signed_URLs 
5 - Grab the URL generated in the previous step and open it
6 - Notice you get an error saying that the `specified key does not exist`. Also notice that the detailed error message says "No such object: my-todo-app/foo/file+with+space.txt"

#### Code example

```ruby
require "google/cloud/storage"

storage = Google::Cloud::Storage.new

bucket = storage.bucket "my-todo-app"
file = bucket.file "foo/file with space.txt"
shared_url = file.signed_url method: "GET",
                             expires: 300 # 5 minutes from now
```

Thanks!

nagarjunac:
I am facing the same issue.  I have found out the cause for it. 
https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
On line no 46 => "CGI.escape node" is substituting 'spaces' in filename with '+' instead of unicode char '%20'.
By overriding that line with  "Addressable::URI.escape node" is working for me.

prapicault:
Thanks @nagarjunac. I had noticed that line of code as well and I tried URI.escape without success. Glad you found the `Addressable::URI`
I'll look into providing a patch for this.

blowmage:
How are you pulling the signed url? Is your client replacing the `+` in the path with `%20`?

prapicault:
> How are you pulling the signed url? Is your client replacing the `+` in the path with `%20`?

I'm copying and pasting into chrome.

blowmage:
Is it consistently erring in every client? Does it work for you in Firefox or curl or wget?

nagarjunac:
Yes it is consistently erring in every client. It is not working in Firefox and curl as well.

prapicault:
While researching this, I've also played with the bucket webclient, and the API https://cloud.google.com/storage/docs/json_api/v1/buckets/get and you can notice that the whitespaces are encoded with `%20`, not with `+`. HTH.

[...]

PR #3081, opened 2019-03-28
Encode object name with URI instead of CGI

prapicault:
Hi, this is a fix for issue #3064.
I've updated the code and the tests, and I have also tested the change in my setup.
The only thing that I have not been able to test is whether the bucket name needs to be encoded using CGI or URI.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
@@ -42,10 +42,7 @@ module Google
           ##
           # The external path to the file.
           def ext_path
-            escaped_path = String(@path).split("/").map do |node|
-              Addressable::URI.escape node
-            end.join("/")
-            "/#{Addressable::URI.escape @bucket}/#{escaped_path}"
+            Addressable::URI.escape "/#{@bucket}/#{@path}"
           end
 
           ##
--- a/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/signer.rb
@@ -43,9 +43,9 @@ module Google
           # The external path to the file.
           def ext_path
             escaped_path = String(@path).split("/").map do |node|
-              CGI.escape node
+              Addressable::URI.escape node
             end.join("/")
-            "/#{CGI.escape @bucket}/#{escaped_path}"
+            "/#{Addressable::URI.escape @bucket}/#{escaped_path}"
           end
 
           ##
--- a/google-cloud-storage/google-cloud-storage.gemspec
+++ b/google-cloud-storage/google-cloud-storage.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "google-api-client", "~> 0.26"
   gem.add_dependency "googleauth", ">= 0.6.2", "< 0.10.0"
   gem.add_dependency "digest-crc", "~> 0.4"
+  gem.add_dependency "addressable", "~> 2.6.0"
 
   gem.add_development_dependency "minitest", "~> 5.10"
   gem.add_development_dependency "minitest-autotest", "~> 1.0"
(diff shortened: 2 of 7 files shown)


quartzmo:
Hi @prapicault, thank you for this PR!

> The only thing that I have not been able to test is whether the bucket name needs to be encoded using CGI or URI.

I do not believe the bucket name needs to be encoded with either, since it contains only [lowercase letters, numbers, dashes (-), underscores (_), and dots (.)](https://cloud.google.com/storage/docs/naming). (See also [Encoding URI path parts](https://cloud.google.com/storage/docs/json_api/#encoding).)

@frankyn Does this sound correct to you? Also, will this PR introduce a breaking change for users?

google-cloud-storage/lib/google/cloud/storage/file/signer.rb
```diff
@@ -43,9 +43,9 @@ def self.from_bucket bucket, path
           # The external path to the file.
           def ext_path
             escaped_path = String(@path).split("/").map do |node|
```
blowmage: This whole method can be replaced with:

```ruby
Addressable::URI.escape "/#{@bucket}/#{@path}"
```

blowmage:
Can you also add a direct dependency for the addressable gem to the `google-cloud-storage.gemspec` file?

google-cloud-storage/lib/google/cloud/storage/file/signer.rb
prapicault: Done

prapicault:
> Can you also add a direct dependency for the addressable gem to the `google-cloud-storage.gemspec` file?
Done

prapicault:
Still on working in the CLA :(

frankyn:
LGTM,  found a related issue in the google-cloud-python lib: https://github.com/googleapis/google-cloud-python/issues/3809

Percent encode the object name -- except '/' which is done by the split. Please don't remove that.

google-cloud-storage/lib/google/cloud/storage/file/signer.rb
```diff
@@ -42,10 +42,7 @@ def self.from_bucket bucket, path
           ##
           # The external path to the file.
           def ext_path
-            escaped_path = String(@path).split("/").map do |node|
-              CGI.escape node
-            end.join("/")
-            "/#{CGI.escape @bucket}/#{escaped_path}"
+            Addressable::URI.escape "/#{@bucket}/#{@path}"
           end
```
quartzmo: @frankyn I think we are relying here on the behavior (seen in the unit tests below) of `Addressable::URI.encode` to not encode forward slashes at all.

frankyn: ah, okay, I take back my last comment about not removing the split. Thanks for clarifying @quartzmo.

quartzmo:
Can you please replace `require "cgi"` with `require "addressable/uri"` in `signer.rb`? Thank you!

blowmage:
I believe `CGI.encode` is still being used in this file, so I would not remove that require.

quartzmo:
True, in `url_escape`. Should it be?

blowmage:
`url_escape` should probably be renamed `param_escape`, and it needs to use CGI. Its different because query strings values are escaped slightly different than paths.

quartzmo:
@prapicault How is it going? Are you blocked by the CLA? Or anything else?

quartzmo:
@prapicault We are planning a release of google-cloud-storage, it would be great to include this change. Is there anything I can do to help?

quartzmo:
@prapicault We would like to release this fix in the next few working days, are you making any progress with the CLA?

quartzmo:
@prapicault I'm going to re-do this fix today in my own PR, but I will credit you for the find and the idea. I hope this is acceptable. We need to release the fix. Thank you!

blowmage:
This has been superseded by #3109.

prapicault:
Sorry for not being able to follow up. I just got the go to sign the CLA.
@quartzmo Thanks for pushing this forward and for the mention.

PR #3085, opened 2019-04-01
Add Pub/Sub reload

blowmage:
Add reload to Pub/Sub Topic and Subscription. This matches how reference/resource objects are used in other gems.

### DIFF
--- a/google-cloud-pubsub/support/doctest_helper.rb
+++ b/google-cloud-pubsub/support/doctest_helper.rb
@@ -412,13 +412,7 @@ YARD::Doctest.configure do |doctest|
       mock_subscriber.expect :get_subscription, subscription_resp, [String, Hash]
     end
   end
-
-  doctest.before "Google::Cloud::PubSub::Subscription#refresh!" do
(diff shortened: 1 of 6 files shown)


google-cloud-pubsub/support/doctest_helper.rb
```diff
@@ -406,6 +406,20 @@ def mock_pubsub
...
   end
 
+  doctest.before "Google::Cloud::PubSub::Subscription#reload!" do
+    mock_pubsub do |mock_publisher, mock_subscriber|
+      mock_subscriber.expect :get_subscription, subscription_resp, [String, Hash]
+      mock_subscriber.expect :get_subscription, subscription_resp, [String, Hash]
+    end
+  end
+
+  doctest.before "Google::Cloud::PubSub::Subscription#refresh!" do
```
quartzmo: We usually skip aliases. (Not a big deal, I will approve anyway in case you are in a hurry.)

blowmage: Skipped.

PR #3086, opened 2019-04-01
Add Subscription#expires_in

### LINKED ISSUE #2775: Pub/Sub ExpirationPolicy?
PR #2764 added ExpirationPolicy to the low level API. Does this need to be exposed in the public API?

blowmage:
Add `expires_in` getter and setter methods that get/set the `ttl` value on ExpirationPolicy.

I decided to expose `expires_in` instead of an ExpirationPolicy because there is only the one field on ExpirationPolicy. If/when ExpirationPolicy adds more fields we can expose this structure while still allowing `expires_in` to directly access that value.

[closes #2775]

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
@@ -288,11 +288,9 @@ module Google
         #   to unset.
         #
         def expires_in= ttl
-          new_expiration_policy = if ttl
-                                    Google::Pubsub::V1::ExpirationPolicy.new(
-                                      ttl: Convert.number_to_duration(ttl)
-                                    )
-                                  end
+          new_expiration_policy = Google::Pubsub::V1::ExpirationPolicy.new(
+            ttl: Convert.number_to_duration(ttl)
+          )
 
           update_grpc = Google::Cloud::PubSub::V1::Subscription.new \
             name: name, expiration_policy: new_expiration_policy
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
@@ -262,7 +262,7 @@ module Google
         # long as any connected subscriber is successfully consuming messages
         # from the subscription or is issuing operations on the subscription.
         #
-        # If `expires_in` is not set, a *default* value of of 31 days will be
+        # If {#expires_in=} is not set, a *default* value of of 31 days will be
         # used. The minimum allowed value is 1 day.
         #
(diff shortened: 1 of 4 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -256,6 +256,47 @@ def labels= new_labels
           @resource_name = nil
         end
 
+        ##
+        # The duration (in seconds) for when a subscription expires after the
+        # subscription goes inactive. A subscription is considered active as
+        # long as any connected subscriber is successfully consuming messages
+        # from the subscription or is issuing operations on the subscription.
+        #
+        # If `expires_in` is not set, a *default* value of of 31 days will be
```
quartzmo: link to `{#expires_in=}` here?

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -256,6 +256,47 @@ def labels= new_labels
...
+          ensure_grpc!
+
+          return nil if @grpc.expiration_policy.nil?
+
+          Convert.duration_to_number @grpc.expiration_policy.ttl
+        end
+
+        ##
+        # Sets the duration (in seconds) for when a subscription expires after
+        # the subscription goes inactive.
```
quartzmo: Add "See also #{expires_in}" here?

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -256,6 +256,47 @@ def labels= new_labels
...
+        #
+        # @param [Numeric, nil] ttl The expiration duration, or `nil` to unset.
```
quartzmo: what is the unit? days? seconds? millis?

quartzmo: Ah, I see now `seconds` in the description, sorry. But wouldn't hurt to repeat it.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: Added.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: Added.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
blowmage: Added.

PR #3153, opened 2019-04-22
remove public_suffix pin and bump rake version

TheRoyalTnetennba:


quartzmo:
This change was originally made in #2130 for reasons discussed here: https://github.com/googleapis/google-cloud-ruby/pull/2130#discussion_r195189450

PR #3237, opened 2019-04-27
fix pubsub tests

TheRoyalTnetennba:
* topics and subscriptions both default to 100. Updated acceptance tests to accomodate this.
* Increase subscriber retries for acknowledge_test to reduce flakiness

### DIFF
--- a/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
+++ b/google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
@@ -46,7 +46,7 @@ describe Google::Cloud::PubSub, :pubsub do
   describe "Topic", :pubsub do
 
     it "should be listed" do
-      topics = pubsub.topics
+      topics = pubsub.topics.all
       topics.each do |topic|
         topic.must_be_kind_of Google::Cloud::PubSub::Topic
       end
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/test/google/cloud/pubsub/subscriber/acknowledge_test.rb
```diff
@@ -97,7 +97,7 @@
     while called < 3
```
blowmage: I would rather see this value increased.

google-cloud-pubsub/acceptance/pubsub/pubsub_test.rb
```diff
@@ -46,15 +46,15 @@ def retrieve_snapshot project, subscription, snapshot_name
   describe "Topic", :pubsub do
 
     it "should be listed" do
-      topics = pubsub.topics
+      topics = pubsub.topics max: 1000
```
blowmage: This will only return the first page, so you would probably want to call `.all` to get them all.

blowmage:
Have you tried running the cleanup task?

TheRoyalTnetennba:
@blowmage is this good to go?

PR #3245, opened 2019-04-29
Fix Spanner session limit bug

### LINKED ISSUE #3243: Spanner Client and SessionLimitError
We are seeing `SessionLimitError`s in production. The errors don't start showing up until we've been running in production for a while without restarting. When the errors do start showing up, they all come from the same process on the same server and go away as soon as we restart that process. We believe this is a problem in how the `google-cloud-spanner` library is managing its session pool.

I'll explain how we're using `google-cloud-spanner`, what we're seeing in production, and what we think is causing the problem.

Here is how we initialize our clients:

```ruby
client = Google::Cloud::Spanner.new(
  # The `client_config` settings are for disabling google-gax's retry behavior. We also
  # monkey patch google-gax to fix a bug that has now been fixed on master, but won't
  # be backported to the current major release. I'm happy to provide the code for the
  # monkey patch if that's helpful, but I don't think it's relevant here. Just wanted to
  # provide this config in the spirit of full disclosure.
  client_config: {
    'interfaces' => {
      'google.spanner.v1.Spanner' => {
        'retry_codes' => {},
        'retry_params' => {},
      },
    },
  },
  credentials: credentials,
  project_id: project_id,
  timeout: timeout.to_f,
).client(
  instance,
  database,
  pool: {
    min: 3,
    max: 10,
  },
)
```

I believe the pool settings are the most relevant part of this:

```ruby
pool: {
  min: 3,
  max: 10,
}
```

Because of the way we run our servers, there is no valid case we know of for having more than 10 sessions open at the same time. `min` is set to 3 to pre-allocate Spanner sessions at startup time. The `fail` option uses its default value of `true`, which is what we want because allocating more than 10 sessions means something is behaving incorrectly.

As for what we're seeing in production, this problem showed up again recently and we were able to pull some data from the process throwing the errors.

Assuming we have a `Google::Cloud::Spanner::Client` instance referenced by a `spanner_client` variable, then we can the `Google::Cloud::Spanner::Pool` object for that client as:

```ruby
session_pool = spanner_client.instance_variable_get(:@pool)
```

On the process throwing the errors, this is what we found:

```ruby
session_pool.instance_variable_get(:@all_sessions).count # => 0
session_pool.instance_variable_get(:@session_queue).count # => 0
session_pool.instance_variable_get(:@transaction_queue).count # => 0
session_pool.instance_variable_get(:@new_sessions_in_process) # => 10
```

Processes that weren't throwing the error look like this:

```ruby
session_pool.instance_variable_get(:@all_sessions).count # => 4
session_pool.instance_variable_get(:@session_queue).count # => 4
session_pool.instance_variable_get(:@transaction_queue).count # => 0
session_pool.instance_variable_get(:@new_sessions_in_process) # => 0
```

The value of `10` for `@new_sessions_in_process` is significant because that's the value we set for the `max` option in the pool config.

Because the `session_queue` array is empty, we try to create a new session, but `can_allocate_more_sessions?` is defined on 1.8.0 as:

```ruby
all_sessions.size + @new_sessions_in_process < @max
```

`@new_sessions_in_process` is already equal to `@max` when we see these errors, so naturally the code will raise `SessionLimitError`.

Why is this happening? We're not sure. We haven't been able to forcibly reproduce this issue, but we have a partial idea which focuses on how`@new_sessions_in_process` doesn't always get decremented when we think it should.

When the `Pool` object is initialized, `@new_sessions_in_process` is set to `@min`, which is `3` in our case. After that, `@new_sessions_in_process` can is only incremented when `checkout_session` and `checkout_transaction` are called. `@new_sessions_in_process` is only decremented when `new_session!` is called.

When `checkout_session` or `checkout_transaction` are called, `@new_sessions_in_process` is incremented and then `new_session!` is called. `new_session!` includes a call to `@client.create_new_session`. If `@client.create_new_session` errors for any reason, `@new_sessions_in_process` doesn't get decremented. If this happens `@max` times, then `@new_sessions_in_process` will equal `@max` and trying to open a new session will throw a `SessionLimitError`.

Here is a reproducible example:

```ruby
# Redefine create_new_session to always raise an exception. This simplifies the
# example significantly.
module Google
  module Cloud
    module Spanner
      class Client
        def create_new_session
          raise 'Example exception'
        end
      end
    end
  end
end

spanner_client = Google::Cloud::Spanner.new(
  credentials: credentials,
  project_id: project_id,
  timeout: timeout,
).client(
  instance,
  database,
  pool: {
    fail: true,
    min: 0,
    max: 10,
  },
)

session_pool = spanner_client.instance_variable_get(:@pool)

# Because `@min` is set to `0`, no sessions are created during initialization.
session_pool.instance_variable_get(:@new_sessions_in_process) # => 0

# Force a new session to be created `@max` times. Since we've redefined
# `create_new_session` to always raise, this will always fail.
10.times do
  begin
    spanner_client.execute('SELECT 1;')
  rescue
    # noop
  end
end

# `@new_sessions_in_process` is now equal to `@max`.
session_pool.instance_variable_get(:@new_sessions_in_process) # => 10

# All future calls will throw `SessionLimitError`.
spanner_client.execute('SELECT 1;') # => Google::Cloud::Spanner::SessionLimitError
```

We believe this problem can be solved with the following change:

```diff
def new_session!
     # don't add if the pool is closed
     return session.release! if @closed

-    @new_sessions_in_process -= 1
     all_sessions << session
   end

   session
+ensure
+  @new_sessions_in_process -= 1
 end
```

We're not sure whether this is part of the problem we're seeing. If this were the only problem then we'd be able to continue using the three sessions that are created at startup time, but per the example output above we took from production, `@all_sessions`, `@session_queue`, and `@transaction_queue` are all empty.

There are two reasons this could happen:

1. The sessions have been idle long enough that they get released.
2. The sessions get checked out and don't get checked back in.

(2) seems unlikely based on looking at the with_session/checkout/checkin code.

(1) seems more likely, but it's unclear how this combines with the `@new_sessions_in_process` issues pointed out above.

That's as far as we've gotten in debugging this issue and we'd be very interested in any thoughts on how to solve this problem and whether we're on the right track.

P.S. Please feel free to change the title of this issue if you can think of a better description for this problem.

#### Environment details

  - OS: `CentOS Linux release 7.6.1810 (Core)`
  - Ruby version: `ruby 2.4.6p354 (2019-04-01 revision 67394) [x86_64-linux]`
  - Gem name and version: `google-cloud-spanner-1.8.0`

#### Steps to reproduce

We haven't been able to intentionally reproduce this, but the longer we run in production without deploying or restarting, the more likely this problem is to show up.

#### Code example

N/A

#### Full backtrace

N/A

blowmage:
As reported in #3243, it is possible for the client to get into a bad state when errors are raised in the API call to create a new session, and handled outside of the client. If this happens too often, then the client will no longer allow new sessions to be created and raise `SessionLimitError`.

This PR attempts to solve this by decrementing the count when an error is raised.

[fixes #3243]

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -243,7 +243,14 @@ module Google
             @new_sessions_in_process += 1
           end
 
-          session = @client.create_new_session
+          begin
+            session = @client.create_new_session
+          rescue StandardError => e
+            @mutex.synchronize do
+              @new_sessions_in_process -= 1
+            end
+            raise e
+          end
 
           @mutex.synchronize do
             @new_sessions_in_process -= 1
--- a/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/pool.rb
@@ -71,7 +71,6 @@ module Google
               return write_transaction.session if write_transaction
 
               if can_allocate_more_sessions?
-                @new_sessions_in_process += 1
                 action = :new
                 break
               end
@@ -127,7 +126,6 @@ module Google
               end
 
               if can_allocate_more_sessions?
-                @new_sessions_in_process += 1
                 action = :new
                 break
               end
@@ -204,7 +202,7 @@ module Google
           # init the thread pool
           @thread_pool = Concurrent::FixedThreadPool.new @threads
           # init the queues
-          @new_sessions_in_process = @min.to_i
+          @new_sessions_in_process = 0
           @all_sessions = []
           @session_queue = []
           @transaction_queue = []
@@ -241,15 +239,17 @@ module Google
         end
 
         def new_session!
+          @mutex.synchronize do
+            @new_sessions_in_process += 1
+          end
+
           session = @client.create_new_session
 
           @mutex.synchronize do
-            # don't add if the pool is closed
-            return session.release! if @closed
-
             @new_sessions_in_process -= 1
             all_sessions << session
           end
+
           session
         end
 
--- a/google-cloud-spanner/test/google/cloud/spanner/pool/new_sessions_in_process_test.rb
+++ b/google-cloud-spanner/test/google/cloud/spanner/pool/new_sessions_in_process_test.rb
@@ -1,0 +1,87 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "helper"
+
+describe Google::Cloud::Spanner::Pool, :new_sessions_in_process, :mock_spanner do
+  let(:instance_id) { "my-instance-id" }
+  let(:database_id) { "my-database-id" }
+  let(:session_id) { "session123" }
+  let(:session_grpc) { Google::Spanner::V1::Session.new name: session_path(instance_id, database_id, session_id) }
+  let(:session) { Google::Cloud::Spanner::Session.from_grpc session_grpc, spanner.service }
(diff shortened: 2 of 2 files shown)


google-cloud-spanner/lib/google/cloud/spanner/pool.rb
```diff
@@ -241,15 +239,24 @@ def shutdown
         end
 
         def new_session!
-          session = @client.create_new_session
-
           @mutex.synchronize do
-            # don't add if the pool is closed
-            return session.release! if @closed
```
blowmage: This is removing the check for the client being closed. It is possible that client is blocking on a new session and when it gets unblocked the client is closed. In that case, should we not create a new session and raise here?

PR #3284, opened 2019-05-03
Fix errors in bigtable code sample

olliebutler:
The code samples for cloud bigtable have a couple of errors:
1. There is a missing comma
2. `entry` is a `Google::Cloud::Bigtable::MutationEntry` not a
`Google::Bigtable::V2::Mutation` and such does not have a method
`delete_from_column` I've changed this to `delete_cells` so the code has the same
gist but is functional :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,9 +213,9 @@ entry = table.new_mutation_entry("user-1")
 entry.set_cell(
   "cf-1",
   "field-1",
-  "XYZ"
+  "XYZ",
   timestamp: Time.now.to_i * 1000 # Time stamp in milli seconds.
-).delete_from_column("cf2", "field02")
+).delete_cells("cf2", "field02")
 
 table.mutate_row(entry)
 ```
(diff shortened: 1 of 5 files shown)


quartzmo:
@olliebutler Thanks for opening this PR! I'm just going to quickly look into whether there are any unit tests that can be updated as well. It is unfortunate that this package did not get doctest coverage like the other veneer packages. I'll open an issue for that.

olliebutler:
No worries, I spotted a similar issue with the bigquery sample code previously [here](https://github.com/googleapis/google-cloud-ruby/pull/2830). When I have some more time I'll go through and check the rest.

quartzmo:
@olliebutler Should this usage of [delete_from_column](https://github.com/googleapis/google-cloud-ruby/blob/4afc82840bdc74928912d9da2644adf7702ca5a7/google-cloud-bigtable/lib/google/cloud/bigtable/mutation_operations.rb#L220) also be updated?

quartzmo:
The comma is also missing in [this `#set_cell` example](https://github.com/googleapis/google-cloud-ruby/blob/4afc82840bdc74928912d9da2644adf7702ca5a7/google-cloud-bigtable/lib/google/cloud/bigtable/mutation_entry.rb#L99).

olliebutler:
@quartzmo Sorry for the delay, it was a holiday weekend in England, I've fixed those additional issues you spotted.

PR #3344, opened 2019-05-08
datastore - use version.rb in place of Gem.loaded_specs

TheRoyalTnetennba:


google-cloud-datastore/lib/google/cloud/datastore/v1/datastore_client.rb
```diff
@@ -126,7 +127,7 @@ def initialize \
               updater_proc = credentials.updater_proc
             end
 
-            package_version = Gem.loaded_specs['google-cloud-datastore'].version.version
+            package_version = package_version = Google::Cloud::Datastore::VERSION
```
quartzmo: why is `package_version` assigned twice?

PR #3416, opened 2019-05-24
Firestore Collection Group queries

quartzmo:
Collection Group queries add support for queries to Firestore that query documents by collection ID rather than by collection path. With this feature, users can now query for documents in collections that share a name, no matter where these collection are located.

Java: googleapis/google-cloud-java#4652
Node: googleapis/nodejs-firestore#578
Python: googleapis/google-cloud-python#7758


quartzmo:
Rebased on master for build fix (#3417).

quartzmo:
@crwilcox Good to merge?

PR #3430, opened 2019-06-03
feat(firestore): add commit_response to Client#transaction

### LINKED ISSUE #3388: Firestore: Return user-defined result from Transaction function
Note: This is a dupe of https://github.com/googleapis/google-cloud-ruby/issues/2103

I would like to ask if we could re-consider the API for `runTransaction` and return a user-provided result. On other platforms, the SDK returns the return value of the last successful run.

You can see a couple of implementations here:
https://github.com/googleapis/nodejs-firestore/blob/master/types/firestore.d.ts#L206
https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java#L84,https://github.com/googleapis/google-cloud-php-firestore/blob/master/src/FirestoreClient.php#L394

Is it possible to update the Ruby client to match?

quartzmo:
BREAKING CHANGE: The default return value of `Client#transaction` has been
changed to **the return value of the yielded block**. Pass `commit_response: true`
for the previous default behavior of returning the `CommitResponse` with timestamp.

[fixes #3388]

### DIFF
--- a/google-cloud-firestore/lib/google/cloud/firestore/client.rb
+++ b/google-cloud-firestore/lib/google/cloud/firestore/client.rb
@@ -595,12 +595,19 @@ module Google
         #
         # @param [Integer] max_retries The maximum number of retries for
         #   transactions failed due to errors. Default is 5. Optional.
+        # @param [Integer] commit_response If true, the return value from this
+        #   method will be a `Google::Cloud::Firestore::CommitResponse` object
+        #   with a `commit_time` attribute. If omitted or false, the return
+        #   value from this method will be the return value of the provided
+        #   yield block. Default is false. Optional.
         #
         # @yield [transaction] The block for reading data and making changes.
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/lib/google/cloud/firestore/client.rb
```diff
@@ -595,12 +595,19 @@ def batch
         #
         # @param [Integer] max_retries The maximum number of retries for
         #   transactions failed due to errors. Default is 5. Optional.
+        # @param [Integer] commit_response If true, the return value from this
```
schmidt-sebastian: s/Integer/Boolean/ ?

quartzmo: Ouch. Fixed. Thank you!

PR #3447, opened 2019-06-07
feat(firestore): add CollectionReference#list_documents

### LINKED ISSUE #2828: Firestore: Add list_documents to CollectionReference
The semantics are a bit different than `CollectionReference#get`, which runs a query for documents in the collection.

From the design document:

> The listDocuments() method returns a list of DocumentReferences that are directly nested under this collection. The SDK fetches all documents from the server, but provides an empty field mask to avoid unnecessary data transfer. We always set the showMissing flag of the ListDocuments RPC to true to support full document traversal. If there are too many documents, our recommendation will be not to call this method.

quartzmo:
[fixes #2828]

Question: Should `CollectionReference#list_documents` expose `order_by`?

/cc @schmidt-sebastian

### DIFF
--- a/google-cloud-firestore/acceptance/firestore/collection_test.rb
+++ b/google-cloud-firestore/acceptance/firestore/collection_test.rb
@@ -41,4 +41,18 @@ describe "Collection", :firestore_acceptance do
 
     doc_snp[:foo].must_equal "hello world"
   end
+
+  it "lists its documents" do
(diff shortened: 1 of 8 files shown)


schmidt-sebastian:
Re: orderBy - We haven't exposed it in other clients and have gotten not (yet) customer requests for it.

PR #3449, opened 2019-06-10
Upgrade Pub/Sub to concurrent-ruby Promises framework

blowmage:
This PR moves Pub/Sub to the concurrent-ruby 1.1, and switches to the Promises framework.

Also, the use of FixedThreadPool has been replaced with CachedThreadPool, which should lower resource utilization when the library is lightly used.


quartzmo:
Related to #2706

blowmage:
@quartzmo I reverted the TimerTask commit. TimerTask is only partially replaced by the new promises framework, and is not deprecated. I think it should continue to be used as it is a really good fit for the timed buffer.

Issue #3413, opened 2019-05-24
Pub/Sub subscriber stops processing messages


mrsimo:
I'm pretty sure I'm having the same problems described in https://github.com/googleapis/google-cloud-ruby/issues/2011. Didn't want to reopen that issue, apologies if it wasn't the right decision.

I run a very simple ruby consumer where I:

* Setup a `Subscriber` and call `start` on it.
* Boot another thread where I start a basic rack http server to expose some custom prometheus metrics.
* Start an endless loop where I print some stuff every few seconds.

I observe that sometimes it just stops processing messages. For instance, you can see a couple pods where CPU usage just decreases for no reason while others keep going:

![Selection_184](https://user-images.githubusercontent.com/19203/58323538-8feb5d80-7e24-11e9-8cc0-5a78bd68fd61.png)

The http thread keeps working, and when the pod stops (because of autoscaling or whatever) it shuts down cleanly.

I added an autoscaler based on CPU a week ago, and that seemed to help, but after a week I see the same behavior. Before the autoscaler they would almost always get stuck unless I ran more consumers than I needed.

My questions, I guess:

* Any idea why this is happening? Could something internal to the pubsub gem be crashing in a way that doesn't raise anything? I've worked with ruby a long time and threads still make me shudder.
* Any way we could cleanly detect if this happens? Would calling [`Subscriber#stopped?`](https://googleapis.github.io/google-cloud-ruby/docs/google-cloud-pubsub/latest/Google/Cloud/PubSub/Subscriber.html#stopped%3F-instance_method) let us know if things have stopped processing?

I would rather not rely on "have messages been processed in the last N seconds" to know if a consumer is working fine, because we run all kinds of consumers in all kinds of environments (like staging) where I'm not sure if there's going to be a solid amount of messages coming in.

Thanks a lot!

blowmage:
Thanks for the issue. I have some questions to help us understand your environment.. What gem version are you using? What Ruby version? What platform? What other settings are being used on the subscriber? Are there any errors sent to `on_error` that coincide with the messages stopping?

mrsimo:
Ah, yes, I should have sent these from the beginning. 

I'm using `google-cloud-pubsub` 0.35.0. Ruby version:

```
$ ruby -v
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux]
```

It runs in a container built with `FROM ruby:2.5-slim` having installed only the `build-essential` package with apt.

[...]

PR #3451, opened 2019-06-10
Remove messages from inventory after callback

blowmage:
This change prevents the Subscriber Inventory from filling up when
messages are never acked or nacked in the user callback.
This might happen due to an error in the user callback code.
Removing a message from the inventory will cause the message to be
redelivered and reprocessed.

### DIFF
--- a/google-cloud-pubsub/test/google/cloud/pubsub/subscriber/inventory_test.rb
+++ b/google-cloud-pubsub/test/google/cloud/pubsub/subscriber/inventory_test.rb
@@ -1,4 +1,4 @@
-# Copyright 2017 Google LLC
+# Copyright 2019 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


quartzmo:
For background, see [related commentary](https://github.com/googleapis/google-cloud-ruby/issues/3413#issuecomment-500492648) in the subthread of #3413.

google-cloud-pubsub/test/google/cloud/pubsub/subscriber/inventory_test.rb
```diff
@@ -0,0 +1,156 @@
+# Copyright 2017 Google LLC
```
quartzmo: 2019

quartzmo:
@blowmage I definitely see the reason for this and I think it makes sense, I'm only just very mildly concerned that there could be some unintended effect for users. Can you add a line of documentation to `Subscription#listen`?

google-cloud-pubsub/test/google/cloud/pubsub/subscriber/inventory_test.rb
blowmage: Fixed.

blowmage:
> Can you add a line of documentation to `Subscription#listen`?

Added.

PR #3452, opened 2019-06-10
Cleanup Subscriber Threading Restart

blowmage:
This PR has several minor refactorings. The most consequential is the removal of the call to `start_streaming!` from the Subscriber's background thread. This method will create a second background thread, but after this method is called the first background thread will retry. This means there is potential for multiple background threads. I haven't been able to observe this however.

There are some other minor changes, refinements to the synchronize block used to keep things thread safe. But the biggest change here is to only use retry to keep the background thread alive.


blowmage:
Reading the code again, it looks like the call to `start_streaming!` was a no-op, because a background thread already exists. So it should be safe to remove.

PR #3499, opened 2019-06-17
Fix lifecycle acceptance test

blowmage:
Apiary now returns Date objects instead of String objects for date fields.


quartzmo:
What do you think about preserving the behavior of our API by converting the Apiary value from Date to String in `Lifecycle::Rule`?

blowmage:
> What do you think about preserving the behavior of our API by converting the Apiary value from Date to String in `Lifecycle::Rule`?

I don't have a strong opinion either way. As long as CI starts passing again.

quartzmo:
The docs actually specify that the value can be either String or Date, so I guess we are OK to expose the Apiary change. Pretty annoying though!

```
@attr [String,Date] created_before A date in RFC 3339 format with
          #   only the date part (for instance, "2013-01-15"). This condition is
          #   satisfied when a file is created before midnight of the specified
          #   date in UTC.
```

blowmage:
> Pretty annoying though!

Agreed.

PR #3500, opened 2019-06-17
Release google-cloud-pubsub 0.37.0

blowmage:
* Add Topic#persistence_regions
* Subscriber changes
  * Fix potential inventory bug
  * Messages are removed after callback
    * This change prevents the Subscriber inventory from filling up
      when messages are never acked or nacked in the user callback.
      This might happen due to an error in the user callback code.
      Removing a message from the inventory will cause the message to
      be redelivered and reprocessed.
  * Update concurrency implementation
    * Use concurrent-ruby Promises framework.
* Update network configuration
* Enable grpc.service_config_disable_resolution

### DIFF
--- a/google-cloud-pubsub/CHANGELOG.md
+++ b/google-cloud-pubsub/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Release History
 
+### 0.37.0 / 2019-06-17
+
+* Add Topic#persistence_regions
+* Subscriber changes
+  * Fix potential inventory bug
+    * It is possible 
+  * Messages are removed after callback
+    * This change prevents the Subscriber inventory from filling up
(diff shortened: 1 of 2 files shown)


google-cloud-pubsub/CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
 # Release History
 
+### 0.37.0 / 2019-06-17
+
+* Add Topic#persistence_regions
+* Subscriber changes
+  * Fix potential inventory bug
+    * It is possible
```
quartzmo: "It is possible"?

Is this entry for #3450?

blowmage: Fixed.

PR #3503, opened 2019-06-18
Add BigQuery Model API

### LINKED ISSUE #3348: BigQuery ML Models API
Expose the BigQuery models API.  It should implement get, list, update, delete operations, but does not direct model creation, as a `CREATE MODEL` query is how a new ML model is created.

[Python implementation](https://googleapis.github.io/google-cloud-python/latest/bigquery/generated/google.cloud.bigquery.model.Model.html#google.cloud.bigquery.model.Model)
[Go implementation](https://code-review.googlesource.com/c/gocloud/+/39090)

blowmage:
[closes #3348]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
@@ -378,9 +378,19 @@ module Google
         #   is nil.
         def self.millis_to_time time_millis
           return nil unless time_millis
-          time_millis = Integer time_millis
-          time_secs = time_millis / 1000.0
-          ::Time.at time_secs
+          ::Time.at Rational(time_millis, 1000)
+        end
+
+        ##
+        # @private
+        #
+        # Converts a Ruby Time object to a primitive time value in milliseconds.
+        #
+        # @return [Integer, nil] The primitive time value in milliseconds, or
+        #   nil if the given argument is nil.
+        def self.time_to_millis time_obj
+          return nil unless time_obj
+          (time_obj.to_i * 1000) + (time_obj.nsec / 1000000)
         end
       end
 
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -17,6 +17,7 @@ require "json"
 require "google/cloud/errors"
 require "google/cloud/bigquery/service"
 require "google/cloud/bigquery/table"
+require "google/cloud/bigquery/model"
 require "google/cloud/bigquery/external"
 require "google/cloud/bigquery/dataset/list"
 require "google/cloud/bigquery/dataset/access"
@@ -666,6 +667,90 @@ module Google
           options = { token: token, max: max }
           gapi = service.list_tables dataset_id, options
           Table::List.from_gapi gapi, service, dataset_id, max
+        end
+
+        ##
+        # Retrieves an existing model by ID.
+        #
+        # @param [String] model_id The ID of a model.
+        # @param [Boolean] skip_lookup Optionally create just a local reference
+        #   object without verifying that the resource exists on the BigQuery
+        #   service. Calls made on this object will raise errors if the resource
+        #   does not exist. Default is `false`. Optional.
+        #
+        # @return [Google::Cloud::Bigquery::Model, nil] Returns `nil` if the
+        #   model does not exist.
+        #
+        # @example
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   model = dataset.model "my_model"
+        #   puts model.model_id
+        #
+        # @example Avoid retrieving the model resource with `skip_lookup`:
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   model = dataset.model "my_model", skip_lookup: true
+        #
+        # @!group Model
+        #
+        def model model_id, skip_lookup: nil
+          ensure_service!
+          if skip_lookup
+            return Model.new_reference project_id, dataset_id, model_id, service
+          end
+          gapi = service.get_model dataset_id, model_id
+          Model.from_gapi_json gapi, service
+        rescue Google::Cloud::NotFoundError
+          nil
+        end
+
+        ##
+        # Retrieves the list of models belonging to the dataset.
+        #
+        # @param [String] token A previously-returned page token representing
+        #   part of the larger set of results to view.
+        # @param [Integer] max Maximum number of models to return.
+        #
+        # @return [Array<Google::Cloud::Bigquery::Model>] An array of models
+        #   (See {Google::Cloud::Bigquery::Model::List})
+        #
+        # @example
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   models = dataset.models
+        #   models.each do |model|
+        #     puts model.model_id
+        #   end
+        #
+        # @example Retrieve all models: (See {Model::List#all})
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   models = dataset.models
+        #   models.all do |model|
+        #     puts model.model_id
+        #   end
+        #
+        # @!group Model
+        #
+        def models token: nil, max: nil
+          ensure_service!
+          options = { token: token, max: max }
+          gapi = service.list_models dataset_id, options
+          Model::List.from_gapi gapi, service, dataset_id, max
         end
 
         ##
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/model.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/model.rb
@@ -16,6 +16,7 @@
 require "google/cloud/errors"
 require "google/cloud/bigquery/service"
 require "google/cloud/bigquery/model/list"
+require "google/cloud/bigquery/standard_sql"
 require "google/cloud/bigquery/convert"
 require "google/apis/bigquery_v2"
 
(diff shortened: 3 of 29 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/model/metrics.rb
```diff
@@ -0,0 +1,702 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+module Google
+  module Cloud
+    module Bigquery
+      class Model
+        # Information about a single training query run for the model.
+        class TrainingRun
```
quartzmo: Personally I would find it easier to keep things straight if this file were `training_run.rb`.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -668,6 +669,90 @@ def tables token: nil, max: nil
...
+        #   model does not exist.
+        #
+        # @example
+        #   require "google/cloud/bigquery"
+        #
+        #   bigquery = Google::Cloud::Bigquery.new
+        #   dataset = bigquery.dataset "my_dataset"
+        #
+        #   model = dataset.model "my_model"
+        #   puts model.name
```
quartzmo: I think `model.model_id` is better for these examples, since it is required to be provided to `CREATE MODEL`.

google-cloud-bigquery/lib/google/cloud/bigquery/model.rb
```diff
@@ -0,0 +1,722 @@
...
+        ##
+        # Load the complete representation of the model if it has been
+        # only partially loaded by a request to the API list method.
+        def ensure_full_data!
+          reload! unless resource_full?
+        end
+
+        ##
+        # Yielded to a block to accumulate changes for a patch request.
+        class Updater < Model
```
quartzmo: If Updater isn't currently used, can we remove it (for now)?

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
blowmage: What do you think about `model_type`?

blowmage:
@shollyman Would you like to look this over?

google-cloud-bigquery/lib/google/cloud/bigquery/model/metrics.rb
blowmage: Done.

google-cloud-bigquery/lib/google/cloud/bigquery/model.rb
blowmage: Removed.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
blowmage: Changed to `model.model_id`.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
```diff
@@ -378,9 +378,50 @@ def self.derive_source_format path
...
+        #   nil if the given argument is nil.
+        def self.time_to_millis time_obj
+          return nil unless time_obj
+          (time_obj.to_i * 1000) + (time_obj.nsec / 1000000)
+        end
+
+        ##
+        # @private
+        #
+        # Converts an array of Google::Apis::BigqueryV2::StandardSqlField to
```
shollyman: I'm not confident this simplification of the type representation makes sense, but BQML models are fairly restrictive in what types are currently allowed as feature and label columns.  With routines, we'll need to expose the StandardSqlStructType/StandardSQLField/StandardSQLDataType as proper resources as it's the types we'll use for exposing function arguments/return values.

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
```diff
@@ -0,0 +1,98 @@
...
+  module Cloud
+    module Bigquery
+      class Model
+        # Aggregate metrics for classification/classifier models. For
+        # multi-class models, the metrics are either macro-averaged or
+        # micro-averaged. When macro-averaged, the metrics are calculated for
+        # each label and then an unweighted average is taken of those values.
+        # When micro-averaged, the metric is calculated globally by counting the
+        # total number of correctly predicted rows.
+        class AggregateClassificationMetrics
```
shollyman: Are you generating these type definitions by hand or are you consuming something? Ex: python consumes the proto representations and just uses the json serialization functionality, java and python expose the discovery types directly with special caveats about volatility.  There's enough churn due to the ML backend change velocity that you'll want to keep out of the business of maintaining them by hand, as there's frequent edits here.

The advantage we have here is that these are output-only messages, since all resource creation is done via CREATE MODEL queries.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
blowmage: Spanner has a very similar structure to describe fields, including nested arrays and structs. In Spanner users sometimes need to specify this type information when it can't be inferred from the data correctly, like when specifying `NULL` while using query parameters. We use this particular structure to so users can specify the values using Ruby primitives, which are easier to create than the verbose objects as defined by protobuf/discovery. We thought that since this structure was useful in Spanner it would also be useful in BigQuery. Is this explanation persuasive at all?

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
blowmage: These files were written by hand, but we can change this approach and generate off the discovery document.

We have so far been very intentional about not leaking the apiary/discovery layer, but for these particular values I am not opposed. I see a few options moving forward. Which would you prefer?

1) Maintain veneer versions of these structures. Changes to the structures in the discovery document will require changes to these files. This can be manual or automated, but will require changes to this gem for new data to be accessed.
2) Return Arrays and Hashes of values instead of documented objects. This can either be the apiary Hash representation, or the API's JSON structure. (`training_options` vs. `trainingOptions`) This means new data can be accessed when available.
3) Return the apiary/Google API Client objects directly, with a caveat that the version of google-api-client used will determine what data is returned. This also means new data can be accessed when available.

google-cloud-bigquery/lib/google/cloud/bigquery/model/confusion_matrix.rb
```diff
@@ -0,0 +1,51 @@
...
+          #
+          # @return [Float]
+          def confidence_threshold
+            @gapi.confidence_threshold
+          end
+
+          # One row per actual label.
+          #
+          # @return [Hash]
+          def rows
```
blowmage: I do not fully understand how this is to be used. I would prefer to convert these rows to Ruby Hash objects, like we do for queries. Are these rows using the structure or `feature_columns` or `label_columns`?

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
shollyman: Does ruby manually pin its discovery dependencies, or does it pick up new versions as they're available?  I want to avoid having anyone having to monitor discovery for changes, and propagating them by hand.

Is option 2 fairly idiomatic for ruby folks, or would they have a stronger preference towards the complex typing we'd get from converting messages to an appropriate object/type?  There's also the option of consuming the [proto message](https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/v2/model.proto) to generate the type in similar fashion to python, but it depends on support for converting between proto types and json representation.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
shollyman: Ahh, I missed the nuance about swapping representations.  In the BQML models usage, we're basically just informing users which columns are supplied to do a prediction, and which columns contain the prediction results.  For this, having the hash with the simplified name/type should be sufficient.

Future work is coming where we'll want users to be able to express things using the backend type representation, so I just wanted to ensure we're able to express the same level of complexity.

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
blowmage: > Does ruby manually pin its discovery dependencies, or does it pick up new versions as they're available?

The google-cloud-bigquery gem maintains a minimum dependency on google-api-client. So anything that the google-cloud-bigquery gem exposes needs to be present in the minimum google-api-client version. Newer google-api-client versions can be used, but they are assumed to be compatible with the minimum version specified. Adding new features in the google-cloud-bigquery gem will bump the dependency on the google-api-client gem to the lowest version that contains it.

If the API returns additional data then it is assumed that those changes are reflected in the apiary discovery document (and proto files). That will kick off a new release of the google-api-client gem that includes those features, which the google-cloud-bigquery gem would then set as the new minimum dependency when it exposes the data. This is most pronounced in option 1.

Option 2 is the most flexible, as it would allow the google-cloud-bigquery gem to expose the data for whatever version of google-api-client was being used, and also move away from google-api-client in the future. Option 3 cements the dependency on google-api-client, and makes it part of the google-cloud-bigquery public API.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
blowmage: We could use the same approach for this data as we use for the training_runs data. Meaning, instead of returning this particular primitive representation we could return the unserialized values or the apiary structured objects.

google-cloud-bigquery/lib/google/cloud/bigquery/model/confusion_matrix.rb
shollyman: Is this a case of wanting a real model you can examine?  The hash approach makes sense here, as the rows are keyed by the label name, but you're dealing with list of entry stats as the value.  I can see about getting some models made public that might be more illustrative of the different types.

blowmage: When I say "convert these rows to Ruby Hash objects, like we do for queries", I don't mean exposing the underlying raw data as being discussed in the other conversation. I mean properly converting values to the correct Ruby objects. For example, a value of the string `"Infinity"` for a FLOAT64 column would be converted to the Ruby object `Float::INFINITY`.

That said, if we don't keep these classes then we don't need to be performing this conversion.

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
shollyman: Sounds like option 2 is the best choice here, as it gives us the ability to keep up with backend changes and not encumber the veneer with the explicit dependency on the discovery library.

blowmage: One thing we can do is bypass the Google API Client object model, and parse the JSON directly. That would allow us to return all the data that the API returns regardless of which version of google-api-client is being used. Is that something that you would like us to do?

google-cloud-bigquery/lib/google/cloud/bigquery/model/confusion_matrix.rb
shollyman: I think the most complex thing is clustering metrics, where the model is going to communicate information about the clusters in terms of centroid objects.  However, the expectation around the statistics objects is for using them to evaluate the model training numerically, so conversion makes sense..  e.g. was the accuracy better in this model than before?  how many items with this label landed in this bucket?

blowmage: This is now deserialized JSON, so there is no need to coerce this data further.

google-cloud-bigquery/lib/google/cloud/bigquery/model/aggregate_classification_metrics.rb
blowmage: I've updated the PR and changed `#training_runs` to return the deserialized JSON from the API, so any changes to the data returned will be visible immediately. PTAL.

google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
blowmage: I've updated the PR and removed the coercion to a Ruby Hash. These methods now return the deserialized JSON from the API, same as `#training_runs`.

blowmage:
@shollyman Can you review the changes?

google-cloud-bigquery/test/google/cloud/bigquery/model/columns_test.rb
```diff
@@ -25,7 +25,20 @@
   let(:model) { Google::Cloud::Bigquery::Model.from_gapi_json model_full_hash, bigquery.service }
 
   it "maps columns to primitive ruby values" do
-    model.feature_columns.must_equal [{ name: "f1", type: { typeKind: "STRING" } }]
-    model.label_columns.must_equal [{ name: "predicted_label", type: { typeKind: "FLOAT64" } }]
+    model.feature_columns.count.must_equal 1
```
blowmage: @shollyman Can you verify this is the type of structure you were looking for regarding the Standard SQL fields and types?

blowmage:
@shollyman Thanks! The plan is to cut a release early next week.

PR #3516, opened 2019-06-20
test: add conformance-tests git submodule

quartzmo:
The googleapis/conformance-tests repo recommends that it should be imported using git submodules:

> Generally, conformance tests are accessed as git submodules in the repositories. 

The repo currently provides conformance test data for Storage, and will soon add Firestore and Bigtable.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "conformance-tests"]
+	path = conformance-tests
+	url = https://github.com/googleapis/conformance-tests
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "conformance-tests"]
-	path = conformance-tests
+[submodule "conformance-tests-storage"]
+	path = google-cloud-storage/conformance
 	url = https://github.com/googleapis/conformance-tests


quartzmo:
Are any additional changes needed for CI/Kokoro?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "conformance-tests"]
+	path = conformance-tests
```
blowmage: I expected a nested share to be put somewhere like `google-cloud-storage/test`, so each gem could be updated independently. Thoughts?

quartzmo: Yes, I thought of that too, but decided to start with the simplest possible import. Really not sure what's best from the perspective of CI, please advise.

blowmage: If you have separate submodules, each gem can pin to a different commit so that each gem can be upgraded independently.

quartzmo: That's a good idea. I was thinking of the more extreme approach, in which CI calls `git submodule update --remote` and we are made aware of (breaking) updates when the build fails. Otherwise, I'm afraid I probably won't remember to periodically check for changes to the tests.

blowmage: Yeah, I don't like that. Still recovering from an unbound rubocop dependency breaking our builds every time a new release was made...

quartzmo: OK, I hear ya. And I agree that if pulls are going to be manual, we should definitely scope the submodules to the gems.

quartzmo:
@blowmage Updated, if these names/paths look good I'll add entry for firestore

quartzmo:
We decided that git submodules scoped to each package would bring in too many files. Replaced by #3525.

PR #3543, opened 2019-06-27
add redirects for devsite

TheRoyalTnetennba:
Do not merge before https://github.com/googleapis/google-cloud-ruby/pull/3690
Relevant files changed:
- Deleted contents of docs folder, aside from js/, css/ and index.html
- Add googleapis.dev redirect logic to the [jsondoc forwarder init](https://github.com/googleapis/google-cloud-ruby/blob/3f0ae3289d5a4c020e756301b91486de60f1291d/js/jsondoc-forwarder-init-0.1.0.js)
- Add jsondoc-forwarder to 404.html


_data/devsite_redirects.yaml
```diff
@@ -0,0 +1,46 @@
+---
+gcloud: gcloud
+google-cloud: google-cloud
+google-cloud-asset: cloudasset
+google-cloud-bigquery: bigquery
```
blowmage: Why are the docs not using their gem names? This seems confusing to me.

TheRoyalTnetennba: Docuploader uses the name field of repo-metadata, which is supposed to be the dns name of the API service. 

I have mixed feelings about this.

I agree that it's more confusing if you arrive there from the library repo. But I think it's probably less confusing if you arrive there from a product page. If all languages follow spec, it would also be less confusing for a user to bounce between languages.

_includes/head.html
```diff
@@ -12,9 +12,12 @@
   <script type="text/javascript" charset="utf-8" src="{{ "/js/plugins.js" | prepend: site.baseurl }}"></script>
   <script type="text/javascript" charset="utf-8" src="{{ "/js/main.js" | prepend: site.baseurl }}"></script>
   <script type="text/javascript" charset="utf-8" src="{{ "/js/vendor/modernizr-2.6.2.min.js" | prepend: site.baseurl }}"></script>
+  <script type="text/javascript" charset="utf-8" src="{{ "/js/vendor/jsondoc-forwarder-0.3.0.js" | prepend: site.baseurl }}"></script>
```
blowmage: Do we really want to load the redirect scripts on every page? Why load the entire existing HTML file and then redirect? Why not delete all the HTML content, have the site load the `404.html` file, and redirect from there?

TheRoyalTnetennba: I want to (at least for now) preserve the TOC pages on the existing gh-pages docs. There isn't anything like that currently for the devsite.

_data/devsite_redirects.yaml
TheRoyalTnetennba: Reverted back to using gem name

_includes/head.html
blowmage: Can't we keep those TOC pages, change the links on them, delete all the other files, and have `404.html` handle all the other redirects?

TheRoyalTnetennba:
holy mother of merge conflicts

_includes/head.html
TheRoyalTnetennba: Done. Gonna wait on resolving merge conflicts until https://github.com/googleapis/google-cloud-ruby/pull/3684 is merged, otherwise they'll keep happening

TheRoyalTnetennba:
Will open in new PR without docs/#{gem_name}/ deletions

PR #3584, opened 2019-07-01
tests(bigtable): add read_rows conformance tests

quartzmo:
Adds conformance tests to google-cloud-bigtable. Use `rake test` to run these tests, since they are included with the ordinary unit tests.


google-cloud-bigtable/test/google/cloud/bigtable/conformance_test.rb
```diff
@@ -0,0 +1,90 @@
...
+                              filter: nil,
+                              rows_limit: nil,
+                              app_profile_id: nil
+                            ]
+
+      results = test.results.to_a
+      if results.map(&:error).include? true
+        expect do
+          table.read_rows.to_a # to_a is needed because Enum lazily invokes read_rows
+        end.must_raise Google::Cloud::Bigtable::InvalidRowStateError
```
jskeet: It looks like this doesn't test the results before the error one. The test should be able to get the expected results before it errors.
I could just be misreading it, of course...

quartzmo: Absolutely right. I'll work on an update.

quartzmo:
@jskeet 

> It looks like this doesn't test the results before the error one. The test should be able to get the expected results before it errors.

I changed the test to iterate over the expected results (instead of the response cells), so that each non-error result is tested against a cell. Please take a close look and let me know what you think. I believe that for each test there is just a single RPC call--does that seem correct?

google-cloud-bigtable/test/google/cloud/bigtable/conformance_test.rb
```diff
@@ -51,29 +50,37 @@ def self.build_test_for test, index
...
-        end.flatten
+        cells = []
+        # Iterate over `results`, fetching rows and cells until each result has been used in assertions
+        test.results.each_with_index do |result|
+          if result.error
+            expect do
+              rows.next
+            end.must_raise Google::Cloud::Bigtable::InvalidRowStateError
+          else
+            if cells.empty?
```
jskeet: We don't have anything like this in the C# code, but I think that's just because our abstraction is "a stream of rows" so we flatten from those rows (via families and columns) to a sequence of cells, and then check all of those.

While this looks different, it looks like it's equivalent.

google-cloud-bigtable/test/google/cloud/bigtable/conformance_test.rb
```diff
@@ -51,29 +50,37 @@ def self.build_test_for test, index
...
 
-      results = test.results.to_a
-      if results.map(&:error).include? true
+      rows = table.read_rows
+      if test.results.empty? # "no data after reset"
         expect do
-          table.read_rows.to_a # to_a is needed because Enum lazily invokes read_rows
-        end.must_raise Google::Cloud::Bigtable::InvalidRowStateError
+          rows.next
+        end.must_raise StopIteration
```
jskeet: Is this basically saying "There are no rows"? Is there an implicit assertion that the loop below raise StopIteration, because that's how the "do" loop works? Sorry for what is probably a very basic question - I don't know Ruby at all.

jskeet: Ah, no - I've just thought, you're currently checking in the loop below that every expected result has a matching actual result, but I don't *think* you're checking that when we get to the end of the expected results, we've got to the end of the actual results.

I suspect you need another

```rb
expect do
  rows.next
end.must_raise StopIteration
```

at the end of the loop (line ~84).

Please let me know if I've misunderstood though!

quartzmo: Thanks @jskeet, that's a very good idea to check for unexpected cells. Unfortunately, I tried it, and it results in the client attempting another RPC:

```
[StopIteration] exception expected, not
Class: <MockExpectationError>
Message: <"No more expects available for :read_rows: [\"projects/test/instances/test-instance/tables/test-table\", {:rows=><Google::Bigtable::V2::RowSet: row_keys: [], row_ranges: []>, :filter=>nil, :rows_limit=>nil, :app_profile_id=>nil}]">
```

So I think I'll have to leave it out.

quartzmo: Looking at it with @blowmage, we managed to add the end of loop check, with a conditional to skip it if an `InvalidRowStateError` has been raised.

PR #3658, opened 2019-07-12
Fix for protobuf 3.9.0

blowmage:
The google-protobuf gem removed the #to_hash method.
The Bigtable, Datastore, and Firestore gems had calls to #to_hash.
This changes those calls to use the correct #to_h method instead.
This fixes these gems so they work with google-protobuf 3.9.0.

Some tests also expected #to_hash, and those tests are updated as well.

### DIFF
--- a/google-cloud-bigtable/test/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client_test.rb
+++ b/google-cloud-bigtable/test/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client_test.rb
@@ -97,7 +97,7 @@ describe Google::Cloud::Bigtable::Admin::V2::BigtableInstanceAdminClient do
         assert_equal(formatted_parent, request.parent)
         assert_equal(instance_id, request.instance_id)
         assert_equal(Google::Gax::to_proto(instance, Google::Bigtable::Admin::V2::Instance), request.instance)
-        assert_equal(clusters, request.clusters)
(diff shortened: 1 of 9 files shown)


blowmage:
I think it is best to get this patched and the Bigtable, Datastore, and Firestore gems released ASAP. We should also update the synth.py files for Bigtable, Container, and Grapheas Client to fix the generated tests.

google-cloud-bigtable/test/google/cloud/bigtable/admin/v2/bigtable_instance_admin_client_test.rb
```diff
@@ -455,7 +455,7 @@ def updater_proc
               formatted_name,
               display_name,
               type,
-              labels
+              labels.to_h
```
blowmage: Reading this again and I'm thinking this change might not be needed.

quartzmo: Do you want to update the PR?

quartzmo: I'm going to merge the PR now; if think this isn't needed, I guess we just omit it from the changes to synth.py?

blowmage: Yeah, that's my thinking.

PR #3678, opened 2019-07-18
Use ThreadPoolExecutor instead of CachedThreadPool

### LINKED ISSUE #3677: Pub/Sub AsyncPublisher creates too many threads
When I publish over a million messages, `Google::Cloud::Pubsub::AsyncPublisher` creates too many threads then it gets slower publishing messages. `Google::Cloud::Pubsub::AsyncPublisher` uses [`Concurrent::CachedThreadPool`](https://ruby-concurrency.github.io/concurrent-ruby/1.1.5/Concurrent/CachedThreadPool.html) with `max_threads` parameter.
https://github.com/googleapis/google-cloud-ruby/blob/7b87b46675ffce0b991f3678bfd4548195f11760/google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb#L196-L199

Although, [the constructor of CachedThreadPool](https://github.com/ruby-concurrency/concurrent-ruby/blob/v1.1.5/lib/concurrent/executor/cached_thread_pool.rb#L42) overrides `max_threads`. This change merged by #3449.

I think `Google::Cloud::Pubsub::AsyncPublisher` should use [`Concurrent::ThreadPoolExecutor`](https://www.rubydoc.info/gems/concurrent-ruby/1.1.5/Concurrent/ThreadPoolExecutor) which doesn't override `max_threads` parameter. By this method, it reduces resource utilization when the library is lightly used.

#### Environment details

  - OS: Linux and macOS
  - Ruby version: 2.6.3
  - Gem name and version: google-cloud-pubsub/v0.27.0 - v0.37.1

#### Steps to reproduce

Run example code and count `create a thread`. You will see many threads are created.

```console
$ time ruby ./publish.rb | uniq -c
   4277 create a thread

real	0m7.032s
user	0m5.880s
sys	0m1.530s
```

When it uses `ThreadPoolExecutor`, it consumes a less threads and completes faster.

```console
$ ruby ./publish.rb | uniq -c
     13 create a thread

real	0m1.598s
user	0m1.310s
sys	0m0.200s
```

#### Code example

```ruby
require 'google/cloud/pubsub'

module ThreadInterceptor
  def initialize(*arg, &block)
    puts 'create a thread'
    super(*arg, &block)
  end
end

class Thread
  prepend ThreadInterceptor
end

c = Google::Cloud::Pubsub.new(project: 'test')
t = c.topic('test', skip_lookup: true)
t = c.create_topic('test') if !t.exists?

10_000.times do |i|
  t.publish_async('') {}
end

t.async_publisher.stop.wait!
```

bgpat:
To fix #3677.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
@@ -193,9 +193,9 @@ module Google
 
         def init_resources!
           @first_published_at   ||= Time.now
-          @publish_thread_pool  ||= Concurrent::CachedThreadPool.new \
+          @publish_thread_pool  ||= Concurrent::ThreadPoolExecutor.new \
             max_threads: @publish_threads
-          @callback_thread_pool ||= Concurrent::CachedThreadPool.new \
+          @callback_thread_pool ||= Concurrent::ThreadPoolExecutor.new \
             max_threads: @callback_threads
           @thread ||= Thread.new { run_background }
         end


blowmage:
Thanks @bgpat!

PR #3685, opened 2019-07-18
Add Ordering Keys

blowmage:
These new feature add the ability to ensure related messages are published in
the order they were provided to the client. And to ensure related messages are
sent to subscribers in the order in which they were published. Messages can be
tagged with an ordering key, a string that identifies related messages for which
publish order should be respected.


google-cloud-pubsub/lib/google/cloud/pubsub/errors.rb
```diff
@@ -0,0 +1,71 @@
...
+require "google/cloud/errors"
+
+module Google
+  module Cloud
+    module PubSub
+      ##
+      # Indicates that the {AsyncPublisher} has been stopped and cannot accept
+      # messages to publish.
+      #
+      class AsyncPublisherStopped < Google::Cloud::Error
```
blowmage: Would it be better to create and inherit these classes from `Google::Cloud::PubSub::Error`?

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
```diff
@@ -0,0 +1,293 @@
...
+          #
+          def add msg, callback
+            synchronize do
+              raise AsyncPublisherStopped if @stopping
+              raise OrderingKeyCanceled, @ordering_key if @canceled
+
+              if @publishing
+                queue_add msg, callback
+                :queued
+              elsif try_add msg, callback
```
kamalaboulhosn: It looks like there is a scenario where messages could get out of order. If I call publish on a message that will exceed the size of a batch, then it will get added to the queue. If I then call publish on a message that does not exceed the batch size, it will get added to the batch. Now these messages will be delivered out of order.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
```diff
@@ -0,0 +1,293 @@
...
+          # This method is only intended to be used by the active publishing
+          # job.
+          #
+          def rebalance!
+            synchronize do
+              return [] if @canceled
+
+              until @queue.empty?
+                item = @queue.first
+                if try_add item.msg, item.callback
```
kamalaboulhosn: What happens if I call publish on a message that is larger than the max batch size I have set? It looks like publishing will just just stop because we'll never be able to add it. In other languages, we consider it okay to publish this message in a batch with just that message.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -224,40 +284,91 @@ def run_background
           end
         end
 
-        def publish_batch!
-          return unless @batch
+        def resolve_batch_for_message msg
+          @batches[msg.ordering_key] ||= Batch.new self, msg.ordering_key
```
kamalaboulhosn: Where do remove the Batch object for ordering keys once we have flushed all of the messages? If we don't, we'll have a slow memory leak if one publishes on some ordering keys and then never publishes on them again.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
```diff
@@ -0,0 +1,103 @@
...
+              # processed from the sequencer, and to return the next message to
+              # be processed. We want to ensure that these operations happen
+              # atomically.
+
+              # The message should be at index 0, so this should be a very quick
+              # operation.
+              if @seq_hash[message.ordering_key].first == message
+                @seq_hash[message.ordering_key].shift
+              else
+                # This indicates something went wrong. Find the first occurrance
```
kamalaboulhosn: occurrence

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
```diff
@@ -0,0 +1,103 @@
...
+                # of the message and remove it.
+                msg_index = @seq_hash[message.ordering_key].index message
+                @seq_hash[message.ordering_key].delete_at msg_index if msg_index
```
kamalaboulhosn: Shouldn't we delete all subsequent messages, too? Or shut down the subscriber? And maybe we need to log an error message? Otherwise, we are going to deliver things out of order.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
```diff
@@ -0,0 +1,103 @@
...
+          ##
+          # @private Implementation attributes.
+          attr_reader :keys, :messages, :process_callback
+
+          ##
+          # @private Create an empty Subscriber::Sequencer object.
+          def initialize &block
+            raise ArgumentError if block.nil?
+
+            @seq_hash = Hash.new { |hash, key| hash[key] = [] }
```
kamalaboulhosn: Where do we delete entries for ordering keys that have no more messages? If we don't do that, we could slowly leak memory.

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -316,6 +316,85 @@ received_messages = sub.pull
...
+                    ordering_key: "task-key"
+
+# Shut down the publisher when ready to stop publishing messages.
+topic.async_publisher.stop.wait!
+```
+
+### Handling error with Ordered Keys
+
+Ordered messages that fail to publish to the Pub/Sub API due to error will have
+the `ordering_key` canceled, and future calls to
```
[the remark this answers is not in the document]
blowmage: it means and future calls to `Topic#publish_async` with the `ordering_key` will raise `OrderingKeyCanceled`. "Canceled" is the terminology used in the documentation I have been given access to describing this behavior.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
blowmage: Messages will not be added out of order. When the publish is complete the batch will pull all available messages from the queue before allowing any message to be added to the batch directly. All this happens in the same lock.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
blowmage: Yes, this is a regression. I'll update the implementation to restore the logic that allowed for this.

google-cloud-pubsub/OVERVIEW.md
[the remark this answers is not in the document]
kamalaboulhosn: I see in the document we provided that we said to cancel outstanding requests, but there isn't a notion of an ordering key being "canceled." If there is documentation that says something to that effect, we should change it. An ordering key for which a publish didn't succeed is in a failed state and all subsequent publishes are canceled until resume is called.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: We had some concurrency bugs during development and this code was used to help diagnose the problem. Once we got all the concurrency issues settled we haven't had any instances where this code was being invoked.

I don't know how the Subscriber should behave in this instance. Should it nack all the remaining messages? Or simply remove them from inventory? What should it do if it receives more messages for the `ordering_key`?

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -360,6 +360,58 @@ def expires_in= ttl
...
+        # enabled, messages published with the same `ordering_key` will be
+        # delivered in the order they were published.
+        #
+        # See {#message_ordering?} and {#disable_message_ordering!}. See
+        # {Topic#publish_async}, {#listen}, and {Message#ordering_key}.
+        #
+        def enable_message_ordering!
+          update_grpc = Google::Cloud::PubSub::V1::Subscription.new \
+            name: name, enable_message_ordering: true
+          @grpc = service.update_subscription update_grpc,
```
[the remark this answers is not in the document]
blowmage: I will remove the update methods.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: The current implementation does not remove empty batches.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: The current implementation does not remove keys.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
kamalaboulhosn: If we are confident this won't happen, then may be we just throw an exception if it does. You are right, we can't just nack the messages currently outstanding, but would have to nack all subsequent messages. In that case, we probably are just better off treating this as a serious exception and stopping everything.

google-cloud-pubsub/OVERVIEW.md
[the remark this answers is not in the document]
blowmage: Updated.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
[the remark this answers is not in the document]
blowmage: Removed.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: Empty batches are now removed on publisher flush.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: Ordered keys are now removed from the sequencer when all messages are delivered.

blowmage:
PR has been rebased on master.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: I am confident this won't happen. I have updated this code to raise a `OrderedMessageDeliveryError` error if it does, which will stop future messages with the same ordering key from being processed by the Subscriber's callback. But like with other concurrent failures, the Subscriber attempts to recover as best as it can. (For example, if the user callback raises an error, the error is handled and indicated to the user but the Subscriber recovers and continues to process messages using the user callback.)

When an unhandled error is raised on a ruby thread, only thread is killed. Only unhandled errors raise on the main thread will kill the process. 

This error is raised on a thread from the thread pool, and not the thread that the Subscriber was created on. We use the `Subscriber#on_error` callback to notify the user when a concurrent error like this has occurred, but we aren't re-raising the error on the Subscriber's thread in an attempt to kill the process. (And there is no guarantee that the Subscriber was created on the main thread.)

We can go the extra step and also stop the subscriber when this happens, but it might be confusing for users to their Subscriber stopped without an error raised on the Subscriber's thread. So for now the implementation is using the `Subscriber#on_error` callback. Is this acceptable?

blowmage: CC @pradn

google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
```diff
@@ -473,17 +477,45 @@ def publish data = nil, attributes = {}
...
+        # respected. The service guarantees that, for a given ordering key and
+        # publisher, messages are sent to subscribers in the order in which they
+        # were published. Ordering does not require sacrificing high throughput
+        # or scalability, as the service automatically distributes messages for
+        # different ordering keys across subscribers.
+        #
+        # To use ordering keys, specify `ordering_key`. Before specifying
+        # `ordering_key` on a message a call to `#enable_message_ordering!` must
+        # be made or an error will be raised.
+        #
```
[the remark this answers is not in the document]
blowmage: I suppose that would depend on what the release schedule for ordered keys is. Will the feature still be alpha by the time this code is released? Or will it be considered beta?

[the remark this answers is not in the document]
kamalaboulhosn: It will be alpha likely until the end of January.

blowmage:
I've rebased this PR to pick up and resolve conflicts with some fixes to Subscriber#stop.

google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb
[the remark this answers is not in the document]
blowmage: Should we say January 2020 in the note? Or should the note be vague?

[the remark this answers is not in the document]
blowmage: Is there a URL we can direct users to about the rollout of ordering keys? Or a form to contact the Pub/Sub team about getting access?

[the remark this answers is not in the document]
kamalaboulhosn: I think it's best just to keep it vague. You can use the language from the Java library: "Ordering is not yet fully supported and requires special project enablements."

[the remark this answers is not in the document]
kamalaboulhosn: We actually reach out to customers via the CEs at this point.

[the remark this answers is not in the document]
blowmage: How's this?

> At the time of this release, ordering keys are not yet fully supported and requires special project enablements. To participate contact the Pub/Sub team to have the feature enabled.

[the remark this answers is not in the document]
blowmage: s/fully supported/publicly enabled?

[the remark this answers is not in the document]
kamalaboulhosn: Sure, anything like that is fine.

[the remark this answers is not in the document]
blowmage: K, note added.

blowmage:
@dazuma @quartzmo I've done some stress and performance testing and this feature is ready for final review. The Pub/Sub team has already approved the approach used, but we need to make sure the code is correctly implemented. I'd like to look closely at the concurrency used in particular.

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -316,6 +316,89 @@ received_messages = sub.pull
...
+messages are sent to subscribers in the order in which they were published.
+Messages can be tagged with an ordering key, a string that identifies related
+messages for which publish order should be respected. The service guarantees
+that, for a given ordering key and publisher, messages are sent to subscribers
+in the order in which they were published. Ordering does not require sacrificing
+high throughput or scalability, as the service automatically distributes
+messages for different ordering keys across subscribers.
+
+Note: At the time of this release, ordering keys are not yet publicly enabled
+and requires special project enablements. To participate contact the Pub/Sub
```
dazuma: How can an external customer "contact the pub/sub team"? If this is limited to certain partners, we should say so and tell them to contact their account manager. Otherwise, if it should be available for any customer, we should provide a link or something, because there isn't otherwise an obvious way for an arbitrary external customer to contact an internal team.

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -316,6 +316,89 @@ received_messages = sub.pull
...
+
+# Publish an ordered message with an ordering key.
+topic.publish_async "task completed",
+                    ordering_key: "task-key"
+
+# Shut down the publisher when ready to stop publishing messages.
+topic.async_publisher.stop.wait!
+```
+
+### Handling error with Ordered Keys
```
dazuma: s/error/errors/ ? Or maybe "an error"? Or is there a specific error being referenced here?
Same question two lines down.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -79,6 +82,15 @@ def initialize topic_name, service, max_bytes: 10000000,
...
 
+          @published_at = nil
+          @publish_thread_pool  = Concurrent::ThreadPoolExecutor.new \
+            max_threads: @publish_threads
+          @callback_thread_pool = Concurrent::ThreadPoolExecutor.new \
+            max_threads: @callback_threads
+          @thread ||= Thread.new { run_background }
+
+          @batches = {}
+
```
dazuma: Initialize `@ordered` to false to prevent uninitialized instance variable warnings

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -224,40 +282,93 @@ def run_background
...
 
+        # rubocop:disable Metrics/AbcSize
+        # rubocop:disable Metrics/MethodLength
+
         def publish_batch_sync topic_name, batch
-          grpc = @service.publish topic_name, batch.messages
-          batch.items.zip Array(grpc.message_ids) do |item, id|
-            next unless item.callback
+          # The only batch methods that are safe to call from the loop are
+          # rebalance! and flush! because they are the only methods that are
```
dazuma: This comment seems inaccurate. `Batch#reset!` is called in the loop, and it is synchronized.

google-cloud-pubsub/OVERVIEW.md
blowmage: @kamalaboulhosn

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -79,6 +82,15 @@ def initialize topic_name, service, max_bytes: 10000000,
           @publish_threads  = (threads[:publish] || 4).to_i
           @callback_threads = (threads[:callback] || 8).to_i
 
+          @published_at = nil
+          @publish_thread_pool  = Concurrent::ThreadPoolExecutor.new \
+            max_threads: @publish_threads
+          @callback_thread_pool = Concurrent::ThreadPoolExecutor.new \
+            max_threads: @callback_threads
+          @thread ||= Thread.new { run_background }
```
dazuma: Don't need `||=`

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -224,40 +282,93 @@ def run_background
...
+          end
+
+          items.each do |item|
             next unless item.callback
 
             publish_result = PublishResult.from_error item.msg, e
             execute_callback_async item.callback, publish_result
           end
+
+          retry unless batch.reset!
```
dazuma: I think you want `retry if batch.reset!`? You want to continue processing _if_ there is more to do?

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
dazuma: Okay, I see how this works. As soon as the batch is full, the `AsyncPublisher` will immediately call `publish_batches!` which marks _all_ batches as publishing (on the caller's thread, and in the `AsyncPublisher`'s lock, so that it is not possible for multiple calling threads to race.) This then causes subsequent adds to go directly to the queue until publishing has completed.

I agree this will work. I'm concerned about fragility, however, since the correctness of the batch class's state (i.e. ordering of messages across the items and the queue) depends on internal logic in the publisher class and the correct use of the publisher class's synchronization. I would prefer to see the batch class be thread-safe within itself, which you could accomplish by having this `add` method itself update any relevant state to indicate where future adds should go.

google-cloud-pubsub/OVERVIEW.md
blowmage: @quartzmo Do you have thoughts on how this could best be worded?

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
```diff
@@ -0,0 +1,119 @@
...
+        # @private The sequencer's job is simple, keep track of all the
+        # streams's recieved message and deliver the messages with an
+        # ordering_key in the order they were recieved. The sequencer ensures
+        # only one callback can be performed at a time per ordering_key.
+        class Sequencer
+          include MonitorMixin
+
+          ##
+          # @private Implementation attributes.
+          attr_reader :keys, :messages, :process_callback
```
dazuma: Are `keys` and `messages` used?

google-cloud-pubsub/OVERVIEW.md
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: Added.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: Yep, this was an early comment before many, many refactorings. Corrected.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: Fixed.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
blowmage: Yes, great catch. It should retry as long as there are items in the (reset) batch.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/sequencer.rb
blowmage: Nope. More old code. Removed.

google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher/batch.rb
dazuma: This was discussed offline. It turns out that _all_ batches need to move into a publishing phase at the same time (as soon as one batch fills up), which makes this more complicated than can be consolidated into a single method on `Batch`. I still think there might be a way to make the logic more solid, but it will take some more thought. Since we're confident of "correctness" at this point, let's accept the code as is and think about refactoring later.

google-cloud-pubsub/OVERVIEW.md
kamalaboulhosn: In other languages we did not specify to "contact the Pub/Sub team." There isn't a public link that can be used for this. I'd rather keep it ambiguous. Can we just go with "Ordering is not yet fully supported and requires special project enablements?" Generally we've been reaching out to customers via CEs or via a list we have of people who have expressed interest over time.

blowmage: Updated and removed the sentence about contacting the team. PTAL.

PR #3694, opened 2019-07-24
feat(bigtable): Accept nil gc_rule arguments for column_family create/update

### LINKED ISSUE #3680: Cloud Bigtable: Unable to create column family without a garbage collection policy
I tried creating a table with a column family that had no garbage collection policy (which would be default,) but I receive an error when it runs.

```
    table = bigtable.create_table @instance_id, table_id
    table.column_family("stats_summary").create()
```

```
  table = bigtable.create_table instance_id, table_id do |column_families|
    column_families.add(column_family)
  end

```
These both fail.

quartzmo:
[closes #3680]

### DIFF
--- a/google-cloud-bigtable/acceptance/bigtable/table_test.rb
+++ b/google-cloud-bigtable/acceptance/bigtable/table_test.rb
@@ -77,8 +77,8 @@ describe "Instance Tables", :bigtable do
     table_id = "test-table-#{random_str}"
 
     table = bigtable.create_table(instance_id, table_id) do |cfs|
-      cfs.add("cf1", Google::Cloud::Bigtable::GcRule.max_versions(1))
-      cfs.add("cf2") # default service value for GcRule
+      cfs.add("cf1") # default service value for GcRule
+      cfs.add("cf2", Google::Cloud::Bigtable::GcRule.max_versions(1))
     end
 
     modifications = []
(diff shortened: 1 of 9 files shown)


google-cloud-bigtable/acceptance/bigtable/table_test.rb
```diff
@@ -78,7 +78,7 @@
 
     table = bigtable.create_table(instance_id, table_id) do |cfs|
       cfs.add("cf1", Google::Cloud::Bigtable::GcRule.max_versions(1))
```
billyjacobson: Could you have cf1 be the one without the rule and cf2 have max versions? Seems more clear to have the default one be first

quartzmo: Done

quartzmo:
@blowmage Can you review when you get the chance, thanks!

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
```diff
@@ -238,8 +238,8 @@ def exists?
         # update, or delete operations.
         #
         # @param name [String] Name of the column family
-        # @param gc_rule [Google::Cloud::Bigtable::GcRule] Optional.
-        #   Required for create and update operations.
+        # @param gc_rule [Google::Cloud::Bigtable::GcRule, nil] A garbage
```
blowmage: I think listing `nil` here might be confusing. What do you think about something like this:

```ruby
# @param gc_rule [Google::Cloud::Bigtable::GcRule] The garbage collection
#   rule to be used for the column family. Optional. The service default value
#   will be used when not specified.
```

quartzmo: I like it. I'll update with that wording everywhere.

quartzmo: Updated.

PR #3708, opened 2019-07-26
Add Table#require_partition_filter

### LINKED ISSUE #3671: Move requirePartitionFilter from TimePartitioning to Table

blowmage:
Add getter and setter methods for the `require_partition_filter` value.

Although there are no acceptance tests for this change, this has been verified manually on a table where time partitioning has been configured.

The following error is raised when setting this value on a table that has not been configured for time partitioning: `Google::Cloud::InvalidArgumentError (invalid: Either interval partition or range partition should be specified.)`

[closes #3671]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
@@ -326,6 +326,52 @@ module Google
         end
 
         ###
+        # Whether queries over this table require a partition filter that can be
+        # used for partition elimination to be specified. See [Partitioned
+        # Tables](https://cloud.google.com/bigquery/docs/partitioned-tables).
+        #
+        # @return [Boolean, nil] `true` when a partition filter will be
+        #   required, `false` otherwise, or `nil` if the object is a reference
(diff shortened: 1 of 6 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/table.rb
```diff
@@ -325,6 +325,52 @@ def time_partitioning_expiration= expiration
...
+        # used for partition elimination to be specified. See [Partitioned
+        # Tables](https://cloud.google.com/bigquery/docs/partitioned-tables).
+        #
+        # @return [Boolean, nil] `true` when a partition filter will be
+        #   required, `false` otherwise, or `nil` if the object is a reference
+        #   (see {#reference?}).
+        #
+        # @!group Attributes
+        #
+        def require_partition_filter
```
quartzmo: I'm sure you considered this, but why not `require_partition_filter?` ?

blowmage: Because it is both a getter and a setter.

PR #3709, opened 2019-07-27
docs(bigtable): Fix timestamp param documentation

### LINKED ISSUE #3676: Cloud Bigtable setCell timestamp field does not set timestamp on cell.
When setting the timestamp on a cell value, the timestamp is ignored. You can test this with the [hello world](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/bigtable/hello_world.rb) 

```
  entry = table.new_mutation_entry "greeting#{i}"
  entry.set_cell(
    column_family,
    column_qualifier,
    value,
    timestamp: Time.now.to_i * 1000
  )

  table.mutate_row entry
```
When I read the results, this is what I see: 
```
greeting0
  cf:greeting                              @ 1970/01/18-21:16:13.999000
    "Hello World!"
----------------------------------------
greeting1
  cf:greeting                              @ 1970/01/18-21:16:14.000000
    "Hello Bigtable!"
----------------------------------------
greeting2
  cf:greeting                              @ 1970/01/18-21:16:14.000000
    "Hello Ruby!"
```

quartzmo:
* Add tests demonstrating correct and incorrect timestamp arguments.
* Add param documentation to clarify that values are in microseconds but must match the granularity of the table, which is currently always milliseconds, meaning that the microsecond values must be rounded to the nearest `1000`.
* Remove `Time` type from `timestamp` parameter types list, since its use raises an error.

[closes #3676]

### DIFF
--- a/google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
+++ b/google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
@@ -47,7 +47,7 @@ describe "DataClient Mutate Row", :bigtable do
     postfix = random_str
     row_key = "setcell-#{postfix}"
     qualifier = "mutate-row-#{postfix}"
-    timestamp = (Time.now.to_f * 1000).to_i * 1000
+    timestamp = (Time.now.to_f * 1000000).round(-3)
 
     # Set cell
     entry = table.new_mutation_entry(row_key)
@@ -135,7 +135,7 @@ describe "DataClient Mutate Row", :bigtable do
     postfix = random_str
     row_key = "setcell-#{postfix}"
     qualifier = "mutate-row-#{postfix}"
-    timestamp_micros = (Time.now.to_f * 1000).to_i * 1000
+    timestamp_micros = (Time.now.to_f * 1000000).round(-3)
 
     table.granularity.must_equal :MILLIS
 
--- a/google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
+++ b/google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
@@ -95,8 +95,8 @@ describe "DataClient Mutate Row", :bigtable do
     postfix = random_str
     row_key = "setcell-#{postfix}"
     qualifier = "mutate-row-#{postfix}"
-    t = Time.now
-    timestamp_micros = t.to_i * 1000000 + t.usec
+    timestamp_micros = (Time.now.to_f * 1000000).round
+    timestamp_micros += 1 if (timestamp_micros % 10).zero?
 
     table.granularity.must_equal :MILLIS
 
@@ -115,8 +115,8 @@ describe "DataClient Mutate Row", :bigtable do
     postfix = random_str
     row_key = "setcell-#{postfix}"
     qualifier = "mutate-row-#{postfix}"
-    t = Time.now
     timestamp_millis = (Time.now.to_f * 1000).to_i
+    timestamp_millis += 1 if (timestamp_millis % 10).zero?
 
     table.granularity.must_equal :MILLIS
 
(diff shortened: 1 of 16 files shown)


blowmage:
Removing `Time` from the documentation makes the documentation reflect the current implementation. Another option is to add support for `Time` objects to the implementation and make it match the documentation. I slightly prefer the latter, but I'm okay with the former (and the PR as it is now) if an argument can be made against supporting `Time`.

quartzmo:
Yes, given the current service behavior (see tests) I'd rather force the user to supply their own microsecond integers rounded to the nearest millisecond.

blowmage:
After an offline discussion I see an argument to correct the docs. We have several code examples in comments that direct users to set the timestamp value to milliseconds instead of microseconds with millisecond granularity. All those instances should also be updated in this PR.

quartzmo:
@blowmage Great catch on the code examples! I'll update them to microseconds values right away.

quartzmo:
@blowmage Updated code examples and unit test fixtures.

google-cloud-bigtable/test/helper.rb
```diff
@@ -259,3 +259,8 @@ def read_rows_acceptance_test_data
 
   tests
 end
+
+# A microseconds integer rounded to the nearest millisecond. For example: `1564257960168000`.
+def timestamp_micros
```
blowmage: I think this method should be in the helper class, not on main.

quartzmo: Thanks, moved.

google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
```diff
@@ -74,6 +74,89 @@
...
+    end.must_raise Google::Protobuf::TypeError
+    err.message.must_match "Expected number type for integral field 'timestamp_micros' (given Time)."
+  end
+
+  it "raises when set_cell with incorrect microsecond integer timestamp" do
+    postfix = random_str
+    row_key = "setcell-#{postfix}"
+    qualifier = "mutate-row-#{postfix}"
+    t = Time.now
+    timestamp_micros = t.to_i * 1000000 + t.usec
```
blowmage: We have a 1 in 1000 chance of `` ending in `000` and this test failing. Perhaps add a check for this here?

```suggestion
    timestamp_micros = (Time.now.to_f * 1000000).round
    timestamp_micros += 1 if (timestamp_micros % 10).zero?
```

google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
```diff
@@ -74,6 +74,89 @@
...
+      table.mutate_row(entry)
+    end.must_raise Google::Gax::RetryError
+    err.message.must_match /Timestamp granularity mismatch. Expected a multiple of 1000 \(millisecond granularity\), but got #{timestamp_micros}/
+  end
+
+  it "raises when set_cell with incorrect millisecond integer timestamp" do
+    postfix = random_str
+    row_key = "setcell-#{postfix}"
+    qualifier = "mutate-row-#{postfix}"
+    t = Time.now
```
blowmage: This line isn't needed anymore.

google-cloud-bigtable/acceptance/bigtable/table/mutate_row_test.rb
```diff
@@ -74,6 +74,88 @@
...
+      table.mutate_row(entry)
+    end.must_raise Google::Gax::RetryError
+    err.message.must_match /Timestamp granularity mismatch. Expected a multiple of 1000 \(millisecond granularity\), but got #{timestamp_micros}/
+  end
+
+  it "raises when set_cell with incorrect millisecond integer timestamp" do
+    postfix = random_str
+    row_key = "setcell-#{postfix}"
+    qualifier = "mutate-row-#{postfix}"
+    timestamp_millis = (Time.now.to_f * 1000).to_i
```
blowmage: I think this test also has a 1 in 1000 chance of not raising.

```suggestion
    timestamp_millis = (Time.now.to_f * 1000).round
    timestamp_millis += 1 if (timestamp_millis % 10).zero?
```

Issue #3084, opened 2019-04-01
Investigate BigQuery Acceptance Test Failures


blowmage:
Nightly CI builds regularly fail due to BigQuery. The failures are usually something like this:

```
  1) Failure:
Google::Cloud::Bigquery::Table::reference::bigquery#test_0008_inserts rows asynchronously and gets its data [/tmpfs/src/github/google-cloud-ruby/google-cloud-bigquery/acceptance/bigquery/table_reference_test.rb:240]:
Expected nil to not be nil?.
  1) Failure:
Google::Cloud::Bigquery::Table::bigquery#test_0020_inserts rows asynchronously and gets its data [/Volumes/BuildData/tmpfs/src/github/google-cloud-ruby/google-cloud-bigquery/acceptance/bigquery/table_test.rb:648]:
Expected: false
  Actual: true
  2) Failure:
Google::Cloud::Bigquery::Table::bigquery#test_0017_inserts rows directly and gets its data [/Volumes/BuildData/tmpfs/src/github/google-cloud-ruby/google-cloud-bigquery/acceptance/bigquery/table_test.rb:501]:
Expected: false
  Actual: true
```

Investigate why these tests are so inconsistent while running on Kokoro. Are delays needed to overcome BigQuery's eventual consistent nature?

PR #3744, opened 2019-08-06
Add assert_data acceptance method

blowmage:
Extract and simplify data checks to an assertion method.

[refs #3084]

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery_helper.rb
+++ b/google-cloud-bigquery/acceptance/bigquery_helper.rb
@@ -123,6 +123,14 @@ module Acceptance
       "kitten-test-data-#{SecureRandom.hex}.json"
     end
 
+    def assert_data data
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/acceptance/bigquery_helper.rb
```diff
@@ -123,6 +123,14 @@ def random_file_destination_name
       "kitten-test-data-#{SecureRandom.hex}.json"
     end
 
+    def assert_data data
+      assert_equal Google::Cloud::Bigquery::Data, data.class
+      refute_nil data.kind
+      refute_nil data.etag
+      refute_nil data.total
+      refute_nil data.count
```
quartzmo: Why the change now to basic assertions?

blowmage: Because it is simpler and I didn't want to fight the Spec DSL to get these tests passing again.

PR #3785, opened 2019-08-13
Release google-cloud-automl 0.2.0

bcoe:
:robot: I have created a release \*beep\* \*boop\* 
---
#### Features

* Update documentation
* Add DocumentDimensions, Layout and PredictResponse#preprocessed_input
  * Update Document
    * Add Document#document_text (TextSnippet)
    * Add Document#layout (Document::Layout)
    * Add Document#document_dimensions (DocumentDimensions)
    * Add Document#page_count
  * Update PredictionServiceClient#predict response
    * Add PredictResponse#preprocessed_input (ExamplePayload)
  * Added classes:
    * Add Document::Layout
    * Add DocumentDimensions
  * Update Documentation
* Deprecate ColumnSpec#top_correlated_columns
  * Deprecate ColumnSpec#top_correlated_columns
  * Add BatchPredictResult#metadata.
  * Add ConfusionMatrix#display_name
  * Add TableSpec#valid_row_count
  * Update documentation
---
### Commits since last release:

* [feat(automl): Add DocumentDimensions, Layout and PredictResponse#preprocessed_input](https://github.com/googleapis/google-cloud-ruby/commit/60d576dfbdc03b459749bc4b425826d7cc4e26da)
<pre><code>  * Update Document
    * Add Document#document_text (TextSnippet)
    * Add Document#layout (Document::Layout)
    * Add Document#document_dimensions (DocumentDimensions)
    * Add Document#page_count
  * Update PredictionServiceClient#predict response
    * Add PredictResponse#preprocessed_input (ExamplePayload)
  * Added classes:
    * Add Document::Layout
    * Add DocumentDimensions
  * Update Documentation</code></pre>
* [feat(automl): Deprecate ColumnSpec#top_correlated_columns](https://github.com/googleapis/google-cloud-ruby/commit/f6ff717d51a91fd88438c2900998b5c1e311e374)
<pre><code>  * Deprecate ColumnSpec#top_correlated_columns
  * Add BatchPredictResult#metadata.
  * Add ConfusionMatrix#display_name
  * Add TableSpec#valid_row_count
  * Update documentation</code></pre>
* [docs(automl): Update links to googleapis.dev](https://github.com/googleapis/google-cloud-ruby/commit/6ac5cc6fc8070e46b687003635e4e5086dac65b1)
* [docs: update links to point to new docsite (#3684)](https://github.com/googleapis/google-cloud-ruby/commit/ac8ce21c294ea111e33142a7ba6da82786cc8935)

### Files edited since last release:

[...]

PR #3808, opened 2019-08-19
chore: skip presubmit tests when commit message contains "[ci skip]"

TheRoyalTnetennba:


blowmage:
Is there a way to set this using conventional commits? Would it make more sense to specify something like `ci: skip` in the commit footer?

TheRoyalTnetennba:
Currently, I'm just setting it to work with release-please. Release-please adds "[ci skip]" to commit messages for trivial changes in https://github.com/googleapis/google-cloud-ruby/pull/3785. It's also worth noting that this won't affect merges to master, only commits on a pull request, so the commit messages don't really have to be in a conventional commit format. 

I'm happy to add additional criteria if you'd like though.

Issue #4011, opened 2019-09-10
Fix pubsub `subscriber.stop` and `subscriber.wait!` documentation. 


josedonizetti:
Since the PR https://github.com/googleapis/google-cloud-ruby/pull/2278 the documentation for the methods `stop`, and `wait!` are incorrect. Both mention the behavior is to wait until all pending messages are processed, which is not the case anymore, pending messages are now purged: 

https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb#L126 

https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber.rb#L149

https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb#L90

Notice that not only pending messages are purged, based on tests I have been doing, in-progress messages are also stopped, which I'm not sure if is a feature, or a bug.

I'm willing to open a pull request to fix the documentation, but I'd like to first understand what is actually expected now from the stop/wait methods. It won't process any pending messages, but how about not finish processing in-progress messages?

blowmage:
Yes, we should correct the documentation. The behavior is now that received but unprocessed messages will purged when the subscriber is stopped, and will be reclaimed by the API. But messages that have been processed (by invoking the subscriber's callback) but not yet updated to the API when stopped will update the API.

pjohnmeyer:
@blowmage: Which of the two behaviors above is supposed to apply to a message that is still _being_ processed? Specifically, this scenario:

- Subscriber callback is called on thread A.
- Before callback completes, stop-wait is called from thread B

It seems in my experimentation that the callback in thread A is being interrupted.

blowmage:
I believe that if the Subscriber is stopped before the callback is completed, the message is purged.

Or, more accurately, if the Subscriber is stopped before the message is acked or mod acked then the message is purged. If is possible to ack or mod ack the message, and then perform other work in the callback. So if the Subscriber is stopped during that post ack/mod ack work then the message should still be updated to the API.

blowmage:
This is fixed in https://github.com/googleapis/google-cloud-ruby/releases/tag/google-cloud-pubsub/v0.39.2

PR #4022, opened 2019-09-12
Do not interrupt Subscriber callbacks when stopping

### LINKED ISSUE #4020: Pub/Sub Subscriber interrupting callbacks when stopping
As discussed in #4011, the current Subscriber implementation interrupts in-process callbacks when a Subscriber is stopped. We should instead allow all in-process callbacks to complete using `Subscriber#wait!`.

blowmage:
This PR fixes an issue described in #4011, where an in-process callback will be interrupted when a Subscriber is stopped. We want to allow for in-process callbacks to complete, but to clear all queued callback threads. Unfortunately, we can't control the queued jobs on the callback thread pool, so this change adds a conditional to the pending callback thread so it will only perform the callback if the subscriber's stream is running (not stopped).

[fixes #4020]

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
@@ -109,7 +109,7 @@ module Google
 
           def wait!
             # Wait for all queued callbacks to be processed.
-            @callback_thread_pool.wait_for_termination
+            @callback_thread_pool.wait_for_termination 60
 
             self
           end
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -107,6 +108,9 @@ def paused?
           end
 
           def wait!
+            # Wait for all queued callbacks to be processed.
+            @callback_thread_pool.wait_for_termination
```
pjohnmeyer: Should this line have a timeout value provided? "When the timeout value is nil the call will block indefinitely."

Ideally a configurable timeout, but elsewhere in this library it is hard-coded to 60.

pjohnmeyer:
Thanks for looking at this. When I ran with this code, the callbacks were allowed to run to completion as intended by this change. However, the ack/nack of the message did not seem to propagate upstream, as I saw the message resent later. Was this your intent with this change, or are more changes necessary?

PR #4031, opened 2019-09-18
Support overriding of service endpoint

### LINKED ISSUE #3763: [translate] Support overriding of endpoint host
Support overriding the endpoint host in the veneer layer, by setting either constructor argument or config field.

See #3742 for an example from bigquery.

blowmage:
[closes #3763]

### DIFF
--- a/google-cloud-translate/lib/google/cloud/translate.rb
+++ b/google-cloud-translate/lib/google/cloud/translate.rb
@@ -68,6 +68,8 @@ module Google
       # @param [Integer] retries Number of times to retry requests on server
       #   error. The default value is `3`. Optional.
       # @param [Integer] timeout Default timeout to use in requests. Optional.
+      # @param [String] endpoint Override of the endpoint host name. Optional.
+      #   If the param is nil, uses the default endpoint.
       # @param [String] project Alias for the `project_id` argument. Deprecated.
       # @param [String] keyfile Alias for the `credentials` argument.
       #   Deprecated.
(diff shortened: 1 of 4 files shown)


google-cloud-translate/lib/google/cloud/translate.rb
```diff
@@ -106,34 +108,40 @@ module Translate
...
-              project_id.to_s, nil, retries: retries, timeout: timeout, key: key
+              project_id.to_s, nil,
+              retries: retries, timeout: timeout, key: key, host: endpoint
             )
           )
         end
 
         scope       ||= configure.scope
-        retries     ||= configure.retries
-        timeout     ||= configure.timeout
```
quartzmo: Nice catch moving these lines up.

google-cloud-translate/lib/google/cloud/translate.rb
```diff
@@ -106,34 +108,40 @@ module Translate
...
 
         unless credentials.is_a? Google::Auth::Credentials
           credentials = Translate::Credentials.new credentials, scope: scope
         end
 
-        project_id ||= credentials.project_id if credentials.respond_to? :project_id
-        project_id = project_id.to_s # Always cast to a string
+        # project_id ||= credentials&.project_id
+        # project_id ||= credentials.project_id if credentials.respond_to? :project_id
+        # project_id = project_id.to_s # Always cast to a string
```
quartzmo: Why are these lines commented out? Should they be removed?

google-cloud-translate/lib/google/cloud/translate.rb
```diff
@@ -179,6 +189,16 @@ def self.default_credentials scope: nil
...
       end
+
+      ##
+      # @private Resolve project.
+      def self.resolve_project_id project_id, credentials
+        # Always cast to a string
+        return project_id.to_s unless credentials.respond_to? :project_id
+
+        # Always cast to a string
+        project_id || credentials.project_id.to_s
```
quartzmo: Given the conditional early return on line 197, I'm having trouble following the logic here that prefers to return `project_id` before `credentials.project_id`.

blowmage: Yeah, blame Rubocop for forcing such a bad layout. The reason it is this way is because we don't want to call the `#project_id` method if it doesn't exist, which is why the early return is there.

google-cloud-translate/lib/google/cloud/translate.rb
blowmage: Yes. Removed.

PR #4040, opened 2019-09-19
Update Pool#init to use BatchCreateSessions

quartzmo:
Replace the loop in `Pool#init` that repeatedly calls `CreateSession` RPC with a single call to the new `BatchCreateSessions` RPC.

This change passes `rake acceptance` for me locally.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -1287,6 +1287,20 @@ module Google
           Session.from_grpc grpc, @project.service
         end
 
+        ##
+        # @private
+        # Creates a batch of new session objects of size `session_count`.
+        def batch_create_new_sessions session_count
+          ensure_service!
+          resp = @project.service.batch_create_sessions \
+            Admin::Database::V1::DatabaseAdminClient.database_path(
+              project_id, instance_id, database_id
+            ),
+            session_count,
+            labels: @session_labels
+          resp.session.map { |grpc| Session.from_grpc grpc, @project.service }
+        end
+
         # @private
         def to_s
           "(project_id: #{project_id}, instance_id: #{instance_id}, " \
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -1289,8 +1289,23 @@ module Google
 
         ##
         # @private
-        # Creates a batch of new session objects of size `session_count`.
-        def batch_create_new_sessions session_count
+        # Creates a batch of new session objects of size `total`.
+        # Makes multiple RPCs if necessary. Returns empty array if total is 0.
+        def batch_create_new_sessions total
+          sessions = []
+          remaining = total
+          while remaining > 0
+            sessions += batch_create_sessions remaining
+            remaining = total - sessions.count
+          end
+          sessions
+        end
+
+        ##
+        # @private
+        # The response may have fewer sessions than requested in the RPC.
+        #
+        def batch_create_sessions session_count
           ensure_service!
           resp = @project.service.batch_create_sessions \
             Admin::Database::V1::DatabaseAdminClient.database_path(
--- a/google-cloud-spanner/lib/google/cloud/spanner/service.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/service.rb
@@ -265,8 +265,8 @@ module Google
           execute do
             # The response may have fewer sessions than requested in the RPC.
             service.batch_create_sessions database_name,
+                                          session_count,
                                           session_template: session,
-                                          session_count: session_count,
                                           options: opts
           end
         end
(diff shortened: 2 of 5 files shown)


quartzmo:
@crwilcox Regarding the requirement that client RPC timeouts for `CreateSession` and `BatchCreateSessions` should be "appropriate", the [current settings](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/v1/spanner_client_config.json#L43-L52) are for 30 and 60 seconds, respectively. Does that sound OK to you?

crwilcox:
@quartzmo that should be alright.
cc: @skuruppu

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -1287,6 +1287,20 @@ def create_new_session
           Session.from_grpc grpc, @project.service
         end
 
+        ##
+        # @private
+        # Creates a batch of new session objects of size `session_count`.
+        def batch_create_new_sessions session_count
+          ensure_service!
+          resp = @project.service.batch_create_sessions \
```
snehashah16: the response may have fewer sessions than requested in the RPC. u should check that and initiate another rpc (with requested_count - response_count) until requested_count == 0

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -1287,6 +1287,20 @@ def create_new_session
...
+            Admin::Database::V1::DatabaseAdminClient.database_path(
+              project_id, instance_id, database_id
+            ),
+            session_count,
+            labels: @session_labels
+          resp.session.map { |grpc| Session.from_grpc grpc, @project.service }
```
snehashah16: also, these sessions should be shuffled in the session pool, because they sessions created from a single BatchCreateSession rpc will all have the same channel stickiness. adding them in the same order of the pool will cause them to use the same channel for all reads / writes request using the session. 
this will cause a few idle channels, and one or two busy channels.

snehashah16: also if possible, use the most recently used session again for the subsequent requests.. 
therefore the pool should be a LIFO data structure (like stack).

this will make sure the sessions are used from backend caches (and keep the cache hit ratio high), which will reduce the read / write latencies on user requests.

snehashah16:
also, is there a user configured option for "number of channels" in ruby ?
If yes, requested sessions should be distributed evenly across these channels. 
If no, then we can just assume to use all sessions on the same channel.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
crwilcox: Ruby is single channel so this is likely unnecessary?

quartzmo:
> also, is there a user configured option for "number of channels" in ruby ?

Afaik, just [one channel per client](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/v1/spanner_client.rb#L217) in all google-cloud-ruby GAPIC libraries.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
quartzmo: Will do, thank you for making us aware of this!!

google-cloud-spanner/lib/google/cloud/spanner/client.rb
quartzmo: OK, I will change the way sessions are checked out from the pool in order to get the desired LIFO behavior.

quartzmo: Done.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
quartzmo: Done.

quartzmo:
@blowmage Can you please re-review? I added a couple of commits based on @snehashah16's comments.

quartzmo:
I tested acceptance tests locally after my most recent commit for the breaking change to the lower-level API, they are now passing again.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
```diff
@@ -259,6 +259,18 @@ def create_session database_name, labels: nil
           end
         end
 
+        def batch_create_sessions database_name, session_count, labels: nil
+          opts = default_options_from_session database_name
+          session = Google::Spanner::V1::Session.new labels: labels if labels
+          execute do
+            # The response may have fewer sessions than requested in the RPC.
```
crwilcox: Isn't this handled by the method being called?

quartzmo: No, in fact it is handled in the higher-level class that calls the method here. (`Client#batch_create_new_sessions`). At this location, the comment is correct.

PR #4053, opened 2019-09-21
docs(bigtable): add doctest coverage and update sample code

### LINKED ISSUE #3285: Add doctest coverage to google-cloud-bigtable.
Add doctest coverage to google-cloud-bigtable.

```sh
google-cloud-bigtable quartzmo$ rake doctest
The google-cloud-bigtable gem does not have doctest tests.
```

Doctest coverage should have prevented the errors fixed in #3284.

quartzmo:
* Add `doctest` task and `support/doctest_helper.rb`.
* Fix syntax errors and typos in code examples.
* Add `console` task.

[closes #3285]

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
@@ -187,7 +187,7 @@ module Google
         #
         #   bigtable = Google::Cloud::Bigtable.new
         #
-        #   instance = bigtable.table("my-instance", "my-table")
+        #   table = bigtable.table("my-instance", "my-table")
         #   table.delete
         #
         def delete
@@ -247,7 +247,7 @@ module Google
         #
         #   bigtable = Google::Cloud::Bigtable.new
         #
-        #   table = bigtable.table("my-instance", my-table)
+        #   table = bigtable.table("my-instance", "my-table")
         #
         #   # OR get table from Instance object.
         #   instance = bigtable.instance("my-instance")
@@ -316,7 +316,7 @@ module Google
         #
         #   modifications = []
         #   modifications << Google::Cloud::Bigtable::ColumnFamily.create_modification(
-        #     "cf1", Google::Cloud::Bigtable::GcRule.max_age(600))
+        #     "cf1", Google::Cloud::Bigtable::GcRule.max_age(600)
         #   )
         #
         #   modifications << Google::Cloud::Bigtable::ColumnFamily.update_modification(
@@ -585,7 +585,7 @@ module Google
         #   table = bigtable.table("my-instance", "my-table")
         #
         #   # Delete rows using row key prefix.
-        #   table.drop_row_range("user-100")
+        #   table.drop_row_range(row_key_prefix: "user-100")
         #
         #   # Delete all data With timeout
         #   table.drop_row_range(delete_all_data: true, timeout: 120) # 120 seconds.
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
@@ -36,22 +36,13 @@ module Google
       #
       #   bigtable = Google::Cloud::Bigtable.new
       #
-      #   table = bigtable.table("my-instance", "my-table", perform_lookup: true)
+      #   table = bigtable.table("my-instance", "my-table")
       #
-      #   table.column_families.each do |cf|
-      #     p cf.name
-      #     p cf.gc_rule
+      #   if table.exists?
+      #     p "Table exists."
+      #   else
+      #     p "Table does not exist"
       #   end
-      #
-      #   # Get column family by name
-      #   cf1 = table.column_families.find_by_name("cf1")
-      #
-      #   # Create column family
-      #   gc_rule = Google::Cloud::Bigtable::GcRule.max_versions(3)
-      #   cf2 = table.column_families.create("cf2", gc_rule)
-      #
-      #   # Delete table
-      #   table.delete
       #
       class Table
         # @!parse extend MutationOperations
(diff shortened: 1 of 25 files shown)


quartzmo:
@crwilcox I found an inconsistency between [this sample code](https://github.com/googleapis/google-cloud-ruby/blob/313d6b8ea04e157150732ac2e7772db7de8077f5/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb#L428-L432) for listing tables with a "full view" argument, and [the method itself](https://github.com/googleapis/google-cloud-ruby/blob/313d6b8ea04e157150732ac2e7772db7de8077f5/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb#L434-L438), which does not support this argument. In the Java design doc there is no mention of this feature. Any idea if it should be provided?

google-cloud-bigtable/lib/google/cloud/bigtable/table/column_family_map.rb
```diff
@@ -23,12 +23,12 @@ class Table
         # It is used to create an instance.
         # @example Add column family with name and garbage collection rule
         #
-        #  column_families = Google::Cloud::Bigtable::Instance::ColumnFamilyMap.new
+        #  column_families = Google::Cloud::Bigtable::Table::ColumnFamilyMap.new
```
blowmage: If the constructor is private, should we show it being created?

quartzmo: Actually I looked into that, it's not marked private.

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
```diff
@@ -145,6 +136,25 @@ def cluster_states
...
+        #
+        #   table = bigtable.table("my-instance", "my-table", perform_lookup: true)
+        #
+        #   table.column_families.each do |cf|
+        #     p cf.name
+        #     p cf.gc_rule
+        #   end
+        #
+        #   # Get column family by name
+        #   cf1 = table.column_families.find_by_name("cf1")
```
blowmage: Does the method `ColumnFamilyMap#find_by_name ` exist?

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
```diff
@@ -145,6 +136,25 @@ def cluster_states
...
+        #
+        #   # Create column family
+        #   gc_rule = Google::Cloud::Bigtable::GcRule.max_versions(3)
+        #   cf2 = table.column_families.create("cf2", gc_rule)
```
blowmage: Does the method `ColumnFamilyMap#create ` exist?

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
quartzmo: No, this example is one of two that are skipped pending follow-up API changes.

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
quartzmo: No, this example is one of two that are skipped pending follow-up API changes.

quartzmo: `table.column_families.add` also doesn't exist, since `Table#column_families` returns an ordinary array.

google-cloud-bigtable/lib/google/cloud/bigtable/table/column_family_map.rb
blowmage: It was marked `@private`, but this PR currently changes it. Look 4 lines below.

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
blowmage: K, I didn't see that these were skipped.

google-cloud-bigtable/lib/google/cloud/bigtable/table/column_family_map.rb
quartzmo: Oops, my mistake, thank you for pointing that out. Are you OK with this solution?

blowmage: I am surprised the public api surface is changing to get doctests running. i would prefer to keep the constructor private, not whatevs.

quartzmo: Given the follow-up work proposed in #4133 and #4134, I anticipate that users will have no need to create instances of `ColumnFamilyMap`. I propose removing this example now and restoring the `@private` tag on the initializer. Sound OK?

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
quartzmo: See #4133 for follow-up that will should changes to this example.

google-cloud-bigtable/lib/google/cloud/bigtable/table.rb
quartzmo: See #4134 for follow-up that will should changes to this example.

google-cloud-bigtable/lib/google/cloud/bigtable/table/column_family_map.rb
blowmage: Yes, either that, or leave the original example code and skip it until the follow up work is completed.

quartzmo: I'll remove the example now. It has very little value since it is just a small fragment of code with no explicit use case (meaning, it does not result in any interaction with the service).

quartzmo: Done.

PR #4068, opened 2019-09-25
Register is a Hash, not Concurrent::Map

blowmage:
This is a simple refactor to bring this instance in line with the other times that the variable is set to an empty hash.


blowmage:
This isn't really a bug fix, as it will behave the same either way. I would consider this to be a refactor more than a performance change.

Issue #3705, opened 2019-07-25
Identify Spanner acceptance tests for multi-use snapshots


quartzmo:
Explicitly name existing acceptance tests (and/or add new acceptance tests) that confirm that the `Snapshot` implementation provides a "multi-use" mode equivalent to that found in google-cloud-python. Snapshots in this mode should not accept `max_staleness` or `min_read_timestamp` isolation. See [google-cloud-python test_system.py](https://github.com/googleapis/google-cloud-python/blob/master/spanner/tests/system/test_system.py#L1157-L1205) for examples.

quartzmo:
Can this be done with tracking tags as in the broader issue #2579?

quartzmo:
@crwilcox Is this still actionable? If so, how should it be prioritized?

Can this be done with tracking tags as in the broader issue #2579?

PR #4071, opened 2019-09-26
test(spanner): acceptance tests for multi-use #3705

jiren:
issue #3705 

- acceptance test for multi-use.  
- Ruby client storing transaction id of snapshot and using it for subsequent operations so by default snapshots created by a ruby client are multi-use.
- snapshot read and update acceptance test are available which are multi-use

### DIFF
--- a/google-cloud-spanner/acceptance/spanner/client/snapshot_test.rb
+++ b/google-cloud-spanner/acceptance/spanner/client/snapshot_test.rb
@@ -301,6 +301,7 @@ describe "Spanner Client", :snapshot, :spanner do
       results.must_be_kind_of Google::Cloud::Spanner::Results
 
       rows = results.rows.to_a
+      rows.count.must_equal default_account_rows.count
       rows.zip(default_account_rows).each do |expected, actual|
         expected[:account_id].must_equal actual[:account_id]
         expected[:username].must_equal actual[:username]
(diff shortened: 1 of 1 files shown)


quartzmo:
@googlebot I consent.

jiren:
> @jiren Thanks for opening this PR. Unfortunately, the assertions for the returned rows in these tests do not work, due to the behavior of `zip`. If the receiver of `zip` is empty, no assertions will be executed in the `each` block:
> 
> ```ruby
> irb(main):001:0> rows = []
> => []
> irb(main):002:0> default = [1,2]
> => [1, 2]
> irb(main):003:0> rows.zip default
> => []
> ```
> 
> I have pushed a commit to your branch with added assertions that verify that the returned rows count is the same as the expected rows:
> 
> ```ruby
> rows.count.must_equal default_account_rows.count
> ```
> 
> With these added assertions, all three tests now fail for me locally:
> 
> ```shell
>   1) Failure:
> Spanner Client::snapshot::spanner#test_0016_multiuse snapshot reads with read timestamp are consistent even when delete happen [/Users/quartzmo/code/google/codez/google-cloud-ruby/google-cloud-spanner/acceptance/spanner/client/snapshot_test.rb:353]:
> Expected: 3
>   Actual: 0
> 
>   2) Failure:
> Spanner Client::snapshot::spanner#test_0015_multiuse snapshot reads are consistent even when delete happen [/Users/quartzmo/code/google/codez/google-cloud-ruby/google-cloud-spanner/acceptance/spanner/client/snapshot_test.rb:317]:
> Expected: 3
>   Actual: 0
> 
>   3) Failure:
> Spanner Client::snapshot::spanner#test_0017_multiuse snapshot reads with exact staleness are consistent even when delete happen [/Users/quartzmo/code/google/codez/google-cloud-ruby/google-cloud-spanner/acceptance/spanner/client/snapshot_test.rb:392]:
> Expected: 3
>   Actual: 0
> ```

Ok, Looking into it.

blowmage:
These tests now pass for me.

PR #4104, opened 2019-10-01
Keep renewing lease

### LINKED ISSUE #4103: pubsub: lease is renewed only once
#### Environment details

  - OS: macOS 10.12.6
  - Ruby version: ruby 2.6.1p33
  - Gem name and version: google-cloud-pubsub v1.0.0

#### Steps to reproduce

  1. Create a pubsub topic and a subscription with ack-deadline of 10 seconds
  2. Publish a message to the topic
  3. If it takes more than 20 seconds for a subscriber to ack the message, Cloud Pub/Sub will redeliver the message because the google-cloud-pubsub gem renews the lease only once

#### Code example

```ruby
# How to run:
#
# $ env PUBSUB_SUBSCRIPTION=... PUBSUB_CREDENTIALS=... ruby this_file.rb
#
require 'google/cloud/pubsub'

subscription = ENV.fetch('PUBSUB_SUBSCRIPTION')
pubsub = Google::Cloud::PubSub.new

subscriber = pubsub.subscription(subscription).listen do |message|
  puts 'Start'
  sleep 21
  puts 'Done'

  message.acknowledge!
end

subscriber.start

at_exit { subscriber.stop }

sleep
```

#### Output

```
Start
Start
Done
Done
```

Expected:

```
Start
Done
```

soylent:
Resolves #4103

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/inventory.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/inventory.rb
@@ -115,8 +115,8 @@ module Google
 
                 unless delay_gap.positive?
                   delay_target = nil
-                  delay_gap = nil # wait until broadcast
                   stream.renew_lease!
+                  next
                 end
 
                 @wait_cond.wait delay_gap


google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/inventory.rb
```diff
@@ -115,8 +115,8 @@ def background_run
 
                 unless delay_gap.positive?
                   delay_target = nil
-                  delay_gap = nil # wait until broadcast
                   stream.renew_lease!
+                  next
```
blowmage: This looks correct and is where I ended up in my attempts to reproduce the bug.

google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/inventory.rb
```diff
@@ -115,8 +115,8 @@ def background_run
 
                 unless delay_gap.positive?
                   delay_target = nil
```
blowmage: The more I think about this, the more I wonder if this should be set to `calc_target` instead of `nil`. When we are dealing with a very short deadline, such as 10 seconds, having the call to `stream.renew_lease!` take a couple of seconds due to size of data or synchronization locking may push the `modify_ack_deadline` RPC call past the deadline. If we set `delay_target` to `calc_target` here we may mitigate that delay. Thoughts?

```suggestion
                   delay_target = calc_target
```

blowmage: I'll do this in a separate PR.

PR #4109, opened 2019-10-01
Fix Ruby 2.7 compatibility in Cloud::Config#mehtod_missing

casperisfine:
Recently an experimental feature has been merged in Ruby's master https://bugs.ruby-lang.org/issues/16150

In short `Symbol#to_s` will return a frozen string, whereas before it would always return a new mutable string.

Because of this `Google::Cloud::Config#method_missing` breaks on `ruby-head`.

The fix is quite straightforward, I simply reworked the code a little bit to avoid mutating the string.

### DIFF
--- a/google-cloud-core/lib/google/cloud/config.rb
+++ b/google-cloud-core/lib/google/cloud/config.rb
@@ -458,8 +458,8 @@ module Google
       def method_missing name, *args
         name_str = name.to_s
         super unless name_str =~ /^[a-zA-Z]\w*=?$/
-        if name_str.chomp! "="
-          self[name_str] = args.first
+        if name_str.end_with? "="
+          self[name_str[0...-1]] = args.first
         else
           self[name]
         end


byroot:
@googlebot I fixed it.

byroot:
@googlebot I consent.

google-cloud-core/lib/google/cloud/config.rb
```diff
@@ -458,8 +460,8 @@ def self.deferred &block
       def method_missing name, *args
         name_str = name.to_s
         super unless name_str =~ /^[a-zA-Z]\w*=?$/
-        if name_str.chomp! "="
-          self[name_str] = args.first
+        if name_str.end_with?("=")
+          self[name_str[0..-2]] = args.first
```
blowmage: FWIW, I find this easier to read:

```suggestion
          self[name_str[0...-1]] = args.first
```

google-cloud-core/lib/google/cloud/config.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
blowmage: We have avoided the magic frozen string comment in this gem so far. Adding it seems unrelated to the Ruby 2.7 compatibility fix.

casperisfine: Indeed, it was an opportunistic change to prevent `end_with?("=")` from allocating a new string on every call. I can remove it if you'd like.

I say it across the project so I assume it was ok to add it.

blowmage:
This PR should be rebased and rerun the build once #4116 is merged.

PR #4147, opened 2019-10-04
Update Bucket#post_object

### LINKED ISSUE #4047: URI escaping in `def ext_url` causes ${filename} variable not to work
#### Environment details

  - Gem name and version: `google-cloud-storage` version 0.24.0

#### Steps to reproduce

  1. Call `post_object` with a key containing `${filename}`.
  2. Inspect `.fields[:key]` on the returned object.
  3. Note that `${filename}` has been changed to `$%7Bfilename%7D`
  4. Try and perform a presigned post in the browser, using the escaped key
  5. The file uploads, but the filename has not been replaced as it's supposed to. The uploaded file has the literal path `$%7Bfilename%7D`

#### Code example

```ruby
client = Google::Cloud::Storage.new project: "xxxxxxxxxx"

key = "${filename}"

post_object = client.bucket("some-bucket").post_object(key,  policy: {
  expiration: (Time.now + 1.hour).to_i,
  conditions: [
    ["starts-with", "$key", "" ],
    ["starts-with", "$Content-Type", "" ],
    { acl: 'public-read' },
    { success_action_status: '201' }
  ]
}

puts post_object.fields[:key]
# => $%7Bfilename%7D
```

#### NOTES

This seems to be caused by `def ext_path` in `lib/google/cloud/storage/file.rb`.

I can make the code work by manually unescaping the key after it's returned from `def post_object`. But this was very tricky to debug and I don't think this should be the expected behavior. 

`${filename}` is documented [here](https://cloud.google.com/storage/docs/xml-api/post-object) as a way to set the key of the uploaded file dynamically. However, the gem is URI escaping it and making it nonfunctional.

quartzmo:
* Avoid encoding special variable name `${filename}`

> You can also use the ${filename} variable if a user is providing a file name.

See [Post Object - Form fields - key](https://cloud.google.com/storage/docs/xml-api/post-object).

[fixes #4047]

### DIFF
--- a/google-cloud-storage/acceptance/storage/bucket_post_object_test.rb
+++ b/google-cloud-storage/acceptance/storage/bucket_post_object_test.rb
@@ -1,0 +1,90 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "storage_helper"
+require 'net/http'
+require 'uri'
+
+describe Google::Cloud::Storage::Bucket, :post_object, :storage do
+  let(:bucket_name) { $bucket_names.first }
+  let :bucket do
+    storage.bucket(bucket_name) ||
+    safe_gcs_execute { storage.create_bucket(bucket_name) }
+  end
+  let(:uri) { URI.parse Google::Cloud::Storage::GOOGLEAPIS_URL }
+  let(:data_file) { "logo.jpg" }
+  let(:data) { File.expand_path("../data/#{data_file}", __dir__) }
+  let :policy do
+    {
+      expiration: (Time.now + 600).iso8601,
+      conditions: [
+        ["starts-with", "$key", ""]
+      ]
+    }
+  end
+
+  it "generates a signed post object" do
+    file_name = "logo-#{SecureRandom.hex(4).downcase}.jpg"
+
+    bucket.file(file_name).must_be :nil?
+
+    signed_post = bucket.post_object file_name, policy: policy
+
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request = Net::HTTP::Post.new(uri.request_uri)
+
+    form_data = [
+      ['file', File.open(data)],
+      ['key', signed_post.fields[:key]],
+      ['GoogleAccessId', signed_post.fields[:GoogleAccessId]],
+      ['policy', signed_post.fields[:policy]],
+      ['signature', signed_post.fields[:signature]]
+    ]
+    request.set_form form_data, 'multipart/form-data'
+
+    response = http.request(request)
+
+    response.code.must_equal "204"
+    bucket.file(file_name).wont_be :nil?
+  end
+
+  it "generates a signed post object with special variable key ${filename}" do
+    # "You can also use the ${filename} variable if a user is providing a file name."
+    #  https://cloud.google.com/storage/docs/xml-api/post-object
+    special_key = "${filename}"
+
+    bucket.file(data_file).must_be :nil?
+
+    signed_post = bucket.post_object special_key, policy: policy
+
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request = Net::HTTP::Post.new(uri.request_uri)
+
+    form_data = [
+      ['file', File.open(data)],
+      ['key', signed_post.fields[:key]],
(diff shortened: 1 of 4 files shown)


quartzmo:
@JesseLovelace Do you know anything about this special variable name `${filename}`? (See docs link in PR description above.) Does this PR look reasonable to you?

quartzmo:
@frankyn Do you know anything about this special variable name `${filename}`? (See docs link in PR description above.) Does this PR look reasonable to you?

frankyn:
Hi @quartzmo, can we hold off on merging this please?

I'm not quite convinced this is fixing something. I'll follow-up on the issue.

frankyn:
Hi folks, summoning @ranylee (from the GCS team), who is looking into this question. She'll get back to us next week.

ranylee:
I confirmed that cloud storage frontend api had implemented the filename token "${filename}" replacement. It checks if the data of "key" payload contains "${filename}", and then expands ${filename} using the client file name in the "file" payload if applicable.

I would suggest to do integration test to verify this fix.

quartzmo:
> cloud storage frontend api had implemented the filename token "${filename}" replacement. It checks if the data of "key" payload contains "${filename}", and then expands ${filename} using the client file name in the "file" payload if applicable.

@maxp-edcast Can you pull this branch and use it to confirm that `${filename}` replacement now works in the GCS service?

frankyn:
@quartzmo could you add an integration test for this fix? (re: "I would suggest to do integration test to verify this fix.")

quartzmo:
> could you add an integration test for this fix? (

@frankyn I worked on it some today, following the examples at [POST Object](https://cloud.google.com/storage/docs/xml-api/post-object). It's probably doable, but is way outside the usual scope of our integration tests and will take considerable time. That's partly why I'm reaching out to @maxp-edcast for a manual confirmation.

frankyn:
Understood, we need to confirm before merging. Thanks @quartzmo.

quartzmo:
@frankyn I'll start work tomorrow on the integration test unless we hear something definite back from @maxp-edcast.

quartzmo:
@frankyn I have manually verified that the following form upload works with this branch and GCS. Uploaded objects have the name of the file chosen in the form.

The client code:

```ruby
    policy = {
      :expiration=>"2020-06-16T11:11:11Z",
      :conditions=>[
        ["starts-with", "$key", ""]
      ]
    }

    signed_post = bucket.post_object "${filename}", policy: policy
```

The form using the values produced above (hidden):

```html
<form action="http://postobjecttest2019110702.storage.googleapis.com" method="post" enctype="multipart/form-data">
  <input type="hidden" name="key" value="postobjecttest2019110702/${filename}">
  <input type="hidden" name="GoogleAccessId" value="...">
  <input type="hidden" name="policy" value="...">
  <input type="hidden" name="signature"
         value="...">
  <input name="file" type="file">
  <input type="submit" value="Upload">
</form>
```

I'm still working on an automated ruby version of this form upload for the integration test.

frankyn:
Thanks @quartzmo! The manual description you provided LGTM. Is there anything I can do to help with  integration tests?

quartzmo:
No, I should have them working soon. I'll ping for a final review after I add that commit.

quartzmo:
@frankyn @blowmage @maxp-edcast PTAL at the [commit adding acceptance tests](https://github.com/googleapis/google-cloud-ruby/pull/4147/commits/92439c6d206b1f239dad7970a754bb79ab725415). If these looks good, I think the PR is ready for final review.

google-cloud-storage/acceptance/storage/bucket_post_object_test.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    signed_post = bucket.post_object special_key, policy: policy
+
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+    request = Net::HTTP::Post.new(uri.request_uri)
+
+    form_data = [
+      ['file', File.open(data)],
+      ['key', signed_post.fields[:key]],
```
frankyn: I was suspecting this to be a new name that didn't match `${filename}`, so it would imply the name can be added in the subsequent request.

quartzmo: As I discovered in my manual testing yesterday (form below), if the `key` field is actually set to `[BUCKET]/${filename}` (literally `${filename}`) and is `hidden`, the user of the form only has to choose a file, and the object created in GCS will have the same name as the uploaded file, without requiring the `key` field also to be set by the user.

GCS documentation for this feature is minimal, but S3 [states](https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/PresignedPost.html):

> The string `${filename}` is automatically replaced with the name of the file provided by the user


```html
<form action="http://storage.googleapis.com" method="post" enctype="multipart/form-data">
  <input type="hidden" name="key" value="postobjecttest2019110702/${filename}">
  <input type="hidden" name="GoogleAccessId" value="...">
  <input type="hidden" name="policy" value="...">
  <input type="hidden" name="signature"
         value="...">
  <input name="file" type="file">
  <input type="submit" value="Upload">
</form>
```

frankyn: I misread the documentation. Thanks for clarifying @quartzmo! 

Approving, let's merge this in.

PR #4157, opened 2019-10-08
Fix Subscriber staying paused after releasing messages

### LINKED ISSUE #4152: Pub/Sub Subscriber halts and never resumes
```rb
=> #<Google::Cloud::PubSub::Subscriber::Stream (inventory: 0, status: sleep)>
[93] pry(main)> g.send :unpause_streaming?
=> true
[94] pry(main)> g.instance_variable_get("@paused")
=> true
```

Stays in this state.

blowmage:
* Correctly reset the `Subscriber` state when releasing messages after the callback either raises an error, or the callback fails to call `acknowledge` or `modify_ack_deadline` on the message. If a `Subscriber` fills it's inventory, and stops pulling additional messages (pauses), before all the callbacks are completed then the `Subscriber` could become stuck in a paused state.
* A paused `Subscriber` will now check whether to unpause after the callback is completed, instead of when `acknowledge` or `modify_ack_deadline` is called on the message.

[fixes #4152]

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
@@ -123,7 +123,6 @@ module Google
             synchronize do
               @inventory.remove ack_ids
               @subscriber.buffer.acknowledge ack_ids
-              unpause_streaming!
             end
 
             true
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/subscriber/stream.rb
```diff
@@ -138,12 +137,27 @@ def modify_ack_deadline deadline, *messages
...
+            ack_ids = coerce_ack_ids messages
+            return if ack_ids.empty?
+
+            synchronize do
+              # Remove from inventory if the message was not explicitly acked or
+              # nacked in the callback
+              @inventory.remove ack_ids
+              # Check whether to unpause the stream only after the callback is
+              # completed and the thread is being reclaimed.
+              unpause_streaming!
```
ce07c3: Do we need to check inventory fill levels before resuming? When the halts occurred I tried to call unpause strraming but it wouldn’t resume. I’ll try this PR out.

ce07c3:
Seems fixed!

Issue #4134, opened 2019-10-03
Bigtable::Table#column_families should accept a block for updates


quartzmo:
```
table.column_families do |cfm|
  cfm.add("cf1", gc_rule)
end
```

This usage should execute an RPC to update the table and reload the new state into the current table instance.

quartzmo:
If `ColumnFamilyMap` is yielded to the block, the method docs for `ColumnFamilyMap` should be updated to demonstrate this use case.

PR #4161, opened 2019-10-08
feat!(bigtable): Update Table#column_families to return frozen ColumnFamilyMap

### LINKED ISSUE #4133: Bigtable::Table#column_families should return frozen hash
`Bigtable::Table#column_families` currently returns an ordinary array of `ColumnFamily` instances. While this data representation is complete (`Bigtable::ColumnFamily` includes a `name` attr), it is inconsistent with [the `column_families` param type in `Project#create_table` and `Instance#create_table`:

```
        # @param column_families [Hash{String => Google::Cloud::Bigtable::ColumnFamily}]
        #   (See {Google::Cloud::Bigtable::Table::ColumnFamilyMap})
        #   If unspecified, you may use a code block to add column families.
```
In this structure, the hash keys are the same as the `ColumnFamily#name`. An alternative solution would be to change the `column_families` param to an array of `ColumnFamily`, but this would have to be checked for duplicate `name` values.

quartzmo:
BREAKING CHANGE: `Table#column_families` return type changed from mutable Array to
frozen `ColumnFamilyMap`.

[closes #4133]

Note: This PR only partially changes `ColumnFamilyMap`. Until the PR for #4134 is completed, `ColumnFamilyMap#add` will continue to store `Google::Bigtable::Admin::V2::ColumnFamily` objects as well.

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
@@ -485,7 +485,7 @@ module Google
         #   # Default view is full view
         #   table = instance.table("my-table", perform_lookup: true)
         #   puts table.name
-        #   puts table.column_families.inspect
+        #   puts table.column_families
         #
         #   # Name-only view
         #   table = instance.table("my-table", view: :NAME_ONLY, perform_lookup: true)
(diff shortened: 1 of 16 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/instance.rb
```diff
@@ -485,7 +485,7 @@ def tables
         #   # Default view is full view
         #   table = instance.table("my-table", perform_lookup: true)
         #   puts table.name
-        #   puts table.column_families
+        #   puts table.column_families.inspect
```
blowmage: Why is `#inspect` being added to these lines?

quartzmo: Thanks for catching that. Removed.

quartzmo:
@dazuma Does this new method return type look OK to you?

PR #4168, opened 2019-10-09
chore: fix config builder

TheRoyalTnetennba:


rakelib/kokoro/kokoro_builder.rb
```diff
@@ -39,8 +39,8 @@ def from_template template, output, gem: nil, base: nil
   def build_kokoro_configs
     gems.each do |gem|
       name = gem.split("google-cloud-").last
-      [:windows, :osx].each do |os_version|
-        [:presubmit, :continuous, :nightly].each do |build_type|
```
blowmage: is `:continuous` is the same thing as `:presubmit`? Will we still get builds for PRs with this change?

TheRoyalTnetennba: We will still get builds for PRs. This change makes it so that running `bundle exec rake kokoro:build` no longer generates the config files required to run all the test in parallel by gem for PRs only.

blowmage: gotcha, thanks.

PR #4201, opened 2019-10-16
feat(bigtable): Add Bigtable::RoutingPolicy

quartzmo:
BREAKING CHANGES: All return types and parameters associated with
AppProfile routing policy have been changed to wrapper types. The
new types have exactly the same API as the old types, so this
change should only affect type introspection.

[refs #4146]


quartzmo:
Rebased on latest master.

PR #4269, opened 2019-10-23
Allow passing an empty array as a value for params

knu:
Currently passing an empty array causes the following error:
```
NoMethodError: undefined method `parameter_type' for nil:NilClass
```
This is due to Convert.to_query_param trying to use the type of the first element as array_type without checking if a given array has an element.

This PR adds the necessary check.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/convert.rb
@@ -199,10 +199,16 @@ module Google
             )
           elsif Array === value
             array_params = value.map { |param| Convert.to_query_param param }
+            array_type =
+              if array_params.empty?
+                Google::Apis::BigqueryV2::QueryParameterType.new(type: "STRING")
+              else
+                array_params.first.parameter_type
+              end
             return Google::Apis::BigqueryV2::QueryParameter.new(
               parameter_type: Google::Apis::BigqueryV2::QueryParameterType.new(
                 type: "ARRAY",
-                array_type: array_params.first.parameter_type
+                array_type: array_type
               ),
               parameter_value: Google::Apis::BigqueryV2::QueryParameterValue.new(
                 array_values: array_params.map(&:parameter_value)
--- a/google-cloud-bigquery/test/google/cloud/bigquery/convert_test.rb
+++ b/google-cloud-bigquery/test/google/cloud/bigquery/convert_test.rb
@@ -63,4 +63,24 @@ describe Google::Cloud::Bigquery::Convert do
       f.must_be :nan?
     end
   end
+
+  describe :to_query_param do
+    it "converts an array correctly" do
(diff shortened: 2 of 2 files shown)


quartzmo:
@knu Thank you for submitting this pull request!

blowmage:
Thank you for your pull request. Unfortunately we have addressed this issue in a different manner. After discussions with the BigQuery team we have decided to add an optional `types` argument. This new argument is used to specify the type of a query parameter when one can't be determined. So while PR has elected to use STRING when a type can't be determined, users will have to specify this as follows:

```ruby
data = dataset.query "SELECT name FROM my_table " \
                     "WHERE url IN UNNEST(@urls)",
                     params: { urls: [] },
                     types: { urls: [:STRING] }
```

Users are also able to specify a non-STRING array:

```ruby
data = dataset.query "SELECT name FROM my_table " \
                     "WHERE id IN UNNEST(@ids)",
                     params: { ids: [] },
                     types: { ids: [:INT64] }
```

pdfabbro:
While I see the code that should handle this here https://github.com/googleapis/nodejs-bigquery/blob/master/src/bigquery.ts passing in a type as documented doesn't seem to work.  I still get the "Parameter types must be provided for null values via the 'types' field in query options.".  Also, the documentation for createQueryJob doesn't show how to pass types (or params for that matter), so it should probably be updated https://googleapis.dev/nodejs/bigquery/latest/BigQuery.html#createQueryJob

If i make a call similar to the following

const createQueryJobHandler = function(req, res) {
  const params = req.body;
  const types = params.types;

  getQueryFromSQLFile(params.queryType).then(query => {
    if (query === "") {
      res.status(401).send({ error: "Invalid query type provided" });
    }

    bigQuery.createQueryJob({
      query,
      params,
      types
    }).then(data => {

and pass zones:[] as a param and types:{zones: ["String"]} as another param it doesn't work.  Is there something I'm missing?

quartzmo:
@pdfabbro Did you intend to post this comment regarding googleapis/nodejs-bigquery here on a googleapis/google-cloud-ruby PR?

pdfabbro:
@quartzmo Well, it's the same general idea, but yes if you have a link to the node version I'd be happy to post it there and delete my comments

quartzmo:
I think maybe you should post it here? Sorry if I'm not understanding something about how this is relevant to the Ruby client.

https://github.com/googleapis/nodejs-bigquery/issues

pdfabbro:
Okay will do.  It's not relevant to the Ruby version, just the google api in general as they all work the same basic way.  Thanks for the link!

quartzmo:
Great, thanks. I'll verify your observations with the Ruby client and open a new issue for this repo (a sibling to the nodejs repo) if needed. Appreciate the heads up!

quartzmo:
The new issue related to null parameter values is googleapis/nodejs-bigquery#915.

PR #4275, opened 2019-10-24
feat(storage): Add force_copy_metadata to File#copy and #rewrite

### LINKED ISSUE #4254: storage: content-type is not copied when providing a block in copy
#### Environment details

  - OS: macOS 10.15
  - Ruby version: 2.6.4
  - Gem name and version: google-cloud-storage 1.21.1

#### Steps to reproduce

Call `File#copy` on an existing file with a `content-type`, and provide a block to set, for example, `cache_control`. The returned file wont have the correct `content-type.`

#### Code example

```ruby

gcs = Google::Cloud::Storage.new(project: '##project')
bucket_name = "##bucket_name"

bucket = gcs.bucket(b, skip_lookup: true)

original_file = bucket.file("test.png")

original_file.content_type
# => "image/png"

test1 = original_file.copy(bucket_name, "test_1.png")
test1.content_type
# => "image/png"

test2 = original_file.copy(bucket_name, "test_2.png") { |f| f.cache_control = "public, max-age: 7200" }
test2.cache_control
# => "public, max-age: 7200"
test2.content_type
# => nil
```

This has been first reported here: https://github.com/renchap/shrine-google_cloud_storage/issues/36
I suspect it comes from https://github.com/googleapis/google-cloud-ruby/blob/2ad4e22280dc8a2e61613251461e8049129deb4b/google-cloud-storage/lib/google/cloud/storage/file.rb#L1201-L1208
and the code overrides the attributes copied from the original file with `nil`, but I did not debug this further.

quartzmo:
feat(storage): Add `force_copy_metadata` to `File#copy` and `#rewrite`.

```ruby 
        # @param [Boolean] force_copy_metadata Optional. If `true` and if updates
        #   are made in a block, the following fields will be copied from the
        #   source file to the destination file (except when changed by updates):
        #
        #   * `cache_control`
        #   * `content_disposition`
        #   * `content_encoding`
        #   * `content_language`
        #   * `content_type`
        #   * `metadata`
        #
        #   If `nil` or `false`, only the updates made in the yielded block will
        #   be applied to the destination object. The default is `nil`.
```

[closes #4254]

API docs for [Objects: rewrite](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite) specify that:

> Request body
> In the request body, optionally supply metadata to apply to the rewritten object using an object resource.

The current implementation passes a patch-type resource representation that only contains the attributes to be **changed**. It seems that a full representation is needed, or else data is lost as observed in the updated acceptance tests in this PR and in #4254. This PR changes the Updater's `gapi` object to a full duplicate of the origin file's `gapi` state.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -951,14 +951,15 @@ module Google
         end
 
         ##
-        # Copies the file to a new location. Metadata from the source object will
-        # be copied to the destination object unless a block is provided.
+        # Copies the file to a new location. Metadata excluding ACL from the source
+        # object will be copied to the destination object unless a block is provided.
         #
-        # If the optional block for updating metadata fields is provided, only the
-        # updates made in this block will appear in the destination object, and
-        # other metadata fields in the destination object will be empty. To copy
-        # all source object metadata while updating destination object fields in a
-        # block, use the `force_copy_metadata: true` flag.
+        # If an optional block for updating is provided, only the updates made in
+        # this block will appear in the destination object, and other metadata
+        # fields in the destination object will not be copied. To copy the other
+        # source file metadata fields while updating destination fields in a
+        # block, use the `force_copy_metadata: true` flag, and the client library
+        # will copy metadata from source metadata into the copy request.
         #
         # If a [customer-supplied encryption
         # key](https://cloud.google.com/storage/docs/encryption#customer-supplied)
@@ -994,10 +995,18 @@ module Google
         #   AES-256 encryption key used to encrypt the file, if one was provided
         #   to {Bucket#create_file}.
         # @param [Boolean] force_copy_metadata Optional. If `true` and if updates
-        #   are made in a block, the metadata for the source object will be copied
-        #   in its entirety to the destination object. Otherwise, only the updates
-        #   made in the yielded block will be applied to the destination object,
-        #   and its other metadata fields will be empty. The default is `nil`.
+        #   are made in a block, the following fields will be copied from the
+        #   source file to the destination file (except when changed by updates):
+        #
+        #   * `cache_control`
+        #   * `content_disposition`
+        #   * `content_encoding`
+        #   * `content_language`
+        #   * `content_type`
+        #   * `metadata`
+        #
+        #   If `nil` or `false`, only the updates made in the yielded block will
+        #   be applied to the destination object. The default is `nil`.
         # @yield [file] a block yielding a delegate object for updating
         #
         # @return [Google::Cloud::Storage::File]
@@ -1064,11 +1073,12 @@ module Google
         # rewrite the file in place. Metadata from the source object will
         # be copied to the destination object unless a block is provided.
         #
-        # If the optional block for updating metadata fields is provided, only the
-        # updates made in this block will appear in the destination object, and
-        # other metadata fields in the destination object will be empty. To copy
-        # all source object metadata while updating destination object fields in a
-        # block, use the `force_copy_metadata: true` flag.
+        # If an optional block for updating is provided, only the updates made in
+        # this block will appear in the destination object, and other metadata
+        # fields in the destination object will not be copied. To copy the other
+        # source file metadata fields while updating destination fields in a
+        # block, use the `force_copy_metadata: true` flag, and the client library
+        # will copy metadata from source metadata into the copy request.
         #
         # If a [customer-supplied encryption
         # key](https://cloud.google.com/storage/docs/encryption#customer-supplied)
@@ -1117,10 +1127,18 @@ module Google
         #   your project requires access to this encryption key. Do not provide
         #   if `new_encryption_key` is used.
         # @param [Boolean] force_copy_metadata Optional. If `true` and if updates
-        #   are made in a block, the metadata for the source object will be copied
-        #   in its entirety to the destination object. Otherwise, only the updates
-        #   made in the yielded block will be applied to the destination object,
-        #   and its other metadata fields will be empty. The default is `nil`.
+        #   are made in a block, the following fields will be copied from the
+        #   source file to the destination file (except when changed by updates):
+        #
+        #   * `cache_control`
+        #   * `content_disposition`
+        #   * `content_encoding`
+        #   * `content_language`
+        #   * `content_type`
+        #   * `metadata`
+        #
+        #   If `nil` or `false`, only the updates made in the yielded block will
+        #   be applied to the destination object. The default is `nil`.
         # @yield [file] a block yielding a delegate object for updating
         #
         # @return [Google::Cloud::Storage::File]
@@ -1226,11 +1244,8 @@ module Google
             yield updater
             updater.check_for_changed_metadata!
             if updater.updates.any?
-              update_gapi = if force_copy_metadata
-                              updater.gapi
-                            else
-                              self.class.gapi_from_attrs updater.gapi, updater.updates
-                            end
+              attributes = force_copy_metadata ? (Updater::COPY_ATTRS + updater.updates).uniq : updater.updates
+              update_gapi = self.class.gapi_from_attrs updater.gapi, attributes
             end
           end
 
@@ -1714,6 +1729,9 @@ module Google
           attr_params = Hash[attributes.map do |attr|
                                [attr, gapi.send(attr)]
                              end]
+          # Sending nil metadata results in an Apiary runtime error:
+          # NoMethodError: undefined method `each' for nil:NilClass
+          attr_params.reject! { |k, v| k == :metadata && v.nil? }
           Google::Apis::StorageV1::Object.new attr_params
         end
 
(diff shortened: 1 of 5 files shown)


quartzmo:
@renchap PTAL.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -1199,11 +1199,11 @@ def rewrite dest_bucket_or_path, dest_path = nil,
 
           update_gapi = nil
           if block_given?
-            updater = Updater.new gapi
+            updater = Updater.new gapi.dup
```
quartzmo: @frankyn Do you think there is any danger in passing the *complete* representation of the `sourceObject` as the request body (Object resource) to `objects.rewrite`? This resource would presumably contain `bucket`, `name`, etc as well as the desired "metadata" fields such as `cacheControl`, `contentType`, etc.

quartzmo: I've tested this solution and it seems to work as desired, but the actual internal behavior of `objects.rewrite` is opaque.

frankyn: How much metadata was being copied in the past? 

I'd rather not magically copy metadata if the user didn't request the copy. @tseaver I think this is similar to the ACL copy in client level.

quartzmo: @frankyn PTAL at the code example in #4254 for current behavior. Our users via shrine-google_cloud_storage find the current behavior of not copying fields such as `content-type` to be problematic.

quartzmo: > How much metadata was being copied in the past? I'd rather not magically copy metadata if the user didn't request the copy.

@renchap Any comment?

renchap: The thing here is that `File#copy` has an inconsistent behaviour. When you pass it a block then metadata are not copied, but if you do not pass a block then they are copied.
I think most of the time people want the metadata to be copied as well, but maybe we can make the default behavior consistent (either always copy, or never copy), and use a method argument to enable/disable metadata copy?

quartzmo: @frankyn: My observation is that @renchap is correct in that if **no block** is provided, the client does **not** pass an object resource as the request body, and the GCS service copies the object with all metadata intact. Only when the the client passes a partial, "PATCH-style" object resource with only *changed* fields populated (in the current block behavior) does the GCS service copy the object **without** all metadata intact. Since the default behavior (no object resource) of [Objects: rewrite](https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite) is to copy all metadata, this behavior can be presumed to be the correct one. My conclusion is that Objects: rewrite was never intended to accept partial, "PATCH-style" object resource representations. Therefore I feel the change in this PR is correct.

quartzmo: I added a [commit](https://github.com/googleapis/google-cloud-ruby/pull/4275/commits/440b78ee010944c54bf9a88d73f77b5a48c60869) demonstrating that if no block is provided, the GCS service copies the object with all metadata intact.

frankyn: Hi folks, I appreciate your patience.

@quartzmo and @renchap, thanks for clarifying.

Objects: rewrite, will only copy object metadata when a resource body isn't provided. If a resource body is provided GCS considers this the expected result which acts as an UPDATE rather than a PATCH, e.g. only modify the parts but keep the rest.

I agree that it's confusing that the rest of the metadata isn't copied, but if if the client forces a PATCH style version of rewrite, behavior expectations will change for the methods File#copy and File#rewrite. Existing users might dependent on this behavior and updating it could break them. 

I'd recommend a flag to copy existing metadata and allow it be modified with a block. In this way the actual API behavior is transparent to users and a user can clearly opt-into this type of request.

```ruby

gcs = Google::Cloud::Storage.new(project: '##project')
bucket_name = "##bucket_name"

bucket = gcs.bucket(b, skip_lookup: true)

original_file = bucket.file("test.png")

original_file.content_type

# force_copy_metadata - copy existing metadata which can be used in tandem with a block to perform a PATCH like request.
#           If no changes to the source object metadata is required then Cloud Storage will copy metadata.

copied_file = original_file.copy(bucket_name, "test_2.png", force_copy_metadata: true) do |f|
   f.cache_control = "public, max-age: 7200"
end
``` 


What do you both think?

quartzmo: An opt-in flag sounds good to me. @renchap?

renchap: Looks good as well, as well as a clear explanation of the behaviour in the documentation of the gem, so people are not confused by what is copied and when (even if this matches the standard GCS behaviour, all users of the Ruby gem may not know the small details of GCS)

quartzmo: Great, I'll add a commit adding documentation of the default behavior, and also adding the flag `force_copy_metadata` to `File#copy` (but not `File#rewrite` unless someone has a reason to do so).

quartzmo: @renchap @frankyn I added `force_copy_metadata` and updated docs in `File#copy` and `#rewrite`.

quartzmo: @frankyn PTAL

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -951,7 +951,14 @@ def download path = nil, verify: :md5, encryption_key: nil, range: nil,
         end
 
         ##
-        # Copy the file to a new location.
+        # Copies the file to a new location. Metadata from the source object will
+        # be copied to the destination object unless a block is provided.
```
frankyn: Metadata excluding ACL from the source object will be copied to...

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -951,7 +951,14 @@ def download path = nil, verify: :md5, encryption_key: nil, range: nil,
...
+        #
+        # If the optional block for updating metadata fields is provided, only the
+        # updates made in this block will appear in the destination object, and
+        # other metadata fields in the destination object will be empty. To copy
+        # all source object metadata while updating destination object fields in a
+        # block, use the `force_copy_metadata: true` flag.
```
frankyn: Add: 
> use the `force_copy_metadata: true` flag, and the library will copy metadata from source metadata into the copy request.

I want the user to understand the library is doing it rather than the API.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -986,6 +993,11 @@ def download path = nil, verify: :md5, encryption_key: nil, range: nil,
         # @param [String] encryption_key Optional. The customer-supplied,
         #   AES-256 encryption key used to encrypt the file, if one was provided
         #   to {Bucket#create_file}.
+        # @param [Boolean] force_copy_metadata Optional. If `true` and if updates
+        #   are made in a block, the metadata for the source object will be copied
+        #   in its entirety to the destination object. Otherwise, only the updates
+        #   made in the yielded block will be applied to the destination object,
+        #   and its other metadata fields will be empty. The default is `nil`.
```
frankyn: I want to make sure, ACLs continue not to be copied.

google-cloud-storage/lib/google/cloud/storage/file.rb
quartzmo: Will do.

google-cloud-storage/lib/google/cloud/storage/file.rb
quartzmo: Will do.

google-cloud-storage/lib/google/cloud/storage/file.rb
quartzmo: I added an assertion to the acceptance test ensuring that the ACL is not copied, and it currently fails. It seems to me that the safest solution is probably a whitelist of fields that should be copied, rather than excluding fields such as `acl`. Here is my proposed whitelist:

* cache_control
* content_disposition
* content_encoding
* content_language
* content_type
* metadata

@frankyn @renchap What do you think?

frankyn: Clarifying, metadata is custom metadata. 
This is a good idea, thanks @quartzmo. We want to be careful not to copy all metadata especially ACL.

quartzmo: OK, thanks @frankyn, I'll go ahead with this whitelist.

quartzmo: Updated, PTAL.

quartzmo:
@renchap PTAL at latest updates to this PR. Does this solution work for you?

renchap:
LGTM 👍

quartzmo:
Released in [1.24.0](https://rubygems.org/gems/google-cloud-storage/versions/1.24.0).

PR #4284, opened 2019-10-25
Update default endpoint for google-cloud-storage

frankyn:
Updating default endpoint prior to discovery document being updated for GCS libraries. 

The change is updating the endpoint from: `https://www.googleapis.com/` to the new endpoint for the JSON API `https://storage.googleapis.com/`. The old endpoint will continue to work. 

This is a temporary change until the auto-generated client is updated when the discovery document is updated.


quartzmo:
@frankyn Can you add a PR description explaining the reason for this change? You want to change the GCS endpoint that will be used by (almost) all users from the current default in `Google::Apis::StorageV1::StorageService`?

quartzmo:
I believe the current endpoint is `https://www.googleapis.com/storage/v1/`

frankyn:
@quartzmo, I can't share more information that this right now. I'll share more context with @dazuma and @TheRoyalTnetennba.

PR #4318, opened 2019-10-29
Add allUsers in Groups hash

pradeepbhadani:
As per https://cloud.google.com/iam/docs/overview#allusers, Google Bigquery has `allUsers` as valid ACL and is different than `allAuthenticatedUsers`.
Adding this will help to identify public ACL for bigquery dataset in tools like inspec which uses this package.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset/access.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset/access.rb
@@ -75,6 +75,7 @@ module Google
             "all"                     => "allAuthenticatedUsers",
             "all_authenticated_users" => "allAuthenticatedUsers",
             "allAuthenticatedUsers"   => "allAuthenticatedUsers"
+            "allUsers"                => "allUsers"
           }.freeze
 
           ##


pradeepbhadani:
@blowmage - I have updated access.rb to incorporate suggested changes.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset/access.rb
```diff
@@ -277,7 +279,7 @@ def add_writer_domain domain
           # Add writer access to a special group.
           #
           # @param [String] group Accepted values are `owners`, `writers`,
-          #   `readers`, and `all`.
+          #   `readers`, `all_authenticated_users`, and `all_users`.
```
blowmage: Could you also update the code examples used in these methods to use the `:all_users` value instead of `:all`?

```ruby
          #   dataset.access do |access|
          #     access.add_writer_special :all_users
          #   end
```

pradeepbhadani: updated code examples

blowmage:
Wonderful! Thank you for your contribution @pradeepbhadani!

pradeepbhadani:
@blowmage  Thanks for accepting my PR. Happy to contribute :)

Issue #4422, opened 2019-11-15
Spanner code generation not adding service address to operation client


jiren:
Here are code locations

DatabaseAdminClient

https://github.com/googleapis/google-cloud-ruby/blob/1d9dbf281a7cf833b554209598c73a734713fa5d/google-cloud-spanner/lib/google/cloud/spanner/admin/database/v1/database_admin_client.rb#L173-L181

InstanceAdminClient

https://github.com/googleapis/google-cloud-ruby/blob/1d9dbf281a7cf833b554209598c73a734713fa5d/google-cloud-spanner/lib/google/cloud/spanner/admin/instance/v1/instance_admin_client.rb#L206-L215

PR #4423, opened 2019-11-15
fix(spanner) : add service address and port for operation client

jiren:
Issue #4422


blowmage:
Can you also run `synthtool` so we can see what this change produces?

jiren:
> Can you also run `synthtool` so we can see what this change produces?

Added generated code.

quartzmo:
@blowmage This has been updated, PTAL.

blowmage:
@dazuma Is this something that should be generalized everywhere?

blowmage:
NM, it looks like it is everywhere but these two instances.

PR #4480, opened 2019-12-05
Allow row inserts to skip insert_id generation

### LINKED ISSUE #4452: BigQuery: provide a way to opt-out of sending row IDs when streaming rows to a BigQuery table
**Is your feature request related to a problem? Please describe.**
To get 1,000,000 rows per second quota, the row ID needs to be omitted. See: https://cloud.google.com/bigquery/quotas#streaming_inserts

 **Describe the solution you'd like**

Continue to provide row IDs by default, as that reduces the number of duplicate rows. Allow the developer to opt-in to the higher quotas by explicitly asking for no row ID. In Python we can check if None is supplied as the row ID and use that. In Go, that's not possible so they use `"NoDedupeID"` as an indicator for this behavior.

 **Describe alternatives you've considered**

There's an argument to be made that row IDs could be omitted by default, but this could result in many more duplicate rows for the majority of developers who do not require the higher quotas.

 **Additional context**

* C# issue: https://github.com/googleapis/google-cloud-dotnet/issues/3769
* Go commit: https://github.com/googleapis/google-cloud-go/commit/1366340a586e4da4c221c5c7dee6f08c977864dc
* Python issue: https://github.com/googleapis/google-cloud-python/issues/9539

blowmage:
* Streaming inserts using an insert_id are not able to be inserted as fast as inserts without an insert_id.
* Add the ability for users to skip insert_id generation in order to speed up the inserts.
* The default behavior continues to generate insert_id values for each row inserted.

[closes #4452]

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -2083,12 +2083,13 @@ module Google
         # @param [String] table_id The ID of the destination table.
         # @param [Hash, Array<Hash>] rows A hash object or array of hash objects
         #   containing the data. Required.
-        # @param [Array<String>] insert_ids A unique ID for each row. BigQuery
-        #   uses this property to detect duplicate insertion requests on a
-        #   best-effort basis. For more information, see [data
-        #   consistency](https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataconsistency).
-        #   Optional. If not provided, the client library will assign a UUID to
-        #   each row before the request is sent.
+        # @param [Array<String|Symbol>, Symbol] insert_ids A unique ID for each row. BigQuery uses this property to
+        #   detect duplicate insertion requests on a best-effort basis. For more information, see [data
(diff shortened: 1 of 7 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -2144,32 +2145,19 @@ def self.new_reference project_id, dataset_id, service
         #
         # @!group Data
         #
-        def insert table_id, rows, insert_ids: nil, skip_invalid: nil, ignore_unknown: nil, autocreate: nil
+        def insert table_id, rows, insert_ids: nil, skip_invalid: nil, ignore_unknown: nil, autocreate: nil, &block
```
quartzmo: I know this PR does not add the yielded block, but it appears to me that the yield is undocumented.

blowmage: I think you are correct.

blowmage: Added docs for this. @quartzmo please re-review.

quartzmo: Thanks!

Issue #4511, opened 2019-12-16
Firestore: DocumentReference not behaving consistently


TheRoyalTnetennba:
#### Environment details

  - OS: Debian
  - Ruby version: 2.6.5
  - Gem name and version: google-cloud-firestore-1.2.1

#### Setup

```ruby
require "google/cloud/firestore"

firestore = Google::Cloud::Firestore.new
col = firestore.col "test_col"
doc = col.doc "test_doc"
doc.create key: "value"

docs = col.list_documents
```

#### Works

```ruby
doc.delete
```

#### Errors

```ruby
docs[0].delete
```

#### Full backtrace

```sh
.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/google-cloud-firestore-1.2.1/lib/google/cloud/firestore/document_reference.rb:455:in `delete': undefined method `batch' for Google::Cloud::Firestore::Service(grahampaye-experiment):Google::Cloud::Firestore::Service (NoMethodError)
```

PR #4561, opened 2020-01-07
fix(firestore): Use client instead of service in DocumentReference::List

quartzmo:
fixes: #4511


blowmage:
Do we need a test to cover the setup described in #4511?

PR #4733, opened 2020-01-23
Add google-cloud-billing-v1

quartzmo:
This client was generated using the `gapic-generator-ruby` microgenerator via the included [`synth.py`](https://github.com/googleapis/google-cloud-ruby/blob/a448d680194b223d31fa29c9dfa07325883371f5/google-cloud-billing-v1/synth.py) script.


quartzmo:
Added google-cloud-billing-v1 to the toplevel Gemfile, all builds green now.

PR #4788, opened 2020-02-07
test(spanner): fix tests to avoid repeated updates to stubbed method calls.

hengfengli:
## Problem

I am working on a feature that needs to add a parameter to `execute_streaming_sql` method. However, this is really annoying because it causes around 250 errors. I need to add this parameter in each single place. 

## Changes

1) Add a helper function `MockSpanner.expect_execute_streaming_sql`, which takes expected results and expected parameters. If parameters are not given, the default values will be expected. 

With this change, I only need to add new parameter in this function (only one place) instead of hundreds of places. 

2) 

Before: 
```ruby
      stubbed_service = ->(project, credentials, timeout: nil, host: nil, client_config: nil) {
        project.must_equal "project-id"
        credentials.must_equal "spanner-credentials"
        timeout.must_be :nil?
        host.must_be :nil?
        client_config.must_be :nil?
        OpenStruct.new project: project
      }
```

After: 
```ruby
      stubbed_service = ->(project, credentials, timeout: nil, host: nil, client_config: nil, **keyword_args) {
        project.must_equal "project-id"
        credentials.must_equal "spanner-credentials"
        timeout.must_be :nil?
        host.must_be :nil?
        client_config.must_be :nil?
        OpenStruct.new project: project
      }
```

With adding `**keyword_args`, when we add new parameter for the stubbed service, we do not need to change hundreds of places. We only need to add a single test to see whether the stubbed service respects the new parameters or not.

### DIFF
--- a/google-cloud-spanner/test/helper.rb
+++ b/google-cloud-spanner/test/helper.rb
@@ -147,14 +147,12 @@ class MockSpanner < Minitest::Spec
   def paged_enum_struct response
     OpenStruct.new page: OpenStruct.new(response: response)
   end
-end
 
-class MockSpannerHelper
-  def self.expect_execute_streaming_sql mock, results_enum, session_name, sql,
+  def expect_execute_streaming_sql results_enum, session_name, sql,
                                    transaction: nil, params: nil, param_types: nil, 
                                    resume_token: nil, partition_token: nil, seqno: nil,
                                    options: nil
-    mock.expect :execute_streaming_sql, results_enum do |session, sql_query, **kargs|
+    spanner.service.mocked_service.expect :execute_streaming_sql, results_enum do |session, sql_query, **kargs|
       session == session_name &&
         sql_query == sql_query &&
         kargs[:transaction] == transaction &&
--- a/google-cloud-spanner/test/helper.rb
+++ b/google-cloud-spanner/test/helper.rb
@@ -149,6 +149,25 @@ class MockSpanner < Minitest::Spec
   end
 end
 
+class MockSpannerHelper
+  def self.expect_execute_streaming_sql mock, results_enum, session_name, sql,
+                                   transaction: nil, params: nil, param_types: nil, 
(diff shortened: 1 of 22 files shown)


google-cloud-spanner/test/helper.rb
```diff
@@ -149,6 +149,25 @@ def paged_enum_struct response
   end
 end
 
+class MockSpannerHelper
+  def self.expect_execute_streaming_sql mock, results_enum, session_name, sql,
```
quartzmo: What do you think about moving this to be an instance method on the existing `MockSpanner` helper class? Then it could be called without a receiver like the other helpers, for example:

```

    mock = Minitest::Mock.new
    expect_execute_streaming_sql mock, results_enum, session.path, sql, transaction: tx_selector, param_types: {}, partition_token: partition_token, options: default_options
    batch_snapshot.session.service.mocked_service = mock
```

quartzmo: Since `MockSpanner` has a reference to the `spanner` client object used in the tests, you should even be able to drop the `mock` parameter from your method as follows, as long as you set the mock on the client in the test before calling `expect`. (There are a couple layers of client in this lib, so it might not work in all cases, but you could add references as needed in `MockSpanner`.)

```
  def expect_execute_streaming_sql results_enum, session_name, sql,
                                   transaction: nil, params: nil, param_types: nil, 
                                   resume_token: nil, partition_token: nil, seqno: nil,
                                   options: nil
      spanner.service.mocked_service.expect :execute_streaming_sql, results_enum do |session, sql_query, **kargs|
  ...
```

and 

```

    mock = Minitest::Mock.new
    batch_snapshot.session.service.mocked_service = mock
    expect_execute_streaming_sql results_enum, session.path, sql, transaction: tx_selector, param_types: {}, partition_token: partition_token, options: default_options
  ...
```

hengfengli: @quartzmo Thanks for pointing out this. I have updated it.

PR #4884, opened 2020-02-27
Support Ruby 2.8 keyword arguments for google-cloud-storage

yahonda:
This pull request supports Ruby 2.8 keyword arguments for google-cloud-storage

[Fully separate positional arguments and keyword arguments](https://github.com/ruby/ruby/pull/2794) support is required to use google-cloud-storage gem with currently developed Ruby 2.8.0-dev.

### Note
* `test/google/cloud/storage_test.rb` failure will be addressed once the newer version of minitest is released including [If a callable mock includes kwargs, Ruby 2.7 warns](https://github.com/seattlerb/minitest/pull/824)

* Workaround to force install grpc gem Refer https://github.com/grpc/grpc/issues/21514#issuecomment-581417788
Note: This commit must be reverted once the newer version of grpc is released to rubygems.org including https://github.com/grpc/grpc/pull/22039

PR #4897, opened 2020-03-03
Bump Rake version to prepare Ruby 3 keyword arguments support

yahonda:
Related to #4884

To prepare Ruby 3 keyword arguments support, there are some other gems also need update. `rake` is one of them.

* Rake 13.0.0 supports Ruby 3 keyword arguments by https://github.com/ruby/rake/pull/326
* Rake 13 drops Ruby 2.1 or older support by https://github.com/ruby/rake/pull/325
* Gems under `google-cloud-ruby` repository only support Ruby 2.4 or higher then dropping Ruby 2.1 support should not be an issue.

I have bumped all of Rake under `google-cloud-ruby` repository.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-gem "rake", "~> 12.3"
+gem "rake"
 # Pin minitest to 5.11.x to avoid warnings emitted by 5.12.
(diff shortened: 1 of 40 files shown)


quartzmo:
@yahonda Thank you for this PR, and sorry this wasn't done already. We really appreciate that you took the time to update all packages.

PR #4907, opened 2020-03-05
feat(spanner): support query options

hengfengli:
This adds the support of providing query options (optimizer version) via three ways (precedence follows the order):

client-level < environment variables < query-level.

The environment variable `SPANNER_OPTIMIZER_VERSION` can be set: 

```bash
export SPANNER_OPTIMIZER_VERSION=1
```


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -97,6 +98,13 @@ def database
           @project.database instance_id, database_id
         end
 
+        # A hash of values to specify the custom
+        # query options for executing SQL query.
```
skuruppu: I think this comment can be reformatted.

hengfengli:
> @hengfengli if you can revert the env var name back to SPANNER_OPTIMIZER_VERSION, that would be great, since we decided to leave it as is to be consistent with the query hint.

Done. I just reverted it.

hengfengli:
@skuruppu I think this can be merged. I only added the support of query options to the last piece `transaction.execute_update` after your approval.

PR #5085, opened 2020-03-19
chore: enable context-aware commits for one API

SurferJeffAtGoogle:


SurferJeffAtGoogle:
> A python global seems like an odd way to configure this. I'll trust that it's intentional for now because I don't have a way of testing this locally.

Thanks for the quick review.  Yes, it's intentional.  Here are some other synth.py's configured the same way: https://github.com/search?q=AUTOSYNTH_MULTIPLE_COMMITS&type=Code

PR #5106, opened 2020-03-21
Official quoting methods for idents and literals

seamusabshere:
Database systems, BigQuery included, need readily-available quoting functions. This adds a Ruby implementation.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery.rb
@@ -116,6 +116,47 @@ module Google
 
         Google::Cloud.configure.bigquery
       end
+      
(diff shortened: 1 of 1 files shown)


seamusabshere:
@googlebot I signed it!

quartzmo:
@seamusabshere Thank you for this PR. We'll take a look at it as soon as possible!

seamusabshere:
I'm not saying this is the correct implementation. For example, `#iso8601` is (I believe) provided by ActiveSupport. I'm more interested to know if you even _want_ a native ruby implementation, or if you do code generation elsewhere.

quartzmo:
@seamusabshere I'm going to close this PR for now and let this remain something that users handle externally. Thanks very much for this useful start however, should we receive more interest in this feature.

PR #5702, opened 2020-04-24
fix(storage): Add missing bucket condition in SignerV4#post_object

quartzmo:
* Update conformance tests to use correct form for custom metadata.
* Update conformance tests to add missing condition `bucket`.
* Add missing `bucket` condition in `SignerV4#post_object`.

refs: googleapis/conformance-tests#29
refs: googleapis/conformance-tests#31


google-cloud-storage/lib/google/cloud/storage/file/signer_v4.rb
```diff
@@ -164,6 +164,7 @@ def escape_special_unicode str
 
           def required_fields issuer, time
             {
+              "bucket" => @bucket_name,
```
frankyn, line 167: @quartzmo does this output a field `bucket` in returned fields? We may have already discussed this but double checking.

frankyn: Gotcha, yes please, could you update it to only include the name in the policy only and not return it as a field.?

quartzmo: Yes, will do.

quartzmo: Fixed in #5803.

PR #5818, opened 2020-05-02
chore(dlp): Move samples from the ruby-docs-samples repo

viacheslav-rostovtsev:
see GoogleCloudPlatform/ruby-docs-samples#567 for the in-place migration of samples to the minitest

move samples for dlp from ruby-docs-samples
update the samples to use the microgenerated library


viacheslav-rostovtsev:
do not merge until https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/567 completes

PR #6491, opened 2020-06-09
feat(spanner): add numeric support

hengfengli:

### DIFF
--- a/google-cloud-spanner/test/google/cloud/spanner/convert/grpc_value_to_object_test.rb
+++ b/google-cloud-spanner/test/google/cloud/spanner/convert/grpc_value_to_object_test.rb
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 require "helper"
+require "bigdecimal"
 
 describe Google::Cloud::Spanner::Convert, :grpc_value_to_object, :mock_spanner do
   # This tests is a sanity check on the implementation of the conversion method.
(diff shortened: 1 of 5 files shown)


google-cloud-spanner/test/google/cloud/spanner/convert/grpc_value_to_object_test.rb
```diff
@@ -125,4 +126,12 @@
     raw = Google::Cloud::Spanner::Convert.grpc_value_to_object value, type
     _(raw).must_equal(Google::Cloud::Spanner::Fields.new([]).struct([]))
   end
+
+  it "converts a NUMERIC value" do
```
thiagotnunes, line 130: Just to verify my understanding:
- We relying on the server to protect against (under/over)flow.
- We are doing that to keep the client thin.
- On the server side we have a precision of 38 cases total.

If any of the above is false we would need to add / change this test.

hengfengli: Yes, we rely on the backend for checking whether the precision is supported. 

The NUMERIC type supports 38 digits of precision and 9 digits of scale. It means the range would be `-99999999999999999999999999999.999999999` to `99999999999999999999999999999.999999999`. 

Same as NUMERIC type in BigQuery (https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#numeric_type).

Issue #7572, opened 2020-08-14
Add support for Firestore Query#limit_to_last


quartzmo:
Original Node issue: googleapis/nodejs-firestore#945

See existing implementation PRs:

Java: googleapis/java-firestore#151
Node: googleapis/nodejs-firestore#954

PR #7575, opened 2020-08-14
feat(firestore): Add Query#limit_to_last

quartzmo:
closes: #7572
closes: #7594

/cc @schmidt-sebastian

### DIFF
--- a/google-cloud-firestore/acceptance/firestore/watch_test.rb
+++ b/google-cloud-firestore/acceptance/firestore/watch_test.rb
@@ -18,8 +18,8 @@ describe "Watch", :firestore_acceptance do
   it "watches a limit query" do
     watch_col = root_col.doc("watch-limit").col("watch-query")
 
-    watch_col.doc("int 1").create(val: 1)
-    watch_col.doc("int 2").create(val: 2)
+    watch_col.doc("int 5").create(val: 5)
+    watch_col.doc("int 4").create(val: 4)
     watch_col.doc("int 3").create(val: 3)
 
     snps = []
@@ -27,7 +27,10 @@ describe "Watch", :firestore_acceptance do
 
     wait_until { snps.count == 1 }
 
-    watch_col.doc("int 0").create(val: 0)
+    firestore.batch do |b|
+      b.create(watch_col.doc("int 2"), { val: 1 })
+      b.create(watch_col.doc("int 1"), { val: 0 })
+    end
 
     wait_until { snps.count == 2 }
 
@@ -35,24 +38,31 @@ describe "Watch", :firestore_acceptance do
 
     _(snps.count).must_equal 2
     _(snps[0].count).must_equal 2
-    _(snps[0].docs.map(&:document_id)).must_equal ["int 1", "int 2"]
+    _(snps[0].docs.map(&:document_id)).must_equal ["int 3", "int 4"]
+    _(snps[0].changes.map(&:type)).must_equal [:added, :added]
+    _(snps[0].changes.map(&:doc).map(&:document_id)).must_equal ["int 3", "int 4"]
     _(snps[1].count).must_equal 2
-    _(snps[1].docs.map(&:document_id)).must_equal ["int 0", "int 1"]
+    _(snps[1].docs.map(&:document_id)).must_equal ["int 1", "int 2"]
+    _(snps[1].changes.map(&:type)).must_equal [:removed, :removed, :added, :added]
+    _(snps[1].changes.map(&:doc).map(&:document_id)).must_equal ["int 3", "int 4", "int 1", "int 2"]
   end
 
   it "watches a limit_to_last query" do
     watch_col = root_col.doc("watch-limit_to_last").col("watch-query")
 
-    watch_col.doc("int 1").create(val: 1)
-    watch_col.doc("int 2").create(val: 2)
     watch_col.doc("int 3").create(val: 3)
+    watch_col.doc("int 2").create(val: 2)
+    watch_col.doc("int 1").create(val: 1)
 
     snps = []
     listener = watch_col.order(:val).limit_to_last(2).listen { |snp| snps << snp }
 
     wait_until { snps.count == 1 }
 
-    watch_col.doc("int 4").create(val: 4)
+    firestore.batch do |b|
+      b.create(watch_col.doc("int 5"), { val: 5 })
+      b.create(watch_col.doc("int 4"), { val: 4 })
+    end
 
     wait_until { snps.count == 2 }
 
@@ -60,9 +70,13 @@ describe "Watch", :firestore_acceptance do
 
     _(snps.count).must_equal 2
     _(snps[0].count).must_equal 2
-    _(snps[0].docs.map(&:document_id)).must_equal ["int 3", "int 2"]
+    _(snps[0].docs.map(&:document_id)).must_equal ["int 2", "int 3"]
+    _(snps[0].changes.map(&:type)).must_equal [:added, :added]
+    _(snps[0].changes.map(&:doc).map(&:document_id)).must_equal ["int 2", "int 3"]
     _(snps[1].count).must_equal 2
-    _(snps[1].docs.map(&:document_id)).must_equal ["int 4", "int 3"]
+    _(snps[1].docs.map(&:document_id)).must_equal ["int 4", "int 5"]
+    _(snps[1].changes.map(&:type)).must_equal [:removed, :removed, :added, :added]
+    _(snps[1].changes.map(&:doc).map(&:document_id)).must_equal ["int 2", "int 3", "int 4", "int 5"]
   end
 
   it "watches a query" do
--- a/google-cloud-firestore/acceptance/firestore/watch_test.rb
+++ b/google-cloud-firestore/acceptance/firestore/watch_test.rb
@@ -15,6 +15,56 @@
 require "firestore_helper"
 
 describe "Watch", :firestore_acceptance do
+  it "watches a limit query" do
+    watch_col = root_col.doc("watch-limit").col("watch-query")
+
+    watch_col.doc("int 1").create(val: 1)
+    watch_col.doc("int 2").create(val: 2)
+    watch_col.doc("int 3").create(val: 3)
+
+    snps = []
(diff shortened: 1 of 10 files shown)


google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -380,7 +384,63 @@ def limit num
...
+          if limit_type == :last # don't reverse order_by more than once
+            raise "limit_to_last may only be called once"
+          end
+
+          # Flip the orderBy directions since we want the last results
+          new_query.order_by.each do |order|
+            order.direction = order.direction.to_sym == :DESCENDING ? :ASCENDING : :DESCENDING
+          end
+
+          # Swap the cursors to match the flipped query ordering.
```
quartzmo, line 435: See [Java impl](https://github.com/schmidt-sebastian/java-firestore/blob/0a5deea81a1ae907015daff52fa14588c00f76e8/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L1070-L1103)

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -380,7 +384,63 @@ def limit num
...
+
+          if new_query.nil? || new_query.order_by.nil? || new_query.order_by.empty?
+            raise "specify at least one order clause before calling limit_to_last"
+          end
+
+          if limit_type == :last # don't reverse order_by more than once
+            raise "limit_to_last may only be called once"
+          end
+
+          # Flip the orderBy directions since we want the last results
```
quartzmo, line 430: See [Java impl](https://github.com/schmidt-sebastian/java-firestore/blob/0a5deea81a1ae907015daff52fa14588c00f76e8/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L1041-L1049)

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -862,6 +927,10 @@ def get
         def listen &callback
           raise ArgumentError, "callback required" if callback.nil?
 
+          if limit_type == :last
+            raise "Queries that include limit_to_last constraints cannot be streamed. Use Query#get instead."
```
quartzmo, line 931: See [Java impl](https://github.com/schmidt-sebastian/java-firestore/blob/0a5deea81a1ae907015daff52fa14588c00f76e8/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L1115-L1118)

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -828,6 +888,11 @@ def get
           return enum_for :get unless block_given?
 
           results = service.run_query parent_path, @query
+
+          # The results for limitToLast queries need to be flipped since we reversed the ordering constraints before
+          # sending the query to the backend.
```
quartzmo, line 893: See [Java impl](https://github.com/schmidt-sebastian/java-firestore/blob/0a5deea81a1ae907015daff52fa14588c00f76e8/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java#L1277-L1285)

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -380,7 +384,63 @@ def limit num
...
+
+        ##
+        # Limits a query to return only the last matching documents.
+        #
+        # You must specify at least one "order by" clause for limitToLast queries.
+        # (See {#order}.)
+        #
+        # Results for `limit_to_last` queries are only available once all documents
+        # are received. Hence, `limit_to_last` queries cannot be streamed using
+        # {#listen}.
```
schmidt-sebastian, line 398: I am not very familiar with this client and its terminology, but both the Node and the Java client support limit-to-last "listens" via their onSnapshot/Watch API. Not supported are the APIs that stream RunQuery results directly to the user.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -380,7 +384,63 @@ def limit num
...
+        #
+        def limit_to_last num
+          new_query = @query.dup
+
+          if new_query.nil? || new_query.order_by.nil? || new_query.order_by.empty?
+            raise "specify at least one order clause before calling limit_to_last"
+          end
+
+          if limit_type == :last # don't reverse order_by more than once
+            raise "limit_to_last may only be called once"
```
schmidt-sebastian, line 427: Could you not just return early here?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -380,7 +384,63 @@ def limit num
...
+          # Flip the orderBy directions since we want the last results
+          new_query.order_by.each do |order|
+            order.direction = order.direction.to_sym == :DESCENDING ? :ASCENDING : :DESCENDING
+          end
+
+          # Swap the cursors to match the flipped query ordering.
+          new_start_at = new_query.end_at.dup
+          new_end_at = new_query.start_at.dup
+          new_query.start_at = new_start_at
+          new_query.end_at = new_end_at
```
schmidt-sebastian, line 439: You also need to switch the "before" direction.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
schmidt-sebastian: As stated above, this limitation is for Query.stream, not Query.onSnapshot. You should be able to flip the direction of the snapshot for the Listen API.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
quartzmo: Thanks for clearing that up. I'll make the change.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
quartzmo: Fixed, thanks!

google-cloud-firestore/lib/google/cloud/firestore/query.rb
schmidt-sebastian: When you get to this, can you also add a test for Listens with limit-to-last? Thanks

quartzmo: Yes, of course!

quartzmo: Updated, thank you.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
quartzmo: Added a conditional to allow multiple calls to `limit_to_last`. Subsequent calls update `limit`, but do not re-flip the cursors.

google-cloud-firestore/lib/google/cloud/firestore/query.rb
quartzmo: Updated, thank you.

google-cloud-firestore/acceptance/firestore/watch_test.rb
```diff
@@ -15,6 +15,56 @@
...
+
+    watch_col.doc("int 4").create(val: 4)
+
+    wait_until { snps.count == 2 }
+
+    listener.stop
+
+    _(snps.count).must_equal 2
+    _(snps[0].count).must_equal 2
+    _(snps[0].docs.map(&:document_id)).must_equal ["int 3", "int 2"]
```
schmidt-sebastian, line 63: Can we make sure this test verifies that the order is "int 2", "int 3"?

quartzmo: Thanks for catching that, will do. I'll also change line 65 below to `["int 3", "int 4"]`

quartzmo: @schmidt-sebastian Can you see the mistake I made in the implementation that causes this result? This is an integration test that currently passes.

schmidt-sebastian: I can take a look at the code if you need me to, but I haven't yet. My strong suspicion is that you are using the flipped query but not re-flipping the results (you always need to reverse the results if you are flipping the orderBy constraints on the user's behalf). The way I do this is Node and Java is that I use the pre-flipped query to process the local results.

quartzmo: That sounds likely, thank you.

quartzmo: @schmidt-sebastian I updated `Inventory` to reverse the results for Query#limit_to_last queries. I'm sorry I missed that the first time. Please take a look at the additional assertions for `changes` ordering to verify that it is correct. (I'm not sure if `changes` ordering should reflect query `orderBy`.)

quartzmo:
Rebased on latest master to get `samples` directory.

google-cloud-firestore/acceptance/firestore/watch_test.rb
schmidt-sebastian: This looks correct now. It is slightly different than what I have implemented since the Watch client in Ruby uses the reversed query Proto for its ordering, rather than the user-specified Query order. In the end this does not matter though, since we only need to match the user's order by clause. With the reversal this is now working as intended.

quartzmo: @schmidt-sebastian Thank you for your help on this PR! I'm going to add samples for this feature (as in GoogleCloudPlatform/python-docs-samples#4196), then ask @dazuma to review for Ruby concerns.

quartzmo: Here are the 3 samples PRs I found:

Go PR: GoogleCloudPlatform/golang-samples#1599
Python PR: GoogleCloudPlatform/python-docs-samples#4196
Node PR: googleapis/nodejs-firestore#1113

quartzmo: Added sample in 13c2253

Issue #7579, opened 2020-08-19
Ruby gem no doc tweak


Rajpratik71:
Optimization tweak for ruby gems no doc while gems installation using
 "--no-document" . Also removal of deprecated "--no-ri" and "--no-rdoc".

Further , in case of Docker Containers , by restricting
--no-document , we can reduce image size.
In term of stats , it depends upon the number of packages
multiplied by their respective doc files size .

 More detail can be found at
 https://guides.rubygems.org/command-reference/#installupdate-options-2

PR #7580, opened 2020-08-19
Ruby gem no doc tweak

Rajpratik71:
Optimization tweak for ruby gems no doc while gems installation using
 "--no-document" . Also removal of deprecated "--no-ri" and "--no-rdoc".

Further , in case of Docker Containers , by restricting
--no-document , we can reduce image size.
In term of stats , it depends upon the number of packages
multiplied by their respective doc files size .

 More detail can be found at
 https://guides.rubygems.org/command-reference/#installupdate-options-2

 Signed-off-by: Pratik raj <rajpratik71@gmail.com>

closes: #7579


quartzmo:
@Rajpratik71 Thank you for this PR!

PR #7591, opened 2020-08-24
feat(spanner): add the support of optimizer statistics package

hengfengli:
Add a new field `optimizer_statistics_package` in query options.


thiagotnunes (approved):
LGTM

Issue #7104, opened 2020-07-29
spanner-client: Retry PDML on "Received unexpected EOS on DATA frame from server" 


thiagotnunes:
This bug is related to the Spanner client library.

For long lived transactions (>= 30 minutes), in the case of large PDML changes, it is possible that the gRPC connection is terminated with an error "Received unexpected EOS on DATA frame from server".

In this case, we need to retry the transaction either with the received resume token obtained on reading the stream or from scratch. This will ensure that the PDML transaction continues to execute until it is successful or a hard timeout is reached.

We have already implemented such change in the Java client library, for more information see this PR: https://github.com/googleapis/java-spanner/pull/360.

In order to test the fix, we can use a large spanner database. Please speak to @thiagotnunes for more details.

thiagotnunes:
@jiren hey 👋 , could you help us debug this problem?

Here you can see the example I am trying to run and the error I am seeing: https://gist.github.com/thiagotnunes/1e18081d589ad690ebc962b2ed780244

I have tried tweaking the `pool` `keepalive` and that helps having the pdml running for a longer period. Let me know if you have any questions or if I can help you in any other way.

jiren:
@thiagotnunes, looking into it.

jiren:
@thiagotnunes I have tried 20 times using provided [gist script](https://gist.github.com/thiagotnunes/1e18081d589ad690ebc962b2ed780244)  PDML statement on 5 million records but not getting any error.
Can you share how many records records present the test database?
Also share How many nodes, replicas are you using?

thiagotnunes:
@jiren thanks for trying this out. 
- The error occurs if the PDML runs for over 30 mins.
- The test database has 500M (child) records.
- In the instance provided there is a single node (3 replicas by default I think?).
- The test I was doing, was updating the full database (500M records).

jiren:
@thiagotnunes Thanks for info. One thing more `ChildTable` is interleaved table?

jiren:
@thiagotnunes Please ignore previous message. I have access to `appdev-soda-spanner-staging` project, I will get all of required information and will try out on your test instance.

thiagotnunes:
@jiren no worries, happy to help with anything!

thiagotnunes:
@jiren while investigating this problem in the PHP client library, it seemed like we could not reproduce the problem. Let me know if this occurs here as well.

jiren:
[...]

PR #7592, opened 2020-08-25
fix(spanner): retry or resume eos and rst_stream errors

thiagotnunes:
Retries or resumes the following errors when executing an execute streaming sql request:
- Unavailable error
- Internal EOS error
- Internal RST_STREAM error

This mirrors what is being done in some of the other clients: 
- Python (https://github.com/googleapis/python-spanner/pull/122/files)
- Java (https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/PartitionedDmlTransaction.java#L98-L116)

closes: #7104


thiagotnunes:
There is a build failure now, because the [following test](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/test/google/cloud/spanner/client/read_resume_buffer_bound_test.rb#L211-L213) expects an exception to be thrown (and for the sql query not to be retried). We need to verify if this is behaviour that we would like to keep or change.

thiagotnunes:
Decided only to retry the internal errors (eos, rst_stream) and unavailable errors, but not aborted errors. This is the current behaviour of the Java library, although it retries aborted errors only for PDML. I think this might be a little tricky to implement in the ruby library, since `results` is used by both `execute_streaming_sql` and `pdmls`.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -198,6 +207,57 @@ def rows
...
+        def retryable? err
+          err.instance_of?(GRPC::Unavailable) ||
+            (err.instance_of?(GRPC::Internal) && err.details.include?(EOS_INTERNAL_ERROR)) ||
+            (err.instance_of?(GRPC::Internal) && err.details.include?(RST_STREAM_INTERNAL_ERROR))
+        end
+
+        ##
+        # @private
+        # Resumes a request, by re-executing it with a resume token.
+        def resume_request! resume_token
```
hengfengli, line 233: Why `exclamation mark` after the method? Do we modify any object in-place? Same for `retry_request! ` as well.

google-cloud-spanner/test/google/cloud/spanner/client/read_resume_buffer_bound_test.rb
```diff
@@ -184,7 +184,7 @@
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash1),
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash2),
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash3),
-      GRPC::Unavailable,
+      GRPC::Aborted,
```
hengfengli, line 187: Why do we delete the `GRPC:: Unavailable`? `Aborted` is newly supported. We probably should keep both `Unavailable` and `Aborted`.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
thiagotnunes: Although it does not modify the object in-place it calls either `execute_streaming_sql` or `read` (in a nutshell performs side effects - not pure function). This is the reasoning behind this choice. Let me know if you'd like me to remove it.

hengfengli (approved):
LGTM. Please wait for @jiren's approval.

google-cloud-spanner/test/google/cloud/spanner/client/read_resume_buffer_bound_test.rb
thiagotnunes: This test was failing, because we are now retrying `Unavailable` errors, such as in the Java / Python clients. Since this test was not specific to handling `Unavailable` errors, but re-throwing errors in the case of a failure, we changed the error to `Aborted`. Let me know if this makes sense.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
hengfengli: But that means even `client.execute_query` should also have `!` at the end. To me,  I would think adding `!` means that it modifies an object in-place.

hengfengli: It's a private method. I am okay with any approach, but it would be better to have a consistent convention in client lib.

thiagotnunes: No worries, I will remove the bangs. I agree with you that having a consistent convention is more important.

thiagotnunes: Done.

jiren:
@thiagotnunes PR looks good. LGMT after resolving review comment.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -198,6 +207,57 @@ def rows
 
         # rubocop:enable all
 
+        ##
+        # @private
+        # Checks if a request can be resumed by inspecting the resume token
+        def resumable? resume_token
+          !resume_token.nil? && !resume_token.empty?
```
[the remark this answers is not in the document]
thiagotnunes: Done.

jiren (approved):

thiagotnunes:
@jiren @hengfengli it seems I found a bug in the implementation and I'd like to confirm with you. In the `results.row` we were retrying on `GRPC::BadStatus` errors (such as `Unavailable` or `Internal`). 

We do, however, use the class `Google::Cloud::Spanner::V1::Client` for executing the GRPC [methods](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb). 
All of the methods of this class have a conversion of `GRPC::BadStatus` errors to `Google::Cloud::Error`s (see [execute_streaming_sql](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb#L913) for instance).

That is why I changed the `rescue` clause of the `results` class to catch the wrapped errors instead.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -143,28 +157,25 @@ def rows
...
               end
-            rescue GRPC::Cancelled, GRPC::DeadlineExceeded, GRPC::Internal,
-                   GRPC::ResourceExhausted, GRPC::Unauthenticated,
-                   GRPC::Unavailable, GRPC::Core::CallError => err
-              if resume_token.nil? || resume_token.empty?
-                # Re-raise if the resume_token is not a valid value.
-                # This can happen if the buffer was flushed.
-                raise Google::Cloud::Error.from_error(err)
-              end
+            rescue GRPC::Core::CallError,
```
thiagotnunes, line 160: Note that `GRPC::Core::CallError` is not a `GRPC::BadStatus` error, so it won't be wrapped.

hengfengli:
@thiagotnunes I think this is caused by the recent generator migration. I wonder why no tests catch this problem.

google-cloud-spanner/test/google/cloud/spanner/client/read_resume_buffer_bound_test.rb
```diff
@@ -184,7 +184,7 @@
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash1),
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash2),
       Google::Cloud::Spanner::V1::PartialResultSet.new(results_hash3),
-      GRPC::Unavailable,
+      Google::Cloud::AbortedError,
```
hengfengli, line 187: Why does the old code work then? Where does it convert to `Google::Cloud::UnavailableError`?

hengfengli: The conversion happens here: https://github.com/googleapis/google-cloud-ruby/blob/e4bcb6ec8b8b525363b4add6f45516256aa33489/google-cloud-spanner/lib/google/cloud/spanner/results.rb#L149-L153

hengfengli: In your changes (google-cloud-spanner/lib/google/cloud/spanner/results.rb), Line 176 will do the same conversion.

hengfengli: But you're right. We can assume that all `GRPC::BadStatus` are converted to `Google::Cloud::Error` in the gapic layer.

hengfengli:
This means that we probably need to fix all `rescue` code trying to catch an `GRPC::BadStatus` from gapic layer. Because now we can only get `Google::Cloud::Error`.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -143,28 +157,25 @@ def rows
...
               else
-                @enum = @service.streaming_read_table \
-                  @session_path, @table, @columns,
-                  @read_options.merge(resume_token: resume_token)
+                raise Google::Cloud::Error.from_error(err)
               end
 
-              # Flush the buffered responses to reset to the resume_token
-              buffered_responses = []
             rescue GRPC::BadStatus => err
```
hengfengli, line 179: This code becomes redundant, right?

thiagotnunes: Yes, it would.

thiagotnunes:
@hengfengli I am re-running the PDML long transaction and seeing if it works with the new code. Will keep you posted.

jiren:
> @thiagotnunes I think this is caused by the recent generator migration. I wonder why no tests catch this problem.

Yes, this is a miss. Not much tests are available for retry and resume. So I think we should add it.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
```diff
@@ -143,28 +157,25 @@ def rows
...
               raise Google::Cloud::Error.from_error(err)
```
jiren, line 180: @thiagotnunes  `rescue GRPC::BadStatus`  need to remove. Generator always going to throw `Google::Cloud::Error` so no need to capture again and cast.

jiren (changes requested):

thiagotnunes:
@jiren @hengfengli so, I don't think that the problem is as simple as I originally thought. The fact is that the Generator client (`Google::Cloud::Spanner::V1::Client`) is returning an `Enumerable` for the `streaming` requests (see [here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb#L875)). 

This means that even though we `rescue` on `GRPC::BadStatus` and transform it to `Google::Cloud::Error` when invoking the streaming method in `Google::Cloud::Spanner::V1::Client` (as in [here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb#L913)), we do not do the same when we call `Enumerable.next` (as in [here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/results.rb#L116)).

Because of this, we can get a either a `Google::Cloud:Error` or a `GRPC:Error`. This is observed when I run the PDML and only retry `Google::Cloud::Error`s or only retry `GRPC::Error`s, I still sometimes get an error (either `Google::Cloud::InternalError` or `GRPC::Internal`) which bubbles up.

So we have a couple of choices here:
1. We can retry both errors in the `Results.row` ([here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/results.rb#L146)). This is the simplest, less risky change, but I think this will add tech debt.
2. We can change `Results.from_enum` to also retry on this [method](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/results.rb#L230), which is the first invocation of RPC. This way we would only have to retry `GRPC::Error`s.
3. We can change the `Results.from_enum` ([here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/lib/google/cloud/spanner/results.rb#L229)), `@enum` field to be wrapped in a code block that `rescues` `GRPC::BadStatus` and transforms them to `Google::Cloud::Error`. This way we only need to handle `Google::Cloud::Error`. 
4. I don't know if this is possible, but could we change `Google::Cloud::Spanner::V1::Client` to return a `Enumerable` that always transforms the `GRPC::BadStatus` to `Google::Cloud::Error` ([here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb#L908) and [here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb#L1236))? With this, we would only need to `rescue` `Google::Cloud::Error`s.
5. Any other choices you can think of?

Personally I would vote for **option 4**, since I think that the intention was to encapsulate GRPC specific errors on the generated client abstraction and only expose `Google::Cloud::Error`s to the users. However, I would like to hear your opinion before implementing anything.

hengfengli:
@thiagotnunes Nice find! My suggestion is that: 

* For this change, we deal with both kind of errors. 
* We file an issue with the ruby gapic generator team (@dazuma). https://github.com/googleapis/gapic-generator-ruby

thiagotnunes:
@hengfengli thanks! I agree with your suggestion. I will make the PR changes here to reflect what we discussed and create an issue with the generator team.

google-cloud-spanner/lib/google/cloud/spanner/results.rb
thiagotnunes: We will need to keep this for the time being (refer to the discussion below).

thiagotnunes:
@jiren @hengfengli updated the code to handle both google cloud errors and grpc errors.

hengfengli:
LGTM.

thiagotnunes:
@hengfengli thanks, created the following issue with the ruby gapic generator team: https://github.com/googleapis/gapic-generator-ruby/issues/509

thiagotnunes:
@jiren could you re-review? The change that you requested is no longer valid, since we need to `rescue` both at the moment. Once the gapic generator team has implemented the feature linked above, we will be able to only `rescue` `Google::Cloud::Error`s.

jiren (approved):

Issue #7599, opened 2020-08-27
Creation of authenticated push subscriptions (OIDC token) without having to edit the subscription


sergioisidoro:
**Our Problem**
We are currently using push subscriptions, and we need to programatically create subscriptions to topics (ie, we cannot create the subscriptions beforehand). 

We bumped into problems when we tried to create authenticated subscriptions with OIDC tokens. Because the gem does not allow passing the `push_config` or OIDC tokens in the `topic.subscribe` method, we need to:

1. Create the subscription (with or without the `endpoint`) 
2. Change the configuration of the subscription containing the `PushConfig` withe the OIDC token

**This means that the IAM doing the subscription needs a subscription editor role, in addition to the create subscription role**. This is a bit of nuisance because we **really** want to have the lowest amount of permissions possible.

 **Describe the solution you'd like**
It would be great if we could either pass the `OICD token` parameters to the topic `subscribe` method, or that we would pass the `PushConfig` object directly.

PR #7600, opened 2020-08-27
PubSub: Pass push config in create subscription method

sergioisidoro:

closes: #7599

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -179,8 +179,6 @@ module Google
                           # Handle if push config is provided when creating the subscription
                           if options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
                             options[:push_config].to_grpc
-                          elsif options[:push_config].is_a? Google::Cloud::PubSub::V1::PushConfig
-                            options[:push_config]
                           else
                             raise ArgumentError, "push_config is not a valid pubsub PushConfig object"
                           end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -179,10 +179,10 @@ module Google
                           # Handle if push config is provided when creating the subscription
                           if options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
                             options[:push_config].to_grpc
-                          elsif !options[:push_config].is_a? Google::Cloud::PubSub::V1::PushConfig
-                            raise ArgumentError "push_config is not a valid pubsub PushConfig object"
-                          else
+                          elsif options[:push_config].is_a? Google::Cloud::PubSub::V1::PushConfig
                             options[:push_config]
+                          else
+                            raise ArgumentError, "push_config is not a valid pubsub PushConfig object"
                           end
                         elsif options[:endpoint]
                           Google::Cloud::PubSub::V1::PushConfig.new \
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -177,11 +177,10 @@ module Google
         def create_subscription topic, subscription_name, options = {}
           push_config = if options[:push_config]
                           # Handle if push config is provided when creating the subscription
-                          if options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
-                            options[:push_config].to_grpc
-                          else
+                          unless options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
                             raise ArgumentError, "push_config is not a valid pubsub PushConfig object"
                           end
+                          options[:push_config].to_grpc
                         elsif options[:endpoint]
                           Google::Cloud::PubSub::V1::PushConfig.new \
                             push_endpoint: options[:endpoint],
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
@@ -176,11 +176,11 @@ module Google
         # Creates a subscription on a given topic for a given subscriber.
         def create_subscription topic, subscription_name, options = {}
           push_config = if options[:push_config]
-                          # Handle if push config is provided at create subscription time
+                          # Handle if push config is provided when creating the subscription
                           if options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
                             options[:push_config].to_grpc
                           elsif !options[:push_config].is_a? Google::Cloud::PubSub::V1::PushConfig
-                            raise ArgumentError "push_config is not a valid pubsub push config object"
+                            raise ArgumentError "push_config is not a valid pubsub PushConfig object"
                           else
                             options[:push_config]
                           end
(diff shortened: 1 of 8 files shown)


sergioisidoro:
@googlebot I fixed it.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
```diff
@@ -175,11 +175,21 @@ def list_subscriptions options = {}
...
         def create_subscription topic, subscription_name, options = {}
-          push_config = if options[:endpoint]
+          push_config = if options[:push_config]
+                          # Handle if push config is provided when creating the subscription
+                          if options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
+                            options[:push_config].to_grpc
+                          elsif !options[:push_config].is_a? Google::Cloud::PubSub::V1::PushConfig
+                            raise ArgumentError "push_config is not a valid pubsub PushConfig object"
+                          else
+                            options[:push_config]
```
quartzmo, line 185: Would this be anything other than a `Hash`? Might be safer to explicitly test for that, and otherwise raise `ArgumentError` in the final `else` block.

google-cloud-pubsub/test/google/cloud/pubsub/topic_test.rb
```diff
@@ -154,6 +154,34 @@
     _(sub).must_be_kind_of Google::Cloud::PubSub::Subscription
   end
 
+  it "creates a subscription with an authenticated push endpoint via push config" do
+    new_sub_name = "new-sub-#{Time.now.to_i}"
+    endpoint = "http://foo.bar/baz"
+
+    push_config = Google::Cloud::PubSub::Subscription::PushConfig.new
```
quartzmo, line 161: Just noting that this currently `@private` constructor would need to be publicly documented.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
sergioisidoro: Although in the end this becomes a hash, I thought it would be better to pass the objects instead, and reject everything else (including the hash).

This if clause is a bit confusing, because the raise is in the condition above.
I'll restructure the if condition into a more orthodox one raising only on the `else` block.

sergioisidoro:
@quartzmo Sorry, I'm unable to run the acceptance tests. 

It seems the tests need a wide range of permissions, and I would need to setup a new Gcloud project with an account with massive permission rights just to test this.
eg. Why do pubsub tests require `storage.buckets.create` ?
```
Google::Apis::ClientError: forbidden: sergio-acceptance-tests@xxxxx.iam.gserviceaccount.com does not have storage.buckets.create access to the Google Cloud project.
```

sergioisidoro:
- Added documentations and examples
- Added an acceptance test (and ran it, everything seemed to went fine)
- Changed the if/else structure to be clear and added tests - (https://github.com/googleapis/google-cloud-ruby/pull/7600/files/1d9dbd206af531334df320feda8a03526147e3a4#diff-d0b5c7f9eb58f0018f1b656372c7a119)

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
sergioisidoro: Thanks for pointing this out. There was an actual bug on the raise branch of this `if` that was discovered when the tests were added :)

google-cloud-pubsub/lib/google/cloud/pubsub/subscription/push_config.rb
```diff
@@ -43,9 +43,24 @@ class Subscription
...
+        #   )
+        #   sub = topic.subscribe "my-subscription", push_config: push_config
+        #
         class PushConfig
           ##
-          # @private
+          # Creates an empty configuration for a push subscription.
+          #
+          # @return [PushConfig]
           def initialize
```
quartzmo, line 64: Feel free to add params accepting the class attributes to `initialize` (keyword args for optional attributes) if you want, for convenience.

sergioisidoro:
* Added param documentations
* Added constructor arguments to push config, a basic test, and some documentation.
* Removed the mentioned non exposed API object.

Let me know what you think @quartzmo 

PS: Some linter rules are quite strict. I spent way too much time shuffling if/else clauses to match the linter suggestions. Maybe relaxing some of these would be nice.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription/push_config.rb
```diff
@@ -43,11 +43,33 @@ class Subscription
...
+          # @param [String] auth_email The GCP service account email associated with the push subscription.
+          #  Push requests carry the identity of this service account.
+          # @param [String] auth_audience A single, case-insensitive string that can be used by
+          #  the webhook to validate the intended audience of this particular token.
+          # @return [PushConfig]
+          def initialize endpoint: nil, auth_email: nil, auth_audience: nil
             @grpc = Google::Cloud::PubSub::V1::PushConfig.new
+
+            self.endpoint = endpoint unless endpoint.nil?
+            set_oidc_token auth_email, auth_audience if auth_email && auth_audience
```
sergioisidoro, line 72: Maybe we should throw an error if one, but not the other argument is provided. 
This silent failure might cause some head scratching.

quartzmo (commented):
@sergioisidoro This is looking good to me. I added just two very minor comments. Thanks for your patience and hard work on this change, I think it is a much needed addition.

google-cloud-pubsub/lib/google/cloud/pubsub/service.rb
```diff
@@ -175,11 +175,18 @@ def list_subscriptions options = {}
         ##
         # Creates a subscription on a given topic for a given subscriber.
         def create_subscription topic, subscription_name, options = {}
-          push_config = if options[:endpoint]
+          push_config = if options[:push_config]
+                          # Handle if push config is provided when creating the subscription
+                          unless options[:push_config].is_a? Google::Cloud::PubSub::Subscription::PushConfig
+                            raise ArgumentError, "push_config is not a valid pubsub PushConfig object"
+                          end
+                          options[:push_config].to_grpc
```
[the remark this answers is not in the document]
sergioisidoro: Good question. I assumed that the service is a semi-private interface, since it's not well documented, and uses non structured, non documented arguments (ie. keyword arguments in `options`). So in that case throwing an error here feels a bit unorthodox.

Also `options[:attributes]` is not passed forward from `topic.subscribe`.

IMHO, PushConfig construction should be handled upstream (eg. in Topic). Maybe in here `push_config` could even be taken out of the `options` and transformed into an argument on its own - it's quite strange that we're providing constructor parameters of `PushConfig` in the service `options`.

But that's a breaking change. 

PR is already getting quite large. I wonder if we should try to merge something here, and try to address this somewhere else. My suggestion would be that in another PR we would:

- Add a deprecation warning to the options[:endpoint] and options[:attributes] in favour of `PushConfig` object construction in the `Service` class.
- Move the construction of the `PushConfig` upstream to `Topic`, and use `to_grpc` in the `Service` class
- Maybe add `push_auth_email`, `push_auth_audience`, and `push_attributes` to the arguments of `topic.subscribe`

At least we could add an Issue to track this situation.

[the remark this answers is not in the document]
quartzmo: > Also options[:attributes] is not passed forward from topic.subscribe.

Thanks for pointing this out. In that case, it's not a problem.

> I assumed that the service is a semi-private interface, since it's not well documented, and uses non structured, non documented arguments (ie. keyword arguments in options)

Yes, `Service` is private. I think in a separate PR I will convert `options` to separate keyword args, for more clarity.

> This PR is already getting quite large. I wonder if we should try to merge something here, and try to address this somewhere else. 

Agreed! This is great, let's merge it!

quartzmo:
Thanks again @sergioisidoro!

quartzmo:
I'll post here once this change has been released, most likely Monday or Tuesday.

Issue #7856, opened 2020-10-02
Storage: x-goog-resumable is not being signed in V4 Signature Tests 


frankyn:
The following test is not signing `x-goog-resumable` header which is being fixed in production to prevent this from succeeding in the future. 

https://github.com/googleapis/google-cloud-ruby/blob/28f5618e5560536832ffde2a009b61112314ed57/google-cloud-storage/acceptance/storage/signed_url_v4_test.rb#L179-L197

PR #7857, opened 2020-10-02
fix: add missing header to v4 test

frankyn:
Fixes: #7856


quartzmo:
I pulled this branch and confirmed that this update passes with our CI project. (It may not work with all projects at this time.)

quartzmo (approved):

Issue #8057, opened 2020-10-26
spanner: Add support of CommitStats in CommitResponse


jiren:
API proto file [reference](https://github.com/googleapis/googleapis/blob/2c99b1e92bd25a3d4d80bedf397b45339b0e46ea/google/spanner/v1/spanner.proto#L868).

```ruby
 require "google/cloud/spanner"

 spanner = Google::Cloud::Spanner.new
 db = spanner.client "my-instance", "my-database"

 commit_resp = db.commit commit_stats: true do |c|
   c.update "users", [{ id: 1, name: "Charlie", active: false }]
   c.insert "users", [{ id: 2, name: "Harvey",  active: true }]
 end

 puts commit_resp.timestamp
 puts commit_resp.stats.mutation_count
 puts commit_resp.stats.overload_delay
```

PR #8058, opened 2020-10-26
feat(spanner): CommitStats in CommitResponse

jiren:

closes: #8057


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -809,7 +810,12 @@ def read table, columns, keys: nil, index: nil, limit: nil,
         #   See [Data
         #   types](https://cloud.google.com/spanner/docs/data-definition-language#data_types).
         #
-        # @return [Time] The timestamp at which the operation committed.
+        # @param [Boolean] commit_stats Include statistics related to the
+        # transaction. Default it is `false`.
+        #
+        # @return [CommitResponse] It include the timestamp at which the
+        #   operation committed. It also include transaction commit stats,
+        #   if passed `commit_stats` value is `true`.
```
skuruppu, line 818: I'm curious, wouldn't this be a breaking change for users? If they are currently calling `upsert`, they are expecting it to return a timestamp. But now we return a `CommitResponse` so wouldn't that throw an error?

E.g. the following code will break:
```
ts = db.upsert "users", [{ id:1, name: "Charlie", active: false }, ...]
# use ts for something else
...
```

google-cloud-spanner/lib/google/cloud/spanner/commit_response/commit_stats.rb
```diff
@@ -0,0 +1,57 @@
...
+    module Spanner
+      class CommitResponse
+        ##
+        # # CommitStats
+        #
+        # Statistical information of a transaction commit.
+        #
+        class CommitStats
+          ##
+          # @private Creates a new CommitS  tats instance.
```
skuruppu, line 29:
```suggestion
          # @private Creates a new CommitStats instance.
```

google-cloud-spanner/lib/google/cloud/spanner/commit_response/commit_stats.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          # Length of time the commit was delayed due to overloaded servers.
+          # @return [Integer]
+          def overload_delay
+            Convert.duration_to_number @grpc.overload_delay
+          end
+
+          ##
+          # @private
+          # Creates a new Commit stats instance from a
```
skuruppu, line 48:
```suggestion
          # Creates a new CommitStats instance from a
```

google-cloud-spanner/lib/google/cloud/spanner/session.rb
```diff
@@ -501,7 +504,9 @@ def partition_read table, columns, transaction, keys: nil,
         # @yield [commit] The block for mutating the data.
         # @yieldparam [Google::Cloud::Spanner::Commit] commit The Commit object.
         #
-        # @return [Time] The timestamp at which the operation committed.
+        # @return [CommitResponse] It include the timestamp at which the
```
skuruppu, line 507: Same here regarding this being a breaking change.

skuruppu (commented):
@jiren I have a concern that this implementation might be causing a breaking change for users who were previously reading the commit timestamp that's returned by most of the methods that are modified. We avoided doing this in most of the other language implementations. Let me know if I misunderstood. But if we could come up with a way to make this a non-breaking change, that would be ideal.

jiren:
> @jiren I have a concern that this implementation might be causing a breaking change for users who were previously reading the commit timestamp that's returned by most of the methods that are modified. We avoided doing this in most of the other language implementations. Let me know if I misunderstood. But if we could come up with a way to make this a non-breaking change, that would be ideal.

Yes, In this case I am planning to extend(delegate in ruby [ref](https://github.com/googleapis/google-cloud-ruby/blob/8fd12a741a3c99a2b42e19d966fbeb2587c35c9c/google-cloud-spanner/lib/google/cloud/spanner/instance/list.rb#L25)) wrapper CommitResponse from Timestamp then it will works. So commit response works as a timestamp with stats attributes.

```ruby
class CommitResponse < DelegateClass(::Time); end
```

skuruppu:
> > @jiren I have a concern that this implementation might be causing a breaking change for users who were previously reading the commit timestamp that's returned by most of the methods that are modified. We avoided doing this in most of the other language implementations. Let me know if I misunderstood. But if we could come up with a way to make this a non-breaking change, that would be ideal.
> 
> Yes, In this case I am planning to extend(delegate in ruby [ref](https://github.com/googleapis/google-cloud-ruby/blob/8fd12a741a3c99a2b42e19d966fbeb2587c35c9c/google-cloud-spanner/lib/google/cloud/spanner/instance/list.rb#L25)) wrapper CommitResponse from Timestamp then it will works. So commit response works as a timestamp with stats attributes.
> 
> ```ruby
> class CommitResponse < DelegateClass(Time); end
> ```

Oh cool, so does that means that users who are already using these APIs will not have to change their code?

hengfengli:
@jiren I am not keen to use `DelegateClass` to let `CommitResponse` inherit from `Time`. This confuses me because `timestamp` should be a field of `CommitResponse`. The relationship here does not look good to me, although it may solve the backward-compatible issue.

skuruppu:
Just to clarify the requirements @jiren, this feature is only meant to be used for debugging queries. So we only expect people to use the flag in development. So we don't want users to have to rewrite their production applications.

In other languages, we added extra methods like `insertWithOptions()` that returns a `CommitResponse`. We might have to do the same here.

google-cloud-spanner/test/google/cloud/spanner/client/transaction_test.rb
```diff
@@ -422,16 +432,78 @@
...
+      tx.update "users", [{ id: 1, name: "Charlie", active: false }]
+    end
+    assert_commit_resp resp, stats: true
+
+    shutdown_client! client
+
+    mock.verify
+  end
+
+  it "execute transaction with call options and return commit stats" do
```
hengfengli, line 481: It's unclear why we add this test. We only need to test the commit stats feature.

google-cloud-spanner/test/google/cloud/spanner/client/commit_test.rb
```diff
@@ -94,11 +126,42 @@
...
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
     spanner.service.mocked_service = mock
 
-    timestamp = client.update "users", [{ id: 1, name: "Charlie", active: false }]
-    _(timestamp).must_equal commit_time
+    resp = client.update "users", [{ id: 1, name: "Charlie", active: false }], commit_stats: true
+    assert_commit_resp resp, stats: true
+
+    resp = client.update "users", { id: 1, name: "Charlie", active: false }, commit_stats: true
```
hengfengli, line 163: Duplicated code?

jiren:
> @jiren I am not keen to use `DelegateClass` to let `CommitResponse` inherit from `Time`. This confuses me because `timestamp` should be a field of `CommitResponse`. The relationship here does not look good to me, although it may solve the backward-compatible issue.
> 
> Instead of changing the return type of existing methods, can we add new methods to return `CommitResponse`? This is what we have done in other languages.

@hengfengli One thing we can do in case we go with the inheritance - add more detail in the documentation 
For the second option adding additional methods, I think it will add a lot more redundant code.
I have one more option return `Time` or `CommitResponse` based on `commit_stats` method argument.

i.e
```
def insert table, rows, commit_stats: nil
   ...
   ...
   if commit_stats 
     return commit_response
   else
     return commit_response.timestamp.
   end
end
```

google-cloud-spanner/test/google/cloud/spanner/client/commit_test.rb
```diff
@@ -94,11 +126,42 @@
...
+          table: "users", columns: %w(id name active),
+          values: [Google::Cloud::Spanner::Convert.object_to_grpc_value([1, "Charlie", false]).list_value]
+        )
+      )
+    ]
+
+    mock = Minitest::Mock.new
+    mock.expect :create_session, session_grpc, [{database: database_path(instance_id, database_id), session: nil}, default_options]
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
```
hengfengli, line 157: Duplicated code?

google-cloud-spanner/test/google/cloud/spanner/client/commit_test.rb
```diff
@@ -94,11 +126,42 @@
 
     mock = Minitest::Mock.new
     mock.expect :create_session, session_grpc, [{database: database_path(instance_id, database_id), session: nil}, default_options]
-    mock.expect :commit, commit_resp, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts }, default_options]
+    mock.expect :commit, commit_resp, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: nil }, default_options]
```
hengfengli, line 129: Why do we test `commit` twice here?

google-cloud-spanner/test/google/cloud/spanner/client/commit_test.rb
```diff
@@ -186,11 +311,42 @@
...
+        )
+      )
+    ]
+
+    mock = Minitest::Mock.new
+    mock.expect :create_session, session_grpc, [{database: database_path(instance_id, database_id), session: nil}, default_options]
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
+    mock.expect :commit, commit_resp_with_stats, [{ session: session_grpc.name, mutations: mutations, transaction_id: nil, single_use_transaction: tx_opts, return_commit_stats: true }, default_options]
+    spanner.service.mocked_service = mock
+
```
hengfengli, line 344: Same here. Why do we test `commit` twice?

google-cloud-spanner/test/google/cloud/spanner/session/commit_test.rb
```diff
@@ -62,20 +71,40 @@
...
       c.replace "users", [{ id: 4, name: "Henry",  active: true }]
       c.delete "users", [1, 2, 3, 4, 5]
     end
-    _(timestamp).must_equal commit_time
+    assert_commit_resp resp
+
+    mock.verify
+  end
+
+  it "commits using a block and return commit stats" do
```
hengfengli, line 94: I feel that this adds a lot of redundant test code. Can we try to avoid these changes?

hengfengli (changes requested):
Thanks for making the changes. There are two major issues: 

1. A breaking change. We need to avoid this. One idea I have is that if the user specifies an `commit_options` then we return `commit_response` instead of `timestamp`. If `commit_options` is not given, then we return `timestamp`. Otherwise, we need to add new methods to just return `commit_response` (this would add a lot of new methods). 
2. Too many redundant test code. They are most duplicated code. Please try to simplify the new added test code. Only test what we just add and avoid too many changes, which is hard for us to review the code.

google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb
```diff
@@ -1332,7 +1332,7 @@ def begin_transaction request, options = nil
             #   @param options [::Gapic::CallOptions, ::Hash]
             #     Overrides the default settings for this call, e.g, timeout, retries, etc. Optional.
             #
-            # @overload commit(session: nil, transaction_id: nil, single_use_transaction: nil, mutations: nil)
+            # @overload commit(session: nil, transaction_id: nil, single_use_transaction: nil, mutations: nil, return_commit_stats: nil)
```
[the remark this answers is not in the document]
jiren: Ok, but unit test will fail on CI.

jiren:
@hengfengli I am adding option for `commit_options`  and based on that I going to return timestamp or stats.

google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb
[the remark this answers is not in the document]
hengfengli: Usually, we run unit tests locally before committing to a PR.

google-cloud-spanner/acceptance/spanner_helper.rb
```diff
@@ -36,7 +36,7 @@ def emulator_enabled?
 end
 
 # Create shared spanner object so we don't create new for each test
-$spanner = Google::Cloud::Spanner.new
+$spanner = Google::Cloud::Spanner.new endpoint: ENV['SPANNER_SERVICE_ENDPOINT']
```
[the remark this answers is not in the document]
jiren: Ok, removing it.

hengfengli:
> @hengfengli I am adding option for `commit_options` and based on that I going to return timestamp or stats.

This sounds good to me.

google-cloud-spanner/test/google/cloud/spanner/session/commit_test.rb
jiren: Yes working on redundant test case code.

google-cloud-spanner-v1/lib/google/cloud/spanner/v1/spanner/client.rb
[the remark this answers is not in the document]
jiren: Ok, I am removing it these auto generated files. While running on local you need to generate Gapic client to pickup changes from googleapis [branch](https://github.com/googleapis/googleapis/tree/spanner_commit_stats). 
FYI Tests on Kokoro CI will fail.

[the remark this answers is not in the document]
hengfengli: Yea, that's fine. Once the generated code is merged into master, we rebase this branch and it will pass.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: Updated for backward compatibility.

google-cloud-spanner/lib/google/cloud/spanner/commit_response/commit_stats.rb
jiren: Done

google-cloud-spanner/acceptance/spanner/client/batch_update_test.rb
```diff
@@ -40,7 +40,7 @@
     prior_results = db.execute_sql "SELECT * FROM accounts"
     _(prior_results.rows.count).must_equal 3
 
-    timestamp = db.transaction do |tx|
+    commit_resp = db.transaction do |tx|
```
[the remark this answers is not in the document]
jiren: Updated for backward compatibility.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -197,6 +197,20 @@
     end
   end
 
+  it "execute transaction and return commit stats" do
+    skip if emulator_enabled?
+
+    commit_resp =  db.transaction commit_stats: true do |tx|
+      tx_results = tx.read "accounts", [:reputation], keys: 1, limit: 1
+      tx_val = tx_results.rows.first[:reputation]
+      sleep 1 # ensure that both threads would have read same value if not read locked
```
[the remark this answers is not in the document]
jiren: added separate test for simple transaction and removed above test.

google-cloud-spanner/test/google/cloud/spanner/client/commit_test.rb
jiren: removed. Added to test rows and single row.

jiren:
@hengfengli @skuruppu Updated for backward compatibility. 
- Added `commit_options : { return commit_stats: } `
- Condition based response - If user provides commit options then return `CommitResponse` object else `Time` obj.

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
```diff
@@ -197,6 +197,27 @@
     end
   end
 
+  it "can execute sql with query options" do
+    query_options = { optimizer_version: "latest" }
+    db.transaction do |tx|
+      tx_results = tx.execute_sql query_reputation, query_options: query_options
+      _(tx_results.rows.first[:reputation]).must_equal 63.5
+    end
+  end
```
hengfengli, line 206: Do we need to add query options test?

google-cloud-spanner/lib/google/cloud/spanner/commit_response/commit_stats.rb
hengfengli: @jiren It seems that this is still not fixed.

hengfengli:
@skuruppu It now looks good to me. Can you please take a quick look?

google-cloud-spanner/acceptance/spanner/client/transaction_test.rb
jiren: removed.

skuruppu (approved):
I think this is reasonable if it's normal to alter the return value in ruby based on inputs. Thanks for making this a non-breaking feature.

skuruppu:
@jiren I have a small request to make. Would you be able to remove any hand-written support for `overload_delay`? This proto field is no longer getting released so when the protos are published, it will no longer be there. I hope it's not too much trouble.

jiren:
> @jiren I have a small request to make. Would you be able to remove any hand-written support for `overload_delay`? This proto field is no longer getting released so when the protos are published, it will no longer be there. I hope it's not too much trouble.

Working on it.

jiren:
> @jiren I have a small request to make. Would you be able to remove any hand-written support for `overload_delay`? This proto field is no longer getting released so when the protos are published, it will no longer be there. I hope it's not too much trouble.

@skuruppu Removed overload_delay code.

hengfengli (approved):

skuruppu:
@jiren the generated code has been merged so please rebase this PR and get it ready for merging.

jiren:
> @jiren the generated code has been merged so please rebase this PR and get it ready for merging.

Done

skuruppu:
@dazuma or @quartzmo, we would really like to get an approval from one of you so that we can merge in this PR.

Issue #8076, opened 2020-10-29
BigQuery: Implement Authorized UDF support for dataset Access


shollyman:
## Overview

Add support for authorized UDFs.

The Access property of a dataset now allows references to **Routine** entities, similar to the existing View support.  Only UDF routines are currently supported.  This feature is enabled and can be implemented immediately.

## References

Example impl: https://github.com/googleapis/google-cloud-go/pull/3084

What we document in the discovery layer for dataset's `access` field:
```
"access": {
          "type": "array",
          "description": "[Optional] An array of objects that define dataset access for one or more entities. You can set this property when inserting or updating a dataset in order to control who is allowed to access the data. If unspecified at dataset creation time, BigQuery adds default dataset access for the following entities: access.specialGroup: projectReaders; access.role: READER; access.specialGroup: projectWriters; access.role: WRITER; access.specialGroup: projectOwners; access.role: OWNER; access.userByEmail: [dataset creator email]; access.role: OWNER;",
          "items": {
            "properties": {
              "view": {
                "$ref": "TableReference",
                "description": "[Pick one] A view from a different dataset to grant access to. Queries executed against that view will have read access to tables in this dataset. The role field is not required when this field is set. If that view is updated by any user, access to the view needs to be granted again via an update operation."
              },
              "groupByEmail": {
                "description": "[Pick one] An email address of a Google Group to grant access to. Maps to IAM policy member \"group:GROUP\".",
                "type": "string"
              },
              "userByEmail": {
                "description": "[Pick one] An email address of a user to grant access to. For example: fred@example.com. Maps to IAM policy member \"user:EMAIL\" or \"serviceAccount:EMAIL\".",
                "type": "string"
              },
              "domain": {
                "description": "[Pick one] A domain to grant access to. Any users signed in with the domain specified will be granted the specified access. Example: \"example.com\". Maps to IAM policy member \"domain:DOMAIN\".",
                "type": "string"
              },
              "iamMember": {
                "type": "string",
                "description": "[Pick one] Some other type of member that appears in the IAM Policy but isn't a user, group, domain, or special group."
              },
              "role": {
[...]

PR #8089, opened 2020-11-03
feat(bigquery): Add routine (UDF) to Dataset::Access

quartzmo:
closes: #8076


shollyman (approved):
Thanks for putting this together!

dazuma (approved):

Issue #8162, opened 2020-11-24
spanner: Add support of request tagging.


jiren:
Adds support for request and transaction tags for read, queries and transactions.

PR #8163, opened 2020-11-24
feat(spanner): support request tagging

jiren:
Adds support for request options with request and transaction tags.

closes: #8162

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/client.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/client.rb
@@ -236,6 +236,10 @@ module Google
         #     Valid values are `:PRIORITY_LOW`, `:PRIORITY_MEDIUM`,
         #     `:PRIORITY_HIGH`. If priority not set then default is
         #     `PRIORITY_UNSPECIFIED` is equivalent to `:PRIORITY_HIGH`.
+        #   * `:tag` (String) A per-request tag which can be applied to
+        #     queries or reads, used for statistics collection. Tag must be a
+        #     valid identifier of the form: `[a-zA-Z][a-zA-Z0-9_\-]` between 2
+        #     and 64 characters in length.
         # @param [Hash] call_options A hash of values to specify the custom
         #   call options, e.g., timeout, retries, etc. Call options are
         #   optional. The following settings can be provided:
@@ -403,6 +407,21 @@ module Google
         #     puts "User #{row[:id]} is #{row[:name]}"
         #   end
         #
+        # @example Query using tag for request query statistics collection.
+        #   require "google/cloud/spanner"
+        #
+        #   spanner = Google::Cloud::Spanner.new
+        #
+        #   db = spanner.client "my-instance", "my-database"
+        #
+        #   request_options = { tag: "Read-Users" }
+        #   results = db.execute_query "SELECT * FROM users",
+        #                              request_options: request_options
+        #
+        #   results.rows.each do |row|
+        #     puts "User #{row[:id]} is #{row[:name]}"
+        #   end
+        #
         def execute_query sql, params: nil, types: nil, single_use: nil,
                           query_options: nil, request_options: nil,
                           call_options: nil
@@ -410,7 +429,8 @@ module Google
           ensure_service!
 
           params, types = Convert.to_input_params_and_types params, types
-
+          request_options = Convert.to_request_options request_options,
+                                                       tag_type: :request_tag
           single_use_tx = single_use_transaction single_use
           results = nil
           @pool.with_session do |session|
@@ -580,6 +600,10 @@ module Google
         #     Valid values are `:PRIORITY_LOW`, `:PRIORITY_MEDIUM`,
         #     `:PRIORITY_HIGH`. If priority not set then default is
         #     `PRIORITY_UNSPECIFIED` is equivalent to `:PRIORITY_HIGH`.
+        #   * `:tag` (String) A per-request tag which can be applied to
+        #     queries or reads, used for statistics collection. Tag must be a
+        #     valid identifier of the form: `[a-zA-Z][a-zA-Z0-9_\-]` between 2
+        #     and 64 characters in length.
         # @param [Hash] call_options A hash of values to specify the custom
         #   call options, e.g., timeout, retries, etc. Call options are
         #   optional. The following settings can be provided:
@@ -657,12 +681,27 @@ module Google
         #    "UPDATE users SET friends = NULL WHERE active = @active",
         #    params: { active: false }, request_options: request_options
         #
+        # @example Query using tag for request query statistics collection.
+        #
+        #   require "google/cloud/spanner"
+        #
+        #   spanner = Google::Cloud::Spanner.new
+        #   db = spanner.client "my-instance", "my-database"
+        #
+        #   request_options = { tag: "Update-Users" }
+        #   row_count = db.execute_partition_update \
+        #     "UPDATE users SET friends = NULL WHERE active = false",
+        #     request_options: request_options
+        #
         def execute_partition_update sql, params: nil, types: nil,
                                      query_options: nil, request_options: nil,
                                      call_options: nil
           ensure_service!
 
           params, types = Convert.to_input_params_and_types params, types
+          request_options = Convert.to_request_options request_options,
+                                                       tag_type: :request_tag
+
           results = nil
           @pool.with_session do |session|
             results = session.execute_query \
@@ -761,6 +800,10 @@ module Google
         #     Valid values are `:PRIORITY_LOW`, `:PRIORITY_MEDIUM`,
         #     `:PRIORITY_HIGH`. If priority not set then default is
         #     `PRIORITY_UNSPECIFIED` is equivalent to `:PRIORITY_HIGH`.
+        #   * `:tag` (String) A per-request tag which can be applied to
+        #     queries or reads, used for statistics collection. Tag must be a
+        #     valid identifier of the form: `[a-zA-Z][a-zA-Z0-9_\-]` between 2
+        #     and 64 characters in length.
         # @param [Hash] call_options A hash of values to specify the custom
         #   call options, e.g., timeout, retries, etc. Call options are
         #   optional. The following settings can be provided:
@@ -840,6 +883,22 @@ module Google
         #     puts "User #{row[:id]} is #{row[:name]}"
         #   end
         #
+        # @example Read using tag for read statistics collection.
+        #
+        #   require "google/cloud/spanner"
+        #
+        #   spanner = Google::Cloud::Spanner.new
+        #
+        #   db = spanner.client "my-instance", "my-database"
+        #
+        #   request_options = { tag: "Read-Users-All" }
+        #   results = db.read "users", [:id, :name],
+        #                     request_options: request_options
+        #
+        #   results.rows.each do |row|
+        #     puts "User #{row[:id]} is #{row[:name]}"
+        #   end
+        #
         def read table, columns, keys: nil, index: nil, limit: nil,
                  single_use: nil, request_options: nil, call_options: nil
           validate_single_use_args! single_use
@@ -847,8 +906,10 @@ module Google
 
           columns = Array(columns).map(&:to_s)
           keys = Convert.to_key_set keys
-
           single_use_tx = single_use_transaction single_use
+          request_options = Convert.to_request_options request_options,
+                                                       tag_type: :request_tag
+
           results = nil
           @pool.with_session do |session|
             results = session.read \
@@ -914,6 +975,9 @@ module Google
         #     Valid values are `:PRIORITY_LOW`, `:PRIORITY_MEDIUM`,
         #     `:PRIORITY_HIGH`. If priority not set then default is
         #     `PRIORITY_UNSPECIFIED` is equivalent to `:PRIORITY_HIGH`.
+        #   * `:tag` (String) A tag used for statistics collection
+        #     about transaction. A tag must be a valid identifier of the format:
+        #     `[a-zA-Z][a-zA-Z0-9_\-]{0,49}`.
         #
         # @return [Time, CommitResponse] The timestamp at which the operation
         #   committed. If commit options are set it returns {CommitResponse}.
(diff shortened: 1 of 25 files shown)


google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -371,19 +375,35 @@ def query_options
...
-                          query_options: nil, call_options: nil
+                          query_options: nil, tag: nil,
+                          call_options: nil
           validate_single_use_args! single_use
           ensure_service!
 
           params, types = Convert.to_input_params_and_types params, types
-
           single_use_tx = single_use_transaction single_use
+          request_options = build_request_options request_tag: tag
```
skuruppu, line 399: What if this `execute_query` is part of a transaction? How does the transaction tag get passed in?

google-cloud-spanner/test/google/cloud/spanner/client/commit_transaction_tag_test.rb
```diff
@@ -0,0 +1,163 @@
+# Copyright 2017 Google LLC
```
skuruppu, line 1:
```suggestion
# Copyright 2020 Google LLC
```

skuruppu (commented):
Just a couple of minor comments.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: Client can execute only single use transaction, so here only request tag will work. Transaction tag on `execute_query` is  works using transaction object. 

```ruby
db.transaction tag: "Tag-1" do |tx|
  results = tx.execute_query "SELECT * FROM users", tag: "Tag-1-1"
end
```

Implementation ref: https://github.com/googleapis/google-cloud-ruby/blob/880f64fe6db2d48ef989fad356a21d23550cbd88/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L1352-L1373

Here on line 1364 `truncation_tag` will set and pass to transaction object to add queries executed by transaction.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -808,6 +865,9 @@ def read table, columns, keys: nil, index: nil, limit: nil,
         #
         #   See [Data
         #   types](https://cloud.google.com/spanner/docs/data-definition-language#data_types).
+        # @param [String] tag A tag used for statistics collection about
+        #   transaction. `transaction_tag` must be a valid identifier of
```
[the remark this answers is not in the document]
jiren: Done

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -1119,6 +1269,10 @@ def commit call_options: nil, &block
         #
         # @param [Numeric] deadline The total amount of time in seconds the
         #   transaction has to succeed. The default is `120`.
+        # @param [String] tag A tag used for statistics collection about
+        #   transaction. The value of transaction_tag should be the same
```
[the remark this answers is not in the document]
jiren: Done

google-cloud-spanner/lib/google/cloud/spanner/transaction.rb
```diff
@@ -1042,6 +1066,15 @@ def tx_selector
...
 
+        def build_request_options request_tag: nil
+          return unless request_tag || transaction_tag
+
+          options = {}
+          options[:transaction_tag] = transaction_tag if transaction_tag
+          options[:request_tag] = request_tag if request_tag
+          options
+        end
+
```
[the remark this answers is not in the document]
jiren: Yes, moved to `Convert` helper file.

google-cloud-spanner/test/google/cloud/spanner/client/commit_transaction_tag_test.rb
jiren: Done

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -1061,6 +1194,9 @@ def delete table, keys = [], call_options: nil
         # this method may be appropriate for latency sensitive and/or high
         # throughput blind changes.
         #
+        # @param [String] tag A tag used for statistics collection about
+        #   transaction. `transaction_tag` must be a valid identifier of
```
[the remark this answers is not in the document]
jiren: Done

google-cloud-spanner/lib/google/cloud/spanner/client.rb
thiagotnunes: Sorry, just to verify I understood it correctly. On the example you provided, the transaction tag will be `Tag-1` and the request tag will be `Tag-1-1`, right?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -371,19 +375,35 @@ def query_options
...
+        #
+        #   results = db.execute_query "SELECT * FROM users", tag: "Tag-1"
+        #
+        #   results.rows.each do |row|
+        #     puts "User #{row[:id]} is #{row[:name]}"
+        #   end
+        #
         def execute_query sql, params: nil, types: nil, single_use: nil,
-                          query_options: nil, call_options: nil
+                          query_options: nil, tag: nil,
```
thiagotnunes, line 392: I wonder if the tag should not be in the `query_options` or `call_options` hashmap, wdyt? Or is this a private method used internally only?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -1092,16 +1228,30 @@ def delete table, keys = [], call_options: nil
...
+        #
+        #   spanner = Google::Cloud::Spanner.new
+        #
+        #   db = spanner.client "my-instance", "my-database"
+        #
+        #   db.commit tag: "Tag-1" do |c|
+        #     c.update "users", [{ id: 1, name: "Charlie", active: false }]
+        #     c.insert "users", [{ id: 2, name: "Harvey",  active: true }]
+        #   end
+        def commit tag: nil, call_options: nil, &block
```
thiagotnunes, line 1242: The commit (transaction) `tag` should be defined in the transaction creation, like in your previous example:

```
db.transaction tag: "Tag-1" do |tx| ...
```

The above transaction tag should be included in the commit statement.

If this is for hand-crafted transactions, I still think we should define the transaction tag when the user creates the transaction initially.

thiagotnunes: Or is this method used internally only?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: Yes "tag-1" is transaction tag and "Tag-1-1" is request tag.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: No, this is not a private method. In my view, we should not use `query_options` or `call_options`.`query_options` for optimization params and `call_options` is not for apis it is for GRPC client. We can use `tag` as `request_options` as per [proto file](https://github.com/googleapis/googleapis/blob/0dd48adfa0aa8f4d042813af248dcbeecc5f7102/google/spanner/v1/spanner.proto#L566) but client design specs "tag" is used. i.e ```roTx.read(table, keySet, columns, Tag.of("tag-2"));``` 

@thiagotnunes  Let me know your thoughts.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: `commit` is public method. 

For hand-crafted transaction :
https://github.com/googleapis/google-cloud-ruby/blob/880f64fe6db2d48ef989fad356a21d23550cbd88/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L1352-L1373

In this 'Tag-1' set to a transaction(L1364) which is use to set a transaction tag in API query calls executed on that transaction.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
thiagotnunes: I think it would be nice to define a `request_options`, because that will make the API more extensible (we have other features coming soon, for which we could piggy back on this new hashmap then).

For the client design spec, we have the `roTx.read(table, keySet, columns, Tag.of("tag-2"));`, but if you check the implementation of that, you can see that `Tag` is actually a flavour of `Option` as well (https://github.com/googleapis/java-spanner/pull/576/files#diff-992d480774d874d97423da09934b0395ef460130558e2fd6f7b1461ba32535c0R154-R166). So, if we have the `request_options` map here, we would not be deviating from the original design.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
thiagotnunes: Got it, so do we need the tag in the `commit` method? I think we should remove that, since it should be set during the transaction creation, right? Otherwise, what would happen if the user sets a transaction tag in the creation and another one in the commit?

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: @thiagotnunes Thanks for detailed information. Making changes according  to it.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: I think no need to remove it. There is a commit method in the client so the user might directly call it, while Transaction calls the internal private service object commit method. 

Client : user can directly use.
https://github.com/googleapis/google-cloud-ruby/blob/bf17920882c3297f66d2249cd023bb2883e707b6/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L1095

Service: Private 
https://github.com/googleapis/google-cloud-ruby/blob/bf17920882c3297f66d2249cd023bb2883e707b6/google-cloud-spanner/lib/google/cloud/spanner/service.rb#L399

google-cloud-spanner/lib/google/cloud/spanner/client.rb
jiren: @thiagotnunes  So method will look like this. Please confirm.

```
 def execute_query sql, params: nil, types: nil, single_use: nil,
                                  query_options: nil, call_options: nil
                                  query_options: nil, request_options: nil,
```
Use:
```
client.execute_query("SELECT ...", request_options: { request_tag: "Tag-1" })
```

For transaction:
```
def transaction deadline: 120, request_options: nil, call_options: nil
```
Use
```
  client.transaction request_options: { transaction_tag: "Tag-1 } do 
  end
```

google-cloud-spanner/lib/google/cloud/spanner/client.rb
skuruppu: Thanks for clarifying.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
skuruppu: I think this is a reasonable design. When you implement it, just make sure that in the documentation for the function, it is clear if the other type of tag is ignored. Something like "if transaction_tag is specified, it will be ignored".

jiren: ok. @skuruppu Do we need to implement ignore tag logic or just documentation is sufficient?

skuruppu (approved):

google-cloud-spanner/lib/google/cloud/spanner/client.rb
thiagotnunes: @jiren I think that the user should never have to decide whether the tag should be a `request_tag` or `transaction_tag`, but that should be done internally. We could change the example to:

```
client.transaction request_options: { tag: "Tag-1 } do 
end
```

```
client.execute_query("SELECT ...", request_options: { tag: "Tag-1" })
```

Wdyt?

jiren: @thiagotnunes yes this will be helpful for users. I will start making changes for this.

jiren: @thiagotnunes Updated as per above examples.

google-cloud-spanner/lib/google/cloud/spanner/client.rb
```diff
@@ -371,19 +377,37 @@ def query_options
...
+        #   results = db.execute_query "SELECT * FROM users",
+        #                              request_options: request_options
+        #
+        #   results.rows.each do |row|
+        #     puts "User #{row[:id]} is #{row[:name]}"
+        #   end
+        #
         def execute_query sql, params: nil, types: nil, single_use: nil,
-                          query_options: nil, call_options: nil
+                          query_options: nil, request_options: nil,
```
thiagotnunes, line 396: Just realised that we added the `request_options` as the sixth parameter, where that position used to be used by the `call_options`. The type of both parameters is a map, which might be silently fail for users who had previous code written and were not naming the arguments.

Do you think it should be worth to move it to be the last argument of the method?
Also, could you make sure that this PR is a breaking change, since it updates this interface? (You can do that through the commit message, see https://www.conventionalcommits.org/en/v1.0.0/#specification BREAKING CHANGE).

jiren: These are named arguments with the default value. In Ruby named arguments are like hash so, the order has no impact while calling.  Here `call_options` are for the GAPIC GRPC client so moved to last, to separate from spanner arguments parameters.

thiagotnunes: Makes sense, thanks for the explanation

thiagotnunes (approved):

larkee:
@jiren Please rebase this

jiren:
> @jiren Please rebase this

@thiagotnunes Updated code.

skuruppu:
@jiren will you be able to resolve the conflicts?

@hengfengli I think we need your approval here before the PR can be merged.

hengfengli (approved):

jiren:
> @jiren will you be able to resolve the conflicts?
> 
> @hengfengli I think we need your approval here before the PR can be merged.

@skuruppu rebased and resolved conflicts.

PR #8169, opened 2020-11-27
feat(spanner):  Point In Time Recovery (PITR)

thiagotnunes:
Implements the PITR-lite support.

With this functionality users will be able to set the retention period for their databases (through the version_retention_period database field).

Users will also be able to create a backup from a consistent timestamp of the database within its version_retention_period.

### DIFF
--- a/google-cloud-spanner/lib/google/cloud/spanner/backup.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/backup.rb
@@ -165,7 +165,7 @@ module Google
         # taken. The version time has microseconds granularity.
         # @return [Time]
         def version_time
-          Convert.timestamp_to_time @grpc.version_time unless @grpc.version_time.nil?
+          Convert.timestamp_to_time @grpc.version_time
         end
 
         ##
--- a/google-cloud-spanner/lib/google/cloud/spanner/backup.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/backup.rb
@@ -161,6 +161,14 @@ module Google
         end
 
         ##
+        # The timestamp when a consistent copy of the database for the backup was
+        # taken. The version time has microseconds granularity.
+        # @return [Time]
+        def version_time
+          Convert.timestamp_to_time @grpc.version_time unless @grpc.version_time.nil?
+        end
+
+        ##
         # Create time is approximately the time when the backup request was
         # received.
         # @return [Time]
--- a/google-cloud-spanner/lib/google/cloud/spanner/database.rb
+++ b/google-cloud-spanner/lib/google/cloud/spanner/database.rb
@@ -85,6 +85,18 @@ module Google
           @grpc.name.split("/")[5]
         end
 
+        # The version retention period for a database.
+        # @return [String]
+        def version_retention_period
+          @grpc.version_retention_period
+        end
+
+        # The earliest available version time for a database.
+        # @return [Time]
+        def earliest_version_time
+          Convert.timestamp_to_time @grpc.earliest_version_time
+        end
+
         ##
         # The full path for the database resource. Values are of the form
         # `projects/<project_id>/instances/<instance_id>/databases/<database_id>`.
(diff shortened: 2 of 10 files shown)


skuruppu (approved):

skuruppu:
@jiren if you could please review this as well, we'd really appreciate it.

Note that it's intentional to include the CMEK protos here since some people wanted to test both the features at the same time.

jiren:
@thiagotnunes Implementation looks good. 

There are errors of doc test on kokoro due to added `encryption_config`.  For this you need add `encryption_config: nil` in doctest mocks. After fixing mocks I think all test will pass. File ref : https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-spanner/support/doctest_helper.rb

i.e
https://github.com/googleapis/google-cloud-ruby/blob/63aa340e32462decba93de423c9ce56fa8107892/google-cloud-spanner/support/doctest_helper.rb#L131-L135

To check test and docs you can run `bundle exec rake ci`

thiagotnunes:
@jiren Thanks for the pointers, I have pushed a fix for the failing tests.

jiren (approved):

thiagotnunes:
@jiren @skuruppu could you provide a review for the last 5 commits? I added the `version_time` to the backup creation and I added the `version_time` to the `backup_info`, which can be obtained in a restore operation.

google-cloud-spanner/lib/google/cloud/spanner/backup.rb
```diff
@@ -160,6 +160,14 @@ def expire_time= time
           raise error
         end
 
+        ##
+        # The timestamp when a consistent copy of the database for the backup was
+        # taken. The version time has microseconds granularity.
+        # @return [Time]
+        def version_time
+          Convert.timestamp_to_time @grpc.version_time unless @grpc.version_time.nil?
```
jiren, line 168: Here no need to check `nil?` value. `Convert` is already checking it. Just use 
`Convert.timestamp_to_time @grpc.version_time`

Ref:
https://github.com/googleapis/google-cloud-ruby/blob/d1aefdee2045edb5ba277d7e46329e3110b41a89/google-cloud-spanner/lib/google/cloud/spanner/convert.rb#L266-L270

google-cloud-spanner/lib/google/cloud/spanner/database.rb
```diff
@@ -422,13 +451,14 @@ def database_operations filter: nil, page_size: nil
         #     backup = job.backup
         #   end
         #
-        def create_backup backup_id, expire_time
+        def create_backup backup_id, expire_time, version_time = nil
```
jiren, line 454: Here version_time is optional so better to use keyword argument.

`def def create_backup backup_id, expire_time, version_time: nil`

google-cloud-spanner/lib/google/cloud/spanner/database/backup_info.rb
```diff
@@ -84,14 +84,23 @@ def source_database_path
...
 
+          ##
+          # The backup contains an externally consistent copy of
+          # `source_database` at the timestamp specified by
+          # the `version_time` received. If no `version_time` was
+          # given during the creation of the backup, the `version_time`
+          # will be the same as the `create_time`.
+          # @return [Time]
+          def version_time
+            Convert.timestamp_to_time @grpc.version_time unless @grpc.version_time.nil?
```
jiren, line 101: Just use `Convert.timestamp_to_time @grpc.version_time`. Convert is already handling nil value.

jiren (changes requested):

google-cloud-spanner/lib/google/cloud/spanner/backup.rb
thiagotnunes: Thanks for spotting this, fixed.

google-cloud-spanner/lib/google/cloud/spanner/database.rb
thiagotnunes: Fixed.

google-cloud-spanner/lib/google/cloud/spanner/database/backup_info.rb
thiagotnunes: Fixed.

jiren (approved):

Issue #8402, opened 2020-12-15
POST Object v2/v4: error: Policy did not reference these fields: commit


chvreddy:
Thanks for stopping by to let us know something could be better!

#### Environment details

Environment details
OS: MAC
Ruby version: ruby 2.4.9p362 (2019-10-02 revision 67824) [x86_64-darwin19]
Gem name and version: gem 'google-cloud-storage', '1.29.2'

ruby

#### Code example

```
def presigned_post(success_action_redirect:)
  conditions = [ ["starts-with", "$key", ""], {"success_action_redirect":success_action_redirect} ]

  result = self.class.bucket. generate_signed_post_policy_v4("#{inbox_prefix}#{Time.current.to_i}_${filename}", conditions: conditions, signer: gcs_signed_url(nil, nil, "POST"), issuer: ENV["GCS_SERVICE_ACCOUNT"] )
  fields = result.fields.merge!(success_action_redirect: success_action_redirect)
  result
end

FORM:
@presigned_post = presigned_post(success_action_redirect: files_processing_banker_final_qc_loan_tapes_url)
<div class="large-7 columns form-element-label">
  <form action="<%= @presigned_post.url %>" method="post" enctype="multipart/form-data">
    <% @presigned_post.fields.each do |name, value| %>
      <input type="hidden" name="<%= name %>" value="<%= value %>"/>
    <% end %>

    <div class="large-6 columns form-element-label">
      <%= label_tag :file_loan, "Choose CSV file to upload" %>
    </div>
    <div class="large-6 columns">
      <%= file_field_tag :file, id: :loan_file, class: 'tiny', accept: '.csv,.CSV' %>
    </div>

    <div class="large-6 columns form-element-label"></div>
    <div class="large-6 columns">
      <%= submit_tag 'Upload', id: 'loan_submit', class: 'button tiny' %>
    </div>
  </form>
</div>
```

**Will pass if uploaded file Greater than 65MB
And when uploaded file less than 65MB it will fail Throwing below error:**

```
<Error>
<Code>InvalidPolicyDocument</Code>
<Message>The content of the form does not meet the conditions specified in the policy document.</Message>
<Details>Policy did not reference these fields: commit</Details>
</Error>
```
For suppose if I add a policy condition with commit

```
conditions = [ ["starts-with", "$key", ""], ["starts-with", "$commit", ""], {"success_action_redirect":success_action_redirect}]
```

**Then it passes for file size less than 65KB and
[...]

PR #8406, opened 2020-12-16
chore(storage): Fix sample storage_generate_signed_post_policy_v4

quartzmo:
refs: #8402

### DIFF
--- a/google-cloud-storage/samples/acceptance/files_test.rb
+++ b/google-cloud-storage/samples/acceptance/files_test.rb
@@ -492,6 +492,7 @@ describe "Files Snippets" do
     assert_includes out, "<input name='policy'"
     assert_includes out, "<input name='x-goog-meta-test' value='data'"
     assert_includes out, "<input type='file' name='file'/>"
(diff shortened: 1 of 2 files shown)


google-cloud-storage/samples/acceptance/files_test.rb
```diff
@@ -492,6 +492,7 @@ def mock_cipher.random_key
     assert_includes out, "<input name='policy'"
     assert_includes out, "<input name='x-goog-meta-test' value='data'"
     assert_includes out, "<input type='file' name='file'/>"
+    assert_includes out, "<input type='submit' value='Upload File'/>"
```
frankyn, line 495: Chris, this wasn't originally added, but would you be open to adding an actual request using the form data into the test?

quartzmo: Added testing with Net::HTTP POST, which is not exactly using the "form data". Let me know if you had something else in mind.

frankyn (approved):

PR #9357, opened 2021-02-19
test(spanner): acceptance test for numeric primary key

jiren:


google-cloud-spanner/acceptance/spanner_helper.rb
```diff
@@ -185,14 +185,35 @@ def items_ddl_statement
         ITEMS
       end
 
+      def numeric_pk_ddl_statement
```
larkee, line 188: These added functions need to check if the emulator is enabled before referencing `NUMERIC`, similar to `stuffs_ddl_statement`. Currently, they are causing failures against the emulator.

skuruppu (approved):

google-cloud-spanner/acceptance/spanner_helper.rb
jiren: @larkee  updated.

larkee (approved):
LGTM, although I'm not very familiar with Ruby..

PR #10827, opened 2021-03-08
remove vars only used once

d-m-u:
It looks like these are only used once, do they need to be saved?


d-m-u:
@googlebot I fixed it.

d-m-u:
@googlebot I fixed it.

PR #10907, opened 2021-03-24
chore(speech): fixed lingering rubocop violations in the samples

trambui09:
Deleted the disabling and enabling of metrics/abcsize and metrics/methodlength.


quartzmo (approved):
🎉

Issue #10924, opened 2021-03-25
Deprecate 'return_immediately' arguments


quartzmo:
`Subscription#pull` (synchronous pull) exposes an `immediate` option with a default value of `true`. The default value is sent when the method is called without arguments, as demonstrated in doc site samples such as the following at https://cloud.google.com/pubsub/docs/pull#ruby_6.

```
require "google/cloud/pubsub"

pubsub = Google::Cloud::Pubsub.new

subscription = pubsub.subscription subscription_id
subscription.pull.each do |message|
  puts "Message pulled: #{message.data}"
  message.acknowledge!
end
```

The proto docs say this flag is deprecated and include a warning:

> Warning: setting this field to `true` is discouraged because it adversely impacts the performance of `Pull` operations. We recommend that users do not set this field.

google-cloud-pubsub should do the same.

This option is only exposed in [Subscription#pull](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb#L744-L751).

/cc @hongalex, @kamalaboulhosn

See also: googleapis/python-pubsub#350

PR #11126, opened 2021-04-08
docs(pubsub): Update Subscription#pull docs and samples to recommend return_immediately: false

quartzmo:
Update `Subscription#pull` docs and samples to recommend `immediate: false` (sending `return_immediately: false`).

The default value sent when calling `Subscription#pull` without arguments will be `return_immediately: true`.

closes: #10924

### DIFF
--- a/google-cloud-pubsub/OVERVIEW.md
+++ b/google-cloud-pubsub/OVERVIEW.md
@@ -205,13 +205,16 @@ sleep
 Messages also can be pulled directly in a one-time operation. (See
 {Google::Cloud::PubSub::Subscription#pull Subscription#pull})
 
+The `immediate: false` option is now recommended to avoid adverse impacts on
+pull operations.
+
 ```ruby
 require "google/cloud/pubsub"
 
 pubsub = Google::Cloud::PubSub.new
 
 sub = pubsub.subscription "my-topic-sub"
-received_messages = sub.pull
+received_messages = sub.pull immediate: false
 ```
 
 A maximum number of messages to pull can be specified:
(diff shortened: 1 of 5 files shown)


dazuma (approved):

hongalex (approved):

google-cloud-pubsub/OVERVIEW.md
```diff
@@ -205,13 +205,16 @@ sleep
 Messages also can be pulled directly in a one-time operation. (See
 {Google::Cloud::PubSub::Subscription#pull Subscription#pull})
 
+The `immediate: false` option is now recommended to avoid adverse impacts on
```
kamalaboulhosn, line 208: I would say "The `immediate: false` option is recommended to avoid adverse impacts on performance of pull operations." That way we are just a little more specific about the adverse impact> Also took out "now" as I think the natural question would be "when did 'now' start?" and so seems like we can just avoid that. If they do get interested in when the docs were updated, they can look at blame and see when it changed.

google-cloud-pubsub/lib/google/cloud/pubsub/subscription.rb
```diff
@@ -733,19 +733,27 @@ def detach
         end
 
         ##
-        # Pulls messages from the server. Returns an empty list if there are no
-        # messages available in the backlog. Raises an ApiError with status
-        # `UNAVAILABLE` if there are too many concurrent pull requests pending
-        # for the given subscription.
+        # Pulls messages from the server, blocking until messages are available
+        # when called with the `immediate: false` option, which is now recommended
+        # to avoid adverse impacts on pull operations. Raises an API error with
```
kamalaboulhosn, line 738: I would make the same change here as I mentioned above.

PR #11234, opened 2021-04-27
docs(storage): Update IAMCredentialsService#sign_service_account_blob inline code samples

quartzmo:
Partial fix for #11224, see [comment](https://github.com/googleapis/google-cloud-ruby/issues/11224#issuecomment-827906734).

refs: #11224

PR #11243, opened 2021-04-29
docs(storage): Update IAMCredentialsService#sign_service_account_blob examples

quartzmo:
closes: #11224 

Replaces: #11234

### DIFF
--- a/google-cloud-storage/acceptance/storage/bucket_generate_signed_post_policy_v4_test.rb
+++ b/google-cloud-storage/acceptance/storage/bucket_generate_signed_post_policy_v4_test.rb
@@ -72,9 +72,9 @@ describe Google::Cloud::Storage::Bucket, :generate_signed_post_policy_v4, :stora
     # Only defined when using a service account
     issuer = iam_client.authorization.issuer
     signer = lambda do |string_to_sign|
-      request = {
-        "payload": string_to_sign,
-      }
+      request = Google::Apis::IamcredentialsV1::SignBlobRequest.new(
+        payload: string_to_sign
+      )
       resource = "projects/-/serviceAccounts/#{issuer}"
       response = iam_client.sign_service_account_blob resource, request
       response.signed_blob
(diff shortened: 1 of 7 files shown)


google-cloud-storage/acceptance/storage/bucket_generate_signed_post_policy_v4_test.rb
```diff
@@ -72,9 +72,9 @@
     # Only defined when using a service account
     issuer = iam_client.authorization.issuer
     signer = lambda do |string_to_sign|
-      request = {
-        "payload": string_to_sign,
-      }
+      request = Google::Apis::IamcredentialsV1::SignBlobRequest.new(
```
frankyn, line 75: What happened here?

quartzmo: @frankyn It appears that using the `Google::Apis::IamcredentialsV1::SignBlobRequest` request object instead of the hash resolves both the new `ArgumentError: unknown keyword: payload` problem and avoids the `` error. I'm guessing you were right about `payload :` and `payload =>`? causing a gotcha / issue with the json body between the two. I can still log the requests if you are interested.

frankyn: that's wild. this wasn't required before, but I'm not opposed to this change. If it's not too much trouble could you log request bodies for both?

frankyn (approved):
Thanks @quartzmo, I'm wondering if there was a change to the google-api-ruby-client which broke this. For now this LGTM, thanks for your patience and addressing this issue!

google-cloud-storage/acceptance/storage/bucket_generate_signed_post_policy_v4_test.rb
quartzmo: I [posted the logs](https://github.com/googleapis/google-cloud-ruby/issues/11224#issuecomment-829345579) to #11224.

Issue #11378, opened 2021-05-10
spanner: add support to create low cost  instance using processing units 


jiren:
- create instance using `processing_units` or `nodes_count`

jiangmichaellll:
Can you assign this to spanner team?

PR #11379, opened 2021-05-10
feat(spanner): create instance using processing units/node count

jiren:

closes: #11378


jiren:
@hengfengli FYI, Generated gapic code is not added

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
```diff
@@ -188,8 +206,55 @@ def labels= labels
           )
         end
 
+        ##
+        # Update changes.
+        #  `display_name`, `lables`, `nodes`, `processing_units` can be
```
hengfengli, line 211: lables -> labels

jiren: done

google-cloud-spanner/lib/google/cloud/spanner/service.rb
```diff
@@ -144,12 +146,12 @@ def create_instance instance_id, name: nil, config: nil, nodes: nil,
           instances.create_instance request, opts
         end
 
-        def update_instance instance, call_options: nil
+        def update_instance instance, field_mask: nil, call_options: nil
```
hengfengli, line 149: This seems to be a breaking change. If the user doesn't provide a `field_mask` in the new version, will `update_instance` updates `display_name`, `node_count`, and `labels`?

hengfengli: Consider to use `%w[display_name node_count labels]` as a default value of `field_mask`.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -217,14 +221,41 @@ def instance instance_id
...
+        #   else
+        #     instance = job.instance
+        #   end
+        #
         def create_instance instance_id, name: nil, config: nil, nodes: nil,
-                            labels: nil
+                            processing_units: nil, labels: nil
           config = config.path if config.respond_to? :path
+
+          if nodes && processing_units
```
hengfengli, line 249: I'd like to avoid adding too many logic in the admin client, can we leave this check to the backend? So we don't need to add this check and its test as well.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
jiren: ok, I will make necessary changes.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -177,7 +177,10 @@ def instance instance_id
         #   configuration. Values can be the `instance_config_id`, the full
         #   path, or an {Instance::Config} object. Required.
         # @param [Integer] nodes The number of nodes allocated to this instance.
-        #   Required.
+        #   Optional. Specifiy either `nodes` or `processing_units`
```
hengfengli, line 180: Specifiy -> Specify

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -177,7 +177,10 @@ def instance instance_id
...
+        # @param [Integer] processing_units The number of processing units
+        #   allocated to this instance. Optional. Specifiy either `nodes`
```
hengfengli, line 182: Same here.

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
```diff
@@ -188,8 +206,55 @@ def labels= labels
...
+        #   job.reload! # API call
+        #   job.done? #=> true
+        #
+        #   if job.error?
+        #     status = job.error
+        #   else
+        #     instance = job.instance
+        #   end
+        #
         def save
```
hengfengli, line 243: Can we provide a `field_mask` to `save` method? So we can avoid the change of line 246-255.

hengfengli: This might not be a good idea. I was just trying to minimize the changes.

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
```diff
@@ -188,8 +206,55 @@ def labels= labels
...
-          job_grpc = service.update_instance @grpc
+          ensure_service!
+
+          field_mask = []
+          @orignal_values.each do |field, value|
+            field_mask << field unless @grpc[field.to_s] == value
+          end
+
+          if field_mask.include?(:node_count) &&
```
hengfengli, line 251: Remove this check as well. Let the backend to return the error and keep a thin client.

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
jiren: Yes we can provide optional argument for `field_mask`, but for usability wise it looks like it is not good. User need to first set the field and then need to pass field mask.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
jiren: Updated logic. Let me know your suggestions.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
jiren: done

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
jiren: done

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
jiren: As per new updated changes by default, it will updates `[display_name node_count labels]`.  If the user has updated fields then it will check for changes for respective fields and build field mask array,

hengfengli: An issue with the current implementation is that users can't change the value twice. Let's say `processing_units` = 500 at the beginning, then we update it to `800`. After that, we change it again to `500` (this will fail because the value is equal to the original value `500`). You need to update `original_values` when the new values are saved.

google-cloud-spanner/lib/google/cloud/spanner/service.rb
hengfengli: Looks good to me.

google-cloud-spanner/lib/google/cloud/spanner/instance.rb
jiren: Updated original value after update request. One catch here is if a job fails then it will not revert back to original values.

hengfengli: Is there any way that we don't store `original_values`? Another problem I notice is that if no thing changes, then `field_mask` will be empty. When we pass an empty array of `field_mask` to `update_instance`, which will use the default config `%w[display_name node_count labels]`.

jiren: @hengfengli Is there any way to convert nodes to processing units? If there is a formula then calculate processing units and update only `processing_units`.

hengfengli: > @hengfengli Is there any way to convert nodes to processing units? If there is a formula then we calculate processing units and update only `processing_units`.

No, I don't think it can work in this way.

hengfengli: Actually, I am wrong. 1 node == 1000 processing_units. If we create 5000 processing unites, then it will be equal to 5 nodes.

hengfengli: But I am a bit hesitate to hardcode this value into ruby client lib. I hope we don't do this.

hengfengli: Can we remove `original_values`? and if `processing_units` is not empty, we use `processing_units`. If `processing_units` is empty, then we use `node_count`. What do you think?

jiren: Processing units less than 1000 then api retuning 0 node count in response. If it is 1000 then returning node count is 1. 
If processing unit set to >1000 then node count is always present.

hengfengli: Yeah, that's correct. But I don't want to write any thresholds in the code. So probably we just keep with the current `original_values` implementation. Instead of calling it `original_values`, may consider `current_values`.

jiren: ok

jiren: renamed original values to `current_values`

hengfengli:
@jiren Thanks a lot for fixing all the comments. It looks good to me now. I will approve it once it can be merged.

hengfengli:
The proto change has been merged. We can start to merge this PR.

hengfengli (approved):

PR #11613, opened 2021-05-19
doc(spanner): document single use as default

olavloite:
Document that Cloud Spanner will use a single use read-only transaction as default for queries when no other option is specified. Also changes the example code for the database client to show how to execute a single query using a single use read-only transaction, instead of a read/write transaction.


olavloite:
cc @hengfengli @jiren

hengfengli (approved):
LGTM.

hengfengli:
@olavloite Same here. Please rebase it. Thanks.

olavloite:
> @olavloite Same here. Please rebase it. Thanks.

done

Issue #11389, opened 2021-05-11
(BigQuery) add support for ParquetOptions for load/external tables


shollyman:
This FR is to add support for some advanced Parquet options (list inference and enums) when loading data or defining an external table.

Additional information has been shared out of band with library maintainers.

PR #11637, opened 2021-05-25
refactor(bigquery): Refactor classes in external.rb to multiple files

quartzmo:
This simple (but verbose) refactoring is a precursor to introducing a new `External` class for Parquet Options (#11389).


dazuma (approved):
LGTM

Issue #11666, opened 2021-06-09
google-cloud-trace v0.4.1 broken on Ruby 3.0


grobie:
#### Environment details

  - OS: Linux
  - Ruby version: 3.0.1p64
  - Gem name and version: google-cloud-trace v0.4.1

#### Steps to reproduce

  1. Add `stackdriver` to Gemfile (following https://cloud.google.com/trace/docs/setup/ruby)
  2. Run vanilla Rails service with Ruby 3.0 in Cloud Run

Expected: See traces reported by stackdriver gem
Actual: No traces at all, no error logs

#### Code example

n/a, even basic Rails trace support is completely broken with Ruby 3.0

#### Full backtrace

There are no logs or any other reporting information available. No traces will be sent by the application.

grobie:
@dazuma as you contributed Ruby 3.0 support for google-cloud-trace in https://github.com/googleapis/google-cloud-ruby/commit/56ca7194af7ab72c352c53dc92d321decd724839.

I digged into the google-cloud-trace gem and was able to find and fix the Ruby 3.0 issue. See #11667.

---

The google-cloud-trace gem is unusable with Ruby 3.0 at the moment. It appears you don't have any integration tests available, as the absence of trace data should have been evident. Do you have any plans to set up some high-level integration tests so catch such issues in the future?

PR #11667, opened 2021-06-09
Fix broken Ruby 3.0 support for GRPC::ActiveCallWithTrace patch

grobie:
In Ruby 3.0 kwargs are explicitly separated from regular arguments and
must be handled accordingly.

https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

closes: #11666


dazuma (approved):
Thanks!

Issue #11419, opened 2021-05-17
pubsub: GetSchema doesn't retrieve schema definition


hongalex:
#### Steps to reproduce

  1. Run `pubsub.schema` to retrieve a schema without a view (`view: :full`)

#### More info

This is analogous to https://github.com/googleapis/java-pubsub/issues/669. By default, get/list schemas should default to the `FULL` view.

quartzmo:
@hongalex Since this was opened as a `bug`, the change should be published as a semver `patch`, correct?

hongalex:
Yeah patch sounds right to me.

PR #11671, opened 2021-06-12
fix(pubsub): Fix Project#schema and #schemas to return full resource

quartzmo:
* Include schema definition in default return values.
* Fix Schema#definition to return nil instead of empty string when not present.

closes: #11419


hongalex (approved):

dazuma (approved):
This is a behavioral change, but should be backwards-compatible.

PR #12006, opened 2021-06-17
feat: [Compute] Compute can now be synthed, improvements for public preview

viacheslav-rostovtsev:
feat: added synth.py
feat: compute now sends rest token in `x-goog-api-client` header
feat: compute has unit tests


dazuma (approved):
Great!

PR #12161, opened 2021-06-24
chore: generate toc when publishing docs to cloudrad

TheRoyalTnetennba:
Creates a menu entry for every class and module. Classes and modules with children that are either classes or modules get a dropdown containing their class/module children.

Add docs.metadata to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ jsondoc/*
 # Ignore YARD stuffs
 .yardoc
 
+# Ignore devsite/cloudrad stuffs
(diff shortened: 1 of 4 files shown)


TheRoyalTnetennba:
outputted toc for google-cloud-vision-v1: https://pastebin.com/jUQW7sZQ

TheRoyalTnetennba:
Jumbled a few strategies together - need to clean up before review

dazuma (approved):

PR #13106, opened 2021-07-12
Set skip_deserialization:false in BigQuery insert_tabledata

Empact:
As InsertResponse depends on @gapi being serialized for #insert_errors, etc.

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb
@@ -264,7 +264,10 @@ module Google
           execute backoff: true do
             service.insert_all_table_data(
               @project, dataset_id, table_id, insert_req,
-              options: { skip_serialization: true }
+              options: {
+                skip_serialization: true,
+                skip_deserialization: false
+              }
             )
           end
         end


quartzmo:
@Empact Thank you for opening this PR. The default value of `skip_deserialization` should already be `false`. (See [google-apis-core/lib/google/apis/options.rb#L111](https://github.com/googleapis/google-api-ruby-client/blob/master/google-apis-core/lib/google/apis/options.rb#L111).) Can you open an issue explaining and demonstrating (ideally, with example code) the problem you're having? (Possibly, for example, the inaccessibility of insert_errors?)

Empact:
The problem occurs when the skip_deserialization value is configured to true. Seems that should either be overridden or the methods that rely on deserialization should raise in that case.

quartzmo:
How is the skip_deserialization value configured to true?

quartzmo:
I see four locations where `skip_deserialization: true` is passed to RPCs in this client. [Here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-bigquery/lib/google/cloud/bigquery/service.rb#L230) for example. I'll do some testing to see if those RPCs somehow change the state of the options in the google-api-ruby-client instance.

Empact:
> How is the skip_deserialization value configured to true?

`Google::Apis::RequestOptions.default.skip_deserialization = true`

quartzmo:
Where is that set?

quartzmo:
Assuming that you are setting that default in your own code, why are you doing it?

Empact:
Does that matter? If the option exists, shouldn't the other code reflect that as well?

quartzmo:
I see your point, and in a perfect world a library like this one would guard against any changes to the configuration of its dependencies. If we do go down that route, the changes would almost certainly be much broader than the single change in this PR. Any information you can provide about your use case would help us determine both the scope and the priority of the work.

quartzmo:
Closing for now. Before continuing work on this PR or a similar solution, an issue should be opened specifying the scope of the work needed to guard the RPCs in this library against any changes to the configuration of googleapis/google-api-ruby-client. The changes would almost certainly be much broader than the single change in this PR, so it will best to do the analysis upfront.

PR #13126, opened 2021-07-20
test(spanner): update optimizer_statistics_package to latest

hengfengli:


hengfengli:
For a note, I only updated the integration tests but not unit tests.

skuruppu (approved):

PR #13456, opened 2021-08-11
feat(spanner): wrapper to create generated admin database client.

skuruppu:
Makes a generated admin client available that also picks up the config that may have setup an emulator endpoint.


google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -552,6 +552,77 @@ def client instance_id, database_id, pool: {}, labels: nil,
...
+        #        })
+        #   job.wait_until_done!
+        #   instance = job.results
+        #
+        def instance_client
+          Admin::Instance::V1::InstanceAdmin::Client.new do |config|
+            config.credentials = service.channel
+            config.timeout = service.timeout if service.timeout
+            config.endpoint = service.host if service.host
+            config.metadata = { "google-cloud-resource-prefix" => "projects/#{@project}" }
```
quartzmo, line 591: Maybe at this point you might yield `config` to the caller of `instance_client` in case they want to do any additional configuration of the `V1::InstanceAdmin::Client` instance?

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -552,6 +552,77 @@ def client instance_id, database_id, pool: {}, labels: nil,
...
+        #       instance: Google::Cloud::Spanner::Admin::Instance::V1::Instance.new({
+        #         display_name: "My Instance",
+        #         config: config_path,
+        #         node_count: 5,
+        #         labels: { "production": :env }
+        #        })
+        #   job.wait_until_done!
+        #   instance = job.results
+        #
+        def instance_client
```
hengfengli, line 586: you may want to call `ensure_service!` here.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -552,6 +552,77 @@ def client instance_id, database_id, pool: {}, labels: nil,
...
+        #
+        #   instance_path = \
+        #     db_admin_client.instance_path project: "my-project",
+        #                                   instance: "my-instance"
+        #   job = db_admin_client.create_database parent: instance_path,
+        #     create_statement: "CREATE DATABASE my-database",
+        #   job.wait_until_done!
+        #   database = job.results
+        #
+        def database_client
```
hengfengli, line 617: Same here as well.

google-cloud-spanner/lib/google/cloud/spanner/project.rb
```diff
@@ -552,6 +552,77 @@ def client instance_id, database_id, pool: {}, labels: nil,
...
+        #
+        #   spanner = Google::Cloud::Spanner.new
+        #
+        #   db_admin_client = spanner.database_client
+        #
+        #   instance_path = \
+        #     db_admin_client.instance_path project: "my-project",
+        #                                   instance: "my-instance"
+        #   job = db_admin_client.create_database parent: instance_path,
+        #     create_statement: "CREATE DATABASE my-database",
```
hengfengli, line 613: I think it should be "CREATE DATABASE \`my-database\`". It does not support `-` without the backticks: https://cloud.google.com/spanner/docs/data-definition-language#database-id-names

hengfengli:
Looks good to me.

skuruppu:
@quartzmo and @hengfengli PTAL.

@dazuma recommended that we follow a similar wrapper to that of https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-kms/lib/google/cloud/kms.rb created by the generator. I adapted it to initialize with the emulator if that setting is detected. I only implemented this for the database client so far and if this looks reasonable, I will implement it for the instance client and revert my first commit for the odd-looking wrappers in the spanner client.

google-cloud-spanner/lib/google/cloud/spanner/admin/database.rb
```diff
@@ -0,0 +1,97 @@
...
+          #
+          def self.new
+            if configure.emulator_host
+              return Admin::Database::V1::DatabaseAdmin::Client.new do |config|
+                config.credentials = :this_channel_is_insecure
+                config.endpoint = configure.emulator_host
+              end
+            end
+
+            Admin::Database::V1::DatabaseAdmin::Client.new
```
quartzmo, line 55: It looks to me as though the config object provided here by `Spanner::Admin::Database.configure` never makes it to the `Admin::Database::V1::DatabaseAdmin::Client` instance, which is quite different than how things work in [`Google::Cloud::Spanner.new`](), where fields such as `lib_name`, etc are actually used to configure the client that is returned. Nor does `Spanner::Admin::Database.new` provide a block parameter that is passed to the client's `new` method, as in [`Kms.key_management_service`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-kms/lib/google/cloud/kms.rb#L85). When using this method to create a client, no instance-scope configuration of that client would be possible here. (Note that `Kms.key_management_service` does not document the block param, which may be a potential problem in that library.)

skuruppu: Thanks for the pointers @quartzmo. I now made the initialization look more like `Google::Cloud::Spanner.new`. However, I don't quite understand what to do with the block parameter now. Is this still needed, and if so, how do I use it? Sorry, I'm completely new to Ruby so if you have any code snippets or examples of how it can be done alongside other input parameters, that would be great.

quartzmo: @skuruppu This is a question of style, but personally I think making this method consistent with `Google::Cloud::Spanner.new` is the right choice. Since introducing the micro-generator and the new clients that accept a block, the cleanest and easiest way to provide **unmodified** user configuration to them is simply to expose a block parameter that is passed to the client's `new` method, as in [`Kms.key_management_service`](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-kms/lib/google/cloud/kms.rb#L85). But in this case, consistency with the long list of keyword args in `Google::Cloud::Spanner.new` is probably better for users who have experience using that method. It's a style call. We should probably leave this thread open for @dazuma to comment.

skuruppu: Ah ok, that makes sense. I would love for @dazuma to chime in here as well then. I will work on the unit/accpetance tests in the meantime to make sure that what I've done works.

google-cloud-spanner/lib/google/cloud/spanner/admin/database.rb
```diff
@@ -0,0 +1,209 @@
...
+          #     * `:initial_delay` (*type:* `Numeric`) - The initial delay in seconds.
+          #     * `:max_delay` (*type:* `Numeric`) - The max delay in seconds.
+          #     * `:multiplier` (*type:* `Numeric`) - The incremental backoff multiplier.
+          #     * `:retry_codes` (*type:* `Array<String>`) -
+          #       The error codes that should trigger a retry.
+          #
+          # @return [::Google::Cloud::Config] The default configuration used by this library
+          #
+          def self.configure
+            yield ::Google::Cloud.configure.spanner if block_given?
```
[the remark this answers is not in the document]
skuruppu: I think the configuration should be the same for the three clients though if we're trying to keep things more or less the same as before.

What are the cases you were thinking of where it might be confusing for users? E.g. they try to have a different timeout for admin clients compared to the data client?

Where I think it's useful to use the same configuration is when using the emulator for example.

hengfengli (approved):
LGTM.

quartzmo (approved):
LGTM

google-cloud-spanner/test/google/cloud/spanner/database/client_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  def test_database_admin
+    emulator_host = "localhost:4567"
+    emulator_check = ->(name) { (name == "SPANNER_EMULATOR_HOST") ? emulator_host : nil }
+    # Clear all environment variables, except SPANNER_EMULATOR_HOST
+    ENV.stub :[], emulator_check do
+      Gapic::ServiceStub.stub :new, :stub do
+          client = Google::Cloud::Spanner::Admin::Database.database_admin project_id: "1234"
+          assert_kind_of Google::Cloud::Spanner::Admin::Database::V1::DatabaseAdmin::Client, client
+          # TODO: Figure out why this fails.
+          # assert_equal :this_channel_is_insecure, client.configure.credentials
```
skuruppu, line 31: @bajajneha27 if you can resolve this TODO with the suggestion that @quartzmo made in #16080, that would be great.

Issue #2260, opened 2018-08-07
Firestore Transaction Retry Behavior


blowmage:
In the Firestore Watch PR #2202 @schmidt-sebastian made the [following comment](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2202#discussion_r206967622) about Transaction retry:

> If a transaction fails with an application-level error (e.g. FAILED_PRECONDITION), it should be retried immediately without backoff. Can you confirm that that is the case? The backend will save the transactions "spot in line" and the transaction is much more likely to succeed if is retried right away.

Make sure Firestore Transactions are performing retries for all appropriate errors and with the correct delay (including incremental backoff).

blowmage:
I am releasing this issue as I am not going to work on it.

quartzmo:
@schmidt-sebastian The Ruby client currently [retries transactions](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-firestore/lib/google/cloud/firestore/client.rb#L626-L645) only for `UNAVAILABLE` errors. This is in addition to the underlying gapic client retry for the `commit` RPC, which is [configured here](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-firestore-v1/lib/google/cloud/firestore/v1/firestore/client.rb#L104) for codes 8 and 14 only (`RESOURCE_EXHAUSTED` and `UNAVAILABLE`). The list of retryable errors in nodejs-firestore was expanded in googleapis/nodejs-firestore#953 and [now includes](https://github.com/googleapis/nodejs-firestore/blob/master/dev/src/transaction.ts#L601-L624) the following additional codes. Should these be added to the top-level transaction retry logic in the Ruby client?

* `ABORTED`
* `CANCELLED`
* `UNKNOWN`
* `DEADLINE_EXCEEDED`
* `INTERNAL`
* `UNAUTHENTICATED`
* `RESOURCE_EXHAUSTED`
* `INVALID_ARGUMENT` (if message matches `transaction has expired`)

schmidt-sebastian:
@quartzmo We have discussed our retries with the backend team and used the above list as a sort of compromise. The list mostly matches what the Android, iOS and Web SDKs do and improves transaction resilience on the servers. I would generally advise to follow the Node implementation as closely as possible. Note that some of these errors result in individual RPCs to be retried while other (such as `ABORTED`) result in a new transaction run.

quartzmo:
[...]

PR #13632, opened 2021-08-19
fix(firestore): Fix transaction retry behavior

quartzmo:
closes: #2260


schmidt-sebastian (approved):
The error codes LGTM. I think the tests could use some de-duplication though :)

dazuma (approved):

PR #13751, opened 2021-08-28
[DO NOT MERGE] demo: spanner clients' changes with explicit headers

viacheslav-rostovtsev:
Demo of how the client will change with the explicit routing headers annotations used in generation

### DIFF
--- a/google-cloud-spanner-admin-database-v1/lib/google/cloud/spanner/admin/database/v1/database_admin/client.rb
+++ b/google-cloud-spanner-admin-database-v1/lib/google/cloud/spanner/admin/database/v1/database_admin/client.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


google-cloud-spanner-admin-database-v1/lib/google/cloud/spanner/admin/database/v1/database_admin/client.rb
```diff
@@ -768,9 +788,25 @@ def set_iam_policy request, options = nil
                     gapic_version: ::Google::Cloud::Spanner::Admin::Database::V1::VERSION
                   metadata[:"x-goog-user-project"] = @quota_project_id if @quota_project_id
 
-                  header_params = {
-                    "resource" => request.resource
-                  }
+                  header_params = {}
+                  if request.resource
```
quartzmo, line 792: The repeated `if request.resource` here could be refactored.

viacheslav-rostovtsev: I'm thinking about it but it's within the realm of 'ugliness' one can expect from the generated code and it makes the code in the generator a bit harder to navigate.
And in some cases you don't want to do it even.
But it's on my radar.

PR #13963, opened 2021-09-13
feat: add smoketest for enum values handling in compute

viacheslav-rostovtsev:
This test verifies that an unknown enum value in proto JSON is being parsed into a label for the default enum value (`0`). 
In proto3 there is always a enum member with a value `0`; in Compute specifically a 0-valued member is always a nonfunctional one, called `UNDEFINED_<ENUM_NAME>`.
For this specific example, for a enum `State`, it's `UNDEFINED_STATE`, so an unknown enum value in proto JSON is parsed into a label `:UNDEFINED_STATE`.


quartzmo:
It looks to me like this should run as a unit test, not a smoke test. Can you add handwritten unit tests for this library? I copy/pasted just the test (without `setup`) to `test/google/cloud/compute/v1/accelerator_types_test.rb` and got this result:

```
  1) Error:
Google::Cloud::Compute::V1::AcceleratorTypes::ClientTest#test_parse_unknown_enum_value:
Google::Protobuf::ParseError: Error occurred during parsing: Error parsing JSON @1:52: Unknown enumerator: 'THIS_VALUE_DOES_NOT_EXIST'
    /Users/quartzmo/code/google/codez/google-cloud-ruby/google-cloud-compute-v1/test/google/cloud/compute/v1/accelerator_types_test.rb:74:in `decode_json'
    /Users/quartzmo/code/google/codez/google-cloud-ruby/google-cloud-compute-v1/test/google/cloud/compute/v1/accelerator_types_test.rb:74:in `test_parse_unknown_enum_value'

```

google-cloud-compute-v1/acceptance/google/cloud/compute/v1/misc_smoke_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  def test_parse_unknown_enum_value
+    deprectation_status_json = Google::Cloud::Compute::V1::DeprecationStatus.new(
+      deleted: "foo",
+      state: :OBSOLETE).to_json
+
+    ds_decoded = ::Google::Cloud::Compute::V1::DeprecationStatus.decode_json deprectation_status_json
+    assert_equal :OBSOLETE, ds_decoded.state
+
+    unknown_enum_json = deprectation_status_json.gsub "OBSOLETE", "THIS_VALUE_DOES_NOT_EXIST"
+    ds_decoded = ::Google::Cloud::Compute::V1::DeprecationStatus.decode_json unknown_enum_json
```
haberman, line 24: I think you need to pass `{:ignore_unknown_fields => true}` to get the ignoring behavior: https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/message.c#L953-L954

Issue #15553, opened 2021-11-02
feat(bigquery): Add session mode support for queries


quartzmo:
There are several components to supporting sessions. Session support should expose these components, allowing users to manually associate queries with a session. It should not present a `Session` decorator or other high-level abstraction.

* The `CreateSession` configuration property, which can be set as part of `jobs.insert` or `jobs.query` to create a session while executing the query.
* The `session_id` for a query is returned back in the `SessionInfo` message in job statistics.
* The `ConnectionProperties` field in query configuration allows users to set key-value pairs associated with queries.  Users can populate the `session_id` key of the property list to associate subsequent queries with a previously created session.

Reference implementations:

* [Go](https://github.com/googleapis/google-cloud-go/pull/4754)
* [Java](https://github.com/googleapis/java-bigquery/pull/1652)

PR #15699, opened 2021-11-04
feat(bigquery): Add session support

quartzmo:
* Add `create_session` and `session_id` params to `Project#query_job`
* Add `session_id` param to `Project#query`
* Add `create_session` and `session_id` params to `Dataset#query_job`
* Add `session_id` param to `Dataset#query`
* Add `Job#session_id`
* Add `QueryJob::Updater#create_session=`
* Add `QueryJob::Updater#session_id=`

closes: #15553

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
@@ -1244,6 +1244,8 @@ module Google
         #   Flattens all nested and repeated fields in the query results. The
         #   default value is `true`. `large_results` parameter must be `true` if
         #   this is set to `false`.
+        # @param [Integer] maximum_billing_tier Deprecated: Change the billing
+        #   tier to allow high-compute queries.
         # @param [Integer] maximum_bytes_billed Limits the bytes billed for this
         #   job. Queries that will have bytes billed beyond this limit will fail
         #   (without incurring a charge). Optional. If unspecified, this will be
@@ -1294,8 +1296,12 @@ module Google
         #   For additional information on migrating, see: [Migrating to
         #   standard SQL - Differences in user-defined JavaScript
         #   functions](https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql#differences_in_user-defined_javascript_functions)
-        # @param [Integer] maximum_billing_tier Deprecated: Change the billing
-        #   tier to allow high-compute queries.
+        # @param [Boolean] create_session If true, creates a new session, where the
+        #   session ID will be a server generated random id. If false, runs query
+        #   with an existing session ID when one is provided in the `session_id`
+        #   param, otherwise runs query in non-session mode. See {Job#session_id}.
+        # @param [String] session_id The ID of an existing session. See also the
+        #   `create_session` param and {Job#session_id}.
         # @yield [job] a job configuration object
         # @yieldparam [Google::Cloud::Bigquery::QueryJob::Updater] job a job
         #   configuration object for setting additional options for the query.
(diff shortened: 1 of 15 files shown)


quartzmo:
acceptance test failure is a flake:

```

Google::Cloud::Bigquery::Dataset::ddl_dml::bigquery#test_0002_creates and populates and drops a table with ddl/dml queries:
Google::Cloud::PermissionDeniedError: Exceeded rate limits: too many table update operations for this table. For more information, see https://cloud.google.com/bigquery/docs/troubleshoot-quotas
```

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1435,16 +1453,52 @@ def routines token: nil, max: nil, filter: nil
...
+                      large_results: nil,
+                      flatten: nil,
+                      maximum_billing_tier: nil,
+                      maximum_bytes_billed: nil,
+                      job_id: nil,
+                      prefix: nil,
+                      labels: nil,
+                      udfs: nil,
+                      create_session: nil,
+                      session_id: nil
```
dazuma, line 1477: Dang, that really makes it clear how many arguments this method takes... 😳
I wonder if we should sort arguments by name in long lists like this.

quartzmo: I know... I hated to add to the list, but I feel these session args really deserve it. See the example for usage.

I'm totally open to sorting. Personally I really like alphabetical order. The established pattern in these libraries has been order of importance initially, then order in which each element was added, since the beginning.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1244,6 +1244,8 @@ def routines token: nil, max: nil, filter: nil
         #   Flattens all nested and repeated fields in the query results. The
         #   default value is `true`. `large_results` parameter must be `true` if
         #   this is set to `false`.
+        # @param [Integer] maximum_billing_tier Deprecated: Change the billing
+        #   tier to allow high-compute queries.
```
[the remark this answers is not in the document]
quartzmo: Yes, will do, thanks.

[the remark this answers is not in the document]
quartzmo: Updated.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
dazuma: Resolving for now, as we've opened https://github.com/googleapis/google-cloud-ruby/issues/16026 to further discussion.

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -1435,16 +1457,52 @@ def routines token: nil, max: nil, filter: nil
...
+                      legacy_sql: nil,
+                      large_results: nil,
+                      flatten: nil,
+                      maximum_billing_tier: nil,
+                      maximum_bytes_billed: nil,
+                      job_id: nil,
+                      prefix: nil,
+                      labels: nil,
+                      udfs: nil,
+                      create_session: nil,
```
dazuma, line 1480: Should this default to `false` rather than `nil`?

quartzmo: Great question! We *try* to never provide defaults (other than `nil`) for parameters that are pass-through in the client (not used in the client.) The reason for this is the same as not providing client-side validation for pass-through arguments. By relying on server-side default values and validations, we provide the same functionality to users (at the cost of an RPC call in the case of validation errors) while avoiding potential conflicts or breaking code changes in the client if the service behavior changes. I wrote "try" above because there are a few regrettable aberrations, such as `priority: "INTERACTIVE"` in this method.

However, I now see that I forgot to document the service default value for this param. I'll add it now. Thanks for the heads-up.

dazuma (approved):

shollyman (commented):
It looks like this change hides the connection properties and only exposes session_id as it's own property?  How much more work is it to expose the connection properties in a more general fashion?

quartzmo:
@shollyman Thanks for asking about this. Typically in the handwritten Ruby clients we try to simplify and flatten the API surface to make things easier for users by managing intermediate value objects that only have a field or two. The current list of connection properties, which is [described](https://cloud.google.com/bigquery/docs/reference/rest/v2/ConnectionProperty) as having only two simple properties (`time_zone` and `session_id`) is a candidate for this. Compare:

```rb
data = dataset.query "SELECT * FROM temptable", session_id: job.session_id
```

with

```rb
data = dataset.query "SELECT * FROM temptable", connection_properties: [{ "session_id": job.session_id }]
```

Composing this `connection_properties` array isn't too bad, but you might not understand from the param docs how it should look and need to locate a sample from which to copy/paste it. Historically we've tried to save users the trouble. Also, if we add `session_id` now, it doesn't prevent us from adding `connection_properties` in the future. The duplication would be especially acceptable if `session_id` is a popular argument that is commonly passed without the other properties.

It really depends on whether more properties will be added, and how many? Are `time_zone` and other possible future properties as likely as `session_id` to be passed with each and every `query` call? If so, I suppose real-world usage might look like this:

```rb
conn_props = [
  { "session_id": job.session_id },
  { "time_zone": time_zone },
  { "new_prop": "abc" }
]

data_a = dataset.query "SELECT * FROM temptable_a", connection_properties: conn_props
data_b = dataset.query "SELECT * FROM temptable_b", connection_properties: conn_props
data_c = dataset.query "SELECT * FROM temptable_c", connection_properties: conn_props
```

Given the examples above, let me know your guesses about the future and your preference!

shollyman:
RE: other properties, I know of a third one coming soon for associating queries to specific workload profiles.  So the number remains small, but is growing.

I think it's reasonable to just expose session now, and expand to a more general connectionproperties feature later, if there's no conflict/oddity with having both specific property and the general access to connection properties hung off the query config.

quartzmo:
@dazuma Any thoughts on this? Does it seem worth it to have a `session_id` keyword arg just for user convenience in addition to the harder-to-use array of hashes `connection_properties` param that will also be needed?

dazuma:
I think it makes sense to keep the separate `session_id` keyword argument, in keeping with our practice of flattening commonly used arguments like this. It doesn't preclude us from adding the more general argument in the future.

quartzmo:
@shollyman Does this sound good? If so, any other concerns on this PR?

meredithslota:
@quartzmo Seth is currently OOO but will reply shortly. :)

quartzmo:
Merging with guidance from @d-torres.

Issue #17988, opened 2022-04-07
google-cloud-trace does not work with rails 7


shouichi:
#### Environment details

  - OS: linux
  - Ruby version: 3.1.1
  - Gem name and version: 0.41.3

#### Steps to reproduce

  1. gem install rails
  2. configure google-cloud-trace
  3. send requests to rails server

grpc log shows that the library sends invalid traces.

```
D, [2022-04-07T06:00:42.994247 #1] DEBUG -- : calling cloudtrace.googleapis.com:/google.devtools.cloudtrace.v1.TraceService/PatchTraces
D, [2022-04-07T06:00:43.182028 #1] DEBUG -- : Failing with status #<struct Struct::Status code=3, details="Invalid timestamp: 54234-08-05 10:32:55.794433593 +0000 UTC", metadata={}, debug_error_string="{\"created\":\"@1649311243.181814787\",\"description\":\"Error received from peer ipv4:173.194.202.95:443\",\"file\":\"src/core/lib/surface/call.cc\",\"file_line\":1063,\"grpc_message\":\"Invalid timestamp: 54234-08-05 10:32:55.794433593 +0000 UTC\",\"grpc_status\":3}">
```

It seems that rails 7 introduced a braking change of `ActiveSupport::Notifications::Event`. Particularly, time from active support events are unix timestamp in millisecond, but google-cloud-trace assumes seconds.

- https://github.com/rails/rails/pull/43502
- https://github.com/googleapis/google-cloud-ruby/blob/c8369f10b2af8d1e294d8203f4fb8d9f4b7eedb0/google-cloud-trace/lib/google/cloud/trace/utils.rb#L35

Thanks.

#### Code example

N/A.

#### Full backtrace

N/A.

PR #17989, opened 2022-04-07
fix(google-cloud-trace): normalize span time based on rails version

shouichi:
Close https://github.com/googleapis/google-cloud-ruby/issues/17988.


shouichi:
@dazuma Could you take a look at this? Because of the problem, tracing is completely broken with rails 7.

NivedhaSenthil (approved):

PR #18018, opened 2022-04-15
edit error reporting service name

minherz:
Change "Cloud Error Reporting" service name to be "Error Reporting", so it will match the official name of the service.


quartzmo (approved):
LGTM

Issue #18270, opened 2022-05-14
google-cloud-bigtable: reads_prefix sample have invalid end_key


jychen7:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/google-cloud-ruby/issues
  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+ruby

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS:
  - Ruby version:
  - Gem name and version:

#### Steps to reproduce

  1. create bigtable rows
  ```
  phone#5c10102#20190502
  phone001#5c10102#20190502
  phong#5c10102#20190502
  ```
  2. [existing ruby code sample](https://cloud.google.com/bigtable/docs/reading-data#prefix) to reads prefix "phone#" returns 2 rows, but should expect 1 row only

#### Code example

```
prefix = "phone#"
range = table.new_row_range.between prefix, prefix.next
table.read_rows(ranges: range)
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

PR #18271, opened 2022-05-14
fix(google-cloud-bigtable): fix reads_prefix sample

jychen7:
closes: https://github.com/googleapis/google-cloud-ruby/issues/18270

#### before this PR

```ruby
prefix = "phone#"
=> "phone#"
prefix.next
=> "phonf#"
```

"phon001" is ordered between "phone#" and "phonf#", so it will wrongly include a lot rows not prefix by "phone#", e.g. "phone001#"

but actually it should expect end key `phone$`, align with [python example](https://cloud.google.com/bigtable/docs/reading-data#prefix).

#### after this PR
```
prefix[0...-1] + prefix[-1].next
=> "phone$"
```


bajajneha27 (approved):
LGTM

NivedhaSenthil (approved):

Issue #18286, opened 2022-05-18
# characters in object keys are not parsed correctly


camiloforero:
Linked is a fix

#### Environment details

  - OS: Ubuntu 22.04
  - Ruby version: 3.0.1
  - Gem name and version:  google-cloud-storage (1.36.1)

#### Steps to reproduce

  1. Upload a file to Google Cloud Storage, which name includes a `#`  character.
  2. Ask for that file's `signed_url`
  3. Put that url on a browser, it won't work, it will say the specified key does not exist

PR #18287, opened 2022-05-18
Patch 1

camiloforero:
The `Addressable::URI.escape` method tries to parse it's input as a full URI, the best it can. However, when a `#` character is included, it incorrectly treats everything before it as the path, and everything after it as a fragment.

As a consequence, the `#` character is not escaped properly, and the `signed_url` method will return a bad url with a `NoSuchKey` error.

Here, instead of using the generic `parse`, we specifically call `encode_component`, treating the path explicitly as a path. In this situation, `#` will be properly encoded as `%23`

closes: #18286

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file/signer_v2.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/signer_v2.rb
@@ -48,7 +48,7 @@ module Google
           #
           def ext_path
             path = "/#{@bucket}/#{@path}"
-            escaped = Addressable::URI.encode_component(path, Addressable::URI::CharacterClasses::PATH)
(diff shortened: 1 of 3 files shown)


bajajneha27:
Hi @camiloforero, thank you for taking time to code this out. This does make sense to me. Do you mind fixing the ci test failure? It's a simple change.
Also, I think we should write some test cases for this, please feel free to do so. If you want me to add tests, I can do that too.

camiloforero:
@bajajneha27 hey, sorry for the delay. I expect to have time to work on the CI test and the test cases tomorrow

camiloforero:
@bajajneha27 alright, it should be ready now

NivedhaSenthil (approved):
fix(storage): update object path parsing to handle hashes in them

Issue #18901, opened 2022-07-27
Out of date SDK auth documentation 


asabourin:
Hi team, just a heads up the documentation at https://googleapis.dev/ruby/google-cloud-storage/latest/file.AUTHENTICATION.html (and equivalent for other products) seems to be out of date regarding authentication via Cloud SDK credentials. We found thanks to the Python SDK docs at https://googleapis.dev/python/google-api-core/latest/auth.html that one needs to run `gcloud auth application-default login` instead of `gcloud auth login` to make local SDK credentials work.

This took us a little while to figure out so hopefully that can be updated easily for everyone else benefit!

Cheers,
Antoine

inspire22:
The Google::Cloud::Translate docs are also horribly out of date.  Looks like the new format (I also had to go read the python docs) is:

> 
> translate = Google::Cloud::Translate.translation_service
> res = translate.translate_text parent:'projects/project_number/locations/global', contents: ['hello, world'], target_language_code: 'la'
> result_text = res.translations.first.translated_text
> from = res.translations.firstdetected_language_code
> 

Kinda frustrating this keeps changing and it's so hard to find any up to date info.

NivedhaSenthil:
@inspire22 Please open a new issue with details for the documentation issue you are facing, so that it can be tracked. Feel free to create a PR with updates. 

Closing this issue as the auth documentation is updated now.

PR #19017, opened 2022-08-14
chore: Fix out of date SDK auth documentation

SandeepTuniki:
Fixes: #18901


dazuma (changes requested):
I can confirm that we should be using `gcloud auth application-default login` (which sets application default credentials for client libraries) instead of `gcloud auth login` (which just sets credentials for gcloud itself).

However, this file is a generated file, and you cannot edit it directly because the generator will just change it back. Instead, update the generator template here: https://github.com/googleapis/gapic-generator-ruby/blob/main/gapic-generator-cloud/templates/cloud/gem/authentication.erb

PR #19108, opened 2022-09-04
feat(storage): add support for non-idempotent operations

SandeepTuniki:
Fixes: b/232351537 (internal)

This PR implements support for non-idempotent retry operations for Storage. It sets the default value of `retries` to 0, and is unmodifiable by the user.


google-cloud-storage/conformance/README.md
```diff
@@ -15,3 +15,5 @@ The output directory is the same is the input directory, so after successful com
 ## Running the tests
 
 The conformance tests are included in the ordinary unit test suite. In the `google-cloud-storage` directory, run `rake test`.
+
+Because the conformance tests are dynamically generated at run time, working with them is more difficult than working with hand-written tests. If you need to execute one or more of these tests in isolation, you can do so by placing the `focus` keyword just above one of the calls to `define_method`. This will isolate a subset of the conformance tests.
```
bajajneha27, line 19: Can we please remove all conformance testing related changes from this PR? We've a separate PR for the same.

google-cloud-storage/lib/google/cloud/storage/service.rb
```diff
@@ -656,6 +656,7 @@ def delete_file_acl bucket_name, file_name, entity, generation: nil,
         # Returns Google::Apis::StorageV1::HmacKey.
         def create_hmac_key service_account_email, project_id: nil,
                             user_project: nil, options: {}
+          options[:retries] = 0
```
bajajneha27, line 659: Can we make it configurable? Give users option to make it retryable if they want. You can refer this [code](https://github.com/googleapis/google-cloud-ruby/pull/18834/files#diff-7b94a6bb6adeaa1a8d4052dc0895cb31ff917367272c55d49f8e21c87db78c5aR138) for the same.

bajajneha27: By default, it'll be non retryable. But users should have option to override it.

bajajneha27: I believe there's one more API which is non idempotent. Can we also include that?

bajajneha27 (changes requested):

PR #19111, opened 2022-09-04
[DO NOT MERGE] google-cloud-bigtable: read_rows return extra rows when DeadlineExceed during enumerate

jychen7:
this PR is just a reproducible step of https://github.com/googleapis/google-cloud-ruby/issues/19112

not reproducible in unit test, because we mock the response chunk there.
It is reproducible in integration test with a real Bigtable instance. I didn't test with Bigtable Emulator yet

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         include:
           - os: ubuntu-latest
             ruby: "2.6"
-            task: "--test"
+            task: "--test --samples-latest"
           - os: ubuntu-latest
             ruby: "2.7"
             task: "--test"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,6 @@ on:
 
 jobs:
   CI:
-    if: ${{ github.repository == 'googleapis/google-cloud-ruby' }}
     strategy:
       matrix:
         include:
--- a/google-cloud-bigtable/samples/acceptance/read_samples_test.rb
+++ b/google-cloud-bigtable/samples/acceptance/read_samples_test.rb
@@ -101,9 +101,10 @@ Column Family stats_summary
 \tconnected_cell: 1 @#{@timestamp}
 \tconnected_wifi: 1 @#{@timestamp}
 \tos_build: PQ2A.190405.004 @#{@timestamp}
+
 OUTPUT
 
-    assert_match expected, out
+    assert_equal expected, out
   end
 
   it "reads_rows_nil" do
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -22,13 +22,12 @@ on:
...
 jobs:
   CI:
-    if: ${{ github.repository == 'googleapis/google-cloud-ruby' }}
     strategy:
       matrix:
         include:
           - os: ubuntu-latest
             ruby: "2.6"
-            task: "--test"
+            task: "--test --samples-latest"
```
jychen7, line 30: enable samples acceptance test temperately

google-cloud-bigtable/samples/acceptance/read_samples_test.rb
```diff
@@ -101,9 +101,95 @@
...
+OUTPUT
+
+    assert_equal expected, out
+  end
+
+  it "reads_rows_empty" do
+    out, _err = capture_io do
+      reads_rows_empty bigtable_instance_id, @table_id
+    end
+    expected = <<~OUTPUT
```
jychen7, line 160: it is arguable whether `read_rows(keys: [])` should return empty or return all rows like `read_rows(keys: nil)`.

I thought it should return empty, but turn out all rows. (This behaviour trigger the bug in `reads_rows_sleep` below)

google-cloud-bigtable/samples/acceptance/read_samples_test.rb
```diff
@@ -101,9 +101,95 @@
...
+\tconnected_cell: 1 @#{@timestamp}
+\tconnected_wifi: 1 @#{@timestamp}
+\tos_build: PQ2A.190405.003 @#{@timestamp}
+
+Reading data for phone#4c410523#20190502:
+Column Family stats_summary
+\tconnected_cell: 1 @#{@timestamp}
+\tconnected_wifi: 1 @#{@timestamp}
+\tos_build: PQ2A.190405.004 @#{@timestamp}
+
```
jychen7, line 189: even though `reads_rows_sleep` only request two, turn out it actually returns more rows after `phone#4c410523#20190502`

https://github.com/googleapis/google-cloud-ruby/blob/535a7f85af1134ece4b84d46fe30b08ccf280f6c/google-cloud-bigtable/samples/read_samples.rb#L94-L97

bajajneha27:
Hi @jychen7 
Are you still working on the PR?

jychen7:
@bajajneha27 nope, this PR is just a reproducible step of https://github.com/googleapis/google-cloud-ruby/issues/19112, it is not a fix

PR #19134, opened 2022-09-08
feat(storage): add retry support for non-idempotent operations

SandeepTuniki:
Fixes: http://b/232351537 (internal)

This PR implements support for non-idempotent retry operations for Storage. It sets the default value of retries to 0, but gives the user option to override it.

Implements the retry option for the following operations:

- `storage.hmacKey.create`
- `storage.notifications.insert`
- `storage.bucket_acl.insert`
- `storage.bucket_acl.delete`
- `storage.default_object_acl.delete`
- `storage.default_object_acl.insert`
- `storage.object_acl.delete`
- `storage.object_acl.insert`

PR #19135, opened 2022-09-08
fix(storage): Correct options checks in retry operations

SandeepTuniki:
An earlier PR #18834 had a minor mistake where it was checking for `options[:retry]` instead of `options[:retries]` for Retry operations. @bajajneha27 and I noticed it during #19134. This PR fixes it.


bajajneha27 (approved):
Thank you!

SandeepTuniki:
The failing specs are unrelated to this PR. They're being fixed at #19147.

PR #19171, opened 2022-09-15
sample(storage): Add sample for configuring retries

diptanshumittal:


google-cloud-storage/samples/acceptance/retry_test.rb
```diff
@@ -0,0 +1,47 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative "helper"
+require_relative "../storage_configure_retries"
+
+describe "Retry Snippets" do
```
bajajneha27, line 18: Minor: can we rename it to `Retry samples` or may be something more descriptive?

Note: Snippets are usually the auto generated ones while samples are hand written code.

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
+# Copyright 2020 Google LLC
```
bajajneha27, line 1: Copyright year should be 2022

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+def configure_retries bucket_name: nil, file_name: nil
+  # The ID of your GCS bucket
+  # bucket_name = "your-unique-bucket-name"
+
+  # The ID of your GCS object
+  # file_name = "your-file-name"
+
+  require "google/cloud/storage"
+
+  # Customize retry with the maximum retry attempt of 5 (default=0).
```
bajajneha27, line 25: default is 3

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+  # Customize retry with a deadline of 500 seconds (default=900 seconds).
```
bajajneha27, line 26: Please verify the default values once.

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+  # Customize retry with a maximum wait time of 45.0 (default=60.0).
+  storage = Google::Cloud::Storage.new(
+    retries: 5,
+    max_elapsed_time: 500,
+    base_interval: 1.5,
+    max_interval: 45,
+    multiplier: 1.2
+  )
+
+  # file = storage.service.get_file bucket_name, filename uses the initialization RETRY_CONFIGURATION by default.
```
bajajneha27, line 38:
```suggestion
  Use the default retry configuration set during the client initialization
```

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+  file = storage.service.get_file bucket_name, file_name
+
+  # Maximum retry attempt can be overridden for each operation using options parameter.
+  storage.service.delete_file bucket_name, file_name, options: { retry: 0 }
```
bajajneha27, line 42:
```suggestion
  storage.service.delete_file bucket_name, file_name, options: { retries: 0 }
```

google-cloud-storage/samples/acceptance/retry_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:storage_client) { Google::Cloud::Storage.new }
+  let(:local_file) { File.expand_path "data/file.txt", __dir__ }
+  let(:remote_file_name) { "path/file_name_#{SecureRandom.hex}.txt" }
+  let(:downloaded_file) { "test_download_#{SecureRandom.hex}" }
```
bajajneha27, line 22: We're not using this variable anywhere.

google-cloud-storage/samples/acceptance/retry_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:bucket) { @bucket }
+  let(:file_content) { "some content" }
```
bajajneha27, line 24: We're not using this either

google-cloud-storage/samples/acceptance/retry_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  after do
+    bucket.files.each(&:delete)
+  end
+
+  it "configure_retries" do
+    bucket.create_file local_file, remote_file_name
+
+    assert_output "Deleted #{remote_file_name} with 0 retries\n" do
```
bajajneha27, line 41: We probably have to modify this. Let's discuss offline.

bajajneha27 (changes requested):

google-cloud-storage/samples/storage_configure_retries.rb
diptanshumittal: Found these lines in the code - 
'''
    RequestOptions.default.retries = 0
    RequestOptions.default.max_elapsed_time = 900
    RequestOptions.default.base_interval = 1
    RequestOptions.default.max_interval = 60
    RequestOptions.default.multiplier = 2 
'''
I was guessing these are the default values only

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,51 @@
...
+    base_interval: 1.5,
+    max_interval: 45,
+    multiplier: 1.2
+  )
+
+  # Use the default retry configuration set during the client initialization above with 5 retries
+  file = storage.service.get_file bucket_name, file_name
+
+  # Maximum retry attempt can be overridden for each operation using options parameter.
+  # We can disable the retry by setting the retries to 0.
```
tritone, line 42: Other language samples show using a RetryAlways policy so that the retry will occur even if the operation is non-idempotent. Could you show the same for Ruby, rather than disabling retries?

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,51 @@
...
+def configure_retries bucket_name: nil, file_name: nil
+  # The ID of your GCS bucket
+  # bucket_name = "your-unique-bucket-name"
+
+  # The ID of your GCS object
+  # file_name = "your-file-name"
+
+  require "google/cloud/storage"
+
+  # Customize retry with the maximum retry attempt of 5 (default=3).
```
tritone, line 25: You don't have to give all the default values here -- they will be in a separate section above the sample in the docs. See https://cloud.google.com/storage/docs/retry-strategy#tools

diptanshumittal: This sample was developed using python sample as reference. And they have mentioned the default values in the comments. So should I still remove the default values ?

tritone: I would still say remove them, they are not present for other languages such as Go and Java. I think the python sample was developed before the full section was written which includes the table above the sample.

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+    base_interval: 1.5,
+    max_interval: 45,
+    multiplier: 1.2
+  )
+
+  # Uses the default retry configuration set during the client initialization above with 5 retries
+  file = storage.service.get_file bucket_name, file_name
+
+  # Maximum retry attempt can be overridden for each operation using options parameter.
+  storage.service.delete_file bucket_name, file_name, options: { retries: 4 }
```
tritone, line 42: Just want to confirm, does this force retries even if the operation is non-idempotent? @bajajneha27

tritone (approved):

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+def configure_retries bucket_name: nil, file_name: nil
+  # The ID of your GCS bucket
+  # bucket_name = "your-unique-bucket-name"
+
+  # The ID of your GCS object
+  # file_name = "your-file-name"
+
+  require "google/cloud/storage"
+
+  # Customize retry with the maximum retry attempt of 5 (default=3).
```
shaffeeullah, line 25: Can these comments be interspersed with the storage client being defined? So users know which comment belongs to which line. Node example of doing this here: https://github.com/googleapis/nodejs-storage/blob/HEAD/samples/configureRetries.js

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+  # Customize retry with a maximum wait time of 45.0 (default=60.0).
+  storage = Google::Cloud::Storage.new(
+    retries: 5,
+    max_elapsed_time: 500,
+    base_interval: 1.5,
+    max_interval: 45,
+    multiplier: 1.2
+  )
+
+  # Uses the default retry configuration set during the client initialization above with 5 retries
```
shaffeeullah, line 38: This wording is a little confusing. This isn't actually the default configuration, right? This is the customized configuration that will be automatically be applied to all operations? If I understand correctly, the default retry configuration would apply if the client was initialized with no retry specification.

google-cloud-storage/samples/storage_configure_retries.rb
```diff
@@ -0,0 +1,50 @@
...
+  file = storage.service.get_file bucket_name, file_name
+
+  # Maximum retry attempt can be overridden for each operation using options parameter.
+  storage.service.delete_file bucket_name, file_name, options: { retries: 4 }
+  puts "Deleted #{file.name} from bucket #{bucket_name}"
```
shaffeeullah, line 43: Since this is a retry sample, this line should also mention the retry configuration. https://github.com/googleapis/nodejs-storage/blob/1d5c8af2bb25f4936dbd799376e4e59edda64505/samples/configureRetries.js#L66-L83

shaffeeullah (changes requested):

google-cloud-storage/samples/storage_configure_retries.rb
diptanshumittal: Got it. Removed the default value.

google-cloud-storage/samples/storage_configure_retries.rb
diptanshumittal: Yes, you are right. I think that the "default" word was causing confusion. I've updated the comment accordingly.

google-cloud-storage/samples/storage_configure_retries.rb
diptanshumittal: Mapped the comments to the code line they belong to.

shaffeeullah (approved):
Thank you!

google-cloud-storage/samples/storage_configure_retries.rb
bajajneha27: Yes, this will override the default retry behaviour and retry regardless of idempotency.

bajajneha27 (approved):

PR #19215, opened 2022-09-23
docs: Add executable example

POPPIN-FUMI:
Hi, I added executable example code.
I hope it helps some ppl.
- [ x] Update code documentation if necessary.


dazuma (commented):
Thanks for your submission! Unfortunately, these READMEs are generated by our client code generator, and they can't be edited manually (or at least, if they are edited, the generator would just overwrite your changes the next time it runs.) If you're interested in contributing some sample code, you can put them in a `samples` directory (similar to what we're doing in some of the other client libraries such as [google-cloud-secret_manager](https://github.com/googleapis/google-cloud-ruby/tree/main/google-cloud-secret_manager/samples).)

PR #19350, opened 2022-10-28
feat(bigquery): Added support for tags in dataset

diptanshumittal:


google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
```diff
@@ -466,6 +467,25 @@ def access
...
+        #
+        # @return [Google::Cloud::Bigquery::Dataset::Tag] The list of tags.
+        #
+        def tags
+          ensure_full_data!
+          return nil if @gapi.tags.nil?
+          tags = []
+          @gapi.tags.each do |gapi|
+            tags.append Tag.from_gapi(gapi)
+          end
```
SandeepTuniki, line 485: You can use `map` method instead of a `loop + append`. Makes the code shorter, easier to read, and avoids mutation.

i.e; replace this: 👇
```rb
tags = []
@gapi.tags.each do |gapi|
  tags.append Tag.from_gapi(gapi)
end
tags
```
with this: 👇
```rb
tags = @gapi.tags.map { |gapi| Tag.from_gapi(gapi) }
```

SandeepTuniki (changes requested):
Its good overall. Requested a  minor change

google-cloud-bigquery/lib/google/cloud/bigquery/dataset.rb
diptanshumittal: Thanks a lot!! will update the code.

bajajneha27 (approved):
LGTM overall.
Nit: can we add unit tests for Tag class as well?

diptanshumittal:
> LGTM overall. Nit: can we add unit tests for Tag class as well?

Added unit test for Tag class

dazuma (approved):

PR #19456, opened 2022-11-17
fix(datastore): fixed request limit

diptanshumittal:
Maximum number of entities returned in a single batch for any query is 300. 
So incase of result limit > 300 in a query, results had to be fetched in multiple batches and we were carrying the same result limit across batches. 
This resulted in entire result set being fetched.

### DIFF
--- a/google-cloud-datastore/acceptance/datastore/datastore_test.rb
+++ b/google-cloud-datastore/acceptance/datastore/datastore_test.rb
@@ -402,12 +402,6 @@ describe Google::Cloud::Datastore::Dataset, :datastore do
       [rickard, eddard, catelyn, arya, sansa, robb, bran, jon]
     end
 
-    let(:post) do
-      Google::Cloud::Datastore::Entity.new.tap do |e|
-        e["title"]       = "How to make the perfect pizza in your grill"
-      end
-    end
-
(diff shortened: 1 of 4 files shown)


diptanshumittal:
> @diptanshumittal is this same codepath when Gql query is parsed?

After the initial batch, further batches of query and gql query follow the same codepath.

google-cloud-datastore/acceptance/datastore/datastore_test.rb
```diff
@@ -424,6 +430,47 @@
...
+      entities_count = 0
+      results = dataset.run query
+      loop do
+        entities_count += results.count
+        break unless results.next?
+        results = results.next
+      end
+      _(entities_count).must_equal limit
+
+      # Testing limit with GQL query
```
NivedhaSenthil, line 455: Can we split this into a separate test ?
It is generally a good practice to assert for one scenario in unit test, will help pin pointing the issue easier in future and also help understanding the functionality.

diptanshumittal: Splitted the test into 2 parts.

NivedhaSenthil (approved):

PR #19457, opened 2022-11-17
feat(firestore): Support query count for Firestore

SandeepTuniki:
This PR implements support for Firestore's Aggregate Query Count feature.


google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
```diff
@@ -0,0 +1,203 @@
...
+        # @private The Google::Cloud::Firestore::V1::StructuredQuery object.
+        attr_reader :query
+
+        ##
+        # @private Array of Google::Cloud::Firestore::V1::StructuredAggregationQuery::Aggregation objects
+        attr_reader :aggregates
+
+        ##
+        # @private Creates a new AggregateQuery
+        def initialize query, aggregates, parent_path, client
```
bajajneha27, line 79: Just a thought: can we prefer keyword arguments over positional arguments?

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
```diff
@@ -0,0 +1,203 @@
...
+        #   end
+        #
+        def get
+          ensure_service!
+
+          return enum_for :get unless block_given?
+
+          results = service.run_aggregate_query @parent_path, structured_aggregation_query
+          results.each do |result|
+            next if result.result.nil?
```
bajajneha27, line 149: Nit: can we change the `results` to something else?

google-cloud-firestore/lib/google/cloud/firestore/query.rb
```diff
@@ -940,6 +941,26 @@ def get
...
+        #   firestore = Google::Cloud::Firestore.new
+        #
+        #   # Get a collection reference
+        #   query = firestore.col "cities"
+        #
+        #   # Create an aggregate query
+        #   aggregate_query = query.aggregate_query
+        #
+        def aggregate_query
+          AggregateQuery.new query, [], parent_path, client
```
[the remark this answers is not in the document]
SandeepTuniki: Good point. I added a factory method in `AggregateQuery` class to be used by the `Query` class. This will avoid exposing the internal property `aggregates`.

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
SandeepTuniki: I renamed `results` to `responses`.

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
SandeepTuniki: Looking through the rest of the codebase of firestore, using positional arguments seems to the norm for constructors (except for situations which have default values). Adding an example below for one of the existing classes.

I followed the existing convention for this new class.
https://github.com/googleapis/google-cloud-ruby/blob/55d181cdcaba45e1725d0161cd5631922df9647d/google-cloud-firestore/lib/google/cloud/firestore/query.rb#L78-L86

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
```diff
@@ -0,0 +1,209 @@
...
+
+        ##
+        # @private Array of Google::Cloud::Firestore::V1::StructuredAggregationQuery::Aggregation objects
+        attr_reader :aggregates
+
+        ##
+        # @private Creates a new AggregateQuery
+        def initialize query, aggregates, parent_path, client
+          aggregates ||= []
+          @query = query
```
bajajneha27, line 81:
```suggestion
          @query = query
          @parent_path = parent_path
          @aggregates = aggregates || []
          @client = client
```

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
bajajneha27: How about we make `aggregates` as optional since we've a default value over there too?

bajajneha27 (approved):

google-cloud-firestore/lib/google/cloud/firestore/aggregate_query.rb
SandeepTuniki: Optional arguments need to be moved to the end of the method signature 😐. I feel `parent_path` & `client` should be at the end.

For now, I converted `aggregates` as a keyword parameter. I can change it later if there's a better pattern for this.

PR #20608, opened 2023-03-03
chore: update language acceptance tests to include REST

viacheslav-rostovtsev:


dazuma (approved):
LGTM.

Fine with changing from spec style to unit style. However, note that, if spec style expectations are causing problems, more recent versions of minitest/spec are less intrusive in terms of adding methods to core classes. (There's a small syntax difference: `expect(foo).must_equal bar` instead of `foo.must_equal bar`.) It's also possible to keep "describe" and "it" blocks from spec style, and mix them with assertions from unit style (and in fact this is kind of our preferred style for most tests in this repo). So you could do that and have less to change for other libraries.

PR #21444, opened 2023-04-14
chore(scheduler): update region tag typo

meredithslota:
This sample has a duplicate region tag (https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-scheduler/samples/create_job.rb) because of a typo; this sample actually _deletes_ a job, so updating the region tag to match.

### DIFF
--- a/google-cloud-scheduler/samples/app.rb
+++ b/google-cloud-scheduler/samples/app.rb
@@ -16,7 +16,7 @@
 
 $stdout.sync = true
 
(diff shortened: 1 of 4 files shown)


meredithslota:
Ok, looks like these should have the prefix `cloudscheduler_*` instead of `cloud_scheduler_*`. I'll update a few more accordingly.

dazuma (approved):

Issue #22896, opened 2023-09-06
Unnecessary warning introduced by recent PR


spickermann:
This recent https://github.com/googleapis/google-cloud-ruby/pull/22866 enabled setting the `endpoint` configuration option for  `Google::Cloud.bigquery` by an `ENV["BIGQUERY_EMULATOR_HOST"]` environment variable. 

At it first glance, this change shouldn't have caused any issues, because when `ENV["BIGQUERY_EMULATOR_HOST"]` is not set, it still returns `nil` like the previous default. But the validator works slightly different when `add_field!` is called with a `Google::Cloud::Config.deferred` block as `initial` value instead of a simple `nil` – even if the block returns `nil` when called.

Previously, when the initial value for `endpoint` was explicitly set to `nil` in [this line](https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-bigquery/lib/google-cloud-bigquery.rb#L142), `config.add_field!` accepted `nil` as a valid value, although the validator was configured to `match: String`.

But with the change from [the PR](https://github.com/googleapis/google-cloud-ruby/pull/22866) `nil` is not considered valid anymore and, the following warning is outputted when the config is loaded:

```
Invalid value nil for key :endpoint. Setting anyway. at /.../gems/google-cloud-bigquery-1.44.0/lib/google-cloud-bigquery.rb:142:in `block in <main>'
```

Which basically means when `ENV["BIGQUERY_EMULATOR_HOST"]` is not set and the config should just use the default, it still would use the default but would always complain about that with a warning – which is a bit annoying.

As far as I can tell, allowing backwards compatibility with the previous behavior that didn't trigger this warning could be achieved by simply changing [line #142 in `google-cloud-ruby/blob/main/google-cloud-bigquery/lib/google-cloud-bigquery.rb`](https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-bigquery/lib/google-cloud-bigquery.rb#L142) from

```ruby
config.add_field! :endpoint, default_endpoint, match: String
```

to

```ruby
config.add_field! :endpoint, default_endpoint, match: String, allow_nil: true
```

I am happy to create a PR for this fix, if that helps you. What do you think?

alex-damian-negru:
Can confirm that this is a very annoying warning & a fix would be greatly appreciated. 🙂

SandeepTuniki:
Thanks for spotting this, @spickermann.

[...]

PR #22904, opened 2023-09-07
fix(bigquery): remove unnecessary warning 

spickermann:
Hi there,

this PR removes the unnecessary `Invalid value nil for key :endpoint. Setting anyway.` warning when the optional env variable `BIGQUERY_EMULATOR_HOST` (which was introduced in https://github.com/googleapis/google-cloud-ruby/pull/22866) is not set.

Allowing `nil` in this case, enables backward compatibility with previous versions which set `nil` explicitly and thereby triggered falling back to the default configuration without any warning message.

closes: #22896


dazuma (approved):

Capncavedan:
thank you @spickermann

PR #23732, opened 2024-01-05
feat(storage): support for object lock / retention

### LINKED ISSUE #23484: storage: add support for object retention
Please implement the object retention feature. The following operations must be supported:

- Create a bucket with object retention enabled by using the `enableObjectRetention` bucket create boolean option at bucket create time
- Get the object retention enabled status of a bucket
- Create an object with retention enabled (on an retention enabled bucket) by populating the `retention` object at object create time (retention object has the fields `mode` which can be Unlocked or Locked, and `retainUntilTime` which is a timestamp)
- Update retention for an existing object with a patch operation, by passing the `overrideUnlockedRetention` boolean patch option
- Get the retention policy of an object
- Remove retention for an existing object the same way as above

Details in go/gcs-object-retention-client-request

You can use the [java implementation](https://github.com/googleapis/java-storage/pull/2277) as a reference.

The kokoro test project should be allowlisted shortly, let me know if you need access to an allowlisted project for your own testing. 

We're aiming to get this in by the November code freeze, let me know if you need any help.

shivgautam:
Fixes #23484


google-cloud-storage/lib/google/cloud/storage/project.rb
```diff
@@ -416,7 +416,8 @@ def create_bucket bucket_name,
                           versioning: nil,
                           requester_pays: nil,
                           user_project: nil,
-                          autoclass_enabled: false
+                          autoclass_enabled: false,
+                          enable_object_retention: nil
```
SandeepTuniki, line 420: Can you document this param (and necessary example snippets) in the YARD comments for this method?

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -865,7 +865,8 @@ def update generation: nil,
                    if_generation_match: nil,
                    if_generation_not_match: nil,
                    if_metageneration_match: nil,
-                   if_metageneration_not_match: nil
+                   if_metageneration_not_match: nil,
+                   override_unlocked_retention: nil
```
SandeepTuniki, line 869: Same comment here about YARD as mentioned in `Project#create_bucket()`

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -1560,6 +1562,46 @@ def delete generation: nil,
...
+          @gapi.retention
+        end
+
+        ##
+        # Update method to update retention parameter of an object / file
+        # It accepts params as a Hash of attributes in the following format:
+        #
+        #     { mode: 'Locked|Unlocked', retain_until_time: '2023-12-19T03:22:23+00:00' }
+        #
+        # @param [Hash(String => String)] new_retention_attributes
```
SandeepTuniki, line 1596: [Optional Nit] This is better read as `@param [Hash(String => String)] new_retention_attributes`.  Has an added benefit of separating parameter names from its description, if we update it in the future.

google-cloud-storage/test/helper.rb
```diff
@@ -114,7 +116,13 @@ def billing_hash(requester_pays)
     { "requesterPays" => requester_pays} unless requester_pays.nil?
   end
 
-  def random_file_hash bucket=random_bucket_name, name=random_file_path, generation="1234567890", kms_key_name="path/to/encryption_key_name", custom_time: nil
+  def file_retention_hash(retention_params)
```
SandeepTuniki, line 119: [Optional Nit] I'd recommend specifying the method declaration as `def file_retention_hash(mode: nil, retain_until_time: nil)`. This ensures the users of this helper method wouldn't need to scan through the method body to know what keys the hash needs to contain.

google-cloud-storage/lib/google/cloud/storage/file.rb
```diff
@@ -1560,6 +1562,46 @@ def delete generation: nil,
...
+        # be deleted or replaced.
+        #
+        # @return [DateTime]
+        def retention_retain_until_time
+          @gapi.retention&.retain_until_time
+        end
+
+        # A collection of object level retention parameters.
+        # The full list of available options are outlined at the [JSON API docs]
+        # (https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request-body).
```
SandeepTuniki, line 1583: Recommend using the curly braces syntax (i.e; `{...}`) for hyperlinks in YARD. See https://rubydoc.info/gems/yard/file/docs/GettingStarted.md#linking-urls-http

SandeepTuniki (changes requested):

google-cloud-storage/lib/google/cloud/storage/file.rb
shivgautam: I don't see a difference. What am I missing?

google-cloud-storage/test/helper.rb
shivgautam: This is for a test helper. The actual code already has this mentioned (in file.rb)

google-cloud-storage/lib/google/cloud/storage/file.rb
shivgautam: I'd generally agree but this file already has this format in multiple places. For e.g.: https://github.com/googleapis/google-cloud-ruby/pull/23732/files/213f6f9bf7a47c7bf9576aa3bcc9bd958682883b#diff-2788eaf7ebb8b45d5b11379afd02ffe326b411ad9d0841f5f0de195073c6869dR1414-R1417

google-cloud-storage/lib/google/cloud/storage/project.rb
shivgautam: Done.

google-cloud-storage/lib/google/cloud/storage/file.rb
SandeepTuniki: YARD treats both of them as the same (hence, the optional nit).

However, the below format makes it easy to separate the parameter from it's description:

```
@param arg_name [datatype] long description text
```

In contrast, the below makes it harder to spot parameter name from description:
```
@param [datatype] arg_name long description text
```

shivgautam: Hmm. Makes sense. But this file has other places with the same format.  For e.g.: https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-storage/lib/google/cloud/storage/file.rb#L359-L366

google-cloud-storage/lib/google/cloud/storage/project.rb
SandeepTuniki: Can you also add some `@example` snippets in the YARD comments?

shivgautam: Done.

SandeepTuniki (approved):
LGTM

Issue #23476, opened 2023-10-26
BigQuery: Add support dataset's storageBillingModel


tamayika:
Dataset does not support to set or change storageBillingModel.
StorageBillingModel is important to save storage cost.
ref: https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets

NivedhaSenthil:
It can be set as shown in the code below.

 ```ruby
  bigquery.create_dataset "dataset_id" do |dataset|
     dataset.gapi.storage_billing_model= "LOGICAL"
  end
```

**Comment by tamayika** [comment]
Thank you for the answer.
But I want to update dataset's storage_billing_model not when creating.
`patch_gapi` is protected, so I can not call the method.

**Comment by SandeepTuniki** [comment]
@tamayika This feature is now available in bigquery gem's 1.48.0 version.

Sample usage:
```rb
require "google/cloud/bigquery"

bigquery = Google::Cloud::Bigquery.new
dataset = bigquery.dataset "my_dataset"

# set the storage_billing_model
dataset.storage_billing_model = "LOGICAL"

# get the storage_billing_model
puts datasset.storage_billing_model
```

PR #24755, opened 2024-02-06
feat(bigquery): Support accessing storage_ billing_model field in a dataset

SandeepTuniki:
closes: #23476

**Note**: Don't merge until the internal discussion around the behaviour of default value is resolved.

### DIFF
--- a/google-cloud-bigquery/test/google/cloud/bigquery/dataset_reference_update_test.rb
+++ b/google-cloud-bigquery/test/google/cloud/bigquery/dataset_reference_update_test.rb
@@ -123,4 +123,23 @@ describe Google::Cloud::Bigquery::Dataset, :reference, :update, :mock_bigquery d
     _(dataset.default_encryption).must_be :frozen?
     mock.verify
   end
+
+  it "updates its storage_billing_model" do
+    storage_billing_model = "LOGICAL"
+
+    mock = Minitest::Mock.new
+    bigquery.service.mocked_service = mock
+    mock.expect :get_dataset, dataset_gapi, [project, dataset_id]
+    updated_gapi = dataset_gapi.dup
+    updated_gapi.storage_billing_model = storage_billing_model
+    patch_gapi = Google::Apis::BigqueryV2::Dataset.new storage_billing_model: storage_billing_model, etag: dataset_gapi.etag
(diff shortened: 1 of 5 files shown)


google-cloud-bigquery/test/google/cloud/bigquery/dataset_reference_update_test.rb
```diff
@@ -123,4 +123,23 @@
...
+
+  it "updates its storage_billing_model" do
+    storage_billing_model = "LOGICAL"
+
+    mock = Minitest::Mock.new
+    bigquery.service.mocked_service = mock
+    mock.expect :get_dataset, dataset_gapi, [project, dataset_id]
+    updated_gapi = dataset_gapi.dup
+    updated_gapi.storage_billing_model = storage_billing_model
+    patch_gapi = Google::Apis::BigqueryV2::Dataset.new storage_billing_model: storage_billing_model, etag: dataset_gapi.etag
```
shivgautam, line 135: since we are doing a dataset_gapi.dup for updated_gapi, it should already contain the etag (which you are setting to the same value). why do we need to set it explicitly

google-cloud-bigquery/test/google/cloud/bigquery/dataset_update_test.rb
```diff
@@ -128,4 +128,22 @@
...
   end
+
+  it "updates its storage_billing_model" do
+    storage_billing_model = "LOGICAL"
+
+    mock = Minitest::Mock.new
+    bigquery.service.mocked_service = mock
+    updated_gapi = dataset_gapi.dup
+    updated_gapi.storage_billing_model = storage_billing_model
+    patch_gapi = Google::Apis::BigqueryV2::Dataset.new storage_billing_model: storage_billing_model, etag: dataset_gapi.etag
```
shivgautam, line 139: same as above. why are we adding etag again?

google-cloud-bigquery/test/google/cloud/bigquery/dataset_reference_update_test.rb
SandeepTuniki: Yes, `updated_gapi` contains etag, but `patch_gapi` is reconstructed from constructor. And `patch_gapi` is what is sent to the backend. So we need to pass it here during initialization.

I tried to remove it, the but the test failed (due to the above reason).

shivgautam (approved):

PR #25391, opened 2024-03-14
fix: fix read rows retry so it doesn't trigger a full table scan in t…

mutianf:
…he end

When a read is done and received a retryable error at the end of the stream, the current implementation will trigger a full table scan because the row resumption logic will strip away all of seen keys & ranges, leaving the retry request empty, which causes a full table scan.

### DIFF
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
@@ -159,10 +159,10 @@ module Google
           rescue *RowsReader::RETRYABLE_ERRORS => e
             rows_reader.retry_count += 1
             raise Google::Cloud::Error.from_error(e) unless rows_reader.retryable?
-            retry_status = rows_reader.retry_options limit, row_set
-            rows_limit = retry_status.rows_limit
-            row_set = retry_status.row_set
-            retry if retry_status.should_retry
+            resumption_option = rows_reader.retry_options limit, row_set
+            rows_limit = resumption_option.rows_limit
+            row_set = resumption_option.row_set
+            retry if resumption_option.is_complete
           end
         end
 
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
@@ -159,8 +159,10 @@ module Google
           rescue *RowsReader::RETRYABLE_ERRORS => e
             rows_reader.retry_count += 1
             raise Google::Cloud::Error.from_error(e) unless rows_reader.retryable?
-            rows_limit, row_set = rows_reader.retry_options limit, row_set
-            retry
+            retry_status = rows_reader.retry_options limit, row_set
+            rows_limit = retry_status.rows_limit
+            row_set = retry_status.row_set
+            retry if retry_status.should_retry
           end
         end
 
@@ -318,6 +320,11 @@ module Google
           if row_ranges
             row_ranges = [row_ranges] unless row_ranges.instance_of? Array
             row_set[:row_ranges] = row_ranges.map(&:to_grpc)
+          end
+
+          # Set the row range to full table scan if the row set is empty
+          if row_set.empty?
+            return Google::Cloud::Bigtable::V2::RowSet.new row_ranges: []
           end
 
           Google::Cloud::Bigtable::V2::RowSet.new row_set
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
@@ -324,7 +324,7 @@ module Google
 
           # Set the row range to full table scan if the row set is empty
           if row_set.empty?
-            return Google::Cloud::Bigtable::V2::RowSet.new row_ranges: []
+            row_set[:row_ranges] = [Google::Cloud::Bigtable::V2::RowRange.new]
           end
 
           Google::Cloud::Bigtable::V2::RowSet.new row_set
--- a/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
+++ b/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
@@ -162,7 +162,7 @@ module Google
             resumption_option = rows_reader.retry_options limit, row_set
             rows_limit = resumption_option.rows_limit
             row_set = resumption_option.row_set
-            retry if resumption_option.is_complete
+            retry unless resumption_option.complete?
           end
         end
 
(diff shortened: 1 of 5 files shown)


google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -116,19 +117,33 @@ def last_key
         #   If not specified, reads from all rows.
         #   A hash of the same form as `Google::Cloud::Bigtable::V2::RowSet`
         #   can also be provided.
-        # @return [Integer, Google::Cloud::Bigtable::V2::RowSet]
+        # @return RetryStatus
         #
         def retry_options rows_limit, row_set
-          return [rows_limit, row_set] unless last_key
+          return RetryStatus.new true, rows_limit, row_set unless last_key
```
mutianf, line 123: rename to ResumptionOptions

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -116,19 +117,33 @@ def last_key
...
+
+          # Check if we've already read read rows_limit number of rows.
+          # If true, return nil to indicate that the read has succeeded.
+          return RetryStatus.new false, nil, nil if rows_limit == @rows_count
```
mutianf, line 127: need to also check rows_limit != 0

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -186,13 +205,50 @@ def start_key_read? range
...
+        def initialize is_complete, rows_limit, row_set
+          @is_complete = is_complete
+          @rows_limit = rows_limit
+          @row_set = row_set
+        end
 
-          end_key && end_key <= last_key
+        ##
+        # returns if this operation should be retried
+        def is_complete
```
meagar, line 238: More idiomatic to call this `complete?`, ie

```
def complete?
  @is_complete
end
```

google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
```diff
@@ -159,8 +159,10 @@ def read_rows keys: nil, ranges: nil, filter: nil, limit: nil, &block
           rescue *RowsReader::RETRYABLE_ERRORS => e
             rows_reader.retry_count += 1
             raise Google::Cloud::Error.from_error(e) unless rows_reader.retryable?
-            rows_limit, row_set = rows_reader.retry_options limit, row_set
-            retry
+            resumption_option = rows_reader.retry_options limit, row_set
+            rows_limit = resumption_option.rows_limit
+            row_set = resumption_option.row_set
+            retry if resumption_option.is_complete
```
meagar, line 165: This seems backwards, wouldn't we want to retry if the read is _not_ complete?

mutianf: got catch, I renamed the variable name from should_retry to is_complete but didnt update all the logic :) fixed it now.

google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb
```diff
@@ -320,6 +322,11 @@ def build_row_set row_keys, row_ranges
             row_set[:row_ranges] = row_ranges.map(&:to_grpc)
           end
 
+          # Set the row range to full table scan if the row set is empty
+          if row_set.empty?
+            return Google::Cloud::Bigtable::V2::RowSet.new row_ranges: []
```
igorbernstein2, line 327: Does this actually work? In other clients we needed to insert an infinite range and then would check if row_ranges is empty. I dont quite understand what having an empty row range list gives you here

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -137,19 +152,23 @@ def retry_options rows_limit, row_set
               end
             end
 
-            delete_indexes.each { |i| row_set.row_ranges.delete_at i }
+            delete_indexes.sort.reverse_each { |i| row_set.row_ranges.delete_at i }
```
igorbernstein2, line 155: why do you need a sort here?

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -170,13 +189,13 @@ def retryable?
...
-                      end
-
-          start_key.empty? || last_key >= start_key
+          if range.start_key_closed.empty? && range.end_key_open.empty?
+            true
+          elsif !range.start_key_closed.empty?
+            last_key >= range.start_key_closed
+          else
+            last_key > range.start_key_open
+          end
```
igorbernstein2, line 198: I think it would cleaner to structure this as:
```
if !range.start_key_closed.empty?
    last_key >= range.start_key_closed
elseif !range.start_key_open.empty?
  last_key > range.start_key_closed
else
  # start is unbounded
  true
end
```

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -186,13 +205,50 @@ def start_key_read? range
...
-                      range.end_key_closed
-                    end
+          if range.end_key_closed.empty? && range.end_key_open.empty?
+            false
+          elsif !range.end_key_closed.empty?
+            range.end_key_closed <= last_key
+          else
+            range.end_key_open <= last_key
+          end
+        end
```
igorbernstein2, line 215: similar as above, please structure it in terms of field presence

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
```diff
@@ -186,13 +205,50 @@ def start_key_read? range
...
+        # returns new new rows_limit for the retry
+        def rows_limit
+          @rows_limit
+        end
+
+        ##
+        # returns new new row_set for the retry
+        def row_set
+          @row_set
         end
```
igorbernstein2, line 252: I think the ruby idiomatic way fora getter is `attr_reader :rows_limit`

igorbernstein2 (changes requested):

google-cloud-bigtable/lib/google/cloud/bigtable/rows_reader.rb
mutianf: Here we delete by index while updating the set. Sorting helps us deletes from the end instead of from the beginning, which would mess up the index we want to delete

igorbernstein2: But the indices are sorted already? that array is populated from `.each_with_index` which will generate indices in ascending order?

Having said that, I think what you want is:
```
row_set.row_ranges.reject! {|r| end_key_read? r }

```

igorbernstein2: Also I think we are missing the bit where the last_key happens to be in the middle of a range?

igorbernstein2: oops github folded over the elseif :(

igorbernstein2 (approved):

PR #25440, opened 2024-03-26
feat: Initial generation of google-apps-chat-v1

aandreassa:


aandreassa:
Being done with the fix in [gapic-generator-ruby](https://github.com/googleapis/gapic-generator-ruby/pull/1061) later on.

PR #25456, opened 2024-03-28
chore: updating blunderbuss config for default assignee

shivgautam:


aandreassa (approved):
Thanks Shiv! I am adding myself too.
Long-term we will switch to 'googleapis/ruby-team' after we fix the group membership.

Issue #23528, opened 2023-11-08
storage: Support custom endpoints for SignURLs


BrennaEpp:
When a custom endpoint is set on the client, Sign URLs and Post Policies should automatically use that domain.

Optional: allow users to set a custom hostname for Sign URL and Post policy methods.

See the following for more information and fill out language section: go/signurl-custom-endpoint-gcs
Please add a comment or ping me if you have any questions.

PR #25469, opened 2024-04-03
feat(storage): Respect custom endpoint for signed_url

bajajneha27:
When a custom endpoint is set on the client, Sign URLs and Post Policies automatically use that domain.

closes: #23528


BrennaEpp (commented):
LGTM but I'll let Ruby folk approve

dazuma (approved):
LGTM

PR #25680, opened 2024-04-13
feat: Add a test proxy for Bigtable client conformance tests

meagar:
Introduce a test proxy required for Bigtable client conformance tests, per https://github.com/googleapis/cloud-bigtable-clients-test, as well as some small (I believe non-breaking) changes in `google-cloud-bigtable` to allow the test proxy to work.

The test harness is written in Go, and provides its own (very simple) mock implementation of Bigtable. The test proxy exposes a number of RPC calls that are thin wrappers around the language-specific client library, so the test harness can tell the proxy to make various permutations of typical Bigtable calls (read row, check-and-mutate, etc) and verify the client's behavior by looking at how it interacts with the mock Bigtable implementation, and the types of responses the proxy returns.

This PR also includes a couple minor fixes to `google-cloud-bigtable` which I think generally update the code to reflect the existing documentation comments (so they require no documentation change).

It's worth mentioning that I tried to fail early with clear error messages in any place in the proxy that the current test suite didn't cause to be executed, and I didn't feel comfortable authoring an implementation that I'd have no way of testing. This should provide good feedback for future maintainers to understand where to make changes, should new tests trip my asserts/aborts.

closes: #25682


google-cloud-bigtable/lib/google/cloud/bigtable/mutation_operations.rb
```diff
@@ -245,7 +245,7 @@ def check_and_mutate_row key, predicate, on_match: nil, otherwise: nil
           response = service.check_and_mutate_row(
             path,
             key,
-            predicate_filter: predicate.to_grpc,
+            predicate_filter: predicate&.to_grpc,
```
meagar, line 248: It seemed important to be able to pass a null/no-op value here, and I couldn't figure out how to do it, except via something very hacky like

```ruby
foo = {}
def foo.to_grpc
  nil
end

table.check_and_mutate_row(key, foo, ...)
```

google-cloud-bigtable/lib/google/cloud/bigtable.rb
```diff
@@ -163,6 +163,7 @@ def self.new_with_emulator project_id, emulator_host, timeout
       def self.resolve_credentials given_credentials, scope
         credentials = given_credentials || default_credentials(scope: scope)
         return credentials if credentials.is_a? Google::Auth::Credentials
+        return credentials if credentials.is_a? GRPC::Core::Channel
```
meagar, line 166: I couldn't figure out how to start an insecure connection to Bigtable, via this high-level entry point to the library. The underlying `Service` object supports it, but the point of the test proxy is to use the language-specific SDK idiomatically. If there's an existing way to create an insecure channel that I missed, please let me know.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,449 @@
...
+    LOGGER.info "ReadRows(#{req.to_h})"
+
+    # At the time of writing, no tests set this field; rather than guess at an implementation,
+    # this is a note for future maintainers to implement this if the field becomes set in the test suite
+    assert req.request.app_profile_id == "", "An app profile ID was specified when non was expected"
+
+    # TODO: req.request.reverse is not supported by the client library; when it is, fix this
+    result = Client.get(req.client_id)
+                   .table(req.request.table_name)
+                   .read_rows(keys: req.request&.rows&.row_keys&.to_a)
```
mutianf, line 206: looks like rows limit is not passed in? And also row ranges? Maybe this is why the TestReadRows_NoRetry_ErrorAfterLastRow failed?

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,449 @@
...
+
+      # Every test uses these values
+      assert project_id == "project", "Unxpected value for `project`"
+      assert instance_id == "instance", "Unexpected value for `instance`"
+
+      # Used later in Client#table
+      self.instance_id = instance_id
+      self.app_profile_id = app_profile_id
+
+      timeout = if per_operation_timeout
```
mutianf, line 104: Is the timeout not configurable for different operations?

meagar: Wouldn't that be working around the intent of the tests? The tests specify the timeout once when the connection is created, and the Ruby client library _claims_ to support setting a default timeout via the `connect` method; if the Ruby library doesn't then honor the timeout, I think that we've exposed a bug, that should be fixed in one place so all the deadline tests start passing.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,449 @@
...
+  # rpc ReadRow(ReadRowRequest) returns (RowResult) {}
+  def read_row req, _call
+    LOGGER.info "ReadRow(#{req.to_h})"
+
+    result = Client.get(req.client_id)
+                   .table(req.table_name)
+                   .read_row(req.row_key)
+
+    if result.nil?
+      # TODO(meagar): Is this the right response?
```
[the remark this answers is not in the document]
meagar: What is the "default" status? When I create a new status and don't supply any values, I get back a status with a code of `0`, which I think is equivalent to "OK"? 

I just tried using that and the tests still pass, so I guess the actual status isn't important.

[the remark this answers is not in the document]
mutianf: > I get back a status with a code of 0, which I think is equivalent to "OK"?

Yes. we can probably just return OK status to be consistent with other test-proxy implementations.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,449 @@
...
+
+  # Reads rows with the client instance.
+  # rpc ReadRow(ReadRowRequest) returns (RowResult) {}
+  # rubocop:disable Metrics/AbcSize
+  def read_rows req, _call
+    LOGGER.info "ReadRows(#{req.to_h})"
+
+    # At the time of writing, no tests set this field; rather than guess at an implementation,
+    # this is a note for future maintainers to implement this if the field becomes set in the test suite
+    assert req.request.app_profile_id == "", "An app profile ID was specified when non was expected"
```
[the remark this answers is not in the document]
mutianf: It is set here in the test: https://github.com/googleapis/cloud-bigtable-clients-test/blob/main/tests/readrows_test.go#L70, which end up setting app profile in createCbtClient here: https://github.com/googleapis/cloud-bigtable-clients-test/blob/main/tests/test_workflow.go#L59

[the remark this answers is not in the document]
meagar: In that specific test, which uses an app profile ID of `test_profile`, what I see from the proxy is this:

Screenshot 2024-04-18 at 2 48 13 PM

The `test_profile` value comes in during CreateClient, and I _do_ set that value aside, and use it to build the client. The `ReadRows` requests all uniformly seem to have `app_profile_id: ""`, so my assertion just conveys the idea that, if any `ReadRows` requests start to specify an app profile, the tests will need to handle this.

google-cloud-bigtable/lib/google/cloud/bigtable.rb
dazuma: Thanks for this. We'll figure out how to fix the rest of the cases later. There is actually a common pattern in the Ruby clients by which passing the symbol `:this_channel_is_insecure` can be used for credentials when creating an insecure channel, but it looks like that code path is not implemented here either.

google-cloud-bigtable/lib/google/cloud/bigtable/mutation_operations.rb
dazuma: Thanks. LGTM.

google-cloud-bigtable/test/test_proxy/.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.2
```
dazuma, line 1: What system uses this file? I'm guessing it's not needed, and even if it's currently being used by something, we'd rather the test infrastructure control the Ruby version(s) rather than pinning it here.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,449 @@
...
+
+    def self.remove id
+      @clients_lock.synchronize do
+        assert(@clients.key?(id), "Client #{id} didn't exist in Client#remove()")
+        @clients.delete id
+      end
+    end
+
+    def close
+      # no-op?
```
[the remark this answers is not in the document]
dazuma: There isn't a "close" function on the client library itself, but you could just implement a simple mechanism here, perhaps by setting `@bigtable = nil` to sever the connection to the client library's object.

google-cloud-bigtable/test/test_proxy/.ruby-version
meagar: This is for rbenv; I can remove it.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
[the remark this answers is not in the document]
mutianf: Ok make sense!

google-cloud-bigtable/test/test_proxy/test_proxy.rb
```diff
@@ -0,0 +1,439 @@
...
+    LOGGER.info "ReadRows(#{req.to_h})"
+
+    # At the time of writing, no tests set this field; rather than guess at an implementation,
+    # this is a note for future maintainers to implement this if the field becomes set in the test suite
+    assert req.request.app_profile_id == "", "An app profile ID was specified when non was expected"
+
+    # TODO: req.request.reverse is not supported by the client library; when it is, fix this
+    result = Client.get(req.client_id)
+                   .table(req.request.table_name)
+                   .read_rows(keys: req.request&.rows&.row_keys&.to_a, limit: req.request.rows_limit)
```
mutianf, line 196: Looks like row ranges and filters are not passed in? https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-bigtable/lib/google/cloud/bigtable/read_operations.rb#L149 

Filter is used in reverse scan test. Even though reverse scan is not supported by ruby right now, not passing in the filter here may cause false negative in the future and make it hard to debug. Row Ranges is used in RowsOutOf order test.

mutianf (approved):

google-cloud-bigtable/test/test_proxy/.ruby-version
dazuma: Please do. Thanks!

google-cloud-bigtable/test/test_proxy/test_proxy.rb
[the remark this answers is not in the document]
dazuma: Okay. If this is a requirement for the bigtable client, please file an issue.

google-cloud-bigtable/test/test_proxy/test_proxy.rb
meagar: Fixed in https://github.com/googleapis/google-cloud-ruby/pull/25680/commits/c075534a0c7d41a40231fe4273c8f4518efaf55b

dazuma (approved):

dazuma:
The conventional commit message should't be `feat:` since this doesn't implement a feature in the library itself. Actually, it probably should be `fix:` since we've included several small bug fixes to the library in addition to test infrastructure.

dazuma:
fix(bigtable): Fixes for using a preconfigured GRPC channel as credentials
fix(bigtable): Fixes for leaving predicate unset in check_and_mutate_row

Issue #26134, opened 2024-06-20
Add a method to big_query load_job to set the "column name character map" option


juseveno:
When you use big_query/load_job to load a CSV file with autodetect schema, the default character map is used to convert column names.

There is currently no method to set this option as in the GCP interface : 

![Capture d’écran 2024-06-20 à 11 16 51](https://github.com/googleapis/google-cloud-ruby/assets/4319193/26e8e881-97f6-4ca3-9c76-24b974f7c919)

It would be useful to allow this option in the ruby library.

PR #26135, opened 2024-06-20
feat(bigquery): add method in load_job to set column name character map

juseveno:
closes: [#<issue_number_goes_here>](https://github.com/googleapis/google-cloud-ruby/issues/26134)


dazuma (changes requested):

dazuma (approved):
Looks great to me! Thanks!

PR #26249, opened 2024-06-28
chore(toys): Add last breaking change info to analyze-clients tool

aandreassa:
Adding last breaking change age to the prerelease analysis.

I am using a bit of a broad regexp for breaking change since it seems to change every so often (`google-cloud-dataproc-v1/CHANGELOG.md` has a different format than something like `google-maps-fleet_engine-v1`).

An example output by the script now:
`google-maps-fleet_engine-v1 0.2.1 (56 days) !!! LAST BREAKING CHANGE: 43 days ago !!!`


aandreassa:
There are only 3 gems that would show any diffs from this check: google-cloud-ai_platform-v1, google-maps-fleet_engine-v1, google-maps-fleet_engine-delivery-v1

dazuma (approved):

PR #26363, opened 2024-07-09
chore: Remove owlbot post-processor for non-snake case path variables

aandreassa:
Generator changes were caught up: https://github.com/googleapis/google-cloud-ruby/pull/26355

This should be a no-op as well.


dazuma (approved):
🎉

PR #26537, opened 2024-07-29
Fix google-cloud-translate README Authentication link

swrobel:
Just a README update, so hopefully this isn't a big deal


aandreassa:
Hi @swrobel, thanks for sending this PR! I wasn't aware of that broken reference there. 

Unfortunately, we rely on generated changes upstream, and this would get reverted automatically. I filed an issue in https://github.com/googleapis/gapic-generator-ruby/issues/1095 to address it there.

Once that is done, it should resolve the link in `google-cloud-translate` and other libraries, such as `google-cloud-app_hub`. I will close this one out.

PR #27388, opened 2024-09-30
chore: add tests for rest error wrapping

viacheslav-rostovtsev:
Error wrapping was tested for GRPC but not REST. This PR adds REST-specific test


dazuma (approved):
Thanks!

PR #27486, opened 2024-10-24
feat: Initial generation of google-ads-ad_manager-v1

aandreassa:
Adding a manual .owlbot.rb for removing 'paths tests' as the service is REST only.

`toys new-library google/ads/admanager/v1 --bootstrap-releases --pull -i`


aandreassa:
I will open a new PR once the namespace is fixed.

PR #27507, opened 2024-10-25
feat(storage): restore soft deleted object using restore token for HNS buckets

shubhangi-google:
Add support for restore token to soft deleted from HNS bucket.

Operation Supported:

- If a restore token is provided as an input to restore_file method along with flename and generation token, it will restore seamlessly.
- If restore token is provided, the method calls works as earlier filename and generation token.
- Invalid restore tokens will result in a 404 Not Found error from the Storage Transfer API.

```ruby
 @example
bucket = storage.bucket "my-hns-bucket"
bucket.restore_file "path/of/file", <generation-of-the-file>, "<restore_token>"
```

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -1843,6 +1843,8 @@ module Google
         #   Name of the file.
         # @param [Fixnum] generation
         #   Selects a specific revision of this object.
+        # @param [String] restore_token
+        #   UUID generated specificly for softdeleted objects of HNS bucket
         # @param [Boolean] copy_source_acl
         #   If true, copies the source file's ACL; otherwise, uses the
         #   bucket's default file ACL. The default is false.
--- a/google-cloud-storage/lib/google/cloud/storage/bucket.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/bucket.rb
@@ -1882,6 +1882,7 @@ module Google
         #
         def restore_file file_path,
                          generation,
+                         restore_token: nil,
                          copy_source_acl: nil,
                          if_generation_match: nil,
                          if_generation_not_match: nil,
@@ -1895,6 +1896,7 @@ module Google
           gapi = service.restore_file name,
                                       file_path,
                                       generation,
+                                      restore_token: restore_token,
                                       copy_source_acl: File::Acl.predefined_rule_for(copy_source_acl),
                                       if_generation_match: if_generation_match,
                                       if_generation_not_match: if_generation_not_match,
--- a/google-cloud-storage/acceptance/storage/bucket_soft_delete_test.rb
+++ b/google-cloud-storage/acceptance/storage/bucket_soft_delete_test.rb
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 require "storage_helper"
-require "pry"
 
 describe Google::Cloud::Storage::Bucket, :soft_delete, :storage do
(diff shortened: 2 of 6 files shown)


google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -1843,6 +1843,8 @@ def create_file file,
         #   Name of the file.
         # @param [Fixnum] generation
         #   Selects a specific revision of this object.
+        # @param [String] restore_token
```
bajajneha27, line 1846: I believe we need to `restore_token` for both `get` and `list` objects operations. Refer to go/hns-bucket-softdelete-support-client-request

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: Hi @bajajneha27 this is already handled we are able to fetch restore token in get and list request
```ruby
@example
bucket.file "try_sub_a/flowers.jpeg", soft_deleted: true, generation: 1729579373155129
=> 
#<Google::Cloud::Storage::File:0x00007f1e2591a7b8
 @gapi=
  #<Google::Apis::StorageV1::Object:0x00007f1e267afdc8
   @bucket="shubhangi_hns5",
   @content_type="image/jpeg",
   @crc32c="1Nykzg==",
   @etag="CLmml6GxoYkDEAE=",
   @generation=1729579373155129,
   @hard_delete_time=#<DateTime: 2024-11-06T12:43:53+00:00 ((2460621j,45833s,248000000n),+0s,2299161j)>,
   @id="shubhangi_hns5/try_sub_a/flowers.jpeg/1729579373155129",
   @kind="storage#object",
   @md5_hash="wISfTGSx0F0r5A+2uS5Hxw==",
   @metageneration=1,
   @name="try_sub_a/flowers.jpeg",
   @restore_token="d3e6275a-46c1-4d65-9f8f-4774e3bf421c",
   @self_link="https://www.googleapis.com/storage/v1/b/shubhangi_hns5/o/try_sub_a%2Fflowers.jpeg",
   @size=11899,
   @soft_delete_time=#<DateTime: 2024-10-30T12:43:53+00:00 ((2460614j,45833s,248000000n),+0s,2299161j)>,
   @storage_class="STANDARD",
   @time_created=#<DateTime: 2024-10-22T06:42:53+00:00 ((2460606j,24173s,245000000n),+0s,2299161j)>,
   @time_storage_class_updated=#<DateTime: 2024-10-22T06:42:53+00:00 ((2460606j,24173s,245000000n),+0s,2299161j)>,
   @updated=#<DateTime: 2024-10-22T06:42:53+00:00 ((2460606j,24173s,245000000n),+0s,2299161j)>>,
 @service=Google::Cloud::Storage::Service(storage-sdk-vendor),

```

bajajneha27: We need to make sure that `file.restore_token` returns the token. We also need to make sure same thing happens for list operation.

shubhangi-google: will add the check in unit test

bajajneha27: Thank you!

JesseLovelace:
We decided not to add this feature to the client libraries, and to direct customers to use the CLI or console instead

PR #27681, opened 2024-11-27
feat(bigquery): load jobs with dest tables in another project

### LINKED ISSUE #27368: Run Load Job in a project that loads data in a table in a different project
This is related to [this](https://github.com/googleapis/google-cloud-ruby/issues/2609) closed issue. 

The problem has been resolved for the #extract, #extract_job, #copy and #copy_job methods, but it's still not possible to create a load job on project A that writes to a table in project B, since the #load_job method of the Client class operates through a Table object that is obtained from the project specified in the client object itself. 

Now, I'm not totally sure the Google BigQuery API actually allows this (which would explain why the problem has not been fixed for the #load_job method). In that case, please let me know and close this ticket.

alvarowolfx:
This PR favors the usage of the `dataset.project_id` method instead of `@service.project` to allow Load jobs to target tables in a dataset that lives in a separated project, other than the project usage on the main client and which is used for billing.

A follow up PR needs to be address more scenarios where we should favor that same usage of the `dataset.project_id` instead of `@service.project`, since is a common use case where operations are run in project A to modify data stored in project B.

I haven't added integration tests, since we would not have write access to two different projects in our CI pipelines. But I tested locally with this given code:

```
# $ export GOOGLE_CLOUD_PROJECT=projectA
```

```ruby

project_id = "projectB"

dataset = bigquery.dataset dataset_id, project_id: project_id # <- this is the important change
gcs_uri  = "gs://cloud-samples-data/bigquery/us-states/us-states.csv"
table_id = "us_states"

load_job = dataset.load_job table_id, gcs_uri, skip_leading: 1 do |schema|
  schema.string "name"
  schema.string "post_abbr"
end
puts "Starting job #{load_job.job_id}"

load_job.wait_until_done! # Waits for table load to complete.
puts "Job finished."

table = dataset.table table_id
puts "Loaded #{table.rows_count} rows to table #{table.id}"
```

Fixes #27368

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
@@ -1373,11 +1373,11 @@ module Google
         # Retrieves an existing dataset by ID.
         #
         # @param [String] dataset_id The ID of a dataset.
-        # @param [String] project_id The GCP Project where the dataset lives.
         # @param [Boolean] skip_lookup Optionally create just a local reference
         #   object without verifying that the resource exists on the BigQuery
         #   service. Calls made on this object will raise errors if the resource
         #   does not exist. Default is `false`. Optional.
+        # @param [String] project_id The GCP Project where the dataset lives.
         #
         # @return [Google::Cloud::Bigquery::Dataset, nil] Returns `nil` if the
         #   dataset does not exist.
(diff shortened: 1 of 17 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/project.rb
```diff
@@ -1372,6 +1373,7 @@ def external url, format: nil
         # Retrieves an existing dataset by ID.
         #
         # @param [String] dataset_id The ID of a dataset.
+        # @param [String] project_id The GCP Project where the dataset lives.
```
quartzmo, line 1376: Should this param come in order of appearance in the method signature, after `skip_lookup`?

quartzmo (approved):

alvarowolfx:
blocked by #27684

dazuma (approved):
This looks good. Thanks much!

Issue #27527, opened 2024-10-29
Firestore: unexpected results when combining `start_at` or `end_at` with `limit_to_last` in queries


sarub0b0:
When combining `start_{at,after}` or `end_{at,before}` with `last_to_limit` in a Firestore query, `start_at` and `end_at` are not applied properly, returning unexpected results.

#### Environment details

  - OS: macOS 14.6.1
  - Ruby version: 3.3.4
  - Gem name and version: google-cloud-firestore 2.16.0

#### Steps to reproduce

  1. Create documents in the `data` collection with an `id` field
       ```ruby
        require "google/cloud/firestore"

        firestore = Google::Cloud::Firestore.new

        10.times do |i|
          doc = firestore.collection("data").doc
          doc.set({ id: i })
        end
       ```
  3. Run the sample code using `end_before` and `limit_to_last`
  4. An empty array is returned

#### Code example

```ruby
# example
require 'google/cloud/firestore'

firestore = Google::Cloud::Firestore.new

col = firestore.col 'data'

query = col.order(:id, :asc).end_before(10).limit_to_last(5)

pp query

docs = query.get

pp docs.map(&:data)
```

#### Full backtrace

```sh
❯ GOOGLE_CLOUD_PROJECT=PROJECT_ID ruby sample.rb
#<Google::Cloud::Firestore::Query:0x0000000121db92b0
 @client=#<Google::Cloud::Firestore::Client:0x00000001220530e8 @service=Google::Cloud::Firestore::Service(PROJECT_ID)((default))>,
 @limit_type=:last,
 @parent_path="projects/PROJECT_ID/databases/(default)/documents",
 @query=
  <Google::Cloud::Firestore::V1::StructuredQuery: from: [<Google::Cloud::Firestore::V1::StructuredQuery::CollectionSelector: collection_id: "data", all_descendants: false>], order_by: [<Google::Cloud::Firestore::V1::StructuredQuery::Order: field: <Google::Cloud::Firestore::V1::StructuredQuery::FieldReference: field_path: "id">, direction: :DESCENDING>], start_at: <Google::Cloud::Firestore::V1::Cursor: values: [<Google::Cloud::Firestore::V1::Value: integer_value: 10>], before: false>, end_at: <Google::Cloud::Firestore::V1::Cursor: values: [<Google::Cloud::Firestore::V1::Value: integer_value: 10>], before: true>, offset: 0, limit: <Google::Protobuf::Int32Value: value: 5>>>
[...]

PR #28137, opened 2024-12-19
fix(firestore): correct cursor flipping in limit_to_last

sarub0b0:
* Fix Google::Cloud::Firestore::Query#limit_to_last

Ensure that cursors are correctly flipped when using limit_to_last in Firestore queries. This change moves the assignment of new_start_at and new_end_at outside of the conditional blocks to ensure they are always set correctly. Added tests to cover scenarios with order, start_at, and end_at in combination with limit_to_last.

closes: #27527


dazuma (approved):
This looks great, thank you!

PR #25340, opened 2024-03-08
feat(storage): add support for soft delete

bajajneha27:
Add support for soft delete.

The operations we need are:

- Set a soft delete policy on a existing bucket
- Get a bucket's soft delete policy
- Get a soft deleted object by doing an object get call with the soft deleted parameter set to true
- Get the soft delete time and hard delete time metadata fields from a soft deleted object
- List soft deleted objects by doing an object list call with the soft deleted parameter set to true
- Restore a soft deleted object

closes: #24793

PR #28760, opened 2025-01-27
sample(storage): Soft deleted Bucket Restore

shubhangi-google:
Add support for restoring soft deleted bucket.

**Operation Supported:**
- Get Bucket Generation
- Get a Soft Deleted Bucket (Also soft-delete time and hard-delete time)
- List Soft Deleted Buckets
- Restore a Soft Deleted Bucket
```ruby
@example
bucket.delete
##fetch bucket generation
 generation= bucket.generation
## list soft_deleted buckets
deleted_buckets= storage.buckets soft_deleted: true
## restore bucket
bucket = storage.restore_bucket "my-bucket", generation
```


shubhangi-google:
Please ignore ci issues in this PR for now, This Pr is for tracking sample failures
we will update the PR when feature Pr is merged

google-cloud-storage/samples/acceptance/buckets_test.rb
```diff
@@ -119,6 +121,43 @@
     end
   end
 
+  describe "storage_soft_deleted_bucket" do
+    let(:new_bucket_name) { random_bucket_name }
+
+    it "get soft deleted bucket, its soft_delete_time and hard_delete_time" do
+      new_bucket = storage_client.create_bucket new_bucket_name
```
bajajneha27, line 128: Does this new bucket have `soft_delete_policy` ?
If not, we won't be able to recover the bucket.

shubhangi-google: @bajajneha27  on console I can see this for one of the ruby sample test bucket
image

bajajneha27: Is the `soft_delete_policy` set by default if you create a new bucket?

If not, you'll have to pass it while creating the new bucket.

shubhangi-google: yes this is set by default

bajajneha27: Hmmm. Can you explicitly check the policy in test too, so that we're sure we've the policy in place.

shubhangi-google: @bajajneha27  tried to set retention period for the bucket explicitly but getting same result
also I can see in this doc (https://cloud.google.com/storage/docs/soft-delete#retention-duration) that default soft deleted policy is set as 7 days for a bucket

bajajneha27: @shubhangi-google , the retention_period that you tried to set in this test is not related to soft delete policy, it's related to retention_policy. These are two different things. See [here](https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-storage/lib/google/cloud/storage/bucket.rb#L769)

I think you can try 

- Setting soft delete policy explicitly, to make sure the bucket has soft deletion enabled. This [PR](https://github.com/googleapis/google-cloud-ruby/pull/25340/files) has example of how you can set soft_delete_policy.
- Check if soft delete policy is actually set.
- Then try fetch the soft deleted bucket.

shubhangi-google: still getting same issue @bajajneha27 

image

google-cloud-storage/samples/acceptance/buckets_test.rb
```diff
@@ -119,6 +121,44 @@
...
+      _(new_bucket.soft_delete_policy.retention_duration_seconds).must_equal 10*24*60*60
+      puts storage_client.service_account_email
+
+      # ensuring bucket is created
+      assert new_bucket.exists?, "Bucket #{new_bucket_name} should exist"
+
+      # fetching bucket
+      check_bucket = storage_client.bucket new_bucket_name
+      puts "new bucket name-- #{check_bucket.name}"
+      puts "new bucket soft_delete_policy-- #{check_bucket.soft_delete_policy}"
```
bajajneha27, line 141: `soft_delete_policy` is `nil` ? 

If the policy is `nil` then that means it's not a soft deleted bucket. You need to make sure of that first.

shubhangi-google: soft deleted policy is not coming as nil
see the log
**new bucket soft_delete_policy-- #<Google::Apis::StorageV1::Bucket::SoftDeletePolicy:0x00007fc1c733a4b0**
we have also checked in test case for the presence of soft deleted policy here https://github.com/googleapis/google-cloud-ruby/pull/28760/files#diff-038ec9ce9ebab1eb9b210c6691a04076bd29214c57b349651b409951341ab908R132

bajajneha27: > we have also checked in test case for the presence of soft deleted policy here https://github.com/googleapis/google-cloud-ruby/pull/28760/files#diff-038ec9ce9ebab1eb9b210c6691a04076bd29214c57b349651b409951341ab908R132

That's for `new_bucket`. Since you just added the policy to the same object, it'll show the value .



> soft deleted policy is not coming as nil see the log **new bucket soft_delete_policy-- #<Google::Apis::StorageV1::Bucket::SoftDeletePolicy:0x00007fc1c733a4b0** 

I don't see this in kokoro samples log. Sending the screenshot offline.

bajajneha27: You can try two things:

1. Enable the DEBUG log to see if your soft_delete_policy is actually being set via the HTTP call. [Sending the command offline]
2. Try setting the soft_delete_policy using the object:
```
soft_delete_policy = Google::Apis::StorageV1::Bucket::SoftDeletePolicy.new
soft_delete_policy.retention_duration_seconds = 10*24*60*60
bucket.soft_delete_policy = soft_delete_policy
```

google-cloud-storage/samples/acceptance/helper.rb
```diff
@@ -25,6 +25,20 @@
 require "ostruct"
 
 
+def grant_storage_permissions bucket_name:
+  storage = Google::Cloud::Storage.new
+  bucket = storage.bucket bucket_name
+
```
cpriti-os: this is not required, remove it.

aandreassa:
@shubhangi-google please reopen the PR if needed. Closing for now.

PR #28815, opened 2025-01-31
docs(storage): updating the encryption key GCS sample

shubhangi-google:
Updating encryption key method in  code sample to avoid confusion


google-cloud-storage/samples/storage_upload_encrypted_file.rb
```diff
@@ -24,7 +24,7 @@ def upload_encrypted_file bucket_name:, local_file_path:, file_name: nil, encryp
   # file_name = "your-file-name"
 
   # The key to encrypt the object with
-  # encryption_key = "TIbv/fjexq+VmtXzAlc63J4z5kFmWJ6NdAPQulQBT7g="
+  # encryption_key = "your-enncryption-key"
```
bajajneha27, line 27: I don't think this alone clarifies that the user can send a simple string. We also need to clarify it in comment like we discussed.

google-cloud-storage/samples/storage_upload_encrypted_file.rb
```diff
@@ -23,8 +23,12 @@ def upload_encrypted_file bucket_name:, local_file_path:, file_name: nil, encryp
   # The ID of your GCS object
   # file_name = "your-file-name"
 
-  # The key to encrypt the object with
-  # encryption_key = "TIbv/fjexq+VmtXzAlc63J4z5kFmWJ6NdAPQulQBT7g="
+
+  # The key used to encrypt the object can be a 32-byte key (raw encrypted data).
+  # This example is provided for demonstration purposes only; it is recommended to generate your own encryption key.
```
bajajneha27, line 28: Can you please clearly mention that they don't need to encode the key.

bajajneha27, line 28: We don't need to write this. Samples, in general, are for their reference

bajajneha27 (approved):

bajajneha27:
There's a CI failure. It doesn't look like it's related to our changes. try re-running it, to begin with.
In general, make sure your CI jobs are passing because we cannot merge the PR without that.

bajajneha27 (approved):

PR #29070, opened 2025-02-12
feat(firestore): Update required version to Ruby 3.0

aandreassa:
feat!: Updated google-cloud-firestore-v1 dependency to gem version 2.x


dazuma (approved):
Looks fine. Please merge with `feat:` for the Ruby version update, and also add `feat!: Updated google-cloud-firestore-v1 dependency to gem version 2.x` (note the `feat!` because it is a breaking change).

PR #29215, opened 2025-02-27
feat(storage): Restart delete resumable upload config

shubhangi-google:
```Ruby
@example   

storage = Google::Cloud::Storage.new upload_chunk_size: chunk_size
#Initiating Resumable upload 
  bucket = storage.bucket bucket_name
 # Restarting  Resumable upload 
  storage = Google::Cloud::Storage.new(upload_chunk_size: chunk_size, upload_url: upload_url)
# Deleting Resumable upload 
  storage = Google::Cloud::Storage.new(upload_chunk_size: chunk_size, upload_url: upload_url, delete_upload: true)

  bucket.create_file file, file_name

```

### DIFF
--- a/google-cloud-storage/lib/google-cloud-storage.rb
+++ b/google-cloud-storage/lib/google-cloud-storage.rb
@@ -142,7 +142,7 @@ module Google
     def self.storage project_id = nil, credentials = nil, scope: nil,
                      retries: nil, timeout: nil, open_timeout: nil, read_timeout: nil, send_timeout: nil,
                      max_elapsed_time: nil, base_interval: nil, max_interval: nil, multiplier: nil,
-                     upload_chunk_size: nil
+                     upload_chunk_size: nil, upload_url: nil, delete_upload: nil
       require "google/cloud/storage"
       Google::Cloud::Storage.new project_id: project_id,
                                  credentials: credentials,
@@ -156,7 +156,9 @@ module Google
                                  base_interval: base_interval,
                                  max_interval: max_interval,
                                  multiplier: multiplier,
-                                 upload_chunk_size: upload_chunk_size
+                                 upload_chunk_size: upload_chunk_size,
+                                 upload_url: upload_url,
+                                 delete_upload: delete_upload
     end
   end
 end
@@ -193,4 +195,6 @@ Google::Cloud.configure.add_config! :storage do |config|
   config.add_field! :upload_chunk_size, nil, match: Integer
   config.add_field! :endpoint, nil, match: String, allow_nil: true
   config.add_field! :universe_domain, nil, match: String, allow_nil: true
+  config.add_field! :upload_url, nil, match: String, allow_nil: true
+  config.add_field! :delete_upload, nil, match: Integer, allow_nil: true
 end
--- a/google-cloud-storage/lib/google/cloud/storage.rb
+++ b/google-cloud-storage/lib/google/cloud/storage.rb
@@ -94,7 +94,9 @@ module Google
                    timeout: nil, open_timeout: nil, read_timeout: nil,
                    send_timeout: nil, endpoint: nil, project: nil, keyfile: nil,
                    max_elapsed_time: nil, base_interval: nil, max_interval: nil,
-                   multiplier: nil, upload_chunk_size: nil, universe_domain: nil
+                   multiplier: nil, upload_chunk_size: nil, universe_domain: nil,
+                   upload_url: nil, delete_upload: nil
+
         scope             ||= configure.scope
         retries           ||= configure.retries
         timeout           ||= configure.timeout
@@ -109,7 +111,8 @@ module Google
         multiplier        ||= configure.multiplier
         upload_chunk_size ||= configure.upload_chunk_size
         universe_domain   ||= configure.universe_domain
-
+        upload_url        ||= configure.upload_url
+        delete_upload     ||= configure.delete_upload
         unless credentials.is_a? Google::Auth::Credentials
           credentials = Storage::Credentials.new credentials, scope: scope
         end
@@ -125,6 +128,7 @@ module Google
             host: endpoint, quota_project: configure.quota_project,
             max_elapsed_time: max_elapsed_time, base_interval: base_interval,
             max_interval: max_interval, multiplier: multiplier, upload_chunk_size: upload_chunk_size,
+            upload_url: upload_url, delete_upload: delete_upload,
             universe_domain: universe_domain
           )
         )
--- a/google-cloud-storage/lib/google/cloud/storage/service.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/service.rb
@@ -49,7 +49,9 @@ module Google
                        timeout: nil, open_timeout: nil, read_timeout: nil,
                        send_timeout: nil, host: nil, quota_project: nil,
                        max_elapsed_time: nil, base_interval: nil, max_interval: nil,
-                       multiplier: nil, upload_chunk_size: nil, universe_domain: nil
+                       multiplier: nil, upload_chunk_size: nil, universe_domain: nil,
+                       upload_url: nil, delete_upload: nil
+
           host ||= Google::Cloud::Storage.configure.endpoint
           @project = project
           @credentials = credentials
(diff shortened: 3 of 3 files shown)


bajajneha27:
> #Initiating Resumable upload 
>   bucket = storage.bucket bucket_name

How is this initiating the resumable upload?

> #Restarting  Resumable upload 
  storage = Google::Cloud::Storage.new(upload_chunk_size: chunk_size, upload_url: upload_url)

Again, how is this resuming the upload and why do I need to pass the `upload_url` in the client initialization ?

> #Deleting Resumable upload 
  storage = Google::Cloud::Storage.new(upload_chunk_size: chunk_size, upload_url: upload_url, delete_upload: true)

I don't think we should be passing `upload_url` or `delete_upload` in the client initialization. They should rather be either separate functions or parameters for `create_file` or equivalent method of uploading an object/file.

google-cloud-storage/lib/google-cloud-storage.rb
```diff
@@ -193,4 +195,6 @@ def self.storage project_id = nil, credentials = nil, scope: nil,
   config.add_field! :upload_chunk_size, nil, match: Integer
   config.add_field! :endpoint, nil, match: String, allow_nil: true
   config.add_field! :universe_domain, nil, match: String, allow_nil: true
+  config.add_field! :upload_url, nil, match: String, allow_nil: true
```
bajajneha27, line 198: Please look at the overall comment I added. I don't think these parameters should be part of config.
And it should not be user's responsibility to fetch the URL. We can either do this on user's behalf or we can ask for the existing object's generation or `upload_id` .

bajajneha27 (changes requested):

bajajneha27:
Also, please disable the acceptance tests that are failing, create a separate bug to fix them and then enable them later.

google-cloud-storage/lib/google-cloud-storage.rb
shubhangi-google: > Please look at the overall comment I added. I don't think these parameters should be part of config. And it should not be user's responsibility to fetch the URL. We can either do this on user's behalf or we can ask for the existing object's generation or `upload_id` .

@bajajneha27
in that case we can ask the user to input upload_id?
if the upload is not completed uploaded object is not being created so we can not get object generation

bajajneha27: Yes, you're right. We can ask user to input `upload_id` .

shubhangi-google:
> Also, please disable the acceptance tests that are failing, create a separate bug to fix them and then enable them later.

Hi @bajajneha27  FYI samples are not always failing after @dazuma had merged some fixes in main branch

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -1410,6 +1410,30 @@ def delete if_metageneration_match: nil, if_metageneration_not_match: nil
                                 user_project: user_project
         end
 
+        ##
+        # Delete ongoing resumable upload
+        # @param [String, ::File] file Path of the file on the filesystem to
```
bajajneha27, line 1415: Why do we need the file path for deleting an ongoing upload?

google-cloud-storage/test/google/cloud/storage/bucket_test.rb
```diff
@@ -1382,6 +1382,44 @@
     mock.verify
   end
 
+   it "resstarts a resumable upload with upload_id" do
```
bajajneha27, line 1385: Typo:
```suggestion
   it "restarts a resumable upload with upload_id" do
```

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: @bajajneha27 since we are not creating any new method specifically for delete upload functionality hence I am reusing **create_file file, file_name, upload_id: upload_id, delete_upload: true**
to perform delete operation hence file param  is required (https://github.com/googleapis/google-cloud-ruby/pull/29215/files#diff-1c445f12b16180bcbdde4cab77c9b5ca55acc3fff5a45c1dd7d2bb4abb64bf76R1431)

shubhangi-google: hi @bajajneha27 any update on the further approach which needs to be taken
https://b.corp.google.com/issues/379010204#comment4

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -1410,6 +1434,28 @@ def delete if_metageneration_match: nil, if_metageneration_not_match: nil
...
+        #
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-bucket"
+        #   bucket.delete_resumable_upload file,upload_id
+
+        def delete_resumable_upload file, upload_id
```
bajajneha27, line 1453: Same question, I asked last time. Why do I need to pass `file` to just delete a resumable upload?

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: since we are making the delete upload  command [here](https://github.com/googleapis/google-api-ruby-client/blob/ebb5d6cd4aca72b1f3b4d3f092ef64805ea7e59d/google-apis-core/lib/google/apis/core/base_service.rb#L377) if upload_source is missing we encounter failure [here](https://github.com/googleapis/google-api-ruby-client/blob/ebb5d6cd4aca72b1f3b4d3f092ef64805ea7e59d/google-apis-core/lib/google/apis/core/storage_upload.rb#L77) 
do you suggest to change this implementation and skip calling the legacy  `execute_or_queue_command` ?

shubhangi-google: @bajajneha27 can you confirm if I need to change the implementation?
I had confirmed the inputs with you in [design doc](https://docs.google.com/document/d/1ntW7ih0im4iX68nDRG5rqHLmRwnYgjGNw-o8jD0rXCU/edit?tab=t.0#heading=h.bkh9z2tfwv8u) before starting the implementation

shubhangi-google:
@bajajneha27 Kokoro failure here is not related to changes made by us
  1) Failure:
Google::Cloud::Storage::Bucket::uniform_bucket_level_access::storage#test_0002_sets uniform_bucket_level_access true and is unable to get the file [acceptance/storage/bucket_uniform_bucket_level_access_test.rb:68]:
Google::Cloud::PermissionDeniedError expected but nothing was raised.
https://btx.cloud.google.com/invocations/adb33245-6f5f-4ba0-b242-dacc2184369e/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-ruby%2Fpresubmit%2Facceptance;config=default/log

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -1410,6 +1433,23 @@ def delete if_metageneration_match: nil, if_metageneration_not_match: nil
...
+        #
+        #   storage = Google::Cloud::Storage.new
+        #
+        #   bucket = storage.bucket "my-bucket"
+        #   bucket.delete_resumable_upload file,upload_id
+
+        def delete_resumable_upload upload_id
+          ensure_service!
+          raise "Upload Id missing" unless upload_id
+          service.delete_resumable_upload name, upload_id, options: { delete_upload: true }
```
bajajneha27: I just thought of this.
Since the method already is separate for deleting the resumable upload, we don't need this option: `delete_upload: true` 
We can just add it directly in core library method.

shubhangi-google: hi @bajajneha27 I have created a new PR https://github.com/googleapis/google-cloud-ruby/pull/30510 as this PR seemed to have some issue after merging main branch

PR #30353, opened 2025-05-13
fix: firestore samples hash string syntax

viacheslav-rostovtsev:


dazuma (changes requested):
See the test output. I think there needs to be a space after the colon. Also, there's one more case that was missed.

dazuma (approved):

PR #30357, opened 2025-05-13
Only load from `cgi` what is required for Ruby 3.5

Earlopain:
In Ruby 3.5 most of the `cgi` gem will be removed. Only the various escape/unescape methods will be retained by default.

https://bugs.ruby-lang.org/issues/21258

`CGI.parse` is used only in tests and can be replaced by `URI.decode_www_form`, which returns an array of arrays. So it just needs to be converted to a hash before.

`CGI.escape` is available by requiring `cgi/escape` and doesn't emit a warning on Ruby 3.5

------

closes: None

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file/signer_v2.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/signer_v2.rb
@@ -15,7 +15,7 @@
 
 require "addressable/uri"
 require "base64"
-require "cgi"
+require "cgi/escape"
 require "openssl"
 require "google/cloud/storage/errors"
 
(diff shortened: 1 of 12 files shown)


Earlopain:
Can someone take a look? The actual change is really small, just changing a require to a more specific one in two places, the rest is test-only.

Thanks!

viacheslav-rostovtsev:
I'm looking at it

Earlopain:
Ruby 4.0.0 preview2 was released in the last few days (3.5 will now be called 4.0). Additionally, Ruby 4.0 will be released in the next 1.5 months. Please take a look, I would appreciate it.

Earlopain:
Ruby 4.0 has been released

PR #30436, opened 2025-05-22
chore: Add h1 heading to migration docs

aandreassa:
Fixes internal issue b/376068354. It is related to doc pages missing a title, such as in https://cloud.google.com/ruby/docs/reference/google-cloud-automl/latest/MIGRATING

Changes generated via `grep -lR "## Migrating to" | xargs sed -i 's/## Migrating/# Migrating/g'`


dazuma (approved):
Use `docs:` when merging, so that a release is triggered and the reference docs get rebuilt.

PR #30681, opened 2025-07-15
Samples(PubSub): Various PubSub Samples

cy-yun:
BEGIN_COMMIT_OVERRIDE
chore(pubsub): Add various cloud ingestion samples
END_COMMIT_OVERRIDE

### DIFF
--- a/google-cloud-pubsub/samples/acceptance/schemas_test.rb
+++ b/google-cloud-pubsub/samples/acceptance/schemas_test.rb
@@ -95,7 +95,7 @@ describe "schemas" do
     out, _err = capture_io do
       revised_schema = commit_proto_schema schema_id: schema_id, proto_file: revision_file
     end
-    assert_includes out, "Schema commited with revision #{revised_schema.revision_id}."
+    assert_includes out, "Schema committed with revision #{revised_schema.revision_id}."
 
     #pubsub_delete_schema_revision
     assert_output "Schema #{schema_id}@#{revised_schema.revision_id} deleted.\n" do
@@ -206,8 +206,8 @@ describe "schemas" do
       out, _err = capture_io do
         schema1 = commit_avro_schema schema_id: schema_id, avsc_file: avsc_file
       end
-      refute_equal out, "Schema commited with revision #{rev_id}."
-      assert_includes out, "Schema commited with revision"
+      refute_equal out, "Schema committed with revision #{rev_id}."
+      assert_includes out, "Schema committed with revision"
 
       # pubsub_list_schema_revisions
       out, _err = capture_io do
@@ -344,8 +344,8 @@ describe "schemas" do
       out, _err = capture_io do
         schema1 = commit_proto_schema schema_id: schema_id, proto_file: revision_file
       end
-      refute_equal out, "Schema commited with revision #{rev_id}."
-      assert_includes out, "Schema commited with revision"
+      refute_equal out, "Schema committed with revision #{rev_id}."
+      assert_includes out, "Schema committed with revision"
 
       # pubsub_list_schema_revisions
       schema2 = nil
--- a/google-cloud-pubsub/samples/pubsub_commit_avro_schema.rb
+++ b/google-cloud-pubsub/samples/pubsub_commit_avro_schema.rb
@@ -23,14 +23,12 @@ def commit_avro_schema schema_id:, avsc_file:
   schemas = pubsub.schemas
 
   schema = schemas.get_schema name: pubsub.schema_path(schema_id)
-
-  definition = File.read avsc_file
-  schema.definition = definition
+  schema.definition = File.read avsc_file
 
   result = schemas.commit_schema name: schema.name,
                                  schema: schema
 
-  puts "Schema commited with revision #{result.revision_id}."
+  puts "Schema committed with revision #{result.revision_id}."
   # [END pubsub_commit_avro_schema]
   result
 end
--- a/google-cloud-pubsub/samples/pubsub_create_bigquery_subscription.rb
+++ b/google-cloud-pubsub/samples/pubsub_create_bigquery_subscription.rb
@@ -18,14 +18,15 @@ require "google/cloud/pubsub"
 # Shows how to create a BigQuery subscription where messages published
 # to a topic populates a BigQuery table.
 #
-def pubsub_create_bigquery_subscription project_id:, topic_id:, subscription_id:, bigquery_table_id:
+def pubsub_create_bigquery_subscription project_id:, topic_id:,
+                                        subscription_id:, bigquery_table_id:
   # [START pubsub_create_bigquery_subscription]
   # project_id = "your-project-id"
   # topic_id = "your-topic-id"
   # subscription_id = "your-subscription-id"
   # bigquery_table_id = "my-project:dataset-id.table-id"
-  pubsub = Google::Cloud::PubSub.new project_id: project_id
 
+  pubsub = Google::Cloud::PubSub.new project_id: project_id
   subscription_admin = pubsub.subscription_admin
 
   subscription = subscription_admin.create_subscription \
--- a/google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
+++ b/google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
@@ -21,14 +21,14 @@ def create_cloud_storage_subscription topic_id:, subscription_id:, bucket:
   # bucket = "your-bucket"
 
   pubsub = Google::Cloud::Pubsub.new
-
   subscription_admin = pubsub.subscription_admin
 
-  subscription = subscription_admin.create_subscription name: pubsub.subscription_path(subscription_id),
-                                                        topic: pubsub.topic_path(topic_id),
-                                                        cloud_storage_config: {
-                                                          bucket: bucket
-                                                        }
+  subscription = subscription_admin.create_subscription \
+    name: pubsub.subscription_path(subscription_id),
+    topic: pubsub.topic_path(topic_id),
+    cloud_storage_config: {
+      bucket: bucket
+    }
 
   puts "Cloud storage subscription #{subscription_id} created."
   # [END pubsub_create_cloud_storage_subscription]
--- a/google-cloud-pubsub/samples/pubsub_create_proto_schema.rb
+++ b/google-cloud-pubsub/samples/pubsub_create_proto_schema.rb
@@ -22,10 +22,10 @@ def create_proto_schema schema_id:, proto_file:
   pubsub = Google::Cloud::PubSub.new
   schemas = pubsub.schemas
 
-  definition = File.read proto_file
-  schema = Google::Cloud::PubSub::V1::Schema.new name: schema_id,
-                                                 type: :PROTOCOL_BUFFER,
-                                                 definition: definition
+  schema = Google::Cloud::PubSub::V1::Schema.new \
+    name: schema_id,
+    type: :PROTOCOL_BUFFER,
+    definition: File.read(proto_file)
 
   schema = schemas.create_schema parent: pubsub.project_path,
                                  schema: schema,
--- a/google-cloud-pubsub/samples/pubsub_create_topic_with_aws_msk_ingestion.rb
+++ b/google-cloud-pubsub/samples/pubsub_create_topic_with_aws_msk_ingestion.rb
@@ -21,11 +21,12 @@ def create_topic_with_aws_msk_ingestion topic_id:,
                                         gcp_service_account:
   # [START pubsub_create_topic_with_aws_msk_ingestion]
   # topic_id = "your-topic-id"
-  # cluster_arn = "arn:aws:kafka:us-east-1:111111111111:" \
-  #               "cluster/cluster-name/11111111-1111-1"
+  # cluster_arn =
+  # "arn:aws:kafka:us-east-1:111111111111:cluster/cluster-name/11111111-1111-1"
   # msk_topic = "msk-topic-name"
   # aws_role_arn = "arn:aws:iam::111111111111:role/role-name"
   # gcp_service_account = "service-account@project.iam.gserviceaccount.com"
+
   pubsub = Google::Cloud::Pubsub.new
   topic_admin = pubsub.topic_admin
 
--- a/google-cloud-pubsub/samples/.rubocop.yml
+++ b/google-cloud-pubsub/samples/.rubocop.yml
@@ -13,3 +13,5 @@ Lint/UselessAssignment:
   Enabled: false
 Metrics/MethodLength:
   Max: 30
+LineLength:
+  Max: 80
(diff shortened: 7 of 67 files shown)


google-cloud-pubsub/samples/acceptance/schemas_test.rb
```diff
@@ -183,6 +184,38 @@
...
+      out, _err = capture_io do
+        schema1 = commit_avro_schema schema_id: schema_id, avsc_file: avsc_file
+      end
+      refute_equal out, "Schema commited with revision #{rev_id}."
+      assert_includes out, "Schema commited with revision"
+
+      # pubsub_list_schema_revisions
+      schema2 = nil
+      out, _err = capture_io do
+        schema2 = commit_avro_schema schema_id: schema_id, avsc_file: avsc_file
```
aandreassa: You don't technically need to call `commit_avro_schema` again here for `schema2`. 

That was done in a different test to be able to test out `list_schema_revisions` samples itself.

You can still use the list_schema_revisions to check if it passes (i.e. via `assert_includes out, schema1.revision_id`), but maybe remove the schema2.

google-cloud-pubsub/samples/acceptance/schemas_test.rb
```diff
@@ -72,6 +73,34 @@
...
+    @schema = schemas.get_schema name: pubsub.schema_path(schema_id)
+    assert @schema
+    assert_equal "projects/#{pubsub.project}/schemas/#{schema_id}", @schema.name
+
+    # pubsub_get_schema
+    assert_output "Schema projects/#{pubsub.project}/schemas/#{schema_id} retrieved.\n" do
+      get_schema schema_id: schema_id
+    end
+
+    # pubsub_list_schemas
```
aandreassa: I'd say remove the rest of this test since those samples were tested above. 

You can still use `pubsub_get_schema` to verify it was created. You don't need to delete the schema since it is done via the `after` routine located at the top of this test.

dazuma:
Drive-by comment: When you finish and merge this, please use `samples:` or `chore:` (instead of `feat:`). Think of conventional commit tags as signals of what kind of library release should be triggered by this change. In this case, sample updates are not features of the library; they have no effect on the library code itself, and should not trigger a library release.

google-cloud-pubsub/samples/acceptance/quickstart_test.rb
```diff
@@ -18,15 +18,17 @@
 describe "quickstart" do
   let(:pubsub) { Google::Cloud::Pubsub.new }
   let(:topic_id) { random_topic_id }
+  let(:topic_admin) { pubsub.topic_admin }
 
+  focus
```
dazuma: Remove

google-cloud-pubsub/samples/pubsub_commit_avro_schema.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  definition = File.read avsc_file
+
+  schema.definition = definition
+
+  result = schemas.commit_schema name: schema.name,
+                                 schema: schema
+
+  puts "Schema commited with revision #{result.revision_id}."
+  result
```
dazuma: Let's move this outside the region tag, so it doesn't show up in the sample as it appears on the website. That text should probably just show the puts (i.e. display the results). Since the `commit_avro_schema` method itself isn't part of the sample as it appears on the website, it doesn't make sense to show its return value.

google-cloud-pubsub/samples/pubsub_commit_avro_schema.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require "google/cloud/pubsub"
+
+def commit_avro_schema schema_id:, avsc_file:
+  # [START pubsub_commit_avro_schema]
+  # schema_id = "your-schema-id"
+  # avsc_file = "path/to/a/avsc_file.avsc"
+
+  pubsub = Google::Cloud::Pubsub.new
+
```
dazuma: There's a lot of vertical whitespace here. Can we maybe group the lines semantically so the vertical whitespace actually means something? For example, I'd consider deleting blank lines 23 and 29.

google-cloud-pubsub/samples/pubsub_create_proto_schema.rb
```diff
@@ -22,7 +22,15 @@ def create_proto_schema schema_id:, proto_file:
   pubsub = Google::Cloud::Pubsub.new
 
   definition = File.read proto_file
-  schema = pubsub.create_schema schema_id, :protocol_buffer, definition
+  schema = Google::Cloud::PubSub::V1::Schema.new name: schema_id,
+                                                 type: :PROTOCOL_BUFFER,
+                                                 definition: definition
+
+  schemas = pubsub.schemas
```
dazuma, line 23: In your `commit_avro_schema` sample, these lines were ordered differently. This line (getting the `schemas` client object) was right after the `pubsub = Google::Cloud::Pubsub.new` line. Here you're doing it later, and separating the `create_schema` call (below) from the lines that create the parameters to pass to it (currently above). Might make sense to reorder these lines, and make it more consistent across samples.

google-cloud-pubsub/samples/pubsub_create_bigquery_subscription.rb
```diff
@@ -12,31 +12,30 @@
...
-  topic = pubsub.topic topic_id
-  subscription = topic.subscribe subscription_id,
-                                 bigquery_config: {
-                                   table: bigquery_table_id,
-                                   write_metadata: true
-                                 }
+
+  subscription_admin = pubsub.subscription_admin
+
+  subscription = subscription_admin.create_subscription name: pubsub.subscription_path(subscription_id),
```
aandreassa: We try keeping these samples to 80width. Please ignore if this one was already fixed in `main`. You can format it like so:
```
subscription = subscription_admin.create_subscription \
    name: pubsub.subscription_path(subscription_id),
    topic: pubsub.topic_path(topic_id),
    bigquery_config: {
      table: bigquery_table_id,
      write_metadata: true
    }
```

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  pubsub = Google::Cloud::Pubsub.new
+
+  subscription_admin = pubsub.subscription_admin
+
+  cloud_storage_config = Google::Cloud::PubSub::V1::CloudStorageConfig.new bucket: bucket
+
+  subscription = subscription_admin.create_subscription name: pubsub.subscription_path(subscription_id),
+                                                        topic: pubsub.topic_path(topic_id),
+                                                        cloud_storage_config: cloud_storage_config
```
aandreassa: I think it may be easier to set `cloud_storage_config` without using the V1 type. I generally tried to avoid introducing these types in the samples unless it got super complicated. It would be like so (similar to bigquery examples):
```
cloud_storage_config: {
    bucket: bucket
}
```

google-cloud-pubsub/samples/pubsub_create_topic_with_aws_msk_ingestion.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  aws_msk = Google::Cloud::PubSub::V1::IngestionDataSourceSettings::AwsMsk.new cluster_arn: cluster_arn,
+                                                                               topic: msk_topic,
+                                                                               aws_role_arn: aws_role_arn,
+                                                                               gcp_service_account: gcp_service_account
+
+  ingestion_data_source_settings = Google::Cloud::PubSub::V1::IngestionDataSourceSettings.new aws_msk: aws_msk
+
+  topic = topic_admin.create_topic name: pubsub.topic_path(topic_id),
+                                   ingestion_data_source_settings: ingestion_data_source_settings
```
aandreassa: It may be easier to see if we just populate it directly down here.

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
```diff
@@ -0,0 +1,35 @@
...
+require "google/cloud/pubsub"
+
+def create_cloud_storage_subscription topic_id:, subscription_id:, bucket:
+  # [START pubsub_create_cloud_storage_subscription]
+  # topic_id        = "your-topic-id"
+  # subscription_id = "your-subscription-id"
+  # bucket = "your-bucket"
+
+  pubsub = Google::Cloud::Pubsub.new
+
```
dazuma: Similarly I would delete line 24. And similar places in samples further down.

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
```diff
@@ -0,0 +1,35 @@
...
+  subscription_admin = pubsub.subscription_admin
+
+  subscription = subscription_admin.create_subscription name: pubsub.subscription_path(subscription_id),
+                                                        topic: pubsub.topic_path(topic_id),
+                                                        cloud_storage_config: {
+                                                          bucket: bucket
+                                                        }
```
dazuma: We'd like for samples not to exceed 80 columns, so they render okay on cloud.google.com web pages without requiring horizontal scrolling.

For the above call, you can do something like this:

```ruby
  subscription = subscription_admin.create_subscription \
    name: pubsub.subscription_path(subscription_id),
    topic: pubsub.topic_path(topic_id),
    cloud_storage_config: {
      bucket: bucket
    }
```

There are several cases of this in various samples throughout this PR.

google-cloud-pubsub/samples/acceptance/topics_test.rb
```diff
@@ -35,6 +43,126 @@
...
+
+  after do
+    @topic = topic_admin.get_topic topic: pubsub.topic_path(topic_id)
+    assert @topic
+    assert_equal "projects/#{pubsub.project}/topics/#{topic_id}", @topic.name
+    topic_admin.delete_topic topic: @topic.name
+  end
+
+  it "supports pubsub_create_topic_with_cloud_storage_ingestion" do
+    # pubsub_create_topic_with_cloud_storage_ingestion
```
dazuma, line 68: As discussed, consider

```ruby
  skip unless ENV["PUBSUB_EMULATOR_HOST"]
```

for each of these tests.

google-cloud-pubsub/samples/acceptance/topics_test.rb
```diff
@@ -35,6 +43,126 @@
...
+
+  before do
+    pubsub_emulator_host = ENV["PUBSUB_EMULATOR_HOST"]
+    if pubsub_emulator_host.nil?
+      raise "PUBSUB_EMULATOR_HOST env variable must be set. Please follow instructions at https://cloud.google.com/pubsub/docs/emulator"
+    end
+  end
+
+  after do
+    @topic = topic_admin.get_topic topic: pubsub.topic_path(topic_id)
```
dazuma: You don't need the `@` if you're just using `topic` as a local variable in this block. The `@` sigil signals an instance variable, which in the case of a test means that it is accessible through the entire test (`before`, `it`, and `after` blocks) since they all execute in the context of the same test object.

google-cloud-pubsub/samples/acceptance/topics_test.rb
```diff
@@ -35,6 +43,126 @@
...
+    end
+
+    subscription_id = random_subscription_id
+    bucket_id = random_bucket_id
+
+    # pubsub_create_cloud_storage_subscription
+    assert_output "Cloud storage subscription #{subscription_id} created.\n" do
+      create_cloud_storage_subscription topic_id: topic_id, subscription_id: subscription_id, bucket: bucket_id 
+    end
+    @subscription = subscription_admin.get_subscription subscription: pubsub.subscription_path(subscription_id)
```
dazuma: Similarly, you can use `subscription` (a local variable) here instead of `@subscription` (an instance variable).

dazuma (changes requested):

cy-yun:
I ran the 6 failing tests locally and 4 of them pass. The 2 that still fail locally are 

  2) Failure:
schemas::AVRO#test_0004_supports pubsub_subscribe_avro_records with JSON encoding [acceptance/schemas_test.rb:213]:
In stdout.
--- expected
+++ actual
@@ -1,3 +1 @@
-"Received a JSON-encoded message:
-{\"name\"=>\"Alaska\", \"post_abbr\"=>\"AK\"}
-"
+""


  3) Failure:
schemas::AVRO#test_0003_supports pubsub_subscribe_avro_records with binary encoding [acceptance/schemas_test.rb:184]:
In stdout.
--- expected
+++ actual
@@ -1,3 +1 @@
-"Received a binary-encoded message:
-{\"name\"=>\"Alaska\", \"post_abbr\"=>\"AK\"}
-"
+""


However, this PR should not have affected either of these failing tests

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
dazuma: not yet done

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
dazuma: not yet done

google-cloud-pubsub/samples/pubsub_commit_avro_schema.rb
```diff
@@ -0,0 +1,36 @@
...
+  # schema_id = "your-schema-id"
+  # avsc_file = "path/to/a/avsc_file.avsc"
+
+  pubsub = Google::Cloud::Pubsub.new
+  schemas = pubsub.schemas
+
+  schema = schemas.get_schema name: pubsub.schema_path(schema_id)
+
+  definition = File.read avsc_file
+  schema.definition = definition
```
dazuma: Nit: Since you don't use `definition` anywhere else, you can forgo the local variable and simply:

```ruby
schema.definition = File.read avsc_file
```

In some of the other samples below, it's maybe a bit more warranted to have the local variable since its usage is buried deeper in an expression. But in this particular case, I think it's hard to argue that the existing two lines are any more readable than the one line I suggest.

(Yes, I'm being nitpicky, but the whole purpose of a sample is to be read by humans, so cleanliness and readability are the highest priority.)

dazuma:
And a couple of the earlier comments not yet resolved (reopened).

google-cloud-pubsub/samples/pubsub_create_cloud_storage_subscription.rb
cy-yun: oops! I think I missed some. I thought I addressed this in most places. Will review once more

google-cloud-pubsub/samples/acceptance/schemas_test.rb
```diff
@@ -170,24 +183,44 @@
...
       # pubsub_subscribe_avro_records
       expect_with_retry "pubsub_subscribe_avro_records" do
-        assert_output "Received a JSON-encoded message:\n{\"name\"=>\"Alaska\", \"post_abbr\"=>\"AK\"}\n" do
+        assert_output "Received a JSON-encoded message:\n{\"name\" => \"Alaska\", \"post_abbr\" => \"AK\"}\n" do
           subscribe_avro_records subscription_id: @subscription.name, avsc_file: nil
         end
       end
     end
+
+    focus
```
dazuma: Remove

dazuma (approved):
Thanks!!

PR #30743, opened 2025-07-28
feat(BigQuery): load job external table options

cy-yun:
https://buganizer.corp.google.com/issues/374142731

### DIFF
--- a/google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
+++ b/google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
@@ -368,6 +368,177 @@ module Google
           end
 
           ##
+          # Specifies a string that represents a null value in a CSV file. For
+          # example, if you specify `\N`, BigQuery interprets `\N` as a null value when
+          # querying a CSV file. The default value is the empty string. If you set this
+          # property to a custom value, BigQuery throws an error if an empty string is
+          # present for all data types except for STRING and BYTE. For STRING and BYTE
+          # columns, BigQuery interprets the empty string as an empty value.
+          #
+          # @return [String, nil] The null marker string. `nil` if not set.
+          #
+          # @example
+          #   require "google/cloud/bigquery"
+          #
+          #   bigquery = Google::Cloud::Bigquery.new
+          #
+          #   csv_url = "gs://bucket/path/to/data.csv"
+          #   csv_table = bigquery.external csv_url do |csv|
+          #     csv.null_marker = "\N"
+          #   end
+          #
+          #   csv_table.null_marker #=> "\N"
+          #
+          def null_marker
+            @gapi.csv_options.null_marker
+          end
+
+          ##
+          # Sets a string that represents a null value in a CSV file. For
+          # example, if you specify `\N`, BigQuery interprets `\N` as a null value when
+          # querying a CSV file. The default value is the empty string. If you set this
+          # property to a custom value, BigQuery throws an error if an empty string is
+          # present for all data types except for STRING and BYTE. For STRING and BYTE
+          # columns, BigQuery interprets the empty string as an empty value.
+          #
+          # @param [String, nil] null_marker The null marker string. `nil` to unset.
+          #
+          # @example
+          #   require "google/cloud/bigquery"
+          #
+          #   bigquery = Google::Cloud::Bigquery.new
+          #
+          #   csv_url = "gs://bucket/path/to/data.csv"
+          #   csv_table = bigquery.external csv_url do |csv|
+          #     csv.null_marker = "\N"
+          #   end
+          #
+          #   csv_table.null_marker #=> "\N"
+          #
+          def null_marker= null_marker
+            frozen_check!
+            @gapi.csv_options.null_marker = null_marker
+          end
+
+          ##
+          # The list of strings represented as SQL NULL value in a CSV file.
+          # null_marker and null_markers can't be set at the same time. If null_marker is
+          # set, null_markers has to be not set. If null_markers is set, null_marker has
+          # to be not set. If both null_marker and null_markers are set at the same time,
+          # a user error would be thrown. Any strings listed in null_markers, including
+          # empty string would be interpreted as SQL NULL. This applies to all column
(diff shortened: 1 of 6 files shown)


google-cloud-bigquery/lib/google/cloud/bigquery/external/data_source.rb
```diff
@@ -744,6 +744,103 @@ def hive_partitioning_source_uri_prefix= source_uri_prefix
             @gapi.hive_partitioning_options.source_uri_prefix = source_uri_prefix
           end
 
+          ##
+          # Time zone used when parsing timestamp values that do not have
+          # specific time zone information (e.g. 2024-04-20 12:34:56). The expected format
+          # is a IANA timezone string (e.g. America/Los_Angeles).
```
dazuma: Surround example strings in backticks so they render in code font and so it's obvious where each example begins and ends. Please also update additional examples below. Format pattern strings are another good candidate for surrounding in backticks, especially because they sometimes include characters that markdown would otherwise attempt to interpret.

google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
```diff
@@ -367,6 +367,65 @@ def skip_leading_rows= row_count
...
+          def null_markers= null_markers
+            frozen_check!
+            @gapi.csv_options.null_markers = null_markers
+          end
+
+          # Controls the strategy used to match loaded columns to the schema. If
+          # not set, a sensible default is chosen based on how the schema is provided. If
+          # autodetect is used, then columns are matched by name. Otherwise, columns are
+          # matched by position. This is done to keep the behavior backward-compatible.
+          # Acceptable values are: POSITION - matches by position. This assumes that the
```
dazuma: I'd recommend using a markdown bulleted list for the list of acceptable values. In general, you can assume that all docstrings are rendered via markdown, and take advantage of that to make your documentation more readable.

google-cloud-bigquery/lib/google/cloud/bigquery/external/data_source.rb
```diff
@@ -744,6 +744,103 @@ def hive_partitioning_source_uri_prefix= source_uri_prefix
...
+          #
+          # @return [String, nil] The IANA time zone name, such as "America/Los_Angeles".
```
dazuma: Under what circumstances would this be nil? Please document that in each case where nil is a type option, both here and elsewhere in this PR.

google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
```diff
@@ -367,6 +367,70 @@ def skip_leading_rows= row_count
             @gapi.csv_options.skip_leading_rows = row_count
           end
 
+          ##
+          # The list of strings represented as SQL NULL value in a CSV file.
+          # null_marker and null_markers can't be set at the same time. If null_marker is
```
dazuma, line 425: I'm a bit confused by this discussion of `null_marker` and `null_markers`. There isn't a `null_marker` attribute on this (`CsvSource`) object. Or are you referring to something else? Can you clarify?

google-cloud-bigquery/lib/google/cloud/bigquery/external/data_source.rb
```diff
@@ -744,6 +744,112 @@ def hive_partitioning_source_uri_prefix= source_uri_prefix
...
+            @gapi.time_zone
+          end
+
+          ##
+          # Sets the time zone used when parsing timestamp values that do not have
+          # specific time zone information (e.g. `2024-04-20 12:34:56`). The expected
+          # format is an IANA timezone string (e.g. `America/Los_Angeles`).
+          #
+          # @param [String] time_zone The IANA time zone name, such as
+          #   `America/Los_Angeles`.
```
dazuma: Is it possible (or reasonable) to pass nil here to "unset" this attribute? (Similar for other attributes below where "nil if not set" is a possibility.)

google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
cy-yun: Screenshot 2025-08-06 at 1 41 18 PM

null_marker is supported by Google::Apis::BigQueryV2::CsvOptions . And it should be here too. I guess it was never added to our library. I assumed it was there because null_marker isn't a new field. I can add null_marker.

google-cloud-bigquery/lib/google/cloud/bigquery/external/data_source.rb
cy-yun: Yes. But that should be true for all other optional parameters that already exist. I'll make the documentation more consistent with existing parameters?

google-cloud-bigquery/lib/google/cloud/bigquery/load_job.rb
```diff
@@ -2585,6 +2669,89 @@ def clustering_fields= fields
             @gapi.configuration.load.clustering.fields = fields
           end
 
+          ##
+          # Sets the format used to parse DATE values. Supports C-style and SQL-style
+          # values.
+          #
+          # @param [String, nil] date_format The date format pattern, such as
```
dazuma, line 2676: The name of your method parameter is `val`, and this needs to match. This is why the YARD build is failing. Here and in the methods below.

google-cloud-bigquery/lib/google/cloud/bigquery/external/csv_source.rb
```diff
@@ -367,6 +367,177 @@ def skip_leading_rows= row_count
             @gapi.csv_options.skip_leading_rows = row_count
           end
 
+          ##
+          # Specifies a string that represents a null value in a CSV file. For
+          # example, if you specify `\N`, BigQuery interprets `\N` as a null value when
```
dazuma, line 372: Shoot, sorry, something I missed. I think you want `\n`, not `\N`. The former is the newline character; the latter has no special meaning in Ruby (and is interpreted as a simple capital-N).

PR #30800, opened 2025-08-07
fix(trace): Correctly normalize inconsistent ActiveSupport timestamps

### LINKED ISSUE #30798: google-cloud-trace drops ActiveSupport::Notifications spans in Rails 8
The normalize_time utility method in google-cloud-trace incorrectly handles timestamps from ActiveSupport::Notifications in Rails 7 and later. The method assumes these timestamps are in milliseconds, when they are actually floats representing seconds. This leads to incorrect time conversion (to the year 1970) and causes the resulting trace spans to be silently dropped by the Google Cloud API.

#### Environment details

  - OS: Linux
  - Ruby version: 3.4.5
  - Gem name and version: google-cloud-trace 0.45.0

#### Steps to reproduce

* Set up a standard Rails 8 application.
* Add the google-cloud-trace gem and configure it in an initializer.
* Subscribe to Rails notifications like sql.active_record
* Make a request to a controller action to trigger the event.
*  Observe that the corresponding span does not appear in the Google Cloud Trace console (because it was created with an invalid timestamp and rejected by the API).

grobie:
The `normalize_time` method in the google-cloud-trace library produced incorrect timestamps when used with certain versions of Rails 7 and with Rails 8.

Because this change was not consistent across all versions and was altered silently (e.g., https://github.com/rails/rails/pull/50779), a simple version check is unreliable. The previous implementation incorrectly assumed any numeric timestamp was in milliseconds, causing faulty time conversions and leading to trace spans being dropped by the API.

It now uses a numeric threshold to reliably differentiate between timestamps in seconds and milliseconds, ensuring correct handling regardless of the specific Rails version.

Fixes #30798.

### DIFF
--- a/google-cloud-trace/lib/google/cloud/trace/notifications.rb
+++ b/google-cloud-trace/lib/google/cloud/trace/notifications.rb
@@ -110,17 +110,19 @@ module Google
           labels
         end
 
+        TIMESTAMP_SECONDS_THRESHOLD = (2**31) - 1
+
         ##
         # @private
         #
-        # active support event's time is:
-        #
-        # - rails >= 7: timestamp in milliseconds
-        # - rails <  7: time
+        # Normalizes potentially inconsistent timestamps from
+        # ActiveSupport::Notifications into a consistent format in seconds since
+        # the Unix epoch. Some older Rails versions passed `Time` objects, while
+        # some versions of Rails 7+ passes a `Float` in milliseconds.
         def self.normalize_time time_or_float
-          return time_or_float if Rails::VERSION::MAJOR < 7
+          return time_or_float unless time_or_float.is_a? Numeric
 
-          Time.at time_or_float / 1000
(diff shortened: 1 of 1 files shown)


google-cloud-trace/lib/google/cloud/trace/notifications.rb
```diff
@@ -110,17 +110,19 @@ def self.payload_to_labels event, maxlen, label_namespace
...
+
         ##
         # @private
         #
-        # active support event's time is:
-        #
-        # - rails >= 7: timestamp in milliseconds
-        # - rails <  7: time
+        # Normalizes potentially inconsistent timestamps from
+        # ActiveSupport::Notifications into a consistent format in seconds since
```
viacheslav-rostovtsev, line 119: It doesn't normalize though. It would return Time if passed Time.

grobie:
@viacheslav-rostovtsev Feel free to close this PR and take it over. I don't have the time to work more on this. We had to patch our own Rails 8 application already to support ActiveSupport notifications again.

PR #30814, opened 2025-08-11
fix(container_analysis): Update google-cloud-pubsub gem dependency to v3.x

aandreassa:
Fixing it manually in place of https://github.com/googleapis/google-cloud-ruby/pull/30745


dazuma (approved):
Thanks for doing this!

PR #31536, opened 2025-09-23
samples(StorageControl): Add samples for anywhereCache

shubhangi-google:
This internal PR adds sample code for the anywhere cache feature, allowing users to:

- [ ] Create anywhere cache
- [ ] Get anywhere cache
- [ ] List anywhere cache
- [ ] Update anywhere cache
- [ ] Pause anywhere cache
- [ ] Disable anywhere cache
- [ ] Resume anywhere cache

this Pr is created as https://github.com/googleapis/google-cloud-ruby/pull/30597 , https://github.com/googleapis/google-cloud-ruby/pull/30930 was corrupted

### DIFF
--- a/google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
+++ b/google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
@@ -60,7 +60,6 @@ def create_anywhere_cache bucket_name:, zone:
     result = storage_control_client.get_anywhere_cache get_request
     min_delay = 30 # 30 seconds
     max_delay = 900 # 15 minutes
-    start_time = Time.now
     while result.state&.downcase != "running"
       unless ["paused", "disabled", "creating"].include? result.state&.downcase
         raise Google::Cloud::Error,
@@ -71,14 +70,10 @@ def create_anywhere_cache bucket_name:, zone:
       min_delay = [min_delay * 2, max_delay].min # Exponential backoff with a max delay
       result = storage_control_client.get_anywhere_cache get_request
     end
-    end_time = Time.now
-    duration = end_time - start_time
-    puts "Total polling time: #{duration.round 2} seconds."
-    message = "Successfully created anywhereCache - #{result.name}."
+    puts "Successfully created anywhereCache - #{result.name}."
   rescue Google::Cloud::Error => e
-    message = "Failed to create AnywhereCache. Error: #{e.message}"
+    puts "Failed to create AnywhereCache. Error: #{e.message}"
   end
-  puts message
 end
 # [END storage_control_create_anywhere_cache]
 create_anywhere_cache bucket_name: ARGV.shift, zone: ARGV.shift if $PROGRAM_NAME == __FILE__
--- a/google-cloud-storage-control/samples/storage_control_update_anywhere_cache.rb
+++ b/google-cloud-storage-control/samples/storage_control_update_anywhere_cache.rb
@@ -58,7 +58,6 @@ def update_anywhere_cache bucket_name:, anywhere_cache_id:
     result = storage_control_client.get_anywhere_cache get_request
     min_delay = 30 # 30 seconds
     max_delay = 600 # 10 minutes
-    start_time = Time.now
     while result.state&.downcase != "running"
       unless ["paused", "disabled", "creating"].include? result.state&.downcase
         raise Google::Cloud::Error,
@@ -69,14 +68,10 @@ def update_anywhere_cache bucket_name:, anywhere_cache_id:
       min_delay = [min_delay * 2, max_delay].min # Exponential backoff with a max delay
       result = storage_control_client.get_anywhere_cache get_request
     end
-    end_time = Time.now
-    duration = end_time - start_time
-    puts "Total waiting time : #{duration.round 2} seconds."
-    message = "Successfully updated anywhereCache - #{result.name}."
+    puts "Successfully updated anywhereCache - #{result.name}."
   rescue Google::Cloud::Error => e
-    message = "Failed to update AnywhereCache. Error: #{e.message}"
+    puts "Failed to update AnywhereCache. Error: #{e.message}"
   end
-  puts message
 end
 # [END storage_control_update_anywhere_cache]
 update_anywhere_cache bucket_name: ARGV.shift, anywhere_cache_id: ARGV.shift if $PROGRAM_NAME == __FILE__


google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
```diff
@@ -0,0 +1,63 @@
...
+    )
+    result = storage_control_client.get_anywhere_cache get_request
+    min_delay = 180 # 3 minutes
+    max_delay = 900 # 15 minutes
+    while result.state != "running"
+      puts "Cache not running yet, current state is #{result.state}. Retrying in #{min_delay} seconds."
+      sleep min_delay
+      min_delay = [min_delay * 2, max_delay].min # Exponential backoff with a max delay
+      result = storage_control_client.get_anywhere_cache get_request
+    end
```
Pulkit0110: What if somehow the cache creation fails on the backend? In that case the state may never change to `running` and it will turn into an infinite loop.

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
```diff
@@ -0,0 +1,63 @@
...
+  # The request creates a new cache in the specified zone.
+  # The cache is created in the specified bucket.
+  begin
+    operation = storage_control_client.create_anywhere_cache request
+    puts "AnywhereCache operation created - #{operation.name}"
+    get_request = Google::Cloud::Storage::Control::V2::GetAnywhereCacheRequest.new(
+      name: name
+    )
+    result = storage_control_client.get_anywhere_cache get_request
+    min_delay = 180 # 3 minutes
```
Pulkit0110: I think `min_delay` should be smaller like 30 seconds or 1 minute.

google-cloud-storage-control/samples/storage_control_disable_anywhere_cache.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  # Create a request.
+  request = Google::Cloud::Storage::Control::V2::DisableAnywhereCacheRequest.new(
+    name: name
+  )
+  # The request disables the cache, but does not delete it.
+  # The cache can be re-enabled later.
+  begin
+    result = storage_control_client.disable_anywhere_cache request
+    puts "AnywhereCache #{result.name} #{result.state}"
```
Pulkit0110: nit: `puts "Successfully disabled anywhereCache for #{result.name}. The current state is #{result.state}"`

google-cloud-storage-control/samples/storage_control_list_anywhere_caches.rb
```diff
@@ -0,0 +1,43 @@
...
+  request = Google::Cloud::Storage::Control::V2::ListAnywhereCachesRequest.new(
+    parent: parent
+  )
+  # The request lists all caches in the specified bucket.
+  # The caches are identified by the specified bucket name.
+  begin
+    result = storage_control_client.list_anywhere_caches request
+    result.response.anywhere_caches.each do |item|
+      puts "AnywhereCache #{item.name} found in list"
+    end
```
Pulkit0110, line 37: Usually when a list response is returned, it contains the result of the first page. So I think `result.response.anywhere_caches.each` only iterates through the first page but not sure if ruby handles pagination on its own. Can you please double check if the code actually iterate through all the pages?

Pulkit0110 (changes requested):

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
shubhangi-google: Based on our [recent logs](https://bit.googleplex.com/#/shubhangisin/6109175002169344), we’ve observed that both the create and delete operations are taking longer than expected. To mitigate potential performance issues, we’ve intentionally set the minimum delay to no less than 3 minutes.
Please confirm if you would still prefer to reduce this to 30 seconds despite the observed latency.

google-cloud-storage-control/samples/storage_control_list_anywhere_caches.rb
shubhangi-google: the code will iterate through all the pages

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
Pulkit0110: Yes I think we should reduce it to 30 seconds because in future there may be optimizations which could reduce this latency. And since we're using exponential backoff, it would only add 2-3 more retry iteration which should be fine.

google-cloud-storage-control/samples/storage_control_list_anywhere_caches.rb
Pulkit0110: Thanks for confirming!

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
shubhangi-google: fixed

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
shubhangi-google: fixed

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
```diff
@@ -0,0 +1,84 @@
...
+              "AnywhereCache operation failed on the backend with state #{result.state&.downcase}."
+      end
+      puts "Cache not running yet, current state is #{result.state&.downcase}. Retrying in #{min_delay} seconds."
+      sleep min_delay
+      min_delay = [min_delay * 2, max_delay].min # Exponential backoff with a max delay
+      result = storage_control_client.get_anywhere_cache get_request
+    end
+    end_time = Time.now
+    duration = end_time - start_time
+    puts "Total polling time: #{duration.round 2} seconds."
```
Pulkit0110: nit: It doesn't add any value to calculate the polling time. You can remove it.

Same for the `update anywhere cache` sample.

google-cloud-storage-control/samples/storage_control_create_anywhere_cache.rb
```diff
@@ -0,0 +1,84 @@
...
+    message = "Successfully created anywhereCache - #{result.name}."
+  rescue Google::Cloud::Error => e
+    message = "Failed to create AnywhereCache. Error: #{e.message}"
+  end
+  puts message
```
Pulkit0110: nit: `message` variable can be removed and `puts` statement can be used directly instead.

Same for the `update anywhere cache` sample.

Pulkit0110 (approved):

PR #31568, opened 2025-10-01
chore: Add integration test for JSON uploads over 5MB

aandreassa:
The limit seems to be 10MB, and it matches what is listed here: https://cloud.google.com/bigquery/quotas

>  Maximum request size | 10 MB | The request size can be up to 10 MB, including additional properties like query parameters.

Example log for 11MB upload: 

``` Generated JSON size: 10 MB
Google::Cloud::Error encountered: Invalid request
Error body: <!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 413 (Request Entity Too Large)!!1</title>
  <style>
…
```

### DIFF
--- a/google-cloud-bigquery/acceptance/bigquery/large_json_test.rb
+++ b/google-cloud-bigquery/acceptance/bigquery/large_json_test.rb
@@ -1,0 +1,61 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


google-cloud-bigquery/acceptance/bigquery/large_json_test.rb
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "bigquery_helper"
+require "securerandom"
+require "json"
+require "httpclient"
```
viacheslav-rostovtsev, line 18: do we use this?

aandreassa:
Not needed anymore. Closing.

PR #31795, opened 2025-10-22
fix(storage): Implement path containment to prevent traversal attacks

shubhangi-google:
This pull request addresses a potential security vulnerability in the file download mechanism by implementing stringent path validation. The primary goal is to prevent malicious users from specifying paths that could lead to directory traversal attacks, ensuring that downloaded files are always written to safe, intended locations relative to the current working directory. This enhancement significantly improves the security posture of the file download feature.

### Highlights

* **Security Enhancement**: Implemented robust path validation within the file download functionality to prevent directory traversal vulnerabilities.
* **New Path Validation Method**: Introduced a new private method, `safe_path_for_download`, which ensures that user-supplied download paths are relative and resolve strictly within the current working directory.
* **Expanded Test Coverage**: Added new acceptance tests to specifically verify that attempts to download files to absolute paths or paths containing directory traversal sequences correctly raise `SecurityError`.

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file.rb
@@ -2297,7 +2297,6 @@ module Google
         #
 
         def safe_path_for_download user_supplied_path
-
           temp_regex = /\A#{Regexp.escape Dir.tmpdir}/
 
           # Allow StringIO to pass through


viacheslav-rostovtsev:
In view of PHP breakage, do not merge any path containment work without my review

viacheslav-rostovtsev (approved):

bajajneha27:
We don't need these changes as we pass specific file path for download.

PR #32121, opened 2025-10-28
docs: add warning about loading unvalidated credentials

aandreassa:


google-cloud-bigquery/AUTHENTICATION.md
```diff
@@ -67,12 +67,14 @@ The environment variables that BigQuery checks for credentials are configured on
 5. `GOOGLE_APPLICATION_CREDENTIALS` - Path to JSON file
 
 ```ruby
+require "googleauth"
```
viacheslav-rostovtsev: Here and other like this with env vars -- it's ok to leave as-is, no need to change them

google-cloud-bigtable/lib/google-cloud-bigtable.rb
```diff
@@ -40,29 +40,31 @@ module Cloud
...
     #   `GRPC::Core::ChannelCredentials` will be used to set up the gRPC client. The channel credentials
     #   should already be composed with a `GRPC::Core::CallCredentials` object.
     #   `Proc` will be used as an updater_proc for the gRPC channel. The proc transforms the
     #   metadata for requests, generally, to give OAuth credentials.
-    # @return [Google::Cloud::Bigtable::Project]
+    #   @note Warning: Passing a `String` to a keyfile path or a `Hash` of credentials
+    #     is deprecated. Providing an unvalidated credential configuration to
+    #     Google APIs can compromise the security of your systems and data.
     #
-    # @example
```
viacheslav-rostovtsev, line 60: This code is strange: `gcloud  = Google::Cloud.new`. does it work? If it works it's probably using ADC and should not be replaced

google-cloud-dns/lib/google/cloud/dns/zone.rb
```diff
@@ -627,7 +627,7 @@ def to_zonefile
         # @param [String] name The owner of the record. For example:
         #   `example.com.`.
         # @param [String] type The identifier of a [supported record
-        #   type](https://cloud.google.com/dns/what-is-cloud-dns).
+        #   type](https://cloud.google.com/dns/docs/records-overview).
```
viacheslav-rostovtsev, line 630: Appreciate the docs URL fixes, I am sure it will be useful for some of our customers

google-cloud-bigtable/lib/google-cloud-bigtable.rb
aandreassa: I am leaving as it was. It uses https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-core/lib/google/cloud.rb#L62

viacheslav-rostovtsev (approved):

PR #32263, opened 2025-12-12
feat(pubsub): Add skip_lookup to Project#publisher to enable lazy loading

### LINKED ISSUE #32247: Allow creating Publisher without GetTopic API lookup to prevent Admin Quota exhaustion
**Is your feature request related to a problem? Please describe.**

This request addresses a regression in functionality introduced in google-cloud-pubsub version 3.0 regarding the initialization of Topic/Publisher references.

In version 2.x, it was possible to create a reference to a Pub/Sub topic without triggering an immediate gRPC API call. However, with the shift to the Publisher class in v3.0, the library now enforces a lookup (specifically topic_admin.get_topic) every time a publisher is instantiated.

This behavior causes significant operational issues in environments with a high volume of topics and services. Specifically, Google Cloud imposes a strict system limit on "Administrator operations per minute." During a rolling restart of our application—where many services attempt to re-initialize publishers for various topics simultaneously—we rapidly exhaust this quota. This results in temporary outages and startup failures purely due to these mandatory, yet redundant, existence checks.

 **Describe the solution you'd like**

I would like to request the restoration of the ability to create a Publisher instance using a known topic name without performing a lookup validation (the GetTopic API call).

This could be achieved by introducing a skip_lookup (or similarly named) option to the factory method.

Below is a proposed implementation sketch demonstrating how this could be achieved in Google::Cloud::PubSub::Project and Google::Cloud::PubSub::Publisher:

In Google::Cloud::PubSub::Project:

```ruby
def publisher topic_name, project: nil, async: nil, skip_lookup: false
  ensure_service!
  options = { project: project, async: async }
  
  # Allow returning a Publisher immediately if we trust the topic name exists
  return Publisher.from_name topic_name, service, options if skip_lookup

  grpc = topic_admin.get_topic topic: service.topic_path(topic_name, options)
  Publisher.from_grpc grpc, service, async: async
end
```

In Google::Cloud::PubSub::Publisher:

```ruby
def name
  return @resource_name if reference?
  @grpc.name
end

def reference?
  @grpc.nil?
end

# Factory method to create a Publisher without a grpc object
def self.from_name name, service, options = {}
  name = service.topic_path name, options
  from_grpc(nil, service, async: options[:async]).tap do |t|
    t.instance_variable_set :@resource_name, name
  end
end
```

 **Describe alternatives you've considered**

1. Remaining on v2.x: This prevents us from utilizing new features and security patches in the long term and creates technical debt.
2. Monkey-patching/Wrapping the library: We have considered implementing a custom wrapper or monkey-patching the library to manually inject the from_name logic proposed above. However, relying on internal implementation details or private forks is fragile and increases the maintenance burden significantly.

 **Additional context**

Our architecture relies on a services mesh communicating via a large number of distinct Pub/Sub topics. The current strict requirement to verify topic existence upon client initialization turns a standard application restart into a "Distributed Denial of Service" attack against our own Google Cloud Project quotas. Restoring the "reference-only" instantiation is critical for reliability at scale.

aandreassa:
Restoring `skip_lookup` capability for a Publisher,  allowing it to lazy load its reference topic resource. 

[Reference from PubSub/v2.x](https://github.com/googleapis/google-cloud-ruby/pull/30349/files#diff-2f76abd3a65deaa2c8faf26d48600a6226e693c340e95f56b3b3fb0575e39cdb)

Fixes #32247

### DIFF
--- a/google-cloud-pubsub/OVERVIEW.md
+++ b/google-cloud-pubsub/OVERVIEW.md
@@ -36,7 +36,7 @@ into Admin Operations and Data Plane Operations.
 * **Data Plane Operations**: For the core functionality of publishing and receiving messages.
 
 To learn more about Pub/Sub, read the [Google Cloud Pub/Sub Overview
-](https://cloud.google.com/pubsub/overview).
+](https://docs.cloud.google.com/pubsub/docs/overview).
 
 ## Admin Operations
 


google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
```diff
@@ -87,9 +87,56 @@ def async_publisher
...
         end
 
+        ##
+        # Determines whether the publisher object was created with a resource
+        # representation from the Pub/Sub service.
+        #
+        # @return [Boolean] `true` when the publisher was created with a resource
+        #   representation, `false` otherwise.
+        #
+        def reference?
```
viacheslav-rostovtsev, line 101: `is_reference?` perhaps

viacheslav-rostovtsev (approved):
LGTM

google-cloud-pubsub/lib/google/cloud/pubsub/publisher.rb
aandreassa: Leaving as `reference` to align with Rubocop. [Naming/PredicateName](https://rubydoc.info/gems/rubocop/0.52.0/RuboCop/Cop/Naming/PredicateName)

PR #32343, opened 2026-01-12
chore: correct github credentials setup in Create New Library Github Action

viacheslav-rostovtsev:


viacheslav-rostovtsev:
tested with successful Create New Library job: https://github.com/googleapis/google-cloud-ruby/actions/runs/20930455777

PR #32404, opened 2026-01-29
feat(pubsub): Support adhoc debug logging

aandreassa:


google-cloud-pubsub/lib/google/cloud/pubsub/async_publisher.rb
```diff
@@ -157,7 +157,7 @@ def publish data = nil, attributes = nil, ordering_key: nil, **extra_attrs, &cal
             end
             batch_action = batch.add msg, callback
             if batch_action == :full
-              publish_batches!
+              publish_batches! reason: "batch full"
```
Hectorhammett, line 160: Is it possible to have an enum in Ruby? Or maybe constants:
```ruby
class Reason
    BATCH_FULL="batch full"
    SHUTDOWN="shutdown"
    MANUAL_FLUSH="manual flush"
    INTERNAL_TIMEOUT="internal timeout"
    UNKNOWN="unknown"
end
```

Then the usage:
```ruby
publish_batch! Reason::MANUAL_FLUSH
```

But it is just a nit, I am usually against "magic strings" and I am fine either way. Maybe in Ruby is not that common to use constants.

aandreassa: We generally don't use them, but I agree we can at least make it some constant. I will follow up

Issue #384, opened 2015-10-20
Rails tutorails with gem "gcloud"


sajjadmurtaza-ror:
Is there any source where i can learn rails with gem "gcloud"

blowmage:
Hi @sajjadmurtaza-ror, thanks for the question. The Google Cloud's [Getting started with Ruby](https://cloud.google.com/ruby/) has an [example app](https://cloud.google.com/ruby/getting-started/hello-world) that uses `gcloud` in a rails app. The tutorial uses named branches, so the datastore service is added in the [2-cloud-datastore](https://github.com/GoogleCloudPlatform/getting-started-ruby/tree/2-cloud-datastore) branch, and storage service is added in the [3-cloud-storage](https://github.com/GoogleCloudPlatform/getting-started-ruby/tree/3-cloud-storage) branch. We'd love to know if this what you are looking for.

sajjadmurtaza-ror:
@blowmage you are great. Its awesome.
But i have lot of question about Gcloud::Datastore. Because am new for it.
an i ask i you don't mind.

jgeewax:
The best place for general questions like this might be StackOverflow.com
-- feel free to use that for high-level questions! :)

sajjadmurtaza-ror:
Actually the Gcloud::Datastore community is not in huge number. Anyway thanks.

blowmage:
Feel free to ask on StackOverflow or email me directly. Thanks!

PR #33452, opened 2026-03-30
feat(storage): Support bucket encryption config

shubhangi-google:
This pull request significantly enhances the Google Cloud Storage client library by introducing comprehensive support for managing bucket encryption enforcement configurations. Users can now define and enforce default encryption behaviors for their buckets and files using customer-managed, customer-supplied, or Google-managed encryption keys. The changes provide robust API methods, update internal mechanisms for handling these configurations, and include practical examples and tests to ensure functionality and ease of use.

### Highlights

* **New Bucket Encryption Configuration Methods**: Added new methods to `Google::Cloud::Storage::Bucket` for managing customer-managed, customer-supplied, and Google-managed encryption enforcement configurations, including getters and setters for `customer_managed_encryption_enforcement_config`, `customer_supplied_encryption_enforcement_config`, and `google_managed_encryption_enforcement_config`.
* **Unified Encryption Update Method**: Introduced a new `update_bucket_encryption_enforcement_config` method to handle updates for different types of bucket encryption enforcement configurations in a unified manner.
* **Enhanced `patch_gapi!` Method**: Modified the internal `patch_gapi!` method to accept an optional `bucket_encryption_config` parameter, allowing more flexible patching of encryption-related attributes.
* **New Samples and Acceptance Tests**: Included new samples and corresponding acceptance tests to demonstrate how to get, set, update, and remove bucket encryption enforcement configurations.
* **Gemfile Updates**: Added `ostruct`, `cgi`, and `irb` gems to the `Gemfile` for samples, addressing gems removed from Ruby core that are required for testing.

This Pr was created as old PR got corrupted - https://github.com/googleapis/google-cloud-ruby/pull/32778, https://github.com/googleapis/google-cloud-ruby/pull/32786


google-cloud-storage/acceptance/storage/bucket_encryption_test.rb
```diff
@@ -71,4 +88,50 @@
...
+      bucket.update_bucket_encryption_enforcement_config google_managed_config
+      _(bucket.google_managed_encryption_enforcement_config.restriction_mode).must_equal "FullyRestricted"
+
+      bucket.reload!
+      _(bucket.customer_supplied_encryption_enforcement_config.restriction_mode).must_equal "FullyRestricted"
+      _(bucket.google_managed_encryption_enforcement_config.restriction_mode).must_equal "FullyRestricted"
+    end
+
+    it "deletes all encryption enforcement configs" do
+      # For the update, need to specify all three configs
```
bajajneha27: Not sure if this statement is correct

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,152 @@ def default_kms_key= new_default_kms_key
...
+        # This configuration defines the
+        # default encryption behavior for the bucket and its files, and it can be used to enforce encryption requirements for the bucket.
+        # For more information, see [Bucket encryption](https://docs.cloud.google.com/storage/docs/encryption/).
+        # @return [Google::Apis::StorageV1::Bucket::Encryption::CustomerManagedEncryptionEnforcementConfig, nil] The bucket's encryption configuration, or `nil` if no encryption configuration has been set.
+        # @example
+        #   require "google/cloud/storage"
+        #   #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   bucket.customer_managed_encryption_enforcement_config #=> Google::Apis::StorageV1::Bucket::Encryption::CustomerManagedEncryptionEnforcementConfig.new
```
bajajneha27: nit: are we also planning to provide the setter using a hash?

bajajneha27: This comment is confusing. This is a getter method, so ideally we should be showing code sample for that.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,152 @@ def default_kms_key= new_default_kms_key
...
+        # for the bucket.
+        # For more information, see [Bucket encryption](https://docs.cloud.google.com/storage/docs/encryption/).
+        # @return [Google::Apis::StorageV1::Bucket::Encryption::CustomerSuppliedEncryptionEnforcementConfig, nil] 
+        # The bucket's encryption configuration, or `nil` if no encryption configuration has been set.
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   bucket.customer_supplied_encryption_enforcement_config #=> Google::Apis::StorageV1::Bucket::Encryption::CustomerSuppliedEncryptionEnforcementConfig.new
```
bajajneha27: Same comment as above.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,152 @@ def default_kms_key= new_default_kms_key
...
+        # requirements for the bucket.
+        # For more information, see [Bucket encryption](https://docs.cloud.google.com/storage/docs/encryption/).
+        # @return [Google::Apis::StorageV1::Bucket::Encryption::GoogleManagedEncryptionEnforcementConfig, nil]
+        # The bucket's encryption configuration, or `nil` if no encryption configuration has been set.
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   bucket.google_managed_encryption_enforcement_config #=> Google::Apis::StorageV1::Bucket::Encryption::GoogleManagedEncryptionEnforcementConfig.new
```
bajajneha27: Same comment here as well

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -3252,13 +3398,18 @@ def ensure_gapi!
 
         def patch_gapi! attributes,
                         if_metageneration_match: nil,
-                        if_metageneration_not_match: nil
+                        if_metageneration_not_match: nil,
+                        bucket_encryption_config: nil
```
bajajneha27: Why are we introducing new parameter here? Why can't we combine this in existing parameter `attributes` ?

google-cloud-storage/samples/Gemfile
```diff
@@ -36,3 +36,7 @@ group :test do
   gem "minitest-hooks", "~> 1.5"
   gem "rake"
 end
+# The following gems have been removed from ruby core and are required for testing.
+gem "ostruct"
```
bajajneha27, line 40: Didn't we change this in another PR?

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
```diff
@@ -88,11 +87,12 @@
...
     it "sets its encryption config to nil" do
       bucket_gapi_with_key = bucket_gapi.dup
-      bucket_gapi_with_key.encryption = encryption_gapi(kms_key)
+      bucket_gapi_with_key.encryption = encryption_gapi(key_name: kms_key)
       bucket_with_key = Google::Cloud::Storage::Bucket.from_gapi bucket_gapi_with_key, storage.service
-      patch_bucket_gapi = Google::Apis::StorageV1::Bucket.new encryption: encryption_gapi(nil)
       mock = Minitest::Mock.new
-      mock.expect :patch_bucket, bucket_gapi, [bucket_name, patch_bucket_gapi], **patch_bucket_args(options: {retries: 0})
+      mock.expect :patch_bucket, bucket_gapi do |name, patch_obj, **args|
+        name == bucket_name && patch_obj.encryption&.default_kms_key_name.nil?
```
bajajneha27: Why are we adding this condition here?

google-cloud-storage/test/helper.rb
```diff
@@ -244,8 +244,23 @@ def download_http_resp gzip: nil
...
   end
 
-  def encryption_gapi key_name
-    Google::Apis::StorageV1::Bucket::Encryption.new default_kms_key_name: key_name
+  def encryption_gapi(key_name: nil, 
+                    customer_managed_config_restriction_mode: nil, 
+                    customer_supplied_config_restriction_mode: nil, 
+                    google_managed_config_restriction_mode: nil)
+  
+    cm_config = customer_managed_config(restriction_mode: customer_managed_config_restriction_mode) if customer_managed_config_restriction_mode
```
bajajneha27: Where is this method `customer_managed_config` defined?

bajajneha27 (changes requested):

google-cloud-storage/acceptance/storage/bucket_encryption_test.rb
shubhangi-google: got reference from [Java pr](https://github.com/googleapis/java-storage/pull/3369/changes#diff-52fc201abe871de221ef35daba734699110b5854fbc2ea48015326d1b1265556R57)

google-cloud-storage/samples/Gemfile
shubhangi-google: we did the changes for main gemfile in another PR , while working on this PR we started getting the same issues for samples so did the same for samples gemfile
can we introduce these changes here or shall I create a separate PR for this?

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: When using the standard `bucket.update` or `@gapi.send(:encryption)` patterns, the Ruby SDK was capturing the entire existing encryption state. This included server-generated fields like `effectiveTime`. When this "polluted" object was sent back via the PATCH API, Google rejected it with an `InvalidArgumentError` because those fields were not in the expected format.

## The Fix
We implemented the [update_bucket_encryption_enforcement_config](url)method to send only the changed data to the patch request.

### How it Works
* **Isolation:** Instead of modifying the existing bucket object, the method initializes a fresh, empty `Google::Apis::StorageV1::Bucket::Encryption` instance.
* **Specific Assignment:** It uses dynamic dispatch (`public_send`) to set only the specific enforcement type requested (e.g., `customer_supplied`).

update_bucket_encryption_enforcement_config sends the correct patch json in bucket_encryption_config

---

### Comparison of JSON Payloads

#### Wrong Request (Polluted)
```json
{
  "encryption": {
    "customerManagedEncryptionEnforcementConfig": {
      "restrictionMode": "NotRestricted"
    },
    "customerSuppliedEncryptionEnforcementConfig": {
      "restrictionMode": "NotRestricted"
    },
    "googleManagedEncryptionEnforcementConfig": {
      "effectiveTime": "2026-03-25T11:40:17.182+00:00",
      "restrictionMode": "FullyRestricted"
    }
  }
}
````
#### Correct request
```json

{
  "encryption": {
    "customerManagedEncryptionEnforcementConfig": {
      "restrictionMode": "NotRestricted"
    }
  }
}

Comment (shubhangi-google) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: @bajajneha27 can you please provide your views on this approach

File: google-cloud-storage/samples/Gemfile (hunk 6 above)
Comment (shubhangi-google) [reply google-cloud-storage/samples/Gemfile:40]: @bajajneha27 can you please suggest here

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb (hunk 2 above)
Comment (shubhangi-google) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: @bajajneha27 did not understand your concerns here 
we are performing the operation as below
#set operation
customer_managed_config =
    Google::Apis::StorageV1::Bucket::Encryption::CustomerManagedEncryptionEnforcementConfig.new(
      restriction_mode: "NotRestricted"
    )
 customer_managed_encryption_enforcement_config = customer_managed_config
 
 #get operation
 bucket.customer_managed_encryption_enforcement_config
 #<Google::Apis::StorageV1::Bucket::Encryption::CustomerManagedEncryptionEnforcementConfig:0x00007f64e9a69dc0
 @effective_time=#<DateTime: 2026-04-02T10:09:23+00:00 ((2461133j,36563s,115000000n),+0s,2299161j)>,
 @restriction_mode="FullyRestricted">

File: google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb (hunk 7 above)
Comment (shubhangi-google) [reply google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb]: updated the code to avoid confusion

File: google-cloud-storage/test/helper.rb (hunk 8 above)
Comment (shubhangi-google) [reply google-cloud-storage/test/helper.rb]: updated the code to avoid confusion

File: google-cloud-storage/acceptance/storage/bucket_encryption_test.rb (hunk 1 above)
Comment (bajajneha27) [reply google-cloud-storage/acceptance/storage/bucket_encryption_test.rb]: @nidhiii-27 Is this correct statement?

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb (hunk 2 above)
Comment (bajajneha27) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: What I meant to ask is that, can we use the setter in following way as well:

```
bucket.customer_managed_config = { restriction_mode: "NotRestricted"
```

Because we usually do that in this codebase.

Comment (bajajneha27) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: A gentle reminder on this

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb [hunk 9]
```diff
@@ -716,6 +716,154 @@ def default_kms_key= new_default_kms_key
...
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #
+        #   new_config = Google::Apis::StorageV1::Bucket::Encryption::GoogleManagedEncryptionEnforcementConfig.new restriction_mode: "FullyRestricted"
+        #   bucket.update_bucket_encryption_enforcement_config new_config
+        #
+        def update_bucket_encryption_enforcement_config incoming_config
+          attr_name = case incoming_config
```
bajajneha27: AFAIU, this code will work only if we update just one config. What if I want to update multiple configs at the same time? For ex. `GoogleManagedEncryptionEnforcementConfig` and `CustomerManagedEncryptionEnforcementConfig`

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,154 @@ def default_kms_key= new_default_kms_key
...
+        # For more information, see [Bucket encryption](https://docs.cloud.google.com/storage/docs/encryption/).
+        # @return [Google::Apis::StorageV1::Bucket::Encryption::CustomerSuppliedEncryptionEnforcementConfig, nil] 
+        # The bucket's encryption configuration, or `nil` if no encryption configuration has been set.
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   
+        #   new_config = Google::Apis::StorageV1::Bucket::Encryption::CustomerSuppliedEncryptionEnforcementConfig.new restriction_mode: "NotRestricted"
```
bajajneha27: The code sample here is for setting the config while the method is a getter here. We should update the sample to show how to use the getter method

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,154 @@ def default_kms_key= new_default_kms_key
...
+        # requirements for the bucket.
+        # For more information, see [Bucket encryption](https://docs.cloud.google.com/storage/docs/encryption/).
+        # @return [Google::Apis::StorageV1::Bucket::Encryption::GoogleManagedEncryptionEnforcementConfig, nil]
+        # The bucket's encryption configuration, or `nil` if no encryption configuration has been set.
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   new_config= Google::Apis::StorageV1::Bucket::Encryption::GoogleManagedEncryptionEnforcementConfig.new restriction_mode: "NotRestricted"
```
bajajneha27: Same comment as above. This is a setter example

google-cloud-storage/lib/google/cloud/storage/bucket.rb
bajajneha27: User can still pass `effectiveTime` if they want to, in the new attribute. 
I think we should not add a separate parameter just for this. We should make use of existing attribute

bajajneha27: Besides, if user wants to update other attributes AND bucket_encryption_configs, that won't work 
https://github.com/googleapis/google-cloud-ruby/pull/33452/changes#diff-1c445f12b16180bcbdde4cab77c9b5ca55acc3fff5a45c1dd7d2bb4abb64bf76R3410-R3414

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
```diff
@@ -88,9 +87,14 @@
 
     it "sets its encryption config to nil" do
       bucket_gapi_with_key = bucket_gapi.dup
-      bucket_gapi_with_key.encryption = encryption_gapi(kms_key)
+      bucket_gapi_with_key.encryption = encryption_gapi(key_name: kms_key)
       bucket_with_key = Google::Cloud::Storage::Bucket.from_gapi bucket_gapi_with_key, storage.service
-      patch_bucket_gapi = Google::Apis::StorageV1::Bucket.new encryption: encryption_gapi(nil)
+      patch_bucket_gapi = Google::Apis::StorageV1::Bucket.new encryption: encryption_gapi(key_name: kms_key)
+      patch_bucket_gapi = Google::Apis::StorageV1::Bucket.new(
```
bajajneha27, line 92: Why are we just overwriting the variable `patch_bucket_gapi` ?

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: Hi @bajajneha27   the  existing patch method is sending all the encryption attribute (the existing one and the updated patch together) 
as shown in this example https://github.com/googleapis/google-cloud-ruby/pull/33452#discussion_r3026578075
here the time format which we are receiving is UTC format from server and API is expecting [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) format  as per [encryption.googleManagedEncryptionEnforcementConfig.effectiveTime](https://docs.cloud.google.com/storage/docs/json_api/v1/buckets#encryption) hence we are getting invalid argument error in response
as a solution I am creating a new method to update the encryption enforcement config which will send only the updated patch
this method is only for handling encryption_config so concern for other attributes will not be valid in this case

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the samples

shubhangi-google: modified the samples

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the code now the code will work for the hash we receive

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the samples

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the samples

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
shubhangi-google: was a mistake corrected it

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the samples

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the samples

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,186 @@ def default_kms_key= new_default_kms_key
...
+        #   Pass `nil` to clear the current configuration.
+        #
+        # @example Enforcing Customer-Managed Encryption
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #
+        #   # Set restriction mode to FullyRestricted
+        #   new_config = { restriction_mode: "FullyRestricted" }
```
bajajneha27: Thanks for updating this. We should also add example of how we can set it using the request object.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,186 @@ def default_kms_key= new_default_kms_key
...
+        #
+        # @return [void]
+        # @raise [Google::Cloud::Error] If the API request fails (e.g., insufficient permissions).
+        def update_bucket_encryption_enforcement_config incoming_config
+          allowed_keys = [
+            :google_managed_encryption_enforcement_config,
+            :customer_managed_encryption_enforcement_config,
+            :customer_supplied_encryption_enforcement_config
+          ]
+          # binding.pry
```
bajajneha27: Looks like this is added by mistake

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,186 @@ def default_kms_key= new_default_kms_key
...
+        # @param new_customer_supplied_encryption_enforcement_config [Hash, nil]
+        #   The configuration hash for encryption enforcement. 
+        #   * `:restriction_mode` (String) - Can be "NotRestricted" or "FullyRestricted".
+        #   Pass `nil` to clear the current configuration.
+        # @example
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   new_config = { restriction_mode: "FullyRestricted" }
```
bajajneha27: Same here. Please add example of how to use request object as well.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
bajajneha27: I don't follow. Why would you get `Invalid Argument` error if you're not passing `effective_time` at all ? 
What happens if I want to pass `bucket_encryption_config` and other attributes as well?

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: Example we want to update customerManagedEncryptionEnforcementConfig restriction mode to "FullyRestricted"
When using the standard bucket.update the [@gapi.send(:encryption)](https://github.com/googleapis/google-cloud-ruby/pull/33452/changes#diff-1c445f12b16180bcbdde4cab77c9b5ca55acc3fff5a45c1dd7d2bb4abb64bf76L3261) patterns, the Ruby SDK is capturing the entire existing encryption state and updating the customerManagedEncryptionEnforcementConfig restriction mode in it and send the updated encryption hash .  which includes effective time we are receiving from API which is in UTC format 
example:
```json
{
  "encryption": {
    "customerManagedEncryptionEnforcementConfig": {
      "restrictionMode": "NotRestricted"
    },
    "customerSuppliedEncryptionEnforcementConfig": {
      "restrictionMode": "NotRestricted"
    },
    "googleManagedEncryptionEnforcementConfig": {
      "effectiveTime": "2026-03-25T11:40:17.182+00:00",
      "restrictionMode": "FullyRestricted"
    }
  }
}
````
when this hash is submitted  to server its returning error 
`Google::Cloud::InvalidArgumentError: invalid: Invalid argument.` because server is expecting the effectiveTime in `1985-04-12T23:20:50.52Z` format and not in UTC format

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb (hunk 14 above)
Comment (shubhangi-google) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: removed

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb (hunk 13 above)
Comment (shubhangi-google) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: done

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb (hunk 15 above)
Comment (shubhangi-google) [reply google-cloud-storage/lib/google/cloud/storage/bucket.rb]: dine

File: google-cloud-storage/lib/google/cloud/storage/bucket.rb [hunk 16]
```diff
@@ -716,6 +716,228 @@ def default_kms_key= new_default_kms_key
...
+            extra_keys = input_keys - allowed_keys
+            unless extra_keys.empty?
+              raise ArgumentError, "Invalid config detected: #{extra_keys.join(', ')}. " \
+                                  "Only #{allowed_keys.join(', ')} are allowed."
+            end
+
+          elsif incoming_config.is_a? Google::Apis::StorageV1::Bucket::Encryption
+            # For objects, ensure at least one of the allowed enforcement configs is present
+            has_any_config = allowed_keys.any? { |key| !incoming_config.send(key).nil? }
+            binding.pry
```
bajajneha27: Looks like this is added by mistake.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,228 @@ def default_kms_key= new_default_kms_key
...
+
+            extra_keys = input_keys - allowed_keys
+            unless extra_keys.empty?
+              raise ArgumentError, "Invalid config detected: #{extra_keys.join(', ')}. " \
+                                  "Only #{allowed_keys.join(', ')} are allowed."
+            end
+
+          elsif incoming_config.is_a? Google::Apis::StorageV1::Bucket::Encryption
+            # For objects, ensure at least one of the allowed enforcement configs is present
+            has_any_config = allowed_keys.any? { |key| !incoming_config.send(key).nil? }
```
bajajneha27: I think we can get rid of these validations. These would be rejected by server if the key doesn't match.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
bajajneha27: I see. So, you don't update the entire hash/object as is. You make sure that you're not sending `effectiveTime` for updating. Because #384-3488 will not work for all cases. If I want to update bucket_encryption_config and other attributes together, it won't work.

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: particular method which this comment was referenced has been removed

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: particular method which this comment was referenced has been removed

google-cloud-storage/acceptance/storage/bucket_encryption_test.rb
shubhangi-google: check this statement was not relevant to ruby so removed it

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: particular method which this comment was referenced has been removed

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: modified the code to use the bucket.update method for updating the bucket encryption configs

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -3386,6 +3552,26 @@ def check_for_mutable_lifecycle!
...
+              :customer_supplied_encryption_enforcement_config
+            ].each do |attr|
+              config = @gapi.encryption.send(attr)
+              next unless config
+              unless config.respond_to?(:to_h)
+                raise ArgumentError, "Encryption config for #{attr} must be a Hash or valid Config object"
+              end
+              clean_config = config.to_h
+              clean_config.delete :effective_time
+              clean_config.delete "effective_time"
```
bajajneha27, line 3570: Repeated lines

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
```diff
@@ -122,6 +125,126 @@
...
+
+      bucket.service.mocked_service = mock
+      _(bucket.customer_managed_encryption_enforcement_config.restriction_mode).must_equal "FullyRestricted"
+      _(bucket.customer_supplied_encryption_enforcement_config.restriction_mode).must_equal "NotRestricted"
+      _(bucket.google_managed_encryption_enforcement_config.restriction_mode).must_equal "NotRestricted"
+
+      bucket.update do |b|
+        b.google_managed_encryption_enforcement_config = incoming_config
+      end
+      # bucket.update_bucket_encryption_enforcement_config incoming_config
```
bajajneha27: We can remove this?

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
```diff
@@ -122,6 +125,126 @@
...
+
+      returned_bucket_gapi = bucket_gapi.dup
+      returned_bucket_gapi.encryption = Google::Apis::StorageV1::Bucket::Encryption.new default_kms_key_name: kms_key
+
+      mock.expect :update_bucket, returned_bucket_gapi do |name, patch_obj, **kwargs|
+        cm_config = patch_obj.encryption.customer_managed_encryption_enforcement_config
+        cs_config = patch_obj.encryption.customer_supplied_encryption_enforcement_config
+        gm_config = patch_obj.encryption.google_managed_encryption_enforcement_config
+
+        name == bucket_name &&
```
bajajneha27: Why are we doing this?

google-cloud-storage/lib/google/cloud/storage/bucket.rb
shubhangi-google: As you know Ruby hashes can be "Symbol-keyed" or "String-keyed." Depending on how the data was parsed from JSON, the effective_time might exist in either format.
`clean_config.delete :effective_time` targets the Symbol key.
`clean_config.delete "effective_time"` targets the String key.
Manually deleting both ensures the field is definitely gone, regardless of hash type.

shubhangi-google:
**Technical Update (for future reference ):** Conditional "Effective Time" Behavior
In our latest testing of the encryption enforcement configs, we confirm that effectiveTime field is state-dependent:

When Enabling Restrictions (Not Restricted → Fully Restricted): The API returns an effectiveTime. This is the server’s way of confirming exactly when the new security policy became active.

When Disabling Restrictions (Fully Restricted → Not Restricted): The API omits the effectiveTime. Since the restriction has been removed, the field is no longer applicable and is stripped from the object hash.

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
shubhangi-google: done

google-cloud-storage/test/google/cloud/storage/bucket_encryption_test.rb
shubhangi-google: updated the mock expect method

bajajneha27 (approved):

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,171 @@ def default_kms_key= new_default_kms_key
...
+        #   Pass `nil` to clear the current configuration.
+        #
+        # @example Enforcing Customer-Managed Encryption
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   # Set restriction mode to FullyRestricted
+        #   new_config = { restriction_mode: "FullyRestricted" }
+        #   bucket.new_google_managed_encryption_enforcement_config = new_config
```
cpriti-os: @shubhangi-google new_ prefix shouldn't be present in this example? Please fix.

shubhangi-google: done

google-cloud-storage/lib/google/cloud/storage/bucket.rb
```diff
@@ -716,6 +716,171 @@ def default_kms_key= new_default_kms_key
...
+        #   Pass `nil` to clear the current configuration.
+        #
+        # @example Enforcing Customer-Managed Encryption
+        #   require "google/cloud/storage"
+        #
+        #   storage = Google::Cloud::Storage.new
+        #   bucket = storage.bucket "my-bucket"
+        #   # Set restriction mode to FullyRestricted
+        #   restricted_config = { restriction_mode: "FullyRestricted" }
+        #   bucket.new_google_managed_encryption_enforcement_config = restricted_config
```
cpriti-os: I meant the new in ` bucket.new_google_managed_encryption_enforcement_config`

shubhangi-google: done

PR #33812, opened 2026-04-08
feat: Update minimum Ruby to v3.2 and support v4.0

aandreassa:


quartzmo (approved):
Compared with #33811. LGTM

PR #33813, opened 2026-04-08
feat(google-cloud-trace): Upgrade to Ruby 3.2 minimum

aandreassa:
Re-enabling a test and fixing the other. 

It was previously skipped in https://github.com/googleapis/google-cloud-ruby/commit/63baac6f5c470ec399f4e42a90b57185be72857c

1. Added a check for polling trace.
2. Added `+span` to filter in order to perform exact name match. Using prefix match as before results in a lot of timeouts.
3. Other test fixtures & gem updates due to Ruby version updates.

https://docs.cloud.google.com/trace/docs/trace-filters


aandreassa:
Re-runnng flaky test

google-cloud-trace/acceptance/trace/trace_test.rb
```diff
@@ -23,9 +23,10 @@
       trace = wait_until do
         tracer.get_trace orig_trace.trace_id
       end
-      trace.must_equal orig_trace
+      _(trace).must_equal orig_trace
     end
 
+    focus
```
quartzmo, line 29: Did you intend to leave `focus` in place here?

quartzmo (changes requested):

google-cloud-trace/acceptance/trace/trace_test.rb
aandreassa: I removed `focus` in the second commit. It shouldn't be there anymore!

quartzmo (approved):

PR #33886, opened 2026-04-24
fix(bigquery): use destination table's project_id for queries

quartzmo:
* Add `destination_table_project_id` helper method.
* Update `data` method to pass `destination_table_project_id` to `service.list_tabledata`.

closes: #33810


aandreassa (approved):
LGTM! Thanks

aandreassa:
The fix is great. Did you mean to update the dependency gems as well?

quartzmo:
> Did you mean to update the dependency gems as well?

That comment was outdated, removed it from commit and PR desc.

PR #33896, opened 2026-04-29
fix(storage): Remove accept encoding gzip for metadata

shubhangi-google:
This pull request modifies the google-cloud-storage library to move the Accept-Encoding: gzip header from a global request.
associated Core library Pr where the header is being added back for media operation: https://github.com/googleapis/google-api-ruby-client/pull/26530


bajajneha27:
LGTM, can we please add some tests? Also is manual testing done for this?

shubhangi-google:
> LGTM, can we please add some tests? Also is manual testing done for this?

@bajajneha27  Added test to ensure accept encoding header is not added by default
yes we did manual testing to ensure header is not added by default and is being added for media operations only via core library

bajajneha27:
Possible to add integration test as well?

Also, until you upgrade core library version, this change won't actually work. We need to do that before we can merge this PR.

bajajneha27 (approved):

PR #33979, opened 2026-05-22
chore: Update dependencies to unblock CI

aandreassa:
These GAPICs need to wait for the next generator run to be unblock their actual releases. 

I manually ran the generator with minimal fixes to unblock CI. 

chore is correct as of these GAPICs have pending PRs.


viacheslav-rostovtsev (approved):
LGTM

PR #34019, opened 2026-06-05
chore: Add pubsub rollback schema sample

torreypayne:
Adds new samples and acceptance tests for Pub/Sub schema evolution features, mirroring the equivalent changes in Java (https://github.com/googleapis/java-pubsub/pull/1469).

### DIFF
--- a/google-cloud-pubsub/samples/acceptance/schemas_test.rb
+++ b/google-cloud-pubsub/samples/acceptance/schemas_test.rb
@@ -29,6 +29,8 @@ require_relative "../pubsub_subscribe_avro_records"
 require_relative "../pubsub_publish_proto_messages"
 require_relative "../pubsub_subscribe_proto_messages"
 require_relative "../pubsub_subscribe_avro_records_with_revisions"
+require_relative "../pubsub_rollback_schema"
+require_relative "../pubsub_update_topic_schema"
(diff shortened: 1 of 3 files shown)


aandreassa (changes requested):
It looks great! Can you please make sure the tests passes? (unrelated, but I wonder if it is just a flaky thing). 

You can retry by adding `kokoro:force-run` label to this PR and it will update eventually.

torreypayne:
> It looks great! Can you please make sure the tests passes? (unrelated, but I wonder if it is just a flaky thing).
> 
> You can retry by adding `kokoro:force-run` label to this PR and it will update eventually.

@aandreassa I landed the fix to the flaky tests and rebased; builds are green!

aandreassa (approved):
Woo! LGTM

aandreassa:
You should see a similar case in this PR when merging (i.e. difference in title vs commit text). 

We do `chore` for samples as they don't need an actual publication to Rubygems.

torreypayne:
> You should see a similar case in this PR when merging (i.e. difference in title vs commit text).
> 
> We do `chore` for samples as they don't need an actual publication to Rubygems.

Updated the commit message to be `chore(...)`! Going to merge

PR #34026, opened 2026-06-08
test(pubsub): fix flakiness and formatting issues in schemas acceptance tests

### LINKED ISSUE #34025: Fix flakiness and formatting issues in google-cloud-pubsub schemas acceptance tests
#### Environment details

  - OS: Linux (Kokoro run) (see: https://btx.cloud.google.com/invocations/6057621a-4719-4302-a701-bf973c011ec7/targets/cloud-devrel%2Fclient-libraries%2Fgoogle-cloud-ruby%2Fpresubmit%2Fsamples/log)
  - Ruby version: 3.2-4.0
  - Gem name and version: google-cloud-pubsub

#### Steps to reproduce

  1. I submitted https://github.com/googleapis/google-cloud-ruby/pull/34019
  2. I noticed test failures in `schemas/acceptance_test.rb`; the PR changes did not overlap with those tests, so these seem to be flaky failures that affect throughput. This issue is to track the fix to address this CI flakiness.

#### Code example

```ruby
    expect_with_retry "pubsub_subscribe_avro_records_with_revisions" do
        assert_output /Received a binary-encoded message:.*Alaska.*Received a binary-encoded message:.*California/m do
          subscribe_avro_records_with_revisions subscription_id: @subscription.name
        end
      end
```

#### Full backtrace

```sh
Finished in 1505.708359s, 0.0345 runs/s, 0.1561 assertions/s.

  1) Failure:
schemas::AVRO#test_0006_supports pubsub_subscribe_avro_records_with_revisions [acceptance/schemas_test.rb:265]:
In stdout.
Expected /Received a binary-encoded message:.*Alaska.*Received a binary-encoded message:.*California/m to match "".

52 runs, 235 assertions, 1 failures, 0 errors, 8 skips
```

torreypayne:
closes: #34025

This PR fixes flakiness and formatting issues in the schemas_test.rb acceptance tests for google-cloud-pubsub samples. These issues were causing CI failures, particularly on newer Ruby versions (3.3+).

### Root Causes & Fixes

**1. Ruby 3.3 Hash Formatting Difference:**

Issue: Avro subscriber tests used assert_output with exact string matching. The expected output had spaces around => (e.g., {"name" => "Alaska"}), but Ruby 3.3 represents Hashes without spaces (e.g., {"name"=>"Alaska"}), causing consistent failures on attempt 1.
Fix: Updated the assertions to use space-tolerant regular expressions (e.g., /Received a binary-encoded message:\n\{"name"\s*=>\s*"Alaska",\s*"post_abbr"\s*=>\s*"AK"\}/).

**2. Subscription Propagation Delay (Flakiness):**

Issue: Messages were being published immediately after subscription creation. Because Pub/Sub subscription creation is eventually consistent, this occasionally led to messages being lost if they were published before the subscription was active in all zones.
Fix: Added sleep 5 after subscription creation in the Avro tests, matching the behavior already present in the Proto tests.

**3. Out-of-Order Message Delivery:**

Issue: The revisions test published two messages (one for Rev A, one for Rev B) and asserted their receipt using a single regex that enforced order (Alaska then California). Since Pub/Sub delivery is unordered by default, they could arrive in reverse order, causing the assertion to fail.
Fix: Changed the assertion to use capture_io and separate assert_match calls for each expected message, making the test order-independent.

**4. Shadowed Errors in expect_with_retry:**

Note: The test helper expect_with_retry retries the test block up to 5 times. If the first attempt failed (due to the issues above) but acknowledged the messages, subsequent attempts would receive nothing and fail with actual: "". Minitest then reported the last failure, shadowing the actual root cause of the initial failure. Fixing the root causes above prevents this retry-failure loop.

**Verification Results**
Verified locally on fix-acceptance-tests-flakiness branch (which is based on main and only contains these fixes):

```bash

GOOGLE_APPLICATION_CREDENTIALS=... GOOGLE_CLOUD_PROJECT=helical-zone-771 bundle exec ruby acceptance/schemas_test.rb
```

Output:

Finished in 392.020107s, 0.0332 runs/s, 0.1377 assertions/s.
13 runs, 54 assertions, 0 failures, 0 errors, 0 skips
All existing tests passed successfully on the first attempt.

### DIFF
--- a/google-cloud-pubsub/samples/acceptance/schemas_test.rb
+++ b/google-cloud-pubsub/samples/acceptance/schemas_test.rb
@@ -161,16 +161,15 @@ describe "schemas" do
       @subscription = subscription_admin.create_subscription name: pubsub.subscription_path(random_subscription_id),
                                                              topic: @topic.name,
                                                              ack_deadline_seconds: 60
-      sleep 5
 
       writer = Avro::IO::DatumWriter.new avro_schema
       buffer = StringIO.new
       writer.write record, Avro::IO::BinaryEncoder.new(buffer)
       publisher = pubsub.publisher @topic.name
-      publisher.publish buffer
 
       # pubsub_subscribe_avro_records
       expect_with_retry "pubsub_subscribe_avro_records" do
+        publisher.publish buffer
         expected = /
           Received\ a\ binary-encoded\ message:\n
           \{"name"\s*=>\s*"Alaska",\s*
@@ -193,13 +192,12 @@ describe "schemas" do
       @subscription = subscription_admin.create_subscription name: pubsub.subscription_path(random_subscription_id),
                                                              topic: @topic.name,
                                                              ack_deadline_seconds: 60
-      sleep 5
 
       publisher = pubsub.publisher @topic.name
-      publisher.publish record.to_json
 
       # pubsub_subscribe_avro_records
       expect_with_retry "pubsub_subscribe_avro_records" do
+        publisher.publish record.to_json
         expected = /
           Received\ a\ JSON-encoded\ message:\n
           \{"name"\s*=>\s*"Alaska",\s*
@@ -252,14 +250,13 @@ describe "schemas" do
       @subscription = subscription_admin.create_subscription name: pubsub.subscription_path(random_subscription_id),
                                                              topic: @topic.name,
                                                              ack_deadline_seconds: 60
-      sleep 5
+
+      publisher = pubsub.publisher @topic.name
 
       # Publish message 1 (Old format - valid for both).
       writer = Avro::IO::DatumWriter.new avro_schema
       buffer = StringIO.new
       writer.write record, Avro::IO::BinaryEncoder.new(buffer)
-      publisher = pubsub.publisher @topic.name
-      publisher.publish buffer
 
       # Publish message 2 (New format - valid only for Rev B).
       avsc_definition_plus = File.read avsc_revision_file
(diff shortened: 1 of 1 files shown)


google-cloud-pubsub/samples/acceptance/schemas_test.rb
```diff
@@ -161,6 +161,7 @@
       @subscription = subscription_admin.create_subscription name: pubsub.subscription_path(random_subscription_id),
                                                              topic: @topic.name,
                                                              ack_deadline_seconds: 60
+      sleep 5
```
quartzmo: I wonder if instead of using a hard-coded sleep (guaranteed to add 5s to the test suite), we could use `expect_with_retry ` from `google-cloud-ruby/google-cloud-pubsub/samples/acceptance/helper.rb` instead? It's used in other tests for this gem. Maybe something like: 

```
# pubsub_subscribe_avro_records
expect_with_retry "pubsub_subscribe_avro_records" do
  # Publish the message inside the retry loop
  publisher = pubsub.publisher @topic.name
  publisher.publish record.to_json

  expected = /.../
  assert_output expected do
    subscribe_avro_records subscription_id: @subscription.name, avsc_file: nil
  end
end
```

torreypayne: ya I don't like the `sleep 5` solution at all and I've been digging into an alternative fix; let me try that out.

torreypayne: @quartzmo LMK what you think about this latest iteration; I pushed the .publish calls into the expect_with_retry blocks and build is green and faster.

quartzmo (approved):

aandreassa (approved):
TY!!

PR #34623, opened 2026-06-17
test(compute-v1): Address flaky tests

aandreassa:
Fixes intermittent test failures caused by `Quota exceeded for quota metric 'Read requests'` on Compute Engine API calls.

### Changes:
1. **Reduce LRO polling frequency**: Increased `max_delay` from `1` to `5` seconds in `samples/quickstart.rb` to prevent tests from hitting the API once per second while waiting for resources (firewalls, instances).
2. **Increase pagination page size**: Set `max_results: 100` instead of `10` in `PaginationSmokeTest` for aggregated lists. This ensures all zones are retrieved in a single request rather than paginating through 10+ pages, reducing request overhead.


aandreassa:
Still figuring this out!

PR #34653, opened 2026-06-22
feat(pubsub): implement streaming keep-alive logic

torreypayne:
## Overview
Implements proactive streaming keep-alive logic and connection health monitoring in `Google::Cloud::PubSub::MessageListener::Stream`, mirroring the design implemented in the .NET Pub/Sub client ([dotnet#15649](https://github.com/googleapis/google-cloud-dotnet/pull/15649)).

Long-running bi-directional gRPC streaming pull connections (`StreamingPull`) can experience silent TCP drops, intermediary network timeouts, or read deadlocks during periods of low message volume. This change introduces background timer tasks to push regular keep-alive requests and actively monitor server Pong timestamps.

## Key Changes
* **Protocol Version Initialization**: Explicitly initializes `protocol_version = 1` on the initial `StreamingPullRequest` protobuf to enable bi-directional stream keep-alive support.
* **Unconditional Keep-Alive Pings**: Configures a background timer task (`@stream_keepalive_task`) to dispatch empty `StreamingPullRequest` pings at regular intervals (default 30 seconds), regardless of current lease inventory volume.
* **Pong Monitoring & Automatic Reconnection**: Introduces `@pong_monitor_task` to inspect timestamps (`@last_ping_at`, `@last_pong_at`). If a keep-alive response is overdue by more than `pong_deadline` seconds (default 15 seconds), the monitor raises `RestartStream` to safely recycle the connection and back off.
* **Concurrency Timestamp Guard**: Guards ping timestamp updates (`@last_ping_at = now if @last_pong_at >= @last_ping_at`) to ensure consecutive un-ponged pings cannot overwrite the timestamp of an overdue request.

## Testing & Validation
* **Unit Test Suite (`keepalive_test.rb`)**: Added targeted unit test coverage asserting protocol version flags, timer intervals, deadline timeouts, and non-disruptive Pong handling.
* **Resiliency & Robustness Suite**: Validated against live GCP test instances (`helical-zone-771`) across simulated TCP socket hangs, sub-millisecond deadline starvation, and post-recovery downstream message delivery.

Fixes b/427319802


robertvoinescu-work (commented):
LGTM from a functional perspective. Just a few minor comments.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -264,6 +316,12 @@ def background_run
               begin
                 # Cannot synchronize the enumerator, causes deadlock
                 response = enum.next
+                synchronize do
+                  @last_pong_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+                  # Reset backoff delay only after successfully reading a frame from enum.next.
+                  # If the connection drops immediately upon reading, @reconnect_delay is preserved.
+                  @reconnect_delay = nil
+                end
                 new_exactly_once_delivery_enabled = response&.subscription_properties&.exactly_once_delivery_enabled
```
robertvoinescu-work: You should add logic here to avoid setting fields off a pong response (an empty array of messages). The subscription properties on pongs are not valid.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -68,24 +69,55 @@ def initialize subscriber
...
+            @last_pong_at = nil
+            @stream_opened = false
+            @reconnect_delay = nil
+
             @stream_keepalive_task = Concurrent::TimerTask.new(
-              execution_interval: 30
+              execution_interval: @keepalive_interval
+            ) do
+              synchronize do
+                if @stream_opened && !@stopped && @request_queue
```
robertvoinescu-work: My understanding is that the `request_queue` contains the ACK responses we send back to Pub/Sub. I believe we should send keep-alive pings regardless of whether there are items in this queue. Currently, in the case of a slow message stream (e.g., receiving a message only every 31 seconds), the service will close the stream due to inactivity. This would force us to establish a new streaming connection every time we want to read.

torreypayne: Yup good call here; the `if \@request_queue` logic just checks for presence of the variable, but does _not_ check for items in the queue before delivery. So the current implementation aligns with your expectation 👍🏿

quartzmo:
@torreypayne 

Great PR! ❤️ 

Three things for you to look into:

### 1. Unpause Race Condition
In the [monitor task check](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L103-L112), you disable the liveness checker when flow control pauses the stream (`!@paused`). The read path is blocked and incoming pongs will sit in the gRPC buffers.

However, there is a potential race condition when the stream is unpaused:
1. While paused, the keepalive task still sends pings but `@last_ping_at` is frozen at the first unponged ping.
2. When [`unpause_streaming!`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L486) is called, it sets `@paused = nil` and signals `@pause_cond`.
3. If the `@pong_monitor_task` thread checks the state (every 6s) *after* `@paused` becomes `nil` but *before* the background thread has had time to wake up, call `enum.next`, read the buffered pong, and update `@last_pong_at`, it will detect a false-positive timeout as follows:
   * `@paused` is `nil` (satisfying the `!@paused` condition at [`stream.rb#L103`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L103)).
   * The elapsed time check (`now - @last_ping_at >= @pong_deadline`) is true (at [`stream.rb#L105`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L105)) because `now - @last_ping_at` is large (representing the pause duration).
   * `@last_pong_at < @last_ping_at` is `true` (at [`stream.rb#L105`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L105)).
4. This will trigger an immediate, false-positive stream restart.

Assuming a background thread wake-up and gRPC read delay of around 10ms alongside a 6-second monitor interval, there is roughly a 0.17% chance of encountering this race on any given unpause event. In a busy production environment where streams pause and unpause frequently, this could potentially scale up to thousands of false-positive stream restarts daily in aggregate. That in turn might lead to duplicate message redeliveries, CPU overhead from renegotiations, and unnecessary warning logs.

Suggested Fix:

Inside `Stream#unpause_streaming!`, reset `@last_pong_at` to the current monotonic time. Additionally, since the method broadcasts to the `@pause_cond` condition variable, we should wrap its body in a `synchronize` block to guarantee thread safety (and to avoid `ThreadError: current fiber not owner` when calling it from tests):

```ruby
def unpause_streaming!
  synchronize do
    return unless unpause_streaming?

    @paused = nil
    @last_pong_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    subscriber.service.logger.log :info, "subscriber-flow-control" do
      "subscriber for #{@subscriber.subscription_name} is unblocking client-side flow control"
    end
    # signal to the background thread that we are unpaused
    @pause_cond.broadcast
  end
end
```

This lock-safe reset ensures that the "last pong" appears newer than the stale ping sent before/during the pause, preventing immediate false-positive restarts while the background thread wakes up to read the buffered pongs.
 

If the stream was indeed dead, the detection will proceed as follows:
1. At the next keepalive task execution (line 90 in [`stream.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L90)), since `@last_pong_at >= @last_ping_at` is true (due to our reset), `@last_ping_at` is updated to the current time, and a new ping is sent.
2. Because the connection is dead, no pong will arrive, leaving `@last_pong_at` stale.
3. Once `pong_deadline` (15s) has passed, the monitor task (line 105 in [`stream.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L105)) will evaluate `now - @last_ping_at >= @pong_deadline && @last_pong_at < @last_ping_at` as true.
4. It will then call `@background_thread&.raise RestartStream` (line 110 in [`stream.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L110)), which restarts the stream in `background_run`'s rescue block (line 379 in [`stream.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L379)).

This delay of at most 15–45 seconds in detecting a dead connection after a long pause is a negligible trade-off to eliminate false-positive restarts.

### 2. TimerTask Lifecycle & Testability
Currently, `@stream_keepalive_task` and `@pong_monitor_task` are initialized inside the constructor [`initialize`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L55), executed in [`start`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L117), and shut down in [`stop`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb#L131).

There are two main issues with this setup:

1. Reusability: Calling `Concurrent::TimerTask#shutdown` permanently terminates its executor thread pool. Once shut down, the task cannot be re-run by calling `.execute` again. If the stream is ever stopped and subsequently restarted, calling `start` again will fail to execute the timers.
2. Testability: The core logic for sending pings and monitoring liveness is embedded directly inside anonymous blocks passed to `Concurrent::TimerTask.new`, where it is very difficult to test in isolation. There is an opportunity here to add fine-grained unit tests in [`stream_test.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/test/google/cloud/pubsub/message_listener/stream_test.rb) to cover this logic.

Suggested Fix:

To resolve both issues:
1. Extract the timer block contents into two private, synchronous helper methods on `Stream`: `send_keepalive_ping!` and `check_liveness!`.
2. Initialize both timer instance variables to `nil` in the constructor.
3. In `start`, instantiate and run the timers directly inline.
4. In `stop`, call `shutdown` and set both variables back to `nil`.

For example:

```ruby
def start
  synchronize do
    break if @background_thread

    @inventory.start

    @stream_keepalive_task = Concurrent::TimerTask.new(
      execution_interval: @keepalive_interval
    ) { send_keepalive_ping! }
    @stream_keepalive_task.execute

    @pong_monitor_task = Concurrent::TimerTask.new(
      execution_interval: [@keepalive_interval / 5.0, 0.01].max
    ) { check_liveness! }
    @pong_monitor_task.execute

    start_streaming!
  end

  self
end

def stop
  synchronize do
    break if @stopped

    # ... other stop logic ...

    @stream_keepalive_task&.shutdown
    @stream_keepalive_task = nil
    @pong_monitor_task&.shutdown
    @pong_monitor_task = nil
  end
end

def send_keepalive_ping!
  synchronize do
    if @stream_opened && !@stopped && @request_queue
      subscriber.service.logger.log :info, "subscriber-streams" do
        "sending keepAlive to stream for subscription #{@subscriber.subscription_name}"
      end
      @last_ping_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) if @last_pong_at >= @last_ping_at
      push Google::Cloud::PubSub::V1::StreamingPullRequest.new
    end
  end
end

def check_liveness!
  synchronize do
    if @stream_opened && @last_ping_at && @last_pong_at && !@stopped && !@paused
      now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
      if now - @last_ping_at >= @pong_deadline && @last_pong_at < @last_ping_at
        subscriber.service.logger.log :error, "subscriber-streams" do
          "Keep-alive pong not received within #{@pong_deadline}s; restarting stream."
        end
        @stream_opened = false
        @background_thread&.raise RestartStream
      end
    end
  end
end
```

By extracting the execution logic to private methods `send_keepalive_ping!` and `check_liveness!`, you can now write direct, synchronous unit tests. In your unit tests, you can instantiate a mock stream, set up specific states (e.g., setting `@last_ping_at`, `@last_pong_at`, and `@paused` to various values), and invoke `check_liveness!` or `send_keepalive_ping!` synchronously to verify their behavior directly, without needing to manage real-time timer threads.

### 3. Test Coverage & CI Flakiness
The current tests in [`keepalive_test.rb`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb) are a mix of serialization checks and timing-based integration tests. 

Suggested testing strategy:

1. Retain serialization checks: The test verifying that `protocol_version = 1` is sent on the initial request is valuable and does not rely on timing or sleeps. This should be kept as a high-level integration test.
2. Retain one end-to-end integration test for stream restart: The test [`restarts stream when keep-alive pong deadline is exceeded`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb#L92) is valuable because it verifies the async timer scheduling, deadline check, and asynchronous thread interruption (`@background_thread&.raise RestartStream`). However, we should increase the polling timeout (e.g., changing `200` to `500` at [`keepalive_test.rb#L105`](https://github.com/googleapis/google-cloud-ruby/blob/pubsub-streaming-keepalive/google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb#L105)) to wait up to 5 seconds instead of 2 seconds, preventing false-positive test flakes in slow CI environments.
3. Replace the other timing-dependent tests: The remaining two timing-dependent integration tests (`sends keep-alive pings periodically...` and `does not restart stream when actively receiving...`) can be replaced with direct, synchronous unit tests on the new private methods. This allows us to verify ping frequency logic and normal liveness checks deterministically without relying on real-time sleeps and thread scheduling.

Additionally, we currently have no coverage for the edge cases described above (the unpause race condition, or the reusability of the timer tasks).

Once the logic is refactored to the `send_keepalive_ping!` and `check_liveness!` methods, you can cover them with unit tests.

Here is an example for the unpause race condition:

```ruby
it "does not trigger false restarts on unpausing even if pause exceeded deadline" do
  # Setup: stream opened, ping sent 30s ago, pong received 35s ago (stale), paused.
  stream = listener.stream_pool.first
  stream.instance_variable_set :@stream_opened, true
  stream.instance_variable_set :@last_ping_at, Process.clock_gettime(Process::CLOCK_MONOTONIC) - 30.0
  stream.instance_variable_set :@last_pong_at, Process.clock_gettime(Process::CLOCK_MONOTONIC) - 35.0
  stream.instance_variable_set :@paused, true
  stream.instance_variable_set :@stopped, false
  
  # 1. When paused, liveness checker should skip verification.
  stream.check_liveness!
  assert stream.instance_variable_get(:@stream_opened) # Stream remains open
  
  # 2. Unpausing must reset @last_pong_at to prevent immediate restart.
  stream.unpause_streaming!
  
  # 3. Direct liveness check immediately after unpausing (simulates the monitor thread racing the background thread).
  # It should NOT close the stream because our reset made @last_pong_at newer than @last_ping_at.
  stream.check_liveness!
  assert stream.instance_variable_get(:@stream_opened) # Stream remains open, no restart!
end
```

And here is an example for the `TimerTask` lifecycle:

```ruby
it "re-creates and executes timer tasks if stopped and restarted" do
  stream = listener.stream_pool.first
  
  # 1. Timers should be active on a started stream
  refute_nil stream.instance_variable_get(:@stream_keepalive_task)
  refute_nil stream.instance_variable_get(:@pong_monitor_task)
  
  first_keepalive = stream.instance_variable_get(:@stream_keepalive_task)
  first_monitor = stream.instance_variable_get(:@pong_monitor_task)
  
  # 2. Stopping the stream must shutdown and nilify the timers
  stream.stop
  assert_nil stream.instance_variable_get(:@stream_keepalive_task)
  assert_nil stream.instance_variable_get(:@pong_monitor_task)
  assert first_keepalive.shutdown?
  assert first_monitor.shutdown?
  
  # 3. Simulating a restart should create completely new timer instances
  stream.instance_variable_set :@background_thread, nil
  stream.instance_variable_set :@stopped, false
  stream.start
  
  new_keepalive = stream.instance_variable_get(:@stream_keepalive_task)
  new_monitor = stream.instance_variable_get(:@pong_monitor_task)
  
  refute_nil new_keepalive
  refute_nil new_monitor
  refute_equal first_keepalive, new_keepalive
  refute_equal first_monitor, new_monitor
end
```

torreypayne:
> @torreypayne
> 
> Great PR! ❤️
> 
> Three things for you to look into:
> ...


@quartzmo Thanks a ton for the thorough review; you identified several areas where I was able to make significant reliability improvements and clarify the keep-alive API. I addressed all 3 points that you identified; of note, I refactored the ping and pong logic into two private methods and added tests on those private methods explicitly in 'keepalive_test.rb', reset '@last_pong_at' when we unpause, and initialize the timerTasks as instance variables that we set/clear in `#start` and `#shutdown`. Please take another look when you have a moment and LMK how this looks.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -68,24 +69,31 @@ def initialize subscriber
...
-            ) do
-              # push empty request every 30 seconds to keep stream alive
-              unless inventory.empty?
-                subscriber.service.logger.log :info, "subscriber-streams" do
-                  "sending keepAlive to stream for subscription #{@subscriber.subscription_name}"
-                end
-                push Google::Cloud::PubSub::V1::StreamingPullRequest.new
-              end
-            end.execute
+            @keepalive_interval = Float(ENV["PUBSUB_TEST_KEEPALIVE_INTERVAL"] || 30)
```
aandreassa: I don't think we should be exposing these test-only environment variables (e.g. `PUBSUB_TEST_KEEPALIVE_INTERVA`) in production code.

`Stream` already has `attr_accessor` `:keepalive_interval`, `:pong_deadline`, we can customize these attributes directly on the stream objects in our test before calling `listener.start`

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -48,6 +48,7 @@ class Stream
           ##
           # @private exactly_once_delivery_enabled.
           attr_reader :exactly_once_delivery_enabled
+          attr_accessor :keepalive_interval, :pong_deadline
```
aandreassa: Please look at other examples in our libraries, but I believe we should split these into multiple lines. It would be great to add docstrings as well.

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/bug_regression_test.rb
```diff
@@ -0,0 +1,73 @@
...
+require "helper"
+
+describe Google::Cloud::PubSub::MessageListener, :bug_regression, :mock_pubsub do
+  let(:topic_name) { "topic-name-goes-here" }
+  let(:sub_name) { "subscription-name-goes-here" }
+  let(:sub_hash) { subscription_hash topic_name, sub_name }
+  let(:sub_grpc) { Google::Cloud::PubSub::V1::Subscription.new(sub_hash) }
+  let(:subscriber) { Google::Cloud::PubSub::Subscriber.from_grpc sub_grpc, pubsub.service }
+
+  it "b/528401453: waits for exponential backoff before retrying on GRPC::Unavailable" do
```
aandreassa: I'd rather not leave the bug IDs unless needed. We should only really use as reminder for follow-ups.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -68,24 +69,31 @@ def initialize subscriber
...
               break if @background_thread
 
               @inventory.start
+              @stream_keepalive_task = Concurrent::TimerTask.new(
+                execution_interval: @keepalive_interval
+              ) { send_keepalive_ping! }
+              @stream_keepalive_task.execute
+
+              @pong_monitor_task = Concurrent::TimerTask.new(
+                execution_interval: [@keepalive_interval / 5.0, 0.01].max
```
aandreassa: It would be great to have some super light comments on logic / why we use certain values.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -416,15 +458,45 @@ def pause_streaming?
...
 
-            @paused = nil
-            subscriber.service.logger.log :info, "subscriber-flow-control" do
-              "subscriber for #{@subscriber.subscription_name} is unblocking client-side flow control"
+              @paused = nil
+              @last_pong_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+              subscriber.service.logger.log :info, "subscriber-flow-control" do
+                "subscriber for #{@subscriber.subscription_name} is unblocking client-side flow control"
+              end
+              # signal to the background thread that we are unpaused
```
aandreassa: Super nit, but capitalize like other comments.

aandreassa (changes requested):

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -483,15 +458,45 @@ def pause_streaming?
...
+                end
+                @last_ping_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) if @last_pong_at >= @last_ping_at
+                push Google::Cloud::PubSub::V1::StreamingPullRequest.new
+              end
+            end
+          end
+
+          def check_liveness!
+            synchronize do
+              if @stream_opened && @last_ping_at && @last_pong_at && !@stopped && !@paused
```
quartzmo: I think we may have lost a good multi-line comment that was above this long conditional. I'm always happy to see comments in concurrent code, explaining what is happening and why. In my opinion you can sprinkle more "play by play" comments throughout this production code.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -219,6 +232,13 @@ class RestartStream < StandardError; end
 
           # rubocop:disable all
 
+          def backoff_and_wait!
+            @reconnect_delay = @reconnect_delay ? [@reconnect_delay * 1.5, 60.0].min : 1.0
+            synchronize do
+              @pause_cond.wait(@reconnect_delay + rand(0.0..0.5)) unless @stopped
```
quartzmo: There is a subtle timing gotcha here. When the stream disconnects due to a network error (e.g. `GRPC::BadStatus` raised inside the `background_run` loop), `@pause_cond.wait` is called here with `@stream_opened` still `true`.

If the `@pong_monitor_task` fires during this backoff sleep, it will call `check_liveness!` (L95), evaluate the long condition on L477 as true (since no pongs are arriving), log the warning, and raise `RestartStream` (L482).

This asynchronous exception immediately interrupts the sleep. The background thread wakes up early, exits the current rescue block, gets rescued by `rescue RestartStream` on L355, and enters a *second* backoff sleep (which multiplies `@reconnect_delay` again).

To prevent the monitor task from interrupting our backoff delays and causing double-backoffs, we should set `@stream_opened = false` under synchronization just above this line:

```ruby
def backoff_and_wait!
  @reconnect_delay = @reconnect_delay ? [@reconnect_delay * 1.5, 60.0].min : 1.0
  synchronize do
    @stream_opened = false # <--- Disable liveness checker during backoff
    @pause_cond.wait(@reconnect_delay + rand(0.0..0.5)) unless @stopped
  end
end
```

quartzmo (changes requested):

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: @quartzmo LMK how these additional comments and context look!

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: Tried to add some more context/detail throughout the PR; LMK what you think

torreypayne: also defined some constants above for more clarity/readability

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
quartzmo: It looks like this problem is now resolved, thanks! But I'm not sure it has test coverage (I might be wrong). Here's a suggested test from Gemini: 

```ruby
it "sets stream_opened to false during backoff to prevent monitor interruption" do
  listener = subscriber.listen streams: 1 do |msg|
  end
  stream = listener.instance_variable_get(:@stream_pool).first
  stream.instance_variable_set :@stream_opened, true
  stream.instance_variable_set :@stopped, false

  wait_called = false
  pause_cond = stream.instance_variable_get(:@pause_cond)
  
  # Stub wait to assert that @stream_opened is false when wait is invoked
  pause_cond.stub :wait, ->(timeout) {
    wait_called = true
    _(stream.instance_variable_get(:@stream_opened)).must_equal false
  } do
    stream.send :backoff_and_wait!
  end

  assert wait_called
end
```

quartzmo (approved):

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb
```diff
@@ -0,0 +1,208 @@
...
+
+    refute_nil new_keepalive
+    refute_nil new_monitor
+    refute_equal first_keepalive, new_keepalive
+    refute_equal first_monitor, new_monitor
+
+    stream.stop
+  end
+
+  it "sets stream_opened to false during backoff to prevent monitor interruption" do
```
torreypayne: @quartzmo added your test!

quartzmo (changes requested):
It appears that the following comment blocks were dropped during the `KeepaliveMonitor` refactoring:

### 1. Restore comments for constants in `keepalive_monitor.rb`:

```ruby
          # Default interval in seconds between keep-alive ping requests.
          DEFAULT_INTERVAL = 30.0
          # Default deadline in seconds to receive a keep-alive pong response.
          DEFAULT_DEADLINE = 15.0
          # Divisor applied to keep-alive interval to calculate monitor polling interval (1/5th of interval).
          MONITOR_DIVISOR = 5.0
          # Minimum floor in seconds (10ms) for the monitor polling interval to prevent CPU spinning.
          MIN_MONITOR_INTERVAL = 0.01
```

### 2. Restore concurrent behavior comments in `keepalive_monitor.rb` inside `send_ping!`:

```ruby
          def send_ping!
            @stream.synchronize do
              # Push unconditional keep-alive pings only when the stream is actively open and request queue is active.
              # Note: ACKs are sent via unary RPCs (TimedUnaryBuffer), not over this stream.
              return unless @stream.stream_opened? && !@stream.stopped? && @stream.request_queue_active?

              @stream.log_info "sending keepAlive to stream for subscription #{@stream.subscriber.subscription_name}"
              # Only advance @last_ping_at if the previous ping was successfully ponged.
              # If a pong is outstanding (@last_pong_at < @last_ping_at), freezing @last_ping_at preserves the start time
              # of the un-ponged cycle so check_liveness! can accurately evaluate the elapsed deadline.
              @last_ping_at = Process.clock_gettime(Process::CLOCK_MONOTONIC) if @last_pong_at >= @last_ping_at
              @stream.send_ping_request!
            end
          end
```

### 3. Restore concurrent behavior comments in `keepalive_monitor.rb` inside `check_liveness!`:

```ruby
          def check_liveness!
            @stream.synchronize do
              # Do not check pong deadline if paused (client flow control inventory full).
              # When paused, background_run waits on condition variable and stops calling enum.next,
              # so incoming server pongs sit buffered in gRPC and last_pong_at stays un-updated.
              return unless @stream.stream_opened? && @last_ping_at && @last_pong_at && !@stream.stopped? && !@stream.paused?

              now = Process.clock_gettime(Process::CLOCK_MONOTONIC)
              if now - @last_ping_at >= @deadline && @last_pong_at < @last_ping_at
                elapsed_pong = (now - @last_pong_at).round(2)
                @stream.log_error "Keep-alive pong not received within #{@deadline}s (last pong #{elapsed_pong}s ago); restarting stream."
                @stream.restart_stream_for_timeout!
              end
            end
          end
```

quartzmo (approved):

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/bug_regression_test.rb
```diff
@@ -0,0 +1,73 @@
...
+      fail "stream did not retry" if retries > 200
+      retries += 1
+      sleep 0.05
+    end
+
+    listener.stop
+    listener.wait!
+
+    elapsed = attempts[1] - attempts[0]
+    puts "\n[b/528401453 Test] Elapsed delay between attempts: #{elapsed.round(3)}s"
```
aandreassa: Can we remove the outputs in this file (or capture it if needed)? It is better to prevent this test from polluting relevant output logs.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -68,24 +85,35 @@ def initialize subscriber
...
-                push Google::Cloud::PubSub::V1::StreamingPullRequest.new
-              end
-            end.execute
+            @keepalive_monitor = KeepaliveMonitor.new self
+            @stream_opened = false
+            @reconnect_delay = nil
+          end
+
+          # keepalive_interval used to send pings and maintain a healthy stream
+          def keepalive_interval
```
aandreassa: Do we need these? Maybe we should keep it minimal and consistent.
You are only using it on tests, and you already call `stream.keepalive_monitor.last_ping_at`, for example.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,137 @@
...
+          DEFAULT_INTERVAL = 30.0
+          # Default deadline in seconds to receive a keep-alive pong response.
+          DEFAULT_DEADLINE = 15.0
+          # Divisor applied to keep-alive interval to calculate monitor polling interval (1/5th of interval).
+          MONITOR_DIVISOR = 5.0
+          # Minimum floor in seconds (10ms) for the monitor polling interval to prevent CPU spinning.
+          MIN_MONITOR_INTERVAL = 0.01
+
+          # keepalive_interval used to send pings and maintain a healthy stream
+          attr_accessor :interval
```
aandreassa: Can we pls have YARD docstrings for these `attr_accessors`?

`rg -U '^\s*#[^\n]*\n\s*attr_accessor'` for examples.

Alternatively, do we need these if they are for testing only? After modernizing in 3.0, we try to keep this library pretty lightweight.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -139,6 +169,55 @@ def wait! timeout = nil
             self
           end
 
+          def stream_opened?
```
aandreassa: I find this confusing. `stream_opened` as in opened in the past and its done, or currently open? You should try using present tense for predicates.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,137 @@
...
+
+require "concurrent"
+
+module Google
+  module Cloud
+    module PubSub
+      class MessageListener
+        ##
+        # @private
+        # Monitors gRPC streaming connection health via periodic bi-directional keep-alive pings and pongs.
```
aandreassa, line 23: Can you please document what this class does at a high level (here or initialize)? I'd love to see more YARD docstrings overall.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -443,6 +571,7 @@ def initial_input_request
               req.client_id = @subscriber.service.client_id
               req.max_outstanding_messages = @inventory.limit
               req.max_outstanding_bytes = @inventory.bytesize
+              req.protocol_version = 1
```
aandreassa: Switch to const?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,137 @@
...
+          attr_accessor :last_ping_at
+          # timestamp of the last pong received
+          attr_accessor :last_pong_at
+
+          # background timer task for sending keepalive pings
+          attr_reader :ping_task
+          # background timer task for monitoring stream liveness
+          attr_reader :monitor_task
+
+          def initialize stream, interval: DEFAULT_INTERVAL, deadline: DEFAULT_DEADLINE
```
aandreassa, line 50: You could do something like (pls double check convention)
```
# Initializes the KeepaliveMonitor.
# 
# @param [Stream] stream The parent streaming pull connection to monitor.
# @param [Float] interval The interval in seconds between keep-alive pings.
# @param [Float] deadline The deadline in seconds for receiving a pong.
```
and document others in-line below.

aandreassa (changes requested):
Implementation looks great, just a couple of Ruby things & docs to think about!

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -21,6 +21,9 @@ class MessageListener
         ##
         # @private
         # Monitors gRPC streaming connection health via periodic bi-directional keep-alive pings and pongs.
+        #
+        # Tracks ping and pong timestamps over the active gRPC stream to detect silent connection drops or freezes,
+        # and triggers a stream restart when server responses exceed the configured pong deadline.
```
torreypayne, line 26: @aandreassa LMK if this additional context is useful/sufficient to address your comment on L23

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
torreypayne: Implemented your suggestion as-is; LMK what you think

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
torreypayne: Take a look and LMK if I should do another pass

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: updated to present tense (`stream_open`); should I add `is_...` to make it more verb-y?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: defined a `PROTOCOL_VERSION` const

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -139,6 +154,56 @@ def wait! timeout = nil
...
+            end
+          end
+
+          def stream_keepalive_task
+            @keepalive_monitor.ping_task
+          end
+
+          def pong_monitor_task
+            @keepalive_monitor.monitor_task
+          end
```
aandreassa: Remove these? I feel like `@keepalive_monitor.ping_task` is self descriptive enough and it would help keep it clean.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
aandreassa: I'm afraid the new `restart_stream_for_timeout!` may cause `self` to bleed into this queue and get stuck. I don't think it would be an issue before. WDYT?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -219,6 +284,15 @@ class RestartStream < StandardError; end
 
           # rubocop:disable all
 
+          def backoff_and_wait!
+            @reconnect_delay = @reconnect_delay ? [@reconnect_delay * RECONNECT_BACKOFF_MULTIPLIER, MAX_RECONNECT_DELAY].min : INITIAL_RECONNECT_DELAY
+            synchronize do
+              # Disable liveness checker during backoff sleep to prevent monitor task from interrupting and double-backing off
+              @stream_open = false
+              @pause_cond.wait(@reconnect_delay + rand(0.0..0.5)) unless @stopped
```
aandreassa: Is there an issue sharing this backoff `@pause_cond` with message processing?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
aandreassa: Looks great! If you want it to be extra, you could add a quick context of why this is useful to users (the reason why we are adding all of this).

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,165 @@
...
+          DEFAULT_DEADLINE = 15.0
+          # Divisor applied to keep-alive interval to calculate monitor polling interval (1/5th of interval).
+          MONITOR_DIVISOR = 5.0
+          # Minimum floor in seconds (10ms) for the monitor polling interval to prevent CPU spinning.
+          MIN_MONITOR_INTERVAL = 0.01
+
+          # @private The keep-alive interval in seconds used to send pings and maintain a healthy stream.
+          attr_accessor :interval
+          # @private The deadline in seconds used to validate whether we received a pong back from the stream.
+          attr_accessor :deadline
```
aandreassa: Do we really need `attr_accessor` for these though? I feel like removing would enforce production immutability as wanted.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,165 @@
...
+          end
+
+          # Starts the background keep-alive ping and liveness monitor timer tasks.
+          #
+          # @return [KeepaliveMonitor] self for chaining.
+          def start
+            @stream.synchronize do
+              return self if @ping_task
+
+              @ping_task = Concurrent::TimerTask.new(execution_interval: @interval) do
```
aandreassa: Do we need flooring here? `[...].max`

aandreassa (changes requested):

google-cloud-pubsub/verify_keepalive.rb
```diff
@@ -0,0 +1,85 @@
+# verify_keepalive.rb
```
aandreassa: Nit: Remove?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -240,16 +300,33 @@ def background_run
               # Always create a new request queue
               @request_queue = EnumeratorQueue.new self
               @request_queue.push initial_input_request
-              old_queue.each { |obj| @request_queue.push obj }
+              old_queue.each do |obj|
+                next if obj == self || obj.is_a?(Stream)
```
aandreassa: Can we simplify this loop? `obj.is_a?(Stream)` is redundant because self is the only `Stream` instance AFAICT.

```
old_queue.each do |obj|
  @request_queue.push obj unless obj == self
end
```

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/keepalive_monitor.rb
```diff
@@ -0,0 +1,165 @@
...
+          DEFAULT_INTERVAL = 30.0
+          # Default deadline in seconds to receive a keep-alive pong response.
+          DEFAULT_DEADLINE = 15.0
+          # Divisor applied to keep-alive interval to calculate monitor polling interval (1/5th of interval).
+          MONITOR_DIVISOR = 5.0
+          # Minimum floor in seconds (10ms) for the monitor polling interval to prevent CPU spinning.
+          MIN_MONITOR_INTERVAL = 0.01
+
+          # @private The keep-alive interval in seconds used to send pings and maintain a healthy stream.
+          attr_reader :interval
```
aandreassa: Sorry if my previous comment wasn't clear! Do we need `attr_reader` or `attr_accessor` for these? I feel we can get by without them since they appear to be used primarily for testing purposes.

`record_pong!` and related methods already update these variables under synchronization. Exposing public accessors could allow background threads to overwrite these timestamps without acquiring the lock, introducing concurrency bugs and data races.

Besides, it will keep the library a bit more lightweight in favor of 3.x

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -60,6 +84,7 @@ def initialize subscriber
             @stopped = nil
             @paused  = nil
             @pause_cond = new_cond
+            @backoff_cond = new_cond
```
aandreassa, line 87: A lightweight comment for each would be helpful since we have two versions now. 

Could you please share what the bug was with only having one? Or point me to a test where you verify the behavior in case I missed it.

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb
```diff
@@ -0,0 +1,243 @@
...
+    subscriber.service.mocked_subscription_admin = stub
+
+    listener = subscriber.listen streams: 1 do |msg|
+    end
+    stream = listener.instance_variable_get(:@stream_pool).first
+    stream.instance_variable_set :@stream_open, true
+    stream.instance_variable_set :@stopped, false
+
+    wait_called = false
+    pause_cond = stream.instance_variable_get(:@backoff_cond)
```
aandreassa: Nit: wrong name

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb
```diff
@@ -0,0 +1,243 @@
...
+      wait_called = true
+      _(stream.instance_variable_get(:@stream_open)).must_equal false
+    } do
+      stream.send :backoff_and_wait!
+    end
+
+    assert wait_called
+  end
+
+  it "does not bleed self sentinel into the new request queue after restart_stream_for_timeout!" do
```
aandreassa, line 199: This test doesn't actually call `restart_stream_for_timeout!` or raise `RestartStream`. It manually pushes stream into the queue and directly calls private `background_run`.

Could we rewrite it to invoke it directly or simulate the monitor timing out? This way, we are testing the actual production path rather than manually stubbing internal variables.

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_integration_test.rb
```diff
@@ -0,0 +1,170 @@
...
+
+      stream = listener.instance_variable_get(:@stream_pool).first
+      stream.keepalive_monitor.instance_variable_set :@interval, 0.05
+      stream.keepalive_monitor.instance_variable_set :@deadline, 0.5
+
+      listener.start
+
+      # Wait for initial message delivery
+      retries = 0
+      until messages_received.any?
```
aandreassa: Instead of duplicating the `until` loop boilerplate code across all these test, can we have a shared `wait_until` helper (or refactor/reuse the `wait_until` helper from `async_publisher_test.rb`)?

aandreassa (changes requested):

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_test.rb
torreypayne: @aandreassa I'm now calling `#check_liveness!` here which invokes the underlying `#restart_stream_for_timeout!` method

google-cloud-pubsub/verify_keepalive.rb
torreypayne: I just removed this script entirely; I'm using it for validation but it's not necessary for the PR.

google-cloud-pubsub/test/google/cloud/pubsub/message_listener/keepalive_integration_test.rb
torreypayne: refactored!

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: The bug: if we share the wait condition, we could prematurely trigger a connection retry; essentially, if the stream was paused due the queue being full, when the queue drops below threshold we could end up retrying a closed connection prematurely. This would manifest as potentially wasted retries and some resources. So good catch on sharing these.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne, line 291: @aandreassa the reason why I can't call `#stop` is actually written explicitly right here; for the stream itself, calling `#stop` _permanently_ tears down the connection and prevents restarts. The KeepaliveMonitor pushing itself onto the queue is essentially signaling EOF to the consumer without fully tearing down the stream. With that being said, I actually think our implementation is good as-is. LMK what you think.

aandreassa (approved):
🚀

PR #34692, opened 2026-06-30
chore(ci): fix triggers for workflows broken by #34682

aandreassa:
The automated changes from #34682 inserted the `permissions` in the wrong place for these two files.

Other workflows do not have this issue as they start with `name:` at the top.

### DIFF
--- a/.github/workflows/conformance.yaml
+++ b/.github/workflows/conformance.yaml
@@ -1,10 +1,10 @@
 # This workflow runs the Bigtable Conformance Test suite against the Bigtable
 # client library.
 # See https://github.com/googleapis/cloud-bigtable-clients-test
-on:
-
 permissions:
   contents: read
+
+on:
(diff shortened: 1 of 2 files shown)


quartzmo (approved):

.github/workflows/conformance.yaml
```diff
@@ -1,10 +1,10 @@
 # This workflow runs the Bigtable Conformance Test suite against the Bigtable
 # client library.
 # See https://github.com/googleapis/cloud-bigtable-clients-test
-on:
-
 permissions:
   contents: read
+
+on:
```
g-husam, line 7: thanks for fixing btw! I just discovered that this happened to some of the workflow files :(

aandreassa: NP! We appreciate the contribution 🙂

PR #34699, opened 2026-07-01
chore(new-library): Change workflow to accept multiple proto paths

aandreassa:
This change allow us to create both the versioned gem (`google-maps-isochrones-v1`) and the wrapper (`google-maps-isochrones`) in one go, essentially removing the merge dependency that currently exists.

The script sorts versioned gems to run first, so in theory it should work well.

Usage:
Set `protoPaths` to `google/maps/isochrones google/maps/isochrones/v1`

(space-delimited list)

### DIFF
--- a/.toys/new-library.rb
+++ b/.toys/new-library.rb
@@ -14,9 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-desc "A tool that generates a new client library"
+desc "A tool that generates new client libraries"
 
 required_arg :proto_namespace
+
+remaining_args :additional_proto_namespaces do
+  desc "Additional proto namespaces to generate"
+end
 
 flag :piper_client, "--piper-client=NAME" do
   desc "Name of the piper client"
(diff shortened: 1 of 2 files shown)


.toys/new-library.rb
```diff
@@ -71,30 +75,39 @@
 def run
   setup
   set :branch_name, "gen/#{gem_name}" unless branch_name
```
torreypayne, line 77: @aandreassa do we also need to change this var to `gem_names` or format the branch in such a way that handles multiple gems?

torreypayne (approved):

torreypayne:
nit: @aandreassa is it possible to share an artifact showing a sample runthrough/output?

.toys/new-library.rb
aandreassa: `branch_name` is just the github branch, which we dont really use for much. I don't think it makes sense to add all of them either and be super long (so it defaults to the first name). 

It may be not be perfect, but it helps us save time here for now!

aandreassa (commented):
Thanks for the review! I tested this out with `cloudproductregistry`, so I will make sure to post here when I do it again.

aandreassa:
@torreypayne this worked on https://github.com/googleapis/google-cloud-ruby/pull/34720

PR #34721, opened 2026-07-07
feat: Ruby 3.2 minimum and Ruby 4.0 support

aandreassa:


torreypayne (approved):
@aandreassa similar question; should we also update the GitHub Actions and related workflows to 4.0 too? And/or remove any instances <3.2 if present?

aandreassa:
I will add in a follow up to update anything & pin versions! @torreypayne

PR #34763, opened 2026-07-14
chore(deps): update core handwritten gems lockfiles (major)

torreypayne:
> ℹ️ **Note**
> 
> This PR body was truncated due to platform limits.

This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [google-apis-core](https://redirect.github.com/google/google-api-ruby-client) ([source](https://redirect.github.com/googleapis/google-api-ruby-client/tree/HEAD/google-apis-core), [changelog](https://redirect.github.com/googleapis/google-api-ruby-client/tree/main/google-apis-core/CHANGELOG.md)) | `0.11.3` → `1.2.4` | age | confidence |
| [minitest](https://minite.st/) ([source](https://redirect.github.com/minitest/minitest), [changelog](https://redirect.github.com/minitest/minitest/blob/master/History.rdoc)) | `"~> 5.14"` → `"~> 6.0"` | age | confidence |
| [minitest](https://minite.st/) ([source](https://redirect.github.com/minitest/minitest), [changelog](https://redirect.github.com/minitest/minitest/blob/master/History.rdoc)) | `"~> 5.25"` → `"~> 6.0"` | age | confidence |
| [minitest](https://minite.st/) ([source](https://redirect.github.com/minitest/minitest), [changelog](https://redirect.github.com/minitest/minitest/blob/master/History.rdoc)) | `"~> 5.16"` → `"~> 6.0"` | age | confidence |
| [retriable](https://redirect.github.com/kamui/retriable) | `"~> 3.1.2"` → `"~> 4.2.0"` | age | confidence |
| [simplecov](https://redirect.github.com/simplecov-ruby/simplecov) ([changelog](https://redirect.github.com/simplecov-ruby/simplecov/blob/main/CHANGELOG.md)) | `"~> 0.22"` → `"~> 1.0"` | age | confidence |
| [simplecov](https://redirect.github.com/simplecov-ruby/simplecov) ([changelog](https://redirect.github.com/simplecov-ruby/simplecov/blob/main/CHANGELOG.md)) | `"~> 0.9"` → `"~> 1.0"` | age | confidence |

---

### Release Notes

<details>
<summary>googleapis/google-api-ruby-client (google-apis-core)</summary>

##### Bug Fixes

- map sharingRateLimitExceeded to RateLimitError ([#&#8203;27178](https://redirect.github.com/googleapis/google-api-ruby-client/issues/27178)) ([80c0656](https://redirect.github.com/googleapis/google-api-ruby-client/commit/80c0656d3933418993f5857b39c6d5bd968a8118)), closes [#&#8203;26531](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26531)

##### Bug Fixes

- upgrade addressable dependency to \~> 2.9 ([#&#8203;26622](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26622))

##### Bug Fixes

- Relax retriable dependency constraint ([#&#8203;26621](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26621)) ([1511d44](https://redirect.github.com/googleapis/google-api-ruby-client/commit/1511d44e3945caebf41160e47cc42f7bdfdb5fe2)), closes [#&#8203;26620](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26620)

##### Bug Fixes

- Add missing multi\_json runtime dependency for representable ([#&#8203;26612](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26612)) ([16afc11](https://redirect.github.com/googleapis/google-api-ruby-client/commit/16afc118f58d64a1394f33b76133d69ce4632575)), closes [#&#8203;26611](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26611) [#&#8203;26613](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26613)

##### Features

- ruby 3.2 minimum ([#&#8203;26608](https://redirect.github.com/googleapis/google-api-ruby-client/issues/26608))

##### Features

- adding checksum header ([#&#8203;25157](https://redirect.github.com/googleapis/google-api-ruby-client/issues/25157))

##### Bug Fixes

- Handle nil response in on\_data callback for Faraday streaming ([#&#8203;24235](https://redirect.github.com/googleapis/google-api-ruby-client/issues/24235))

##### Bug Fixes

- Compute and send content-length header when posting a CompositeIO ([#&#8203;23864](https://redirect.github.com/googleapis/google-api-ruby-client/issues/23864))

### [`v1.0.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#100-2025-08-03)

This is a major release that replaces the underlying httpclient library with Faraday ([#&#8203;23524](https://redirect.github.com/googleapis/google-api-ruby-client/issues/23524)). This will ensure the client libraries are on a more stable and better maintained foundation moving forward.

For most users, this change should be transparent. However, if your application depends on the httpclient interfaces, you can retain compatibility with httpclient by pinning the `google-apis-core` gem to `~> 0.18` in your Gemfile. Httpclient-based versions of this gem will remain on the 0.x release train, while Faraday-based versions will occupy the 1.x release train. We will push critical fixes and security updates to both branches for one year until August 2026, but new feature work will take place only on the 1.x branch.

### [`v0.18.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0180-2025-05-22)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.17.0...0.18.0)

##### Features

- Restart & delete resumable upload ([#&#8203;21896](https://redirect.github.com/googleapis/google-api-ruby-client/issues/21896))

### [`v0.17.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0170-2025-04-30)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.16.0...0.17.0)

##### Features

- ruby 3.1 minimum, 3.4 default ([#&#8203;22594](https://redirect.github.com/googleapis/google-api-ruby-client/issues/22594))

##### Bug Fixes

- Fixed a method redefined warning ([#&#8203;21572](https://redirect.github.com/googleapis/google-api-ruby-client/issues/21572))
- Ensure compatibility with frozen string literals ([#&#8203;21648](https://redirect.github.com/googleapis/google-api-ruby-client/issues/21648))

##### Features

- add ECONNRESET error as retriable error ([#&#8203;20354](https://redirect.github.com/googleapis/google-api-ruby-client/issues/20354))

##### Bug Fixes

- remove rexml from dependencies ([#&#8203;19971](https://redirect.github.com/googleapis/google-api-ruby-client/issues/19971))

### [`v0.15.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0150-2024-05-13)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.14.1...0.15.0)

##### Features

- Introduce api\_version to discovery clients ([#&#8203;18969](https://redirect.github.com/googleapis/google-api-ruby-client/issues/18969))

### [`v0.14.1`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0141-2024-03-13)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.14.0...0.14.1)

##### Bug Fixes

- fixes uninitialized Pathname issue ([#&#8203;18480](https://redirect.github.com/googleapis/google-api-ruby-client/issues/18480))

### [`v0.14.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0140-2024-02-23)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.13.0...0.14.0)

##### Features

- Update minimum Ruby version to 2.7 ([#&#8203;17896](https://redirect.github.com/googleapis/google-api-ruby-client/issues/17896))

##### Bug Fixes

- allow BaseService#root\_url to be an Addressable::URI ([#&#8203;17895](https://redirect.github.com/googleapis/google-api-ruby-client/issues/17895))

### [`v0.13.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0130-2024-01-26)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.12.0...0.13.0)

##### Features

- Verify credential universe domain against configured universe domain ([#&#8203;17569](https://redirect.github.com/googleapis/google-api-ruby-client/issues/17569))

### [`v0.12.0`](https://redirect.github.com/googleapis/google-api-ruby-client/blob/HEAD/google-apis-core/CHANGELOG.md#0120-2024-01-22)

[Compare Source](https://redirect.github.com/googleapis/google-api-ruby-client/compare/0.11.3...0.12.0)

##### Features

- Support for universe\_domain ([#&#8203;17131](https://redirect.github.com/googleapis/google-api-ruby-client/issues/17131))

</details>

<details>
<summary>minitest/minitest (minitest)</summary>

### [`v6.0.6`](https://redirect.github.com/minitest/minitest/compare/v6.0.5...v6.0.6)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.5...v6.0.6)

### [`v6.0.5`](https://redirect.github.com/minitest/minitest/compare/v6.0.4...v6.0.5)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.4...v6.0.5)

### [`v6.0.4`](https://redirect.github.com/minitest/minitest/compare/v6.0.3...v6.0.4)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.3...v6.0.4)

### [`v6.0.3`](https://redirect.github.com/minitest/minitest/compare/v6.0.2...v6.0.3)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.2...v6.0.3)

### [`v6.0.2`](https://redirect.github.com/minitest/minitest/compare/v6.0.1...v6.0.2)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.1...v6.0.2)

### [`v6.0.1`](https://redirect.github.com/minitest/minitest/compare/v6.0.0...v6.0.1)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v6.0.0...v6.0.1)

### [`v6.0.0`](https://redirect.github.com/minitest/minitest/compare/v5.27.0...v6.0.0)

[Compare Source](https://redirect.github.com/minitest/minitest/compare/v5.27.0...v6.0.0)

</details>

<details>
<summary>kamui/retriable (retriable)</summary>

### [`v4.2.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#420)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v4.1.1...v4.2.0)

##### Bug fixes

- The `Kernel` extension methods (`require "retriable/core_ext/kernel"`) are now
  private, matching idiomatic `Kernel` helpers like `puts` and `rand`.
  Previously `retriable` and `retriable_with_context` were public instance
  methods, so they leaked onto every object's public API and could be invoked
  with an explicit receiver (e.g. `"foo".retriable { ... }`). They remain
  callable in the documented receiver-less form.
  ([#&#8203;146](https://redirect.github.com/kamui/retriable/pull/146))
- `Retriable.with_context` (and `Kernel#retriable_with_context`) now raises
  `ArgumentError` when called without a block, matching `with_override`.
  Previously a missing block was silently ignored: the call returned `nil` and
  the intended block never ran, hiding a caller bug. Behavior change: code that
  relied on the silent no-op will now raise.
- `Config#validate!` now validates the structure of each entry in `contexts`,
  so configured contexts are checked on every `Retriable.retriable`/
  `with_context` call rather than only when a given context is first used. A
  context whose options contain an unknown key (including a nested `contexts`
  key) now raises `ArgumentError, "<key> is not a valid option"`, matching the
  `with_override` path. Non-Hash `contexts` and non-Hash per-context values
  remain leniently treated as empty options (no behavior change). Option
  *values* are still validated lazily at retry time, unchanged.

##### Docs

- Document that `on_retry` receives `next_interval: nil` on the final rescued
  attempt, when Retriable is about to give up because `tries` are exhausted.
  `on_retry` still fires before `on_give_up` (unchanged behavior); the `nil`
  contract is now called out in the `on_retry` documentation so handlers guard
  arithmetic or logging on `next_interval`.

##### Performance

- `Config#initialize` no longer allocates a throwaway `ExponentialBackoff` (and
  runs its redundant `validate!`) just to read default values. Defaults now live
  in a frozen `ExponentialBackoff::DEFAULTS` constant, removing an allocation and
  redundant validation from the `retriable` hot path.
  ([#&#8203;149](https://redirect.github.com/kamui/retriable/pull/149))

### [`v4.1.1`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#411)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v4.1.0...v4.1.1)

##### Bug fixes

- `retry_if`, `on_retry`, and `on_give_up` are now validated to be callable
  (respond to `#call`) or falsy. A non-callable truthy value raises
  `ArgumentError` at configuration time instead of a later `NoMethodError` on a
  retry path. ([#&#8203;140](https://redirect.github.com/kamui/retriable/pull/140))

##### Internal

- Add RBS type signatures for the public API (`Retriable.configure`, `config`,
  `retriable`, `with_override`, `with_context`, and `Retriable::Config`) and
  validate them in CI with `rbs validate`.
  ([#&#8203;142](https://redirect.github.com/kamui/retriable/pull/142))
- Enforce a minimum test coverage floor and add a `bundler-audit` dependency
  audit job to CI. ([#&#8203;143](https://redirect.github.com/kamui/retriable/pull/143))
- Remove an unused `CC_TEST_REPORTER_ID` from the CI workflow.
  ([#&#8203;141](https://redirect.github.com/kamui/retriable/pull/141))

### [`v4.1.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#410)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v4.0.0...v4.1.0)

##### Bug fixes

- A per-call or `with_context` `tries:` now clears an inherited `intervals:` from
  global config or a context, matching the documented precedence. Previously
  `Retriable.retriable(tries: 1)` was silently ignored when `intervals` was
  configured, running `intervals.size + 1` times. Passing both `intervals:` and
  `tries:` in the same call still lets `intervals:` win.

### [`v4.0.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#400)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.8.0...v4.0.0)

**This is a major release with breaking changes. Please read carefully before upgrading.**

##### Breaking changes

- Removed `timeout:` option. The `timeout:` option has been removed from `Retriable.retriable`, `Retriable.configure`, and `Retriable.with_override`. It was a thin wrapper around Ruby's `Timeout.timeout`, which has well-documented safety issues: it interrupts execution at arbitrary lines and can corrupt internal state in libraries that are not interrupt-safe (mutexes, file handles, network sockets, allocator state). This was first raised against this gem in [#&#8203;96](https://redirect.github.com/kamui/retriable/issues/96) in 2021; Retriable 3.8.0 deprecated the option, and 4.0 removes the footgun entirely. As a side effect, the historical bug where Retriable's own internal `Timeout::Error` was silently retried by default is no longer reachable, since Retriable no longer raises a timeout itself. User-raised `Timeout::Error` (for example, from a `Timeout.timeout` block you write inside the retried block) is still matched by the default `on: [StandardError]` because `Timeout::Error < RuntimeError < StandardError`. Passing `timeout:` to `Retriable.retriable` or `Retriable.with_override` now raises `ArgumentError`; setting `config.timeout` in `Retriable.configure` now raises `NoMethodError` because the configuration attribute has been removed. See the [4.0 migration section in the README](README.md#migration-from-3x-to-40) for replacement patterns.
- Minimum Ruby version is now 3.2. Support for Ruby 2.x, 3.0, and 3.1 has been dropped in Retriable 4.0. If you need Retriable on Ruby 2.3.0-3.1.x, the 3.8.x line (`~> 3.8`) remains available.

##### Features

- Add [`on_give_up`](README.md#callbacks) callback that runs when Retriable stops retrying after a rescued retriable exception. Receives `(exception, try, elapsed_time, next_interval, reason)`, where `reason` is `:tries_exhausted` or `:max_elapsed_time`. Does not fire for non-retriable exceptions or `retry_if` rejections. Pass `on_give_up: false` to suppress a configured handler for a single call.
- Accept a [`Set` of `Exception` classes](README.md#configuring-which-options-to-retry-with-on) as the `on:` option, in addition to a single class, an `Array`, or a `Hash`.

##### Internal

- Switched `Retriable.retriable`, `Retriable.with_context`, and the `Kernel` extension methods to Ruby 3.1+ anonymous block forwarding. No user-visible behavior change.

### [`v3.8.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#380)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.7.0...v3.8.0)

##### Deprecations

- Deprecated the `timeout:` option ahead of its removal in Retriable 4.0. Non-nil timeout values supplied through `Retriable.configure`, `Retriable.retriable(...)`, or `Retriable.with_override(...)` now emit a deprecation warning while keeping the existing runtime behavior unchanged. On Ruby 2.7+ the warning is emitted via `Kernel.warn(..., category: :deprecated)`, so callers can silence it through the standard Ruby controls (`Warning[:deprecated] = false`, `ruby -W:no-deprecated`, or a custom `Warning.warn`). To keep the notice from drowning busy applications, it is emitted at most once per process; suppression via `Warning[:deprecated]` leaves the warner armed for the next call that re-enables the category. Prefer library-native timeout settings, or wrap the retried block in `Timeout.timeout(...)` directly if you still need that behavior. See the README migration guidance for details.

### [`v3.7.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#370)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.6.1...v3.7.0)

- Feature: Opt-in unbounded retries via `tries: Float::INFINITY`. Requires a finite `max_elapsed_time` as a safety bound and is incompatible with custom `intervals:`. Both invalid configurations raise `ArgumentError` from `Config#validate!`.

### [`v3.6.1`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#361)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.6.0...v3.6.1)

- Fix: Validate the `on:` option before retrying. Previously, passing a non-`Exception` value such as `Object`, `Kernel`, or a plain `Module` (which appear in every `Exception`'s ancestor chain) would silently retry process-critical exceptions like `SystemExit` and `Interrupt`. The `on:` option now requires an `Exception` subclass, an array of them, or a hash whose keys are such classes and whose values are `nil`, a `Regexp`, or an array of `Regexp`s. Invalid shapes raise `ArgumentError` before the block runs.
- Fix: Validate `with_override(contexts:)` shape before applying overrides. `contexts` may be `nil` or a hash, and each per-context override must be a hash.
- Docs: Document that `on_retry: false` disables a callback set in `Retriable.configure` for a single call.

### [`v3.6.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#360)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.5.1...v3.6.0)

- Breaking: `Retriable.override` and `Retriable.reset_override` are removed and replaced by block-scoped `Retriable.with_override(opts) { ... }`. The new API requires a block, restores the previous override (or absence of override) when the block exits via `ensure`, and is thread-local — overrides set in one thread do not affect other threads, and child threads do not inherit them. Fibers within a thread still share the thread's active override. Nested `with_override` calls correctly restore the outer override on inner exit. See the README and `docs/testing.md` for migration and testing patterns. This replaces the override API introduced in 3.5.0.

### [`v3.5.1`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#351)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.5.0...v3.5.1)

- Fix: Validate retry timing and count options before use to reject invalid retry configurations. `tries` must now be a positive integer unless a custom `intervals` array is provided.

### [`v3.5.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#350)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.4.1...v3.5.0)

- Fix: Do not count skipped sleep intervals against `max_elapsed_time` when `sleep_disabled` is true.
- Add `override` and `reset_override` APIs to force retry settings over local call options when needed (for example, test short-circuiting).

### [`v3.4.1`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#341)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.4.0...v3.4.1)

- Fix: Use `Process.clock_gettime(CLOCK_MONOTONIC)` for elapsed time tracking so retry timing is immune to wall-clock adjustments (NTP, manual changes).
- Fix: Handle `max_elapsed_time: nil` gracefully instead of raising `NoMethodError`.
- Remove dead `* 1.0` float coercion in `ExponentialBackoff#randomize`.

### [`v3.4.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#340)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.3.0...v3.4.0)

- Add `retry_if` option to support custom retry predicates, including checks against wrapped `exception.cause` values.

### [`v3.3.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#330)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.2.1...v3.3.0)

- Refactor `Retriable.retriable` internals into focused private helpers to improve readability while preserving behavior.
- Modernize `.rubocop.yml` with explicit modern defaults to enable new cops while preserving existing project style policies.

### [`v3.2.1`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#321)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.2.0...v3.2.1)

- Remove executables from gemspec as it was polluting the path for some users. Thanks [@&#8203;hsbt](https://redirect.github.com/hsbt).

### [`v3.2.0`](https://redirect.github.com/kamui/retriable/blob/HEAD/CHANGELOG.md#320)

[Compare Source](https://redirect.github.com/kamui/retriable/compare/v3.1.2...v3.2.0)

- Require ruby 2.3+.
- Fix: Ensure `tries` value is overridden by `intervals` parameter if both are provided and add a test for this. This is always what the README stated but the code didn't actually do it.
- Fix: Some rubocop offenses.

</details>

<details>
<summary>simplecov-ruby/simplecov (simplecov)</summary>

### [`v1.0.0`](https://redirect.github.com/simplecov-ruby/simplecov/blob/HEAD/CHANGELOG.md#100-2026-07-12)

[Compare Source](https://redirect.github.com/simplecov-ruby/simplecov/compare/v0.22.0...v1.0.0)

\==================

First stable release of the 1.0 line. The entries below consolidate release candidates rc1 through rc5 and describe all changes since 0.22.1.

#### Breaking Changes

- Dropped support for Ruby 3.1 and JRuby 9.4. The minimum is now Ruby 3.2 (and JRuby 10, which reports `RUBY_VERSION` 3.4). Ruby 3.1 reached end of life in March 2025, and a recent `i18n` release calls `Fiber[]`, a Ruby 3.2 API, at load time, so suites that load Rails no longer run on 3.1. Raising `required_ruby_version` to `>= 3.2` also excludes JRuby 9.4, which reports `RUBY_VERSION` 3.1.x. See [#&#8203;1171](https://redirect.github.com/simplecov-ruby/simplecov/issues/1171).
- JSON formatter: group stats changed from `{ "covered_percent": 80.0 }` to full stats shape `{ "covered": 8, "missed": 2, "total": 10, "percent": 80.0, "strength": 0.0 }`. The key `covered_percent` is renamed to `percent`.
- JSON formatter: `simplecov_json_formatter` gem is now built in. `require "simplecov_json_formatter"` continues to work via a shim.
- `StringFilter` now matches at path-segment boundaries. `"lib"` matches `/lib/` but no longer matches `/library/`. Use a `Regexp` filter for substring matching.
- `SourceFile#project_filename` now returns a truly relative path with no leading separator (e.g. `lib/foo.rb` instead of `/lib/foo.rb`). This also removes the leading `/` from file path keys in `coverage.json` and from the filename in `minimum_coverage_by_file` error messages. Anchored `RegexFilter`s that relied on a leading `/` (e.g. `%r{^/lib/}`) should be rewritten (e.g. `%r{\Alib/}`).
- Removed `docile` gem dependency. The `SimpleCov.configure` block is now evaluated via `instance_exec` with instance variable proxying.
- Removed automatic activation of `JSONFormatter` when the `CC_TEST_REPORTER_ID` environment variable is set. The default `HTMLFormatter` now emits `coverage.json` alongside the HTML report (using `JSONFormatter.build_hash` to serialize the same payload `JSONFormatter` writes), so the env-var special case is no longer needed. Because of this, listing `JSONFormatter` alongside `HTMLFormatter` is redundant and can be removed.
- `SimpleCov.start` now loads the `test_frameworks` profile by default, which filters paths under `test/`, `spec/`, `features/`, and `autotest/`. Running the suite always executes 100% of the test files themselves, which inflated the overall percentage and obscured application coverage. To opt back in (e.g. to surface dead test helpers), drop the filter with `remove_filter %r{\A(test|features|spec|autotest)/}`. See [#&#8203;816](https://redirect.github.com/simplecov-ruby/simplecov/issues/816).
- HTML and JSON formatters now write the "Coverage report generated for X to Y" status line (and the per-criterion totals beneath it) to stderr instead of stdout. The message is a diagnostic, not the program's output, and routing it to stdout polluted pipelines like `rspec -f json`. Suppress it entirely with `silent: true` on the formatter; redirect with `2>&1` if you want the old behavior. See [#&#8203;1060](https://redirect.github.com/simplecov-ruby/simplecov/issues/1060).
- Under `parallel_tests`, SimpleCov now waits in the **first** started process (via `ParallelTests.first_process?`) rather than the last. This matches the convention `parallel_tests`'s own README recommends for "do something once after all workers finish" hooks, so user code that has its own `ParallelTests.wait_for_other_processes_to_finish` in an `RSpec.after(:suite)` (or equivalent) no longer deadlocks against SimpleCov's wait when both pick the same process. As a side benefit, the previous `PARALLEL_TEST_GROUPS=1` workaround for `last_process?`'s `"" == "1"` mismatch ([#&#8203;1066](https://redirect.github.com/simplecov-ruby/simplecov/issues/1066)) is no longer needed — `first_process?` handles that case naturally. **Migration:** the rare project that wired its own wait via `ParallelTests.last_process?` now hits the symmetric deadlock and must switch to `first_process?`. See [#&#8203;922](https://redirect.github.com/simplecov-ruby/simplecov/issues/922).
- Removed `SimpleCov.coverage_criterion`. It was a reader/writer for a value nothing in SimpleCov ever consumed, so it duplicated `primary_coverage` without affecting any behavior. Use `primary_coverage` to choose the report's leading criterion (or the `coverage :branch, primary: true` form).

#### Deprecations

- The configuration API has been redesigned around a smaller, more consistent set of verbs. The legacy methods continue to work but each emits a deprecation warning that names its replacement; a future release will remove them. Warnings are deduplicated by call site, so a deprecated method called in a loop or a configuration block re-evaluated once per parallel worker or spec file warns at most once per source location (see [#&#8203;1204](https://redirect.github.com/simplecov-ruby/simplecov/issues/1204)). See the "Migrating from the legacy configuration API" section in the README for the full migration table and a before/after example.
  - `add_filter` → `skip` (identical matcher grammar; no behavior change)
  - `add_group` → `group` (identical matcher grammar; no behavior change)
  - `track_files` → `cover` (`cover` includes unloaded files like `track_files` did **and** restricts the report to the matching set; pass every directory you want reported, e.g. `cover "lib/**/*.rb", "app/**/*.rb"`, to keep the old additive-only behavior)
  - `use_merging` → `merging` (same value)
  - `enable_for_subprocesses` → `merge_subprocesses` (same value)
  - `enable_coverage_for_eval` → `enable_coverage :eval` (folds into the same call that enables `:line` / `:branch` / `:method`)
  - `print_error_status` (reader) → `print_errors` (the `print_error_status=` writer is unaffected for now)
- Calling `SimpleCov.start` from `.simplecov` is deprecated. Coverage tracking still begins for backward compatibility, but a one-time deprecation warning fires pointing the user at moving the call into `spec_helper.rb` / `test_helper.rb`; a future release will require the explicit `SimpleCov.start` from a test helper. The migration goes hand-in-hand with the bugfix below: once `SimpleCov.start` lives in the test helper, the parent process that auto-loads `.simplecov` never starts tracking and the empty-report-overwrite scenario can't arise. See [#&#8203;581](https://redirect.github.com/simplecov-ruby/simplecov/issues/581).
- `# :nocov:` toggle comments (and the configurable `SimpleCov.nocov_token` / `SimpleCov.skip_token`) are deprecated in favor of the new `# simplecov:disable` / `# simplecov:enable` directives. Each file that still uses `# :nocov:` emits a one-time deprecation warning to stderr at load time pointing at the recommended replacement, and any call to `SimpleCov.nocov_token` or `SimpleCov.skip_token` (getter or setter) likewise warns. The directive will be removed in a future release.
- `SimpleCov::SourceFile#branches_coverage_percent` and `#methods_coverage_percent` are deprecated in favor of the uniform `covered_percent(:branch)` / `covered_percent(:method)`. `covered_percent` (and `covered_strength`) now take a criterion argument (defaulting to `:line`), so the same call reaches any criterion instead of line being the unprefixed default while branch and method had their own differently-named methods. `coverage_statistics` also now accepts a criterion (e.g. `coverage_statistics(:branch)`) to return that one `CoverageStatistics` rather than the whole Hash.
- `minimum_coverage_by_file` and `minimum_coverage_by_group` are deprecated in favor of the `coverage` method's `minimum_per_file` / `minimum_per_group` verbs. The legacy methods overloaded a single hash to carry both per-criterion defaults and per-path / per-group overrides, with `minimum_coverage_by_file` further distinguishing Symbol keys (criterion defaults) from String / Regexp keys (path overrides) and accepting either a bare number or a per-criterion hash as the value. The `coverage` block fixes the criterion so every threshold is a plain percentage with an `only:` target. The setter form emits a deprecation warning naming the replacement; the no-arg getter (read internally) is unchanged. Replace e.g. `minimum_coverage_by_file line: 70, 'app/x.rb' => 100` with `coverage(:line) { minimum_per_file 70; minimum_per_file 100, only: 'app/x.rb' }`. See the "Per-criterion thresholds with `coverage`" README section.

#### Enhancements

- `simplecov uncovered` gained `--criterion line|branch|method` (default `line`) so the lowest-coverage listing can rank by branch or method coverage, not just line.
- Added the criterion-first `coverage` configuration method — a uniform way to configure each coverage criterion (`:line`, `:branch`, `:method`) in one place. `coverage :line do minimum 90; minimum_per_file 80; maximum_drop 5 end` (or the one-liner `coverage :branch, minimum: 80`) enables the criterion and declares its thresholds with identical syntax regardless of criterion, because the criterion is fixed by the enclosing call rather than smuggled into the argument as the historical "a bare number means line coverage, every other criterion needs a Hash" special case. Verbs: `minimum`, `maximum`, `exact`, `maximum_drop`, `minimum_per_file` (with `only:` String-path / Regexp overrides), and `minimum_per_group`. Options: `primary:` (the report's leading criterion), `oneshot:` (oneshot-lines mode for `:line`), and `:eval`. The flat `minimum_coverage` family remains as suite-wide sugar. Thresholds feed the same internal stores, so exit-code enforcement is unchanged. See the "Per-criterion thresholds with `coverage`" section in the README.
- JSON formatter: `coverage.json` now carries a top-level `$schema` field holding the URL of the versioned canonical JSON Schema the document conforms to, plus a human-readable `meta.schema_version` (`"major.minor"`, currently `"1.0"`). The versioned canonical lives at `schemas/coverage-v1.0.schema.json` and is immutable per version, an unversioned convenience alias at `schemas/coverage.schema.json` always tracks the latest. Downstream tools can validate inputs, generate types, or pin to a known shape, and the document-level `$schema` makes each payload self-describing. The schema version is independent of the gem version: additive changes bump minor, removals or shape changes bump major and ship as a new `schemas/coverage-vX.0.schema.json` file so prior-version consumers stay valid. `meta.commit` carries the git commit SHA the report was generated against (or null outside a git checkout), so tools can recover the exact source from history even when `source_in_json false` omits the per-file source arrays.
- Added `SimpleCov::ParallelAdapters` — a pluggable adapter interface for parallel test runners. SimpleCov's coordination with parallel test runners (deciding which worker does final-result work, waiting for siblings, knowing how many resultsets to expect) now routes through an adapter chain rather than hard-coding the `parallel_tests` gem's API. Two adapters ship: `ParallelTestsAdapter` wraps the historical grosser/parallel\_tests gem (precise, gem-API-based); `GenericAdapter` handles any runner that follows the `TEST_ENV_NUMBER` / `PARALLEL_TEST_GROUPS` env-var convention without shipping a Ruby API. The practical impact: **parallel\_rspec (and any similar env-var-only runner) now works out of the box** — previously every worker thought it was the "final" one and they clobbered each other's resultsets. Custom runners can register their own adapter via `SimpleCov::ParallelAdapters.register MyAdapter`, where `MyAdapter` subclasses `SimpleCov::ParallelAdapters::Base` and overrides the four contract methods (`active?`, `first_worker?`, `wait_for_siblings`, `expected_worker_count`). See [#&#8203;1065](https://redirect.github.com/simplecov-ruby/simplecov/issues/1065).
- Added `SimpleCov.ignore_branches` for opting out of synthetic `:else` branches that Ruby's `Coverage` library reports for constructs with no literal `else` keyword — exhaustive `case/in` pattern matches, `case/when` without `else`, `||=` / `&&=`, and `if` / `unless` without `else`. Variadic; only `:implicit_else` is supported today, with room for future synthetic branch types. Calling it without (or before) `enable_coverage :branch` is harmless — the setting is stored and applies once branch coverage is enabled. Explicit `else` arms still count. See [#&#8203;1033](https://redirect.github.com/simplecov-ruby/simplecov/issues/1033).
- Added `SimpleCov.cover` for declaring a positive coverage scope (the long-requested allowlist counterpart to `add_filter`). Accepts string globs, Regexps, blocks, or arrays of those; multiple calls union. When any `cover` matcher is configured the report drops every source file that doesn't match at least one of them, and string-glob matchers also expand on disk so files that exist but were never required during the run still appear in the report (at 0% coverage). Resolves the long-standing requests in [#&#8203;696](https://redirect.github.com/simplecov-ruby/simplecov/issues/696) and [#&#8203;869](https://redirect.github.com/simplecov-ruby/simplecov/issues/869). The companion `SimpleCov.no_default_skips` opts out of the filters that `SimpleCov.start` installs (hidden files, `vendor/bundle/`, test directories) so users who want to opt out wholesale don't have to call `clear_filters` themselves.
- `SimpleCov.formatter false` (and the equivalent `SimpleCov.formatters []`) now opts out of formatting entirely instead of raising `ConfigurationError`. `SimpleCov::Result#format!` returns `nil` when no formatter is configured. Intended for worker processes in big parallel CI runs (hundreds of jobs) where only a final `SimpleCov.collate` step needs a report — every other worker just drops its `.resultset.json` and exits without paying for HTML or multi-formatter output. See [#&#8203;964](https://redirect.github.com/simplecov-ruby/simplecov/issues/964).
- Setting `TEST_ENV_NUMBER` and `PARALLEL_TEST_GROUPS` no longer triggers a spurious `"SimpleCov guessed you were running inside parallel tests but couldn't load it"` warning when the `parallel_tests` gem isn't actually installed. SimpleCov now treats a `LoadError` on the auto-require as "the user set those env vars for some other reason (custom subprocess coordination, CI sharding) and isn't using parallel\_tests" and silently skips. The new `SimpleCov.parallel_tests true` / `false` setting forces the auto-require on or off when finer control is needed. See [#&#8203;1018](https://redirect.github.com/simplecov-ruby/simplecov/issues/1018).
- `SimpleCov.minimum_coverage_by_file` now accepts per-path overrides alongside the existing per-criterion defaults: pass String or Regexp keys to declare file- or directory-specific thresholds, e.g. `minimum_coverage_by_file line: 70, 'app/mailers/request_mailer.rb' => 100`. A String ending in `/` matches as a directory prefix; otherwise it must equal the project-relative path. Regexp keys match against the project-relative path. Per-path values may be a Numeric (primary criterion) or a per-criterion Hash; for each file the effective threshold is the defaults merged with any matching overrides (later overrides win per criterion, overrides win over defaults). The new overrides surface in `coverage.json` under the existing `errors.minimum_coverage_by_file` block. See [#&#8203;575](https://redirect.github.com/simplecov-ruby/simplecov/issues/575).
- Added `SimpleCov.maximum_coverage` (and the convenience `SimpleCov.expected_coverage`, which sets `minimum_coverage` and `maximum_coverage` to the same value) so the suite can be pinned to an exact coverage figure. A drop fails per the minimum; an unexpected increase also fails, prompting you to bump the threshold up rather than silently absorbing the improvement. Accepts the same Numeric / per-criterion Hash forms as `minimum_coverage`. Exits with status 4 (`SimpleCov::ExitCodes::MAXIMUM_COVERAGE`) when violated, and surfaces in `coverage.json` under `errors.maximum_coverage`. Comparisons floor the actual percent to two decimal places, so `expected_coverage 95.42` still passes when the actual is e.g. 95.4287. See [#&#8203;187](https://redirect.github.com/simplecov-ruby/simplecov/issues/187).
- Added a bundled `strict` profile (`SimpleCov.start "strict"`) that enables line, branch, and method coverage and pins the minimum threshold for each at 100%. Drops to line-only on engines without branch/method support (JRuby). See [#&#8203;1061](https://redirect.github.com/simplecov-ruby/simplecov/issues/1061).
- `SimpleCov.coverage_path` is now explicitly settable rather than always computed from `SimpleCov.root + SimpleCov.coverage_dir`. Setting it pins the report destination regardless of later `root` / `coverage_dir` changes — useful for out-of-tree build directories (CMake/CTest etc.) where the coverage report doesn't live under the source root. See [#&#8203;716](https://redirect.github.com/simplecov-ruby/simplecov/issues/716).
- The "Coverage report generated for X to Y" status line now prints the report path relative to the current working directory when it lives under cwd, and includes the entry-point filename — `coverage/index.html` from the HTML formatter, `coverage/coverage.json` from the JSON formatter — so the line points at a concrete file (and is clickable in terminals that hyperlink paths). Paths outside cwd stay absolute. See [#&#8203;197](https://redirect.github.com/simplecov-ruby/simplecov/issues/197).
- Added `SimpleCov.disable_coverage(criterion)` so a project can opt out of line coverage entirely — e.g. `enable_coverage :branch; disable_coverage :line` for a branch-only run. `SimpleCov.start` now raises `SimpleCov::ConfigurationError` when every criterion has been disabled. The formatter summary and JSON output emit only the criteria that were actually measured, so a branch-only run produces no `Line coverage:` line, no `lines` key in `coverage.json`, and no zero-padded line numbers in the HTML report. See [#&#8203;845](https://redirect.github.com/simplecov-ruby/simplecov/issues/845).
- Added `SimpleCov.remove_filter(arg)` to drop a specific filter (matching by `filter_argument`) and `SimpleCov.clear_filters` to wipe the entire chain. Useful for selectively turning off one of the defaults loaded by `SimpleCov.start` — e.g. `remove_filter(/\A\..*/)` to stop hiding paths that begin with a dot. The README's "Default filters" section enumerates what's loaded by default and how to disable each piece. See [#&#8203;803](https://redirect.github.com/simplecov-ruby/simplecov/issues/803).
- Terminal output is now colorized when stderr is a TTY: coverage percentages in the formatter summary line and threshold-violation messages are rendered green (>= 90%), yellow (>= 75%), or red (< 75%) — matching the HTML report's thresholds. The "SimpleCov failed with exit N…" line is red and the "Stopped processing SimpleCov…" line is yellow. Respects `NO_COLOR` (force off, per no-color.org) and `FORCE_COLOR` (force on); `NO_COLOR` wins if both are set. See [#&#8203;1157](https://redirect.github.com/simplecov-ruby/simplecov/issues/1157).
- CLI subcommands `coverage`, `report`, `uncovered`, and `diff` now colorize their printed percentages by the same threshold (and `diff` colors regressions red, improvements green). Auto-detect based on whether stdout is a TTY; the same `NO_COLOR` / `FORCE_COLOR` env vars apply. Each subcommand also accepts a `--no-color` flag as a per-invocation override.
- Added `# simplecov:disable` / `# simplecov:enable` directive comments for selectively skipping `line`, `branch`, and `method` coverage. Block form (own line) opens a region until the matching `# simplecov:enable`; inline form (trailing a code line) skips just that line. Categories may be combined (`# simplecov:disable line, branch`); omitting categories targets all three. Any trailing text is treated as a free-form reason and discarded (e.g. `# simplecov:disable line legacy adapter`). Directive markers inside string literals or heredocs are ignored.
- Added `SimpleCov.source_in_json` (default true) to make the per-file `source` array in `coverage.json` opt-out. Tools that read the project's source files from disk don't need the embedded copy, and on larger projects it dominates the JSON payload. The HTML report's `coverage_data.js` still embeds source unconditionally because the client-side viewer renders source from there. See [#&#8203;1143](https://redirect.github.com/simplecov-ruby/simplecov/issues/1143).
- JSON formatter: `meta.timestamp` is now emitted with millisecond precision (`iso8601(3)`) so the concurrent-overwrite warning can distinguish writes within the same wall-clock second
- JSON formatter: added `total` section with aggregate coverage statistics (covered, missed, total, percent, strength) for line, branch, and method coverage. Line stats additionally include `omitted` (count of blank/comment lines, i.e. lines that cannot be covered)
- JSON formatter: per-file output now includes `total_lines`, `lines_covered_percent`, and when enabled: `branches_covered_percent`, `methods` array, and `methods_covered_percent`
- JSON formatter: group stats now include full statistics for all enabled coverage types, not just line coverage percent
- JSON formatter: added `silent:` keyword to `JSONFormatter.new` to suppress console output
- Merged `simplecov-html` formatter into the main gem. A backward-compatibility shim ensures `require "simplecov-html"` still works.
- Merged `simplecov_json_formatter` into the main gem. A backward-compatibility shim ensures `require "simplecov_json_formatter"` still works.
- `CommandGuesser` now appends the framework name to parallel test data (e.g. `"RSpec (1/2)"` instead of `"(1/2)"`)
- `SimpleCov::Result.new` is roughly 7× faster for already-string-keyed input (the `SimpleCov.collate` hot path). The previous implementation deep-cloned each file's coverage data with `JSON.parse(JSON.dump(coverage))` per source file — a useful normalization for live `Coverage.result` symbol keys, but pure overhead for resultsets loaded from disk that already have string keys. `Result` now stringifies the outer hash keys with `transform_keys` only when needed; the inner branch/method-key shape is already handled by `SourceFile#restore_ruby_data_structure`. See [#&#8203;916](https://redirect.github.com/simplecov-ruby/simplecov/issues/916).
- Added `SimpleCov.parallel_wait_timeout` (default 60 seconds), which controls how long the process that writes the final report waits for the other parallel-test workers to finish writing their resultsets before it merges. Raise it when one worker runs much heavier test files and routinely finishes well after the others, so its coverage is included in the merge and the minimum and maximum coverage checks run against the full total instead of being skipped against a partial one. See [#&#8203;1171](https://redirect.github.com/simplecov-ruby/simplecov/issues/1171).
- Added `SimpleCov.finalize_merge` to separate storing mergeable worker resultsets from owning final report finalization. Parallel workers that write to explicit custom coverage destinations can now store their shard `.resultset.json` files without waiting on sibling shards they cannot see; an explicit `SimpleCov.collate` cleanup step then formats the merged report, enforces thresholds, and writes `.last_run.json`. SimpleCov infers this external-finalization mode only for recognized multi-worker parallel runs with merging enabled and a custom coverage destination, and emits a configuration warning until users set `finalize_merge false` (or `true`) explicitly. See [#&#8203;1215](https://redirect.github.com/simplecov-ruby/simplecov/issues/1215).
- The gem no longer packages `CHANGELOG.md`, trimming the largest avoidable part of the payload (the changelog grew with every release). The `changelog_uri` gem metadata already links the copy on GitHub.
- HTML report: file lists now window at 1,000 rows. Lists with more file rows matching the active filters show the first 1,000 and a "Showing the first 1,000 of N files. Show all" row at the bottom. Browsers lay out every visible table row again after each re-sort, which froze multi-thousand-file reports for seconds per header click. The window is presentation only (sorting, filtering, and the totals row still cover every file) and "Show all" restores the full list for the rest of the session. Measured on a synthetic 8,000-file report in headless Chrome, initial render drops from 3.8s to 0.8s and each re-sort from about 700ms to under 200ms. See [#&#8203;1171](https://redirect.github.com/simplecov-ruby/simplecov/issues/1171).
- HTML report: re-sorting and filtering large file lists is faster, and slow re-sorts are signaled instead of silently freezing the page. Sort values are cached after the first read, re-clicking the active sort column reverses the current order outright, name comparisons go through a cached collator, and file lists past 500 rows dim behind a "Sorting…" overlay that also absorbs stray clicks while the browser re-lays-out the table. Filter keystrokes are cheaper too: the coverage-bar width equalizer now writes one CSS custom property per measurement probe instead of inline styles on every bar, and the filename filter caches lowercased names instead of re-reading the DOM. See [#&#8203;1171](https://redirect.github.com/simplecov-ruby/simplecov/issues/1171).
- HTML report: file lists sort on load by the primary coverage criterion (`SimpleCov.primary_coverage`, line by default), ascending, so the least-covered files lead the list. To carry the primary criterion to the report, `coverage.json` includes a `primary_coverage` field in `meta`, also documented in the coverage schema. See [#&#8203;1171](https://redirect.github.com/simplecov-ruby/simplecov/issues/1171).

#### Bugfixes

- Added `:eval_generated` tokens to `SimpleCov.ignore_branches` and the new `SimpleCov.ignore_methods` so projects using macros like Rails' `delegate` (or any pattern that calls `module_eval(body, __FILE__, __LINE__)`) can drop the synthetic branch and method entries those macros inject. Ruby's `Coverage` attributes eval'd code to the caller's `__FILE__` / `__LINE__`, so a `delegate :foo, to: :bar` line surfaces as if it had a `def foo` and an `if` branch right there. Detection uses Prism to walk the static source and treats any Coverage entry whose start\_line lacks a real `def` keyword (for methods) or branch construct (for branches) as eval-generated. Opt in with `ignore_methods :eval_generated` and / or `ignore_branches :eval_generated`. Prism ships with Ruby 3.3+; on older Rubies `gem install prism` enables the filter, otherwise the setting is a no-op. See [#&#8203;1046](https://redirect.github.com/simplecov-ruby/simplecov/issues/1046).
- Files added via `cover` / `track_files` that were never `require`'d during the run now contribute branch and method entries to the report, not just lines. Previously `SimulateCoverage` left those fields as empty hashes (because parsing source ourselves felt risky), which made unloaded files invisible to the branch and method denominators while their lines DID count — so a `cover "{app,lib}/**/*.rb"` glob over files without specs silently inflated branch% relative to line% (the OP's reproduction was via SonarQube, which surfaces the asymmetry more visibly than the SimpleCov HTML report). Branches and methods are now enumerated statically via `SimpleCov::StaticCoverageExtractor`, which uses Prism to walk the AST and emits Coverage-shaped tuples without loading the file. The shape matches what Ruby's own `Coverage` library reports for the same source: `:if` / `:case` / `:while` / `:until` constructs plus their `:then` / `:else` / `:when` / `:in` / `:body` arms, with the synthetic `:else` for case-without-explicit-else that the `ignore_branches :implicit_else` setting (see Enhancements) targets. Prism is bundled with Ruby 3.3+; on older Rubies `gem install prism` enables the fix, otherwise SimulateCoverage falls back to the previous "empty hashes" behavior. See [#&#8203;1059](https://redirect.github.com/simplecov-ruby/simplecov/issues/1059).
- HTML report: two groups whose names share an alphanumeric suffix but differ only in a leading non-letter (e.g. `">100LOC"` / `"<10LOC"`, or any pair using different special characters) no longer render into the same DOM container. The JS that built HTML ids from group names stripped every non-letter prefix and then every remaining non-alphanumeric char, so both names sanitized to `"LOC"` and the second group silently replaced the first in the rendered tabs. The new encoding (`"g-" + each-non-id-char-as-hex`) preserves uniqueness across all input shapes. See [#&#8203;1038](https://redirect.github.com/simplecov-ruby/simplecov/issues/1038).
- HTML report: filenames containing `"` or `'` characters are now escaped when rendered into `title="..."` attributes. The previous DOM-based `escapeHTML` only escaped `&`, `<`, and `>`, so a project with such filenames could break out of the attribute. The replacement encodes all five HTML-attribute-sensitive characters via a `replace` callback (also avoids allocating a DOM node per call).
- HTML report: files with literally 0% branch or method coverage now display 0% instead of 100%. The per-file row rendering used `f.branches_covered_percent || 100.0` (and the method equivalent), which treated a real `0` as falsy and substituted the disabled-criterion fallback. The check now distinguishes "criterion disabled" (`undefined`) from "criterion measured zero" (`0`).
- `SimpleCov::Result` now warns when it drops source files because their absolute paths aren't on the local filesystem, instead of silently producing an empty `0 / 0 (100.00%)` report. The most common trigger is `SimpleCov.collate` invoked from a machine or working directory different from where the individual resultsets were generated — when *every* entry is missing the warning explicitly names that case and points at the issue; when only some are missing the warning is quieter and lists up to five paths with a `(+N more)` suffix. See [#&#8203;980](https://redirect.github.com/simplecov-ruby/simplecov/issues/980).
- Files added via `track_files` that were never loaded now use the same line classification as loaded files. Previously, `SimulateCoverage` ran the file through `LinesClassifier`, which marks every non-blank, non-comment line as relevant — so a multi-line method chain `@x = a.foo.bar` reported 4 relevant lines for the unloaded copy and 2 for the loaded copy, throwing off per-file and overall percentages. `SimulateCoverage` now uses `Coverage.line_stub` (the same stub Ruby would have produced if the file were required), then overlays `# :nocov:` toggles and `# simplecov:disable line` directive ranges that the runtime doesn't know about. The two paths now agree on every shape: multi-line statements, `end` keywords, blank lines, and SimpleCov-specific exclusion comments. Some projects will see their `tracked_files` percentages shift as a result. See [#&#8203;654](https://redirect.github.com/simplecov-ruby/simplecov/issues/654).
- Fix the parent-process / subprocess race where a Rakefile (or Rails `Bundler.require`) caused `.simplecov` to auto-load `SimpleCov.start` in the rake parent, which then shelled out to a test runner subprocess; the subprocess wrote a correct report, then the parent's `at_exit` would clobber it with an empty 0% report. Three layers of defense now apply: (1) `.simplecov` is treated as configuration only and no longer starts tracking from the parent (see Deprecations); (2) `Result

> ✂ **Note**
> 
> PR body was truncated to here.

</details>

---

### Configuration

📅 **Schedule**: (UTC)

- Branch creation
  - At any time (no schedule defined)
- Automerge
  - At any time (no schedule defined)

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR becomes conflicted, or you tick the rebase/retry checkbox.

👻 **Immortal**: This PR will be recreated if closed unmerged. Get [config help](https://redirect.github.com/renovatebot/renovate/discussions) if that's undesired.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Mend Renovate](https://redirect.github.com/renovatebot/renovate).


torreypayne:
### ⚠️ Artifact update problem

Renovate failed to update an artifact related to this branch. You probably do not want to merge this PR as-is.

♻ Renovate will retry this branch, including artifacts, only when one of the following happens:

 - any of the package files in this branch needs updating, or 
 - the branch becomes conflicted, or
 - you click the rebase/retry checkbox if found above, or
 - you rename this PR's title to start with "rebase!" to trigger it manually

The artifact failure details are included below:

##### File name: google-cloud-pubsub/Gemfile.lock

```
Writing lockfile to /tmp/renovate/repos/github/googleapis/google-cloud-ruby/google-cloud-pubsub/Gemfile.lock
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...

Could not find compatible versions

Because every version of google-cloud-pubsub depends on retriable ~> 3.1
  and Gemfile depends on google-cloud-pubsub >= 0,
  retriable ~> 3.1 is required.
So, because Gemfile depends on retriable ~> 4.2.0,
  version solving has failed.

```

PR #34956, opened 2026-07-17
feat(storage): Support for idempotency_token_header

shubhangi-google:
adding support for X-Goog-Gcs-Idempotency-Token
linked core library implementation - https://github.com/googleapis/google-api-ruby-client/pull/27338


shubhangi-google:
The failing test cases will pass once core library Pr is released

PR #34844, opened 2026-07-16
chore(storage): fix order dependency in get_object_contexts test

aandreassa:
Currently this fails if the order is incorrect, introducing flakiness:
```
1) Failure:
Files Snippets::get_object_contexts#test_0001_fetches all object contexts [acceptance/files_test.rb:341]:
In stdout.
--- expected
+++ actual
@@ -1,4 +1,4 @@
 "Custom Contexts for path/file_name_9d512d3ea21775ecf8d953f9f31138f6.txt are:
-Key: my-custom-key, Value: my-custom-value
 Key: my-custom-key-2, Value: my-custom-value-2
+Key: my-custom-key, Value: my-custom-value
 "

```

Noticed in https://github.com/googleapis/google-cloud-ruby/pull/34751

chore(storage): pass user_project when resetting requester_pays in samples cleanup

Fixes an order dependency where resetting `requester_pays` in the `after` cleanup hook failed
due to a missing `user_project` parameter, breaking subsequent tests.

Example error:
```
  1) Error:
Files Snippets#test_0001_list_files:
Google::Cloud::InvalidArgumentError: required: Bucket is a requester pays bucket but no user project provided.
    /workspace/google-cloud-storage/lib/google/cloud/storage/service.rb:1037:in 'Google::Cloud::Storage::Service#execute'
    /workspace/google-cloud-storage/lib/google/cloud/storage/service.rb:167:in 'Google::Cloud::Storage::Service#patch_bucket'
    /workspace/google-cloud-storage/lib/google/cloud/storage/bucket.rb:3467:in 'Google::Cloud::Storage::Bucket#patch_gapi!'
    /workspace/google-cloud-storage/lib/google/cloud/storage/bucket.rb:656:in 'Google::Cloud::Storage::Bucket#requester_pays='
    acceptance/files_test.rb:80:in 'block (2 levels) in <top (required)>'
```

PR #35000, opened 2026-07-23
chore(storage): Fix flaky test list_object_context

shubhangi-google:
PR Fixes following sample failure:
```
 1) Failure:
 Files Snippets::list_object_contexts#test_0001_filters out files on the basis of custom context key [acceptance/files_test.rb:351]:
In stdout.
--- expected
+++ actual
@@ -1,2 +1 @@
-"File: path/file_name_11435b5825845b3fdbe1af94a783ea46.txt has context key: my-custom-key
-"
+""
```


aandreassa:
@shubhangi-google I already started working on this. https://github.com/googleapis/google-cloud-ruby/pull/34844

shubhangi-google:
> @shubhangi-google I already started working on this. #34844

@aandreassa  saw your PR this PR addresses another flaky bug related to list object contexts
I have updated my PR and have changed the heading and description to avoid confusion

google-cloud-storage/samples/acceptance/files_test.rb
cpriti-os: can we do this at a lesser interval, does it take 5s for consistency? This high wait time could add up eventually and lead to very high test execution time, try 0.2 10 times?

cpriti-os (changes requested):

google-cloud-storage/samples/acceptance/files_test.rb
```diff
@@ -349,25 +348,40 @@ def mock_cipher.random_key
...
     end
 
+    
     it "filters out files on the basis of custom context key" do
-      assert_output "File: #{remote_file_name} has context key: #{custom_context_key1}\n" do
-        list_object_contexts bucket_name: bucket.name, custom_context_key: custom_context_key1
+      out = nil
+      10.times do
+        out, _err = capture_io do
+          list_object_contexts bucket_name: bucket.name, custom_context_key: custom_context_key1
```
cpriti-os, line 366: We are setting the same context key in other tests as well, if its possible that list can return those files then this test could still fail. It would be better to use keys with some random part to reliably run this test.

cpriti-os (changes requested):

google-cloud-storage/samples/acceptance/files_test.rb
shubhangi-google: done

cpriti-os (approved):
The changes look good! Using `SecureRandom.hex` and reducing the wait time should resolve the flakiness. 

LGTM!

PR #35013, opened 2026-07-24
chore(google-ads-ad_manager-v1): Remove Gemlock.file from static config

aandreassa:
This seems to be the only missing one. 

It missed it in https://github.com/googleapis/google-cloud-ruby/pull/34997 before reverting via https://github.com/googleapis/google-cloud-ruby/pull/34998


aandreassa:
@torreypayne it should be the last thing that got through!

torreypayne (approved):

PR #35031, opened 2026-07-28
chore(storage): Error handling for open and unlinked files

### LINKED ISSUE #34987: storage: Error during CRC check when uploading an open + unlinked file
#### Context

The `google-cloud-storage` library attempts to reopen any `File` that is attempted to be uploaded. See this line:
https://github.com/shubhangi-google/google-cloud-ruby/blob/042468dcccc83a802ae4993e76ebea203827d3cb/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb#L84

**This is a problem when the file no longer exists**.

When would that actually happen? The `puma` web server notably unlinks files immediately after storing, while keeping the file descriptor open:
https://github.com/puma/puma/blob/v8.0.2/lib/puma/client.rb#L507-L511

Although this trips people up, it is very much intentional and in fact recommended by the Ruby documentation.
https://github.com/puma/puma/issues/3597
https://github.com/aws/aws-sdk-ruby/issues/3163
https://github.com/aws/aws-sdk-ruby/pull/3164
https://docs.ruby-lang.org/en/3.4/Tempfile.html#class-Tempfile-label-Unlink+after+creation

It would be tremendously helpful if the `google-cloud-storage` client would not assume that a `File` or `Tempfile` is still linked when it is being uploaded.

#### Environment details

  - OS: Alpine Linux, Debian Linux
  - Ruby version: 3.4
  - Gem name and version: google-cloud-storage (1.62.0)

#### Steps to reproduce

  1. Use Puma web server
  2. Upload files above configured memory limit
  3. Observe that google-cloud-storage fails to compute digest

#### Code example

This is a minimal example that mimics Puma's behaviour but it is a much simpler reproduction.

```ruby
source "https://rubygems.org"

gem "google-cloud-storage", "1.62.0"
```

```ruby
require "tempfile"

storage = Google::Cloud::Storage.anonymous
bucket = storage.bucket("unused", skip_lookup: true)

Tempfile.create("gcs-repro", binmode: true) do |file|
  file.write("still readable")
  file.rewind

  path = file.path
  File.unlink(path)

  puts "Path exists: #{File.exist?(path)}"
  puts "Contents: #{file.read.inspect}"
  file.rewind

  bucket.create_file(file, "test")
end
```

#### Full backtrace

```sh
3.4-slim: Pulling from library/ruby
b16cb8fde57a: Pulling fs layer
ead741ce7b27: Pulling fs layer
59f54fbcd984: Pulling fs layer
b160435c969d: Pulling fs layer
52a2f7711503: Pulling fs layer
b16cb8fde57a: Download complete
ead741ce7b27: Download complete
b160435c969d: Download complete
4371fa0b2e33: Download complete
79dd7f77afe3: Download complete
52a2f7711503: Download complete
59f54fbcd984: Download complete
59f54fbcd984: Pull complete
ead741ce7b27: Pull complete
b160435c969d: Pull complete
b16cb8fde57a: Pull complete
52a2f7711503: Pull complete
Digest: sha256:614edae6a80eb2a7cf1984f03a6d814f523a48355e8f96deb5ddd25faa86353e
Status: Downloaded newer image for ruby:3.4-slim
/usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/file/verifier.rb:84:in 'File#initialize': No such file or directory @ rb_sysopen - /tmp/gcs-repro20260721-29-v1priq (Errno::ENOENT)
	from /usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/file/verifier.rb:84:in 'IO.open'
	from /usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/file/verifier.rb:84:in 'Google::Cloud::Storage::File::Verifier._digest_for'
	from /usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/file/verifier.rb:59:in 'Google::Cloud::Storage::File::Verifier.crc32c_for'
	from /usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/bucket.rb:3507:in 'Google::Cloud::Storage::Bucket#crc32c_for'
	from /usr/local/bundle/gems/google-cloud-storage-1.62.0/lib/google/cloud/storage/bucket.rb:2007:in 'Google::Cloud::Storage::Bucket#create_file'
	from -e:1:in 'block in <main>'
	from /usr/local/lib/ruby/3.4.0/tempfile.rb:576:in 'Tempfile.create_with_filename'
	from /usr/local/lib/ruby/3.4.0/tempfile.rb:562:in 'Tempfile.create'
	from -e:1:in '<main>'
path exists: false
contents: still readable
```

shubhangi-google:
Fixes issue where uploading unlinked temporary files (such as large payloads handled by the Puma web server) fails during the CRC32c digest computation.
closes https://github.com/googleapis/google-cloud-ruby/issues/34987

### DIFF
--- a/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
@@ -81,18 +81,18 @@ module Google
           def self._digest_for local_file, digest_class
 
             if local_file.respond_to?(:read)
-              # Case 1: Input is an open stream (File, StringIO, ActionDispatch::Http::UploadedFile, etc.).
+              # Case 1: Input is an open stream (File or StringIO).
               local_file.rewind if local_file.respond_to?(:rewind)
-              digest = digest_class.new
-              while (chunk = local_file.read(16 * 1024))
-                digest.update(chunk)
-              end
+              digest = digest_class.base64digest local_file.read
               local_file.rewind if local_file.respond_to?(:rewind)
-              digest.base64digest
+              digest
             elsif local_file.respond_to?(:to_path) || local_file.is_a?(String)
-              # Case 2: Input is a file path (String, Pathname).
-              digest_class.file(Pathname(local_file).to_path).base64digest
+              # Case 2: Input is a file path (String, Pathname, or object that responds to :to_path).
+              ::File.open Pathname(local_file).to_path, "rb" do |f|
+                digest_class.file(f).base64digest
+              end
             end
+
           end
         end
       end
--- a/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
@@ -68,7 +68,7 @@ module Google
           #    the entire file into memory. The stream is rewound before and after
           #    reading to ensure its position is not permanently changed.
           # 2. A file path (String or Pathname): It efficiently streams the file
-          #    directly via the digest C-extension.
+          #    to compute the digest without loading the entire file into memory.
           #
           # @param local_file [String, Pathname, IO] The local file path or IO
           #   stream for which to compute the digest.
--- a/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
@@ -92,7 +92,6 @@ module Google
                 digest_class.file(f).base64digest
               end
             end
-
           end
         end
       end
--- a/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
@@ -79,13 +79,18 @@ module Google
           # @return [String] The base64-encoded digest of the file's content.
           #
           def self._digest_for local_file, digest_class
-
             if local_file.respond_to?(:read)
               # Case 1: Input is an open stream (File or StringIO).
               local_file.rewind if local_file.respond_to?(:rewind)
-              digest = digest_class.base64digest local_file.read
+              
+              digest = digest_class.new
+              buf = ""
+              while local_file.read(16_384, buf)
+                digest.update buf
+              end
+              
               local_file.rewind if local_file.respond_to?(:rewind)
-              digest
+              digest.base64digest
             elsif local_file.respond_to?(:to_path) || local_file.is_a?(String)
               # Case 2: Input is a file path (String, Pathname, or object that responds to :to_path).
               ::File.open Pathname(local_file).to_path, "rb" do |f|
--- a/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
+++ b/google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
@@ -63,33 +63,35 @@ module Google
           # Computes a base64-encoded digest for a local file or IO stream.
           #
           # This method handles two types of inputs for `local_file`:
-          # 1. A file path (String or Pathname): It efficiently streams the file
-          #    to compute the digest without loading the entire file into memory.
-          # 2. An IO-like stream (e.g., File, StringIO): It reads the stream's
-          #    content to compute the digest. The stream is rewound before and after
+          # 1. An IO-like stream (e.g., File, Tempfile, StringIO): It reads the
+          #    stream's content in chunks to compute the digest without loading
+          #    the entire file into memory. The stream is rewound before and after
           #    reading to ensure its position is not permanently changed.
+          # 2. A file path (String or Pathname): It efficiently streams the file
+          #    directly via the digest C-extension.
           #
           # @param local_file [String, Pathname, IO] The local file path or IO
           #   stream for which to compute the digest.
           # @param digest_class [Class] The digest class to use for the
-          #   calculation (e.g., `Digest::MD5`). It must respond to `.file` and
-          #   `.base64digest`.
+          #   calculation (e.g., `Digest::MD5`). It must respond to `.new`,
+          #   `.file` and `.base64digest`.
           #
           # @return [String] The base64-encoded digest of the file's content.
           #
           def self._digest_for local_file, digest_class
 
-            if local_file.respond_to?(:to_path) || local_file.is_a?(String)
-              # Case 1: Input is a file path (String, Pathname, or object that responds to :to_path).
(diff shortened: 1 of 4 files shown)


google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
```diff
@@ -63,33 +63,35 @@ def self.crc32c_for local_file
           # Computes a base64-encoded digest for a local file or IO stream.
           #
           # This method handles two types of inputs for `local_file`:
-          # 1. A file path (String or Pathname): It efficiently streams the file
```
bajajneha27, line 66: Why are we removing this?

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
```diff
@@ -63,33 +63,35 @@ def self.crc32c_for local_file
...
-          #    to compute the digest without loading the entire file into memory.
-          # 2. An IO-like stream (e.g., File, StringIO): It reads the stream's
-          #    content to compute the digest. The stream is rewound before and after
+          # 1. An IO-like stream (e.g., File, Tempfile, StringIO): It reads the
+          #    stream's content in chunks to compute the digest without loading
+          #    the entire file into memory. The stream is rewound before and after
           #    reading to ensure its position is not permanently changed.
+          # 2. A file path (String or Pathname): It efficiently streams the file
```
bajajneha27, line 70: I'm not sure if this is true

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
```diff
@@ -63,33 +63,35 @@ def self.crc32c_for local_file
...
           #
           # @return [String] The base64-encoded digest of the file's content.
           #
           def self._digest_for local_file, digest_class
 
-            if local_file.respond_to?(:to_path) || local_file.is_a?(String)
-              # Case 1: Input is a file path (String, Pathname, or object that responds to :to_path).
-              ::File.open Pathname(local_file).to_path, "rb" do |f|
-                digest_class.file(f).base64digest
+            if local_file.respond_to?(:read)
```
bajajneha27, line 82: Why are we removing existing code?

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
shubhangi-google: we are swapping the order of Case 1 and Case 2 to match the updated logic below. We need to check if the object is an IO-like stream (i.e., responds to :read) before checking if it's a file path. If we check for the file path first, an unlinked Tempfile will still return true for :to_path, but attempting to open that path will raise an Errno::ENOENT error. Checking if we can read it directly (:read) first fixes this.

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
shubhangi-google: we are swapping the existing code  to check for the :read capability (the IO-stream case) before falling back to the :to_path capability (the file path case).

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
shubhangi-google: reverted to previous version

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
bajajneha27: Still we're removing some of the cases. Can we try to keep our changes as minimum as possible?

bajajneha27 (changes requested):

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
shubhangi-google: done kept relevant changes only

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
```diff
@@ -63,33 +63,34 @@ def self.crc32c_for local_file
...
           # @return [String] The base64-encoded digest of the file's content.
           #
           def self._digest_for local_file, digest_class
 
-            if local_file.respond_to?(:to_path) || local_file.is_a?(String)
-              # Case 1: Input is a file path (String, Pathname, or object that responds to :to_path).
+            if local_file.respond_to?(:read)
+              # Case 1: Input is an open stream (File or StringIO).
+              local_file.rewind if local_file.respond_to?(:rewind)
+              digest = digest_class.base64digest local_file.read
```
torreypayne: @shubhangi-google one thing to note about this solution:

```
digest = digest_class.base64digest local_file.read  # <-- Entire file loaded into RAM!
```

The existing production flow does `digest_class.file(f).base64digest` which loads the file in chunks and uses O(1) memory. I'm wondering if it makes sense to do something like:

```
local_file.rewind if local_file.respond_to? :rewind
digest = digest_class.new
buf = ""
while local_file.read(16_384, buf)
  digest.update buf
end
local_file.rewind if local_file.respond_to? :rewind
digest.base64digest
...
```

torreypayne (approved):

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
shubhangi-google: done

google-cloud-storage/lib/google/cloud/storage/file/verifier.rb
```diff
@@ -63,33 +63,39 @@ def self.crc32c_for local_file
...
           def self._digest_for local_file, digest_class
-
-            if local_file.respond_to?(:to_path) || local_file.is_a?(String)
-              # Case 1: Input is a file path (String, Pathname, or object that responds to :to_path).
+            if local_file.respond_to?(:read)
+              # Case 1: Input is an open stream (File or StringIO).
+              local_file.rewind if local_file.respond_to?(:rewind)
+              
+              digest = digest_class.new
+              buf = ""
```
bajajneha27, line 87: Maybe we can make this more idiomatic Ruby.

```Ruby
local_file.each(16_384) do |chunk|
  digest.update(chunk)
end
```

PR #35106, opened 2026-08-04
chore(renovate): restrict update schedule to weekends and group minor and patch dependencies

torreypayne:
This PR optimizes `.github/renovate.json5` configuration to throttle and group dependency updates as requested in b/525027506:

1. **Weekend Schedule Window**: Restricts Renovate execution schedule to `"after 8pm on Friday and before 6am on Monday"` (replacing `"schedule:monthly"` in `extends`), ensuring updates are generated on weekends.
2. **Dependency Grouping**: Groups all minor and patch updates together into a single collective `"all minor and patch dependencies"` PR.
3. **Lockfile Maintenance**: Keeps automated `Gemfile.lock` maintenance enabled across all monorepo gems (`lockFileMaintenance`), which inherits the root weekend schedule.

Fixes b/525027506

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,9 +2,11 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     // Use pre-determined best practices, including monorepo groupings & GitHub Action digest pinning
-    "config:best-practices",
-    // Create version bump PRs globally on a monthly basis (first day of the month)
-    "schedule:monthly"
+    "config:best-practices"
+  ],
+  // Restrict Renovate's execution to weekend window
+  "schedule": [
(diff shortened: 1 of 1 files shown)


aandreassa (changes requested):
Should we just start with lockfile maintenance for all first? 

Once we see that working, then we would need to touch handwrittens only with renovate. We don't want it to change things that come from the generator (or they will get reverted).

torreypayne:
> Should we just start with lockfile maintenance for all first?
> 
> Once we see that working, then we would need to touch handwrittens only with renovate. We don't want it to change things that come from the generator (or they will get reverted).

Ya I'm open to that! I just wanted to start looking into this, but can hold off.

PR #36179, opened 2026-08-05
chore: update API sources and regenerate

JoeWang1127:


librarian.yaml
```diff
@@ -15,8 +15,8 @@ language: ruby
 version: v0.33.0
 sources:
   googleapis:
-    commit: 6145fa8cc2b137bf4c0ed114e2e39c1157ea9722
-    sha256: 489923d12c766c9e323b42de057037977b03f20c8a874bc8553d29f9dd3f0e3b
+    commit: 57394718a71457aa37ab002352b22081976bdd91
```
codyoss, line 18: Should there be library changes related to this sha change?

JoeWang1127: We only onboard two libraries so far so it's fine these two libraries have no change.

codyoss (approved):

PR #36232, opened 2026-08-11
fix: onboarding library batch 3

JoeWang1127:
See go/sdk-librarian-ruby-tracker


librarian.yaml
```diff
@@ -661,14 +665,14 @@ libraries:
       wrapper_of:
         - v1:0.12
   - name: google-cloud-bigquery-data_transfer-v1
+    version: 1.7.1
```
aandreassa, line 668: It would be good to figure out why this keeps happening. Maybe release-please needs to update the correct line?

aandreassa (approved):

PR #36299, opened 2026-08-19
chore: update librarian to v0.38.0

quirogas:


suztomo (approved):
@JoeWang1127 There seems to have been a step that missed `librarian tidy`. This pull request fixes the order.

PR #36310, opened 2026-08-20
fix: onboarding library batch 4

suztomo:
See go/sdk-librarian-ruby-tracker


google-cloud-alloy_db-v1/.gitignore
```diff
@@ -1,3 +1,5 @@
+# Ignore bundler lockfiles
+Gemfile.lock
```
torreypayne, line 2: I'm curious whether you have pulled the latest `gapic-generator-ruby` main? We updated the generator to no longer ignore lockfiles since we need to pin dependencies for our CI/CD pipeline; see b/509981628

suztomo: That resolves many questions. I was usng old gapic-generator-ruby. Thank you.

PR #36479, opened 2026-08-27
chore(pubsub): Add inventory wait and stream shutdown nack

aandreassa:
Adds internal primitives in `Inventory` and `Stream` in preparation for exposing subscriber shutdown options in a follow-up change:

1. Added `Inventory#wait_until_empty(timeout)` to block until messages finish processing or time out.
2. Updated `Stream#stop` and `Stream#wait!` to support immediate nacking and waiting for message processing to complete.
3. Added unit tests in `inventory_test.rb` and `stream_test.rb`.

### DIFF
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener.rb
@@ -71,6 +71,7 @@ module Google
         attr_reader :message_ordering
         attr_reader :callback_threads
         attr_reader :push_threads
+        attr_reader :shutdown_behavior
 
         ##
         # @private Implementation attributes.
@@ -91,6 +92,7 @@ module Google
           @subscription_name = subscription_name
           @deadline = deadline || 60
           @streams = streams || 1
+          @shutdown_behavior = :wait_for_processing
           coerce_inventory inventory
           @message_ordering = message_ordering
           @callback_threads = Integer(threads[:callback] || 8)
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
@@ -121,14 +121,14 @@ module Google
 
               if timeout
                 target_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
-                while !@inventory.empty?
+                while !@inventory.empty? && !@stopped
                   remaining = target_time - Process.clock_gettime(Process::CLOCK_MONOTONIC)
                   break if remaining <= 0
 
                   @wait_cond.wait remaining
                 end
               else
-                @wait_cond.wait_while { !@inventory.empty? }
+                @wait_cond.wait_while { !@inventory.empty? && !@stopped }
               end
               @inventory.empty?
             end
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
@@ -109,6 +109,31 @@ module Google
             end
           end
 
+          ##
+          # @private
+          # Blocks until the inventory is empty or until timeout expires.
+          #
+          # @param [Numeric, nil] timeout Maximum time in seconds to wait, or nil to wait indefinitely.
+          # @return [Boolean] true if inventory became empty, false if timed out.
+          def wait_until_empty timeout = nil
+            synchronize do
+              return true if @inventory.empty?
+
+              if timeout
+                target_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+                while !@inventory.empty?
+                  remaining = target_time - Process.clock_gettime(Process::CLOCK_MONOTONIC)
+                  break if remaining <= 0
+
+                  @wait_cond.wait remaining
+                end
+              else
+                @wait_cond.wait_while { !@inventory.empty? }
+              end
+              @inventory.empty?
+            end
+          end
+
           def start
             @background_thread ||= Thread.new { background_run }
 
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
@@ -50,7 +50,7 @@ module Google
           end
 
           def ack_ids
-            @inventory.keys
+            synchronize { @inventory.keys }
           end
 
           def add *rec_msgs
--- a/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
+++ b/google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
@@ -134,11 +134,11 @@ module Google
 
               @keepalive_monitor.stop
 
-              # When :nack_immediately, release all queued messages immediately.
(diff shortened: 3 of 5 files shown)


google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/inventory.rb
```diff
@@ -109,6 +109,31 @@ def empty?
...
+              if timeout
+                target_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) + timeout
+                while !@inventory.empty?
+                  remaining = target_time - Process.clock_gettime(Process::CLOCK_MONOTONIC)
+                  break if remaining <= 0
+
+                  @wait_cond.wait remaining
+                end
+              else
+                @wait_cond.wait_while { !@inventory.empty? }
```
torreypayne: Should we be waiting for `|| @stopped == true` as an additional condition? I walked through the `#stop` calls with Gemini and if the waiting thread wakes up, checks `!@inventory.empty?`, and messages remain (e.g., an unhandled error or stuck callback), it re-enters wait and sleeps indefinitely. Teardown cannot interrupt it. My only feedback is nitty edge-cases so just bringing these up to consider.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -157,10 +184,19 @@ def running?
...
 
+          ##
+          # @private
+          # Blocks until all received messages are processed, or until timeout expires.
+          #
+          # @param [Numeric, nil] timeout The maximum seconds to wait, or nil to wait indefinitely.
+          # @return [Stream] self for chaining.
           def wait! timeout = nil
-            # Wait for all queued callbacks to be processed.
+            emptied = @inventory.wait_until_empty timeout
```
torreypayne, line 195: since you're passing the full timeout value to both methods, we could have up to 2 x timeout wait here

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -157,10 +184,19 @@ def running?
...
+            nack_unprocessed_messages! unless emptied
```
torreypayne, line 196: @callback_thread_pool may still be actively running user code for those messages. When those callbacks subsequently complete and attempt to ack!, nack!, or hit release, they will operate on ack IDs that were already nacked and removed from inventory. The thread pool should be shut down / cancelled before nacking in this failure path

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener.rb
```diff
@@ -71,6 +71,7 @@ class MessageListener
         attr_reader :message_ordering
         attr_reader :callback_threads
         attr_reader :push_threads
+        attr_reader :shutdown_behavior
```
torreypayne, line 74: Should we be accepting this param in the initializer so the tests can set their own value? Also should we be setting it as a `@private` var? I'm still getting used to YARD

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
torreypayne: Because of the switch to `nack_immediately`, LLM thinks we need to consider the following:

`running?` is `!stopped?`. Once `stream.stop` is called, `running?` evaluates to `false`.
Failure Mode: For ordered subscriptions (`@sequencer`), as each active callback finishes, `@sequencer.next` is skipped. Any remaining messages queued behind that ordering key in `@sequencer` are never dispatched to callbacks and remain stuck in `@inventory`. Consequently, `inventory.wait_until_empty` hangs forever (or until timeout), defeating `:wait_for_processing`.

Fix: Update the guard to check !nack_immediately? rather than `running?`:

```
if @sequencer && !nack_immediately?
```

torreypayne (changes requested):
Can you also include some context from the design docs on what this is solving/how it rolls up to the larger initiative?

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener.rb
aandreassa: Let's discuss this in the follow-up PR. This was the easiest way I found to split the behavior between the two changes. We can fix there.

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
aandreassa: You are right, fixed it!

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
aandreassa: Fixed, ty!

aandreassa:
@torreypayne thanks for the feedback. What kind of context were you hoping for?

Today, stopping a subscriber abandons buffered messages in memory without NACKing them, leaving them in limbo until server-side ack deadlines expire, while queued callbacks are abruptly dropped.

This initiative introduces two standard behaviors:

- `nack_immediately`: Actively NACKs all held messages on shutdown so other workers can process them immediately without latency.
 -`wait_for_processing`: True graceful shutdown, stops pulling new messages but drains in-memory messages before exiting.
 
 I am adding some docs in the last PR as well!

aandreassa:
Rebased! Please note the second commit so its hopefully easier to review requested changes @torreypayne

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -129,22 +134,44 @@ def stop
...
 
             self
           end
 
+          ##
+          # @private
+          # Nacks all messages currently held in inventory.
+          def nack_unprocessed_messages!
+            synchronize do
+              ack_ids = @inventory.ack_ids
```
torreypayne, line 152: We should synchronize the `@inventory#keys` calls:

```
def ack_ids
  synchronize { @inventory.keys }
end
```

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
```diff
@@ -129,22 +134,44 @@ def stop
...
+              return if ack_ids.empty?
+
+              @subscriber.buffer.modify_ack_deadline 0, ack_ids
```
torreypayne, line 155: These won't flush immediately unless you explicitly `#flush`:

```
@subscriber.buffer.modify_ack_deadline 0, ack_ids
@subscriber.buffer.flush!
@inventory.remove ack_ids
```

torreypayne: I think this might also affect the tests too so I'd double-check that afterwards

torreypayne (approved):

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
aandreassa: Awesome, thanks for catching it!

google-cloud-pubsub/lib/google/cloud/pubsub/message_listener/stream.rb
aandreassa: Thank you! I updated the test a bit to catch that case.

